Issue #8897, opened 2019-06-21
Writing quantity with user-defined unit to fits table breaks if unit is registered


mhvk:
From #8883: one can write a user defined unit to a fits table, and generally this works though one gets an appropriate warning:
```
import numpy as np
from astropy import units as u
from astropy.table import QTable

bandpass_sol_lum = u.def_unit('bandpass_sol_lum') 
lum_array = np.ones((5,5)) * bandpass_sol_lum
qt = QTable()
qt['Luminosity'] = lum_array 
qt.write('output_lum.fits',overwrite=True,format='fits')
# WARNING: The unit 'bandpass_sol_lum' could not be saved to FITS format [astropy.io.fits.convenience]

u.add_enabled_units(bandpass_sol_lum)
QTable.read('output_lum.fits', format='fits')

<QTable length=5>
 Luminosity [5] 
bandpass_sol_lum
    float64     
----------------
      1.0 .. 1.0
      1.0 .. 1.0
      1.0 .. 1.0
      1.0 .. 1.0
      1.0 .. 1.0
```
So far, so good. But when one then tries to write the fits file again, it fails:
```
qt.write('output_lum.fits',overwrite=True,format='fits')
WARNING: The unit 'bandpass_sol_lum' could not be saved to FITS format [astropy.io.fits.convenience]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-c1bb5ecc6cf2> in <module>()
----> 1 qt.write('output_lum.fits',overwrite=True,format='fits')

/data/mhvk/venv/numpy-dev/lib/python3.7/site-packages/numpy-1.17.0.dev0+0dcd300-py3.7-linux-x86_64.egg/astropy/table/connect.py in __call__(self, *args, **kwargs)
    112         instance = self._instance
    113         with serialize_method_as(instance, serialize_method):
--> 114             registry.write(instance, *args, **kwargs)

/data/mhvk/venv/numpy-dev/lib/python3.7/site-packages/numpy-1.17.0.dev0+0dcd300-py3.7-linux-x86_64.egg/astropy/io/registry.py in write(data, format, *args, **kwargs)
    564 
    565     writer = get_writer(format, data.__class__)
--> 566     writer(data, *args, **kwargs)
    567 
    568 

[...]

Issue #8887, opened 2019-06-20
Truly deprecate pedantic option in VO table


pllim:
Follow up of #8715, which was milestoned to 4.0. Remove `pending=True` to emit `AstropyDeprecationWarning`. Also open a follow-up issue to remove `pedantic` altogether after a deprecation period.

PR #8906, opened 2019-06-23
Only create unit from the columns unit if the to_string worked.

MSeifert04:
Otherwise the unit would be None which would result in an unexpected
(and unnecessary) Exception.

Related: #8897

I'm not sure how much this actually fixes, however since `col.unit` is always `None` in case the `try` block fails - but it seems to make sense only to do the `Unit` call when it succeeded.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -313,6 +313,9 @@ astropy.io.fits
 - Fixes bug where an invalid TRPOS<n> keyword was being generated for FITS
   time column when no location was available. [#8784]
 
+- Fixed a wrong exception when converting a Table with a unit that is not FITS
+  compliant and not convertible to a string using ``format='fits'``. [#8906]
+
 astropy.io.registry
 ^^^^^^^^^^^^^^^^^^^
 
--- a/astropy/io/fits/convenience.py
+++ b/astropy/io/fits/convenience.py
@@ -527,9 +527,10 @@ def table_to_hdu(table, character_as_bytes=False):
                 warnings.warn(
                     "The unit '{0}' could not be saved to FITS format".format(
                         unit.to_string()), AstropyUserWarning)
-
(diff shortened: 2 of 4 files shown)


MSeifert04:
Hm, the failures seem real but I don't know what the problem could be...

Probably the most sensible way would be replace the test with a test that only covers the "fixed" case here.

mhvk:
p.s. @MSeifert04 - if you haven't got time to check/write regression check for #8887, I can do it. Just don't want to spend the time if you are already doing it.

p.p.s. Nice find!

MSeifert04:
@mhvk I think (hope) you mixed up some issues. 😄 

I initially thought this would fix #8897, and it kinda did (locally 😅). The problem is really messy because a defined and enabled unit is somehow represented as IrreducibleUnit instead of as `UnrecognizedUnit`. An irreducible unit fails when converting with `to_string('fits')` but roundtrips at least partially - see the first commit here - because the table serialization infrastructure puts it in special comments. The unrecognized unit however works with the `to_string` but isn't properly "read" afterwards.

However the first commit which tried to make a regression test for #8897 failed in some CIs and I haven't had time to investigate it so I reversed the regression test and added one which simply did the isolated test for the non-FITS IrreducibleUnit.

Long story short: If you have time to add the regression test please go ahead. 👍

mhvk:
Yes, mistaken issue number...

I hadn't seen that you had tried the test in the original commit. The python-version dependent failure in travis is very surprising; I can see why you went for the "well, at least this solves something" option!!!

I actually think that is sensible enough that I will just merge this PR, which I feel is obviously correct, and then experiment a bit on the new master (saves me separately downloading your PR...)

MSeifert04:
Thanks @mhvk. :+1:

PR #8957, opened 2019-07-04
Table operations always return unmasked table

### LINKED ISSUE #8942: Fix Table operations for masking post #8789
Table operations like `join` and `vstack` are still setting the global `Table.masked` attribute if any columns are masked, but this is not desired.  This is basically another follow-on to #8789.

taldcroft:
Closes #8942 

This implementation is not the fastest since it just does a post-facto fix of the table.  But this is simple (i.e. reliable) and not a terrible performance hit.

### DIFF
--- a/astropy/table/operations.py
+++ b/astropy/table/operations.py
@@ -17,7 +17,7 @@ from collections.abc import Mapping, Sequence
 import numpy as np
 
 from astropy.utils import metadata
-from .table import Table, QTable, Row, Column
+from .table import Table, QTable, Row, Column, MaskedColumn
 from astropy.units import Quantity
 
 from . import _np_utils
@@ -216,8 +216,8 @@ def setdiff(table1, table2, keys=None):
     t12 = _join(t1, t2, join_type='left', keys=keys,
                 metadata_conflicts='silent')
 
-    # If t12 is masked then that means some rows were in table1 but not table2.
-    if t12.masked:
+    # If t12 index2 is masked then that means some rows were in table1 but not table2.
+    if hasattr(t12['__index2__'], 'mask'):
         # Define bool mask of table1 rows not in table2
         diff = t12['__index2__'].mask
         # Get the row indices of table1 for those rows
@@ -700,8 +700,7 @@ def _join(left, right, keys=None, join_type='inner',
     masked, n_out, left_out, left_mask, right_out, right_mask = \
         _np_utils.join_inner(idxs, idx_sort, len_left, int_join_type)
 
-    # If either of the inputs are masked then the output is masked
-    if left.masked or right.masked:
+    if left.has_masked_columns or right.has_masked_columns:
         masked = True
     masked = bool(masked)
 
@@ -743,7 +742,7 @@ def _join(left, right, keys=None, join_type='inner',
         # Finally add the joined column to the output table.
         out[out_name] = array[name][array_out]
 
-        # If the output table is masked then set the output column masking
+        # If the output column is masked then set the output column masking
         # accordingly.  Check for columns that don't support a mask attribute.
         if masked and np.any(array_mask):
             # array_mask is 1-d corresponding to length of output column.  We need
@@ -754,7 +753,7 @@ def _join(left, right, keys=None, join_type='inner',
             # Now broadcast to the correct final shape
             array_mask = np.broadcast_to(array_mask, out[out_name].shape)
 
-            if array.masked:
+            if hasattr(array[name], 'mask'): # array.masked:
                 array_mask = array_mask | array[name].mask[array_out]
             try:
                 out[out_name][array_mask] = out[out_name].info.mask_val
@@ -767,6 +766,12 @@ def _join(left, right, keys=None, join_type='inner',
     # If col_name_map supplied as a dict input, then update.
     if isinstance(_col_name_map, Mapping):
         _col_name_map.update(col_name_map)
+
+    if out.masked:
+        out._set_masked(False)
+        for col in out.itercols():
+            if isinstance(col, MaskedColumn) and not np.any(col.mask):
+                out[col.name] = Column(col, copy=False)
 
     return out
 
--- a/astropy/table/operations.py
+++ b/astropy/table/operations.py
@@ -744,7 +744,7 @@ def _join(left, right, keys=None, join_type='inner',
             # If col is a Column but not MaskedColumn then upgrade at this point
             # because masking is required.
             if isinstance(col, Column) and not isinstance(col, MaskedColumn):
-                col = col.view(MaskedColumn)
+                col = out.MaskedColumn(col, copy=False)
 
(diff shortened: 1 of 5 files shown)


astropy/table/operations.py
```diff
@@ -767,9 +769,19 @@ def _join(left, right, keys=None, join_type='inner',
     if isinstance(_col_name_map, Mapping):
         _col_name_map.update(col_name_map)
 
+    unmask_table_columns(out)
+
     return out
 
 
+def unmask_table_columns(tbl):
```
taldcroft: Oops, needs docstring (though this is not publicly exposed).

astropy/table/operations.py
```diff
@@ -767,9 +769,19 @@ def _join(left, right, keys=None, join_type='inner',
...
+    if tbl.masked:
+        tbl._set_masked(False)
+        for col in tbl.itercols():
+            if isinstance(col, MaskedColumn) and not np.any(col.mask):
+                tbl[col.name] = Column(col, copy=False)
```
taldcroft: It looks like `tbl[col.name] = tbl[col.name].view(Column)` works the same and is 3 times faster (91 vs 248 us).  Should that be fine @mhvk?

mhvk: Indeed, what we really want here is the unmasked data; in `MaskedArray`, this is exposed as both `data` and `_data`, and all that does is `self.view(self._baseclass)`. Now, we cannot use this directly since the `_baseclass` is `BaseColumn`, not `Column` (not completely sure why...). But that means your view does the right thing.

The one thing that this does not deal with is possible subclasses. Maybe easiest is to just let those fend for themselves (i.e., do `type(col) is MaskedColumn`, or `type(col._baseclass) is BaseColumn).

Though, a more general question: is there a risk that this change of nature of the column will surprise users? They have a table with some masked columns, do something, and suddenly the `.mask` property on their column disappears?

taldcroft: @mhvk - those are good points that need consideration.  Esp. the last one, for instance hstack'ing two same-length tables and having your masked column turn unmasked.  That does seem less than perfect.  OTOH right now the opposite can happen in a join or vstack where unmasked columns become masked for no good reason.  But there at least you don't lose functionality.

This probably helps you understand why I don't like the whole two-class solution to masking..  There are pros and cons, but the cons are not trivial.  If / when the `MaskedQuantity` class comes back to life as a duck array we will have a fair bit of pain in the table code supporting it.  Quantity won't be a generic mixin, but will need special handling.  We may need to somehow generalize the Column/MaskedColumn dichotomy to allow for multiple Class / MaskedClass variants.  Or something...

taldcroft: (In case it wasn't clear, I'm thinking about better solutions in the mean time.)

mhvk: I've been thinking on and off about the `Masked` versions and very much see your point that it is a pain to have different classes. I've still got some hope for a generic `info.mask` attribute - as long as we keep this super-simple (e.g., reductions would be lead to masked values if even one element was masked). But that's for another discussion!

taldcroft: OK, new implementation of the `join` update that doesn't use such a big hammer.  I'm much happier with this.  Now I have a path for the other ones, but each operation type will be different.

taldcroft:
@mhvk - so I found a problem with viewing as `MaskedColumn`.  Basically if the original object already has `info` in `__dict__` then that propagates and will be invalid.  Is this something that can or should be fixed?
```
In [2]: c = Column([1])
In [3]: mc = c.view(MaskedColumn)
In [4]: mc.info.mask_val
Out[4]: masked

In [5]: c = Column([1])
In [6]: c.info.description = 'hi'
In [7]: mc = c.view(MaskedColumn)
In [8]: mc.info.mask_val
~/git/astropy/astropy/utils/data_info.py in __getattr__(self, attr)
    302         try:
--> 303             value = self._attrs[attr]
    304         except KeyError:

KeyError: 'mask_val'

<ipython-input-8-4e60b3a96079> in <module>()
----> 1 mc.info.mask_val

~/git/astropy/astropy/utils/data_info.py in __getattr__(self, attr)
    303             value = self._attrs[attr]
    304         except KeyError:
--> 305             super().__getattribute__(attr)  # Generate AttributeError
    306 
    307         # Weak ref for parent table

AttributeError: 'ColumnInfo' object has no attribute 'mask_val'
```

astropy/table/operations.py
```diff
@@ -946,10 +946,10 @@ def _hstack(arrays, join_type='outer', uniq_col_name='{col_name}_{table_name}',
     # If there are any output rows where one or more input arrays are missing
     # then the output must be masked.  If any input arrays are masked then
     # output is masked.
-    masked = any(getattr(arr, 'masked', False) for arr in arrays) or len(set(arr_lens)) > 1
+    # masked = any(getattr(arr, 'masked', False) for arr in arrays) or len(set(arr_lens)) > 1
```
taldcroft: Rats, that's a leftover.

astropy/table/table.py
```diff
@@ -905,10 +908,7 @@ def _convert_data_to_col(self, data, copy=True, default_name=None, dtype=None, n
             # gets upgraded to MaskedColumn, but the converse (pre-4.0) behavior
             # of downgrading from MaskedColumn to Column (for non-masked table)
             # does not happen.
-            if issubclass(self.ColumnClass, data.__class__):
-                col_cls = self.ColumnClass
-            else:
-                col_cls = data.__class__
+            col_cls = self._get_col_cls_for_table(data)
```
taldcroft: This refactor (here and below) and change in behavior was necessary to get the hstack conversion working.  But in general it is a good thing for more consistent behavior that works with the advertised table API of allowing custom default `Column` and `MaskedColumn` classes.

astropy/table/table.py
```diff
@@ -964,7 +989,9 @@ def _convert_col_for_table(self, col):
         override this method.
         """
         if isinstance(col, Column) and not isinstance(col, self.ColumnClass):
-            col = self.ColumnClass(col)  # copy attributes and reference data
+            col_cls = self._get_col_cls_for_table(col)
+            col = col_cls(col, copy=False)
```
taldcroft: I think this needs a check that `col_cls` is different from `col.__class__`.

taldcroft:
@mhvk - I think this is basically ready if you have time for another look.  The travis fail is flake8 (I started using pycharm and it's pep8 warnings are too subtle).  This PR grew to include some relatively deep changes related to masking and column initialization, and fixes a previously-unseen regression.

mhvk:
Darn, that `info` copying in `__array_finalize__` is problematic - at least, I think the problem here is that the `info` is of the wrong class for the new column. It seems safest for now to just initialize a class and worry about this later.

(Won't be able to review immediately, but will try to do it soon.)

taldcroft:
@mhvk - thanks again for review, your comments pushed this PR to a much better solution (again!).

Issue #8962, opened 2019-07-06
Add Table `zip_columns` method for faster row iteration


taldcroft:
One way to substantially improve performance when iterating over rows is to avoid explicit use of `Row` for access.  Instead of:
```
for row in tbl:
   function(row['a'], row['b'], row['z'])
```
Do:
```
for a, b, z in zip(tbl['a'], tbl['b'], tbl['z']):
   function(a, b, z)
```
This is much faster because it ends up using C-level iteration instead of a lot of Python code to make a `Row` and then access it.

In many cases of iterating over table rows, one only cares about a small subset of the column values.  So we can make the above nicer with:
```
for a, b, z in tbl.zip_columns('a', 'b', 'z'):
   function(a, b, z)
```

~~One question is if the name is confusing.  Technically it should be called `zip_columns`, but I'm not sure that would make more sense to the average user.~~ [Changed to `zip_columns`].

@himanshupathak21061998 - this might be a good straightforward table performance issue to try.

himanshupathak21061998:
Sure, I will try to fix this.

saimn:
This seems a great idea, I frequently use this to iterate on column values. 
About the name, I think I prefer `zip_columns` which is closer to what it actually does (calling zip on given columns). `iterrows` is more confusing because there are several possible way to iterate on rows: it could return `Row` object, or just a tuple, or a dict.

taldcroft:
@saimn - I hear you about the name, I'm just wondering about the median user and how familiar is `zip`.  My thoughts often go to tab completion, seeing `iterrows` in the list next to `itercols`, and then `t.iterrows?`.  I'm not super concerned about potential ambiguity in the output since it is so easy to discover the answer.  If you are reading code then `for a, c, e in tbl.iterrows('a', 'c', 'e'):` gives you are pretty strong clue about what's happening.

That said, I can go either way, esp. if others weigh in.

One thing that occurred to me is that `tbl.iterrows()` could be short-hand for `tbl.iterrows(*tbl.colnames)`, i.e. zip over all the columns.

mhvk:
On the name: I'd agree with @saimn that I would actually expect to get `Row` objects out, just like `itercols` gives `Column` objects. Maybe `itervalues`?

Alternatively, would it be possible to improve the table iteration generally, by improving how `Row` works? E.g., if I now do
```
t = Table([np.arange(3.), np.ones(3)], names=['a', 'b'])
a,b  = t[1]
```
[...]

PR #8969, opened 2019-07-09
Add `iterrows()` for faster row iteration

himanshupathak21061998:
fix: #8962
Function with example added @taldcroft please review it.

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -2174,6 +2174,43 @@ class Table:
         if hasattr(self, '_groups'):
             del self._groups
 
+    def iterrows(self, col_names=None):
+        """
+        Return zipped columns for row-wise iteration
+
+        Parameters
+        ----------
+        col_names : list
+        List of column names
+
+         Examples
+        --------
+        Create a table with three columns 'a', 'b' and 'c'::
+
+            >>> t = Table([[1, 2, 3], [0.1, 0.2, 0.3], ['x', 'y', 'z']],
+            ...           names=('a', 'b', 'c'))
+
+        To iterate row-wise using column names
+            >>> for a,b in t.iterrows(['a', 'b'])
+            ...     print(a,b)
+            1 0.1
+            2 0.2
+            3 0.3
+
+        """
+        if col_names is None:
+            raise ValueError('No input Provided')
+
+        for x in col_names:
+            if x not in self.colnames:
+                raise ValueError('Enter Valid Column Names')
+
+        col = []
+        for x in col_names:
+            col.append(self[x])
+        zipped = zip(*col)
+        return zipped
+
     def remove_column(self, name):
         """
         Remove a column from the table.
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -2205,11 +2205,9 @@ class Table:
             if name not in self.colnames:
                 raise ValueError('{} is not a valid column name'.format(name))
 
-        col = []
-        for name in names:
-            col.append(self[name])
-        zipped = zip(*col)
-        return zipped
+        cols = (self[name] for name in names)
+        out = zip(*cols)
+        return out
 
     def remove_column(self, name):
         """
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -2174,16 +2174,16 @@ class Table:
         if hasattr(self, '_groups'):
             del self._groups
 
-    def iterrows(self, col_names=None):
+    def zip_columns(self, *names):
         """
         Return zipped columns for row-wise iteration
 
         Parameters
         ----------
-        col_names : list
-        List of column names
+        names : list
+            List of column names
 
-         Examples
+        Examples
         --------
         Create a table with three columns 'a', 'b' and 'c'::
 
@@ -2191,23 +2191,23 @@ class Table:
             ...           names=('a', 'b', 'c'))
 
         To iterate row-wise using column names
-            >>> for a,b in t.iterrows(['a', 'b'])
+            >>> for a,b in t.zip_colunms('a', 'b'):
             ...     print(a,b)
             1 0.1
             2 0.2
             3 0.3
 
         """
-        if col_names is None:
-            raise ValueError('No input Provided')
+        if len(names) == 0:
+            raise ValueError('at least one column name must be provided')
 
-        for x in col_names:
-            if x not in self.colnames:
-                raise ValueError('Enter Valid Column Names')
+        for name in names:
+            if name not in self.colnames:
+                raise ValueError('{} is not a valid column name'.format(name))
 
         col = []
-        for x in col_names:
-            col.append(self[x])
+        for name in names:
+            col.append(self[name])
         zipped = zip(*col)
         return zipped
 
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -2174,36 +2174,49 @@ class Table:
         if hasattr(self, '_groups'):
             del self._groups
 
-    def zip_columns(self, *names):
+    def iterrows(self, *names):
         """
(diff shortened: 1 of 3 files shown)


astropy/table/table.py
```diff
@@ -2132,6 +2132,43 @@ def remove_rows(self, row_specifier):
         if hasattr(self, '_groups'):
             del self._groups
 
+    def iterrows(self, col_names=None):
```
taldcroft: Call this arg `names` and make it a required argument.  Also, it should be `*names` here to match the API that I mentioned.

astropy/table/table.py
```diff
@@ -2132,6 +2132,43 @@ def remove_rows(self, row_specifier):
...
+        """
+        Return zipped columns for row-wise iteration
+
+        Parameters
+        ----------
+        col_names : list
+        List of column names
```
taldcroft: This line needs to be indented by 4 spaces.

astropy/table/table.py
```diff
@@ -2132,6 +2132,43 @@ def remove_rows(self, row_specifier):
...
+
+         Examples
+        --------
+        Create a table with three columns 'a', 'b' and 'c'::
+
+            >>> t = Table([[1, 2, 3], [0.1, 0.2, 0.3], ['x', 'y', 'z']],
```
taldcroft: Use float values like 1.0, 2.5, 5.0 that have exact float representation.  That prevents any possibility of docstring fails on different architectures.

astropy/table/table.py
```diff
@@ -2132,6 +2132,43 @@ def remove_rows(self, row_specifier):
...
+            ...     print(a,b)
+            1 0.1
+            2 0.2
+            3 0.3
+
+        """
+        if col_names is None:
+            raise ValueError('No input Provided')
+
+        for x in col_names:
```
taldcroft: Use `for name in names`.  Generally try to use singular/plural when iterating over a list.

astropy/table/table.py
```diff
@@ -2132,6 +2132,43 @@ def remove_rows(self, row_specifier):
...
+        To iterate row-wise using column names
+            >>> for a,b in t.iterrows(['a', 'b'])
+            ...     print(a,b)
+            1 0.1
+            2 0.2
+            3 0.3
+
+        """
+        if col_names is None:
+            raise ValueError('No input Provided')
```
taldcroft:
```
if len(names) == 0:
    raise ValueError('at least one column name must be provided')
```
Exception messages should start with lower case.

astropy/table/table.py
```diff
@@ -2132,6 +2132,43 @@ def remove_rows(self, row_specifier):
...
+
+        for x in col_names:
+            if x not in self.colnames:
+                raise ValueError('Enter Valid Column Names')
```
taldcroft: Try to provide useful information in the exception message, e.g. `f"'{name}' is not a valid column name"`.

astropy/table/table.py
```diff
@@ -2132,6 +2132,43 @@ def remove_rows(self, row_specifier):
...
+
+        col = []
+        for x in col_names:
+            col.append(self[x])
+        zipped = zip(*col)
```
taldcroft: Probably just
```
cols = (self[name] for name in names)
out = zip(*cols)
return out
```
Note the use of an iterator instead of a list, and using plural `cols` for something that is a list.  You should do that in general.

astropy/table/tests/test_table.py
```diff
@@ -2392,3 +2392,15 @@ class MyTable(table.Table):
         assert type(t['c']) is MyMaskedColumn
         assert type(t['d']) is MySubColumn
         assert type(t['e']) is MySubMaskedColumn  # sub-class not downgraded
+
+
+def test_iterrows():
+    t = table.Table([(2, 3, 2, 1), (8, 7, 6, 5)], names=('a', 'b'))
+    assert t.iterrows(['a']) == zip(t['a'])
```
taldcroft: I don't think this test is meaningful as it is just comparing two zip generators.  You need to write out a loop like the doc string example and prove that things are working as expected.

astropy/table/tests/test_table.py
```diff
@@ -2392,3 +2392,15 @@ class MyTable(table.Table):
         assert type(t['c']) is MyMaskedColumn
         assert type(t['d']) is MySubColumn
         assert type(t['e']) is MySubMaskedColumn  # sub-class not downgraded
+
+
+def test_iterrows():
+    t = table.Table([(2, 3, 2, 1), (8, 7, 6, 5)], names=('a', 'b'))
```
taldcroft: Use values that are easily verified by a human to be right, e.g. `[(1, 2, 3), (4, 5, 6)]`.

astropy/table/tests/test_table.py
```diff
@@ -2392,3 +2392,15 @@ class MyTable(table.Table):
...
+    assert t.iterrows(['a']) == zip(t['a'])
+    assert t.iterrows(['a', 'b']) == zip(t['a'], t['b'])
+    with pytest.raises(ValueError) as err:
+        t.iterrows()
+        err.value.args[0] == 'No input Provided'
```
taldcroft: This needs to be de-dented by 4 spaces (even with the `with pytest...`) because the way it is written the `err.value` line is never reached because of the exception in `t.iterrows()`.  Same below.

taldcroft:
@himanshupathak21061998 - one hint is to add `[skip ci]` at the end of commits before you are ready to run the full set of CI tests (which uses plenty of energy).  So if you have only partially addressed comments but want to put the commits up to github (a good idea!) do that `[skip ci]` trick.

himanshupathak21061998:
Next time, I will take care of it thanks @taldcroft

astropy/table/table.py
himanshupathak21061998: SorryI did not checked this comment before updating

himanshupathak21061998:
@taldcroft  Now Everything is done as suggested thanks

astrofrog:
Just out of curiosity, why the change in name to ``zip_columns``? It doesn't sound very intuitive to me as a row iteration method.

saimn:
@astrofrog - see the discussion in #8962

taldcroft:
Oh no, now we have a tie, 2 votes for `iterrows` and 2 for `zip_columns`.

himanshupathak21061998:
I think we have to wait for more votes :grin:

astrofrog:
Given the discussion in #8962 I think ``iterrows`` doesn't make sense since that should return Row objects, so I retract my objection to ``zip_columns``.

taldcroft:
What about
```
def iterrows(*names, return_type='tuple'):
```
where `return_type` could be `Row`, `tuple` or `dict` (could be supplied as strings or types).

taldcroft:
@astrofrog et al. - coming back to this after a break, I feel more strongly that `iterrows` is the right now.  It matches `itercols` and there is nothing written in stone that a table "row" needs to be `astropy.table.Row`.  There are many possible representations of a "row", and here `iterrows` chooses the most performant.  

In addition I think that `zip_columns` is making unnecessary reference to the implementation and relies on having a good understanding of Python `zip` to make sense.

The docstring now makes it very clear that this returns a tuple for each row.

@himanshupathak21061998 - in order to get this very nice and high-impact addition into 4.0, I went ahead and made the change and did a few other bits of cleanup.  Thanks for doing most of the work!

taldcroft:
@mhvk - since I made the final commit in this PR I'm no longer an independent reviewer.  Can you take a quick look, it should be pretty simple / fast.

himanshupathak21061998:
Thanks, @taldcroft I was thinking of doing this :smiley:

PR #8972, opened 2019-07-09
Change Table data-to-col convert strategy from white-list to "try"

### LINKED ISSUE #8971: Small regression in Table to create a column
A small regression in the creation of columns with a scalar, probably related to the recent changes (ping @taldcroft). It works with a Python scalar but breaks with a Numpy scalar:

```
In [1]: %astropy                                                                               
Numpy 1.17.0rc1
Astropy 4.0.dev25172

In [4]: t = Table([[1, 2, 3], [1.1, 2.2, 3.3]])                                                

In [5]: t                                                                                      
Out[5]: 
<Table length=3>
 col0   col1 
int64 float64
----- -------
    1     1.1
    2     2.2
    3     3.3

In [6]: t['foo'] = 1                                                                           

In [7]: t['bar'] = np.int64(1)                                                                 
<ipython-input-7-a0126752c187> in <module>
----> 1 t['bar'] = np.int64(1)

~/dev/astropy/astropy/table/table.py in __setitem__(self, item, value)
   1495         # If that column doesn't already exist then create it now.
   1496         if isinstance(item, str) and item not in self.colnames:
-> 1497             self.add_column(value, name=item, copy=True)
   1498 
   1499         else:

~/dev/astropy/astropy/table/table.py in add_column(self, col, index, name, rename_duplicate, copy, default_name)
   1760         # of scalars to the correct shape for adding to table.
   1761         col = self._convert_data_to_col(col, name=name, copy=copy,
-> 1762                                         default_name=default_name)
   1763 
   1764         # Make col data shape correct for scalars.  The second test is to allow

~/dev/astropy/astropy/table/table.py in _convert_data_to_col(self, data, copy, default_name, dtype, name)
    930 
    931         else:
--> 932             raise ValueError('Elements in list initialization must be '
    933                              'either Column or list-like')
    934 

ValueError: Elements in list initialization must be either Column or list-like
```

taldcroft:
The regression in #8971 is addressed by this change, but the change is actually a bit broader and needs some consideration.  Previously the `_convert_data_to_col` method was basically white-listing all possible valid data types (and of course missing a numpy scalar). 

 Instead this approach checks a few cases that need special handling and at the end just tries to initialize a Column (or whatever subclass) with the data.  I think this is a better approach but I wonder if there are things that can now slip through and make trouble.

Will add a test for the #8971 regression but wanted to see if @mhvk has any thoughts.  I can't quite think of any obvious way to generate the exception at this point.

Fixes #8971

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -925,15 +925,17 @@ class Table:
                        if issubclass(self.ColumnClass, self.MaskedColumn)
                        else self.MaskedColumn)
 
-        elif isinstance(data, np.ndarray) or isiterable(data):
+        else:
+            # `data` is none of the above, so just go for it and try init'ing Column
             col_cls = self.ColumnClass
 
-        else:
-            raise ValueError('Elements in list initialization must be '
-                             'either Column or list-like')
+        try:
+            col = col_cls(name=name, data=data, dtype=dtype,
+                          copy=copy, copy_indices=self._init_indices)
+        except Exception:
+            # Broad exception class since we don't know what might go wrong
+            raise ValueError('unable to convert data to Column for Table')
 
-        col = col_cls(name=name, data=data, dtype=dtype,
-                      copy=copy, copy_indices=self._init_indices)
         col = self._convert_col_for_table(col)
 
         return col
--- a/astropy/table/tests/test_table.py
+++ b/astropy/table/tests/test_table.py
@@ -2392,3 +2392,14 @@ def test_custom_masked_column_in_nonmasked_table():
         assert type(t['c']) is MyMaskedColumn
         assert type(t['d']) is MySubColumn
         assert type(t['e']) is MySubMaskedColumn  # sub-class not downgraded
+
+
+def test_data_to_col_convert_strategy():
+    """Test the update to how data_to_col works (#8972), using the regression
+    example from #8971.
+    """
+    t = table.Table([[0, 1]])
+    t['a'] = 1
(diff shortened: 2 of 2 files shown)


mhvk:
I like the approach, but you are right, very hard to test, since everything strange just gets turned into an object array.

taldcroft:
But it turns out the behavior is not new:
```
In [1]: astro
astropy=3.2.1

In [2]: t = simple_table()

In [3]: t['d'] = {1:2}

In [4]: t
Out[4]: 
<Table length=3>
  a      b     c     d   
int64 float64 str1 object
----- ------- ---- ------
    1     1.0    c {1: 2}
    2     2.0    d {1: 2}
    3     3.0    e {1: 2}
```

mhvk:
Good, so by adopting this we actually regress least!

hamogu:
I've independently discovered the regression in #8971 in https://travis-ci.org/Chandra-MARX/marxs/jobs/558412814 and I'm glad to see that that had been reported and that there is already a PR to fix it. 

I think this PR is the right way to do it and in fact, I've used the behavior of "making an object array this way" before on purpose so I fully support this choice and I suggest to merge this as is. (Since I'm not a table maintainer, I won't leave this as a "review" in github.)

taldcroft:
@mhvk - I added a test here, ready for review.  I think code coverage is complaining about the exception not being hit.  I guess we can there with `Table([[1,2]], dtype=['hi there'])`.  Is it worth burning another CI test cycle for that?

Issue #8978, opened 2019-07-10
Pitfalls of "unknown units' support in VOUnits


dhomeier:
The full support of the [**VOUnit** standard](http://ivoa.net/documents/VOUnits/) with #2901 has led to some confusion about its acceptance of unrecognised or unknown units (§ `2.11` of the standard) as in #6302 and #8959.

Essentially `u.Unit("name", format='vounit')` is just following quite literally the 1.0 standard by allowing, even implicit, construction of arbitrary 'unknown units'. These are not instances of outr `units`'s `UnrecognizedUnit` class, since per the standard they still accept SI prefixes. Handling is further complicated by the requirement that any valid SI prefix shall always be recognised in a plain string, while unit names in literal single quotes must not be further parsed. As per the example in the standard

- `"furlong"` is parsed as "femto-urlong", with `.bases[0]._represents` `Unit("1e-15 urlong")`
- `"'furlong'"` is parsed as "'furlong'", being an `IrreducibleUnit`
- `"f'furlong'"` would be parsed as "femto-'furlong'".

Again, this is rather confusing to most users not intimately familiar with the VO standards, and since the quoted custom units are parsed without any notification, others with only a one-time warning, prone to coding errors. In Astropy itself there is some ambiguity at least in  `CHANGES.rst`, which only mentions the inline definition of units in quotes, and in the test of `u.Unit('ANGSTROM', format='vounit')` which seems to suggest that it should raise a `ValueError`, but actually passes without an exception.

My recommendation would be to raise all warnings by one level, i.e. issue a warning on uses of the quoted syntax like `"m'ANGSTROM'"`, `"'kibibytes'"` as well, and allow other custom units only per explicit definition. The most straightforward way would be to accept names like `"furlong/week"` only with `parse_strict='warn'` or `'silent'` set as for the other parsers, but since these cases are handled by `generic.py`, that would break the current compliance with the prefix treatment (the units would become `UnrecognizedUnit` instances).
One solution might be adding a fourth `parse_strict` setting to return a regular unit instead, but maybe a better way would be adding an additional syntax in `votable.py` to recognise double-quoted strings, to make this work:
```
myunit = u.Unit("'furlong'", format='vounit')
WARNING: UnitsWarning: Unit 'furlong' is not recognized in the VOUnit standard.  [astropy.units.format.vounit]
>>> myunit.names
['furlong', "'furlong'"]
>>> isinstance(myunit, u.IrreducibleUnit)
True
```
```
[...]

Issue #6302, opened 2017-06-29
Unit ignores 'parse_strict' keyword when format is 'vounit'


astrogreen:
The units package doesn't seem to do any validation when the format is 'vounit' other than occasionally issuing warnings.
```python
>>> import astropy.units as units

>>> units.Unit("Angstroem")
ValueError: 'Angstroem' did not parse as unit: At col 0, Angstroem is not a valid unit. Did you mean Angstrom or angstrom?

>>> units.Unit("Angstroem", format='vounit')
Unit("Angstroem")

>>> units.Unit("Angstroem", format='vounit', parse_strict='raise')
Unit("Angstroem")

```

I would expect the vounit formatting to require that the unit string match the VO standard? I also note that the signature for `unit.Unit` does not actually have a `parse_strict` keyword (I thought this would be an error.)

Furthermore, it seems that the unit created using the 'vounit' format is not "unrecognized":
```python
>>> u = units.Unit("Angstroem", format='vounit')

>>> isinstance(u, units.UnrecognizedUnit)
False
>>> type(v)
astropy.units.core.IrreducibleUnit

```

Other formats, such as 'fits' and 'cds' seem to behave as expected.

pllim:
The actual logic is in the `_parse_unit()` method within `vounit.py`. That part of the code has not been touched for years and was mainly written by @mdboom who is no longer working on this.

As for `parse_strict`, it is defined in [__call__() of unit's metaclass](https://github.com/astropy/astropy/blob/master/astropy/units/core.py#L1749).

@mhvk -- What do you think?

mhvk:
Definitely a bug! Thanks for reporting, @astrogreen. If you are up to investigating this further, a PR would obviously be very welcome! (Though this is not trivial code.)

pllim:
There are more recent discussions over at #13017 , so I am closing this as duplicate.

PR #8979, opened 2019-07-11
Introduce double-quote syntax to inline define custom units in VOUnit

dhomeier:
Adressing #8978, this is changing the (undocumented) inline definition of arbitrary custom (unknown) units possible as
```
u.Unit('swallow', format='votable')
```
to be explicitly requested by enclosing the whole unit string in double quotes as
```
u.Unit('"swallow"', format='votable')
```
This is still complementary to the existing support of single-quoted unit names, i.e. _will_ parse SI prefixes according to the VOUnits standard - unlike  `'furlong'`, `"furlong"` will be parsed as `femto-urlong`.

There are some remaining complications because custom units are placed in the dictionary of known `units` and cannot be replaced by a re-definition, i.e. depending on the order in which
```
u.Unit('"furlong/week"', format='vounit')
```
```
u.Unit("'furlong' / 'week'", format='vounit')
```
are called, one has either `'furlong'` or `urlong` and `femto-urlong` already defined, and the second call will fail. The standard describes in 2.2

> A **symbol** within a unit-component _should_ be parsed as follows:
> 1. If it corresponds to a known **base symbol**, then it _must_ be recognised as such (for example the Pa must be parsed as the known Pascal, and never as the peta-year).
> 2. If the symbol starts with a multiplicative prefix, then this is recognised independently of whether  the resulting base symbol is a known or unknown unit (...)

In that sense, if the new unit becomes a "known base symbol" once defined, case 1. should take precedence over identifying any prefixes, as is currently the case.
But one might discuss whether the prefix search in `_def_custom_unit` should rather be moved ahead of the loop over `unit in cls._custom_units`...

It should also be noted that both `u.Unit('"furlong"', format='vounit')` and `u.Unit("'furlong'", format='vounit')` create different units from `u.Unit('furlong', format='vounit', parse_strict='warn')`, which returns a `u.UnrecognizedUnit` instance.

### DIFF
--- a/astropy/units/core.py
+++ b/astropy/units/core.py
@@ -536,7 +536,9 @@ class UnitBase:
         useful as a dictionary key.
         """
         unit = self.decompose()
-        r = zip([x.name for x in unit.bases], unit.powers)
+        # custom units are stored under their quoted (alias) name
+        r = zip([x.name if f"'{x.name:}'" not in x.aliases else
+                 f"'{x.name:}'" for x in unit.bases], unit.powers)
         # bases and powers are already sorted in a unique way
         # r.sort()
         r = tuple(r)
@@ -876,6 +878,8 @@ class UnitBase:
             if physical_type != 'unknown':
                 unit_str = "'{}' ({})".format(
                     unit_str, physical_type)
+            elif hasattr(unit, 'long_names') and len(unit.long_names) > 0:
+                unit_str = f"'{unit_str}' (custom unit \"{unit.long_names[0]}\")"
             else:
                 unit_str = f"'{unit_str}'"
             return unit_str
--- a/astropy/units/format/vounit.py
+++ b/astropy/units/format/vounit.py
@@ -121,9 +121,9 @@ class VOUnit(generic.Generic):
                     raise ValueError()
                 raise ValueError("Unit {0!r} not supported by the VOUnit "
                                  "standard. {1}".format(
-                                 unit, utils.did_you_mean_units(
-                                     unit, cls._units, cls._deprecated_units,
-                                     cls._to_decomposed_alternative)))
+                                     unit, utils.did_you_mean_units(
+                                         unit, cls._units, cls._deprecated_units,
+                                         cls._to_decomposed_alternative)))
 
             warnings.warn(
                 "Unit {!r} not supported by the VOUnit "
@@ -198,7 +198,7 @@ class VOUnit(generic.Generic):
                     return core.PrefixUnit(
                         [prefix + x for x in base_unit.names],
                         core.CompositeUnit(factor, [base_unit], [1],
-                                        _error_check=False),
+                                           _error_check=False),
                         format={'vounit': prefix + base_unit.names[-1]},
                         namespace=cls._custom_units)
 
--- a/astropy/units/format/vounit.py
+++ b/astropy/units/format/vounit.py
@@ -24,6 +24,7 @@ class VOUnit(generic.Generic):
         r"^[YZEPTGMkhdcmunpfazy]?'((?!\d)\w)+'$")
     _custom_unit_regex = re.compile(r"^((?!\d)\w)+$")
     _custom_units = {}
+    _custom_inline = False
 
     @staticmethod
     def _generate_unit_names():
@@ -91,7 +92,11 @@ class VOUnit(generic.Generic):
             raise core.UnitsError(
                 "'{}' contains multiple slashes, which is "
                 "disallowed by the VOUnit standard".format(s))
-        result = cls._do_parse(s, debug=debug)
+        if len(s) > 2 and s[0] == s[-1] == '"':
+            cls._custom_inline = True
+            result = cls._do_parse(s.strip('"'), debug=debug)
+        else:
+            result = cls._do_parse(s, debug=debug)
         if hasattr(result, 'function_unit'):
             raise ValueError("Function units are not yet supported in "
                              "VOUnit.")
@@ -101,10 +106,23 @@ class VOUnit(generic.Generic):
     def _parse_unit(cls, unit, detailed_exception=True):
         if unit not in cls._units:
             if cls._explicit_custom_unit_regex.match(unit):
+                warnings.warn(
+                    "Unit {0!r} is not recognized in the VOUnit "
+                    "standard. {1}".format(
+                        unit.split("'")[1], utils.did_you_mean_units(
+                            unit, cls._units, cls._deprecated_units,
+                            cls._to_decomposed_alternative)),
+                    core.UnitsWarning)
                 return cls._def_custom_unit(unit)
 
-            if not cls._custom_unit_regex.match(unit):
-                raise ValueError()
+            if not (cls._custom_inline and cls._custom_unit_regex.match(unit)):
+                if not detailed_exception:
+                    raise ValueError()
+                raise ValueError("Unit {0!r} not supported by the VOUnit "
+                                 "standard. {1}".format(
+                                 unit, utils.did_you_mean_units(
+                                     unit, cls._units, cls._deprecated_units,
+                                     cls._to_decomposed_alternative)))
 
             warnings.warn(
                 "Unit {!r} not supported by the VOUnit "
--- a/astropy/units/format/vounit.py
+++ b/astropy/units/format/vounit.py
@@ -96,6 +96,7 @@ class VOUnit(generic.Generic):
             cls._custom_inline = True
             result = cls._do_parse(s.strip('"'), debug=debug)
         else:
+            cls._custom_inline = False
             result = cls._do_parse(s, debug=debug)
         if hasattr(result, 'function_unit'):
             raise ValueError("Function units are not yet supported in "
--- a/astropy/units/physical.py
+++ b/astropy/units/physical.py
@@ -60,7 +60,7 @@ def get_physical_type(unit):
     -------
     physical : str
         The name of the physical quantity, or unknown if not
-        known.
+        known (quoted custom units are always considered unknown).
     """
     r = unit._get_physical_type_id()
     return _physical_unit_mapping.get(r, 'unknown')
--- a/astropy/units/tests/test_format.py
+++ b/astropy/units/tests/test_format.py
@@ -419,13 +419,6 @@ def test_scaled_dimensionless():
 
 
 def test_deprecated_did_you_mean_units():
-    # First clear custom unit namespace to ensure consistent test outcome.
-    u.format.VOUnit._custom_units.clear()
-    with pytest.raises(ValueError,
-                       match='not supported by the VOUnit standard') as e:
-        u.Unit('ANGSTROM', format='vounit')
-    assert str(e.value).count('nm') == 1
-
     with pytest.raises(ValueError,
                        match=r'Crab .deprecated. or mCrab .deprecated..$'):
         u.Unit('crab', format='ogip')
@@ -437,6 +430,11 @@ def test_deprecated_did_you_mean_units():
                        match='not supported by the CDS SAC standard') as e:
         u.Unit('ANGSTROM', format='cds')
     assert str(e.value).count('Angstrom') == 5
+
+    with pytest.raises(ValueError,
+                       match='not supported by the VOUnit standard') as e:
+        u.Unit('ANGSTROM', format='vounit')
+    assert str(e.value).count('nm') == 1
 
     with catch_warnings() as w:
         u.Unit('angstrom', format='vounit')
--- a/astropy/units/tests/test_format.py
+++ b/astropy/units/tests/test_format.py
@@ -419,24 +419,6 @@ def test_scaled_dimensionless():
 
 
 def test_deprecated_did_you_mean_units():
-<<<<<<< HEAD
-    with pytest.raises(ValueError) as exc_info:
-        u.Unit('ANGSTROM', format='fits')
-    assert 'Did you mean Angstrom or angstrom?' in str(exc_info.value)
-
-    with pytest.raises(ValueError) as exc_info:
-        u.Unit('crab', format='ogip')
-    assert 'Crab (deprecated)' in str(exc_info.value)
-    assert 'mCrab (deprecated)' in str(exc_info.value)
-
-    with catch_warnings() as w:
-        u.Unit('ANGSTROM', format='vounit')
-
-    assert len(w) == 1
-    assert 'angstrom (deprecated)' in str(w[0].message)
-    assert '0.1nm' in str(w[0].message)
-    assert str(w[0].message).count('0.1nm') == 1
-=======
     with pytest.raises(ValueError,
                        match=r'Crab .deprecated. or mCrab .deprecated..$'):
         u.Unit('crab', format='ogip')
@@ -453,7 +435,6 @@ def test_deprecated_did_you_mean_units():
                        match='not supported by the VOUnit standard') as e:
         u.Unit('ANGSTROM', format='vounit')
     assert str(e.value).count('nm') == 1
->>>>>>> Introduce double-quote syntax to mark custom units in VOUnit
 
     with catch_warnings() as w:
         u.Unit('angstrom', format='vounit')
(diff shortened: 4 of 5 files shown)


mhvk:
I like this a lot! I actually didn't know about this behaviour and was put off enough by `furlong` being `femto-urlong` to put a `Yikes!` in a comment in #9005 (which I fear also means this will need a rebase...).

The one thing I find a bit odd is the special-casing of double quotes - shouldn't single quotes be acceptable as well?

dhomeier:
> The one thing I find a bit odd is the special-casing of double quotes - shouldn't single quotes be acceptable as well?

The special treatment of single quotes (apostrophe, ASCII 27_{16}) is actually mandated by the standard (2.2#3 and 2.11)
> In the VOUnits syntax, base symbols may be put between single quotes '...' (a significant divergence from the other syntaxes). Such symbols must be parsed as unrecognised unit symbols which are not further examined.

which dictates the different treatment of `'furlong'` and `"furlong"` = `furlong` = `femto-furlong` as in the example in #8978.

I agree that the different treatment and meaning of `"''"` and `'""'` can be confusing, and I partly chose this implementation to avoid the extra work of introducing a new keyword that had to be passed down all the way to the parsers. But I still find it relatively convenient, and at least the behaviour is sort of consistent with shells' treatment of single vs. double quotes (re. variable resolution etc.).

I am more concerned about the fact how order matters in defining new unit names (first call to a new name gets the slot in the `units` dictionary). I am afraid the standard does not specify how to proceed in such cases. But I guess one could say it the user makes up new units it's up to them to ensure everything works in a consistent way.

dhomeier:
Hopefully the rebase should not be too painful, the only place likely to conflict will be the test for the (now corrected) behaviour of `u.Unit('ANGSTROM', format='vounit')`. And this PR needs to be updated to use f-string syntax anyway.

mhvk:
OK, thanks. I think I am starting to understand. Your PR ensures that:
1. `Unit('furlong')` will now error.
2. `Unit('"furlong"') will get you the old behaviour of (1): get femto-urlong, i.e., a `PrefixUnit` of `an `IrreducibleUnit`, including the old warning.
3. `Unit("'furlong'")` will continue to explicitly define a new `IrreducibleUnit`, but you added a warning.

This works for me except that I think we should not warn: after all, the user is explicitly asking for something.

With your change, getting `UnrecognizedUnit` for case (1) with `parse_strict='warn'` or `'silent'` also makes sense.

p.s. Definitely will need something in the documentation!!

dhomeier:
Yes. I've been thinking about whether cases 2+3 warrant a warning, but left it in for now, as it is possible that a user passes a string with the extra quotes by chance, without knowing about their specific meaning. Probably rather unlikely, though. And it might trick users into adding the `parse_strict='silent'` option, with undesired side-effects.

mhvk:
Let's remove the warnings... (certainly for (3), where there wasn't one before)

For the regular case, perhaps the new `ValueError` could point to the `"` behaviour?

Actually, one thing: with your changes, is it still possible to get the femto-urlong with the same string on old and new astropy? I think `Unit("f'urlong'")` will continue to work, right? I ask since it is helpful for code that needs to support multiple versions of astropy to have a way to do so without version checks (but it is not so bad if they have to make their code more explicit). Probably another reason to be sure to put a link to further docs in the `ValueError`.

Another worry: suppose people are just reading `votable` or so; are we now going to break interpretation of those units?

mhvk:
Hmm, perhaps the last worry means that we should just add `"` as an explicit no warning option, but otherwise stick with things as they are?

dhomeier:
> Another worry: suppose people are just reading `votable` or so; are we now going to break interpretation of those units?

Do you mean in the sense of "just reading 3rd party data", possibly through a higher-level interface such as `votable`? Not sure how likely breakage is here, perhaps should have a look at how comprehensive the present votable tests are.

> Hmm, perhaps the last worry means that we should just add " as an explicit no warning option, but otherwise stick with things as they are?

I.e. continue to accept any unknown unit strings with only a warning? That may ultimately be a political decision, but I'd rather vote against it. Again, the standard is not terribly explicit about this, but e.g. 2.2

> A library which implements this specification **should** be able to distinguish known and unknown units, and identify deviations from the restrictions on their use, below. It **should** be able to communicate such information to a caller, but it **should not** unilaterally reject unit strings which use unknown units or use known units in disapproved ways (of course, a higher-level application is free to reject unit strings for any reason it pleases).

certainly does not mandate that the parser should accept such units _unconditionally_ (i.e. without signalling it through some switch). From the user side, judging from issues like #6302 the almost silent
acceptance of arbitrary unit strings, in contrast to `fits` or other formats, seems to create more confusion than their rejection. But of course we may only hear from the other side after the behaviour has changed.

dhomeier:
Then, adding a `DeprecationWarning` to the old usage may be an alternative, but in my experience people who did not care about existing warnings are unlikely to pay attention to depreciation warnings until the usage actually is longer supported...

dhomeier:
> Actually, one thing: with your changes, is it still possible to get the femto-urlong with the same string on old and new astropy? I think `Unit("f'urlong'")` will continue to work, right? I ask since it is helpful for code that needs to support multiple versions of astropy to have a way to do so without version checks (but it is not so bad if they have to make their code more explicit). Probably another reason to be sure to put a link to further docs in the `ValueError`.

Yes, since this is tested in `test_vounit_custom` and in fact required by the standard, it does work. What's not working and was concerning me is that the unit cannot be defined both implicitly and explicitly, but that does not work in the existing implementation either - raising the same
```
ValueError: 'f'urlong'' did not parse as vounit unit: At col 0, Object with name 'urlong' already exists in given namespace (Unit("urlong")).
```

And yes, definitely should add a reference to the quoting syntax to warnings/error messages.

dhomeier:
Just had a look into the conflict resolver - I think the only ambiguous case would be your
```
    # Yikes, this becomes "femto-urlong"...  But at least there's a warning.
    with catch_warnings() as w:
        x = u.Unit("furlong/week", format="vounit")
```
I could add the `catch_warnings` context to the first test of `u.Unit('"furlong/week"', format="vounit")` and put this one into a `pytest.raises` one (if we do decide to keep the `ValueError`), but essentially this would just be repeating the `test_deprecated_did_you_mean_units` checks.

mhvk:
A compromise would be to change to `ValueError` for the default and use `parse_strict='warn'` in any `vo` code.

Separately, currently `to_string('vounit')` does what I feel is "the right thing" - always use the explicit single-quote form. Maybe good to explicitly test that in your tests as well.

mhvk:
On the conflict - just go ahead and remove if it is a duplication!

mhvk:
I'm confused about your `ValueError: 'f'urlong'' did not parse as vounit unit: At col 0, Object with name 'urlong' already exists in given namespace (Unit("urlong")).` - is this different from what you have on top?

dhomeier:
> A compromise would be to change to `ValueError` for the default and use `parse_strict='warn'` in any `vo` code.

But with the current implementation `parse_strict='warn'` returns an `UnrecognizedUnit` instead, which is generally not what we want. It should be possible to change this to let `'warn'` (and `'silent'`) instead make the `vounit` format accept unknown units in any string. A bit trickier to implement, but probably possible, I think.

dhomeier:
> Separately, currently `to_string('vounit')` does what I feel is "the right thing" - always use the explicit single-quote form. Maybe good to explicitly test that in your tests as well.

That's effectively what `test_custom_vounit` already does, just not in the simpler case of `'furlong'`.

dhomeier:
For the `ValueError`, the full sample code (omitting full backtrace) is
(old)
```
> x = u.Unit('furlong', format='vounit')
WARNING: UnitsWarning: Unit 'furlong' not supported by the VOUnit standard. [astropy.units.format.vounit]

> x = u.Unit("f'urlong'", format='vounit')
ValueError: 'f'urlong'' did not parse as vounit unit: At col 0, Object with name 'urlong' already exists in given namespace (Unit("urlong")).

> x = u.Unit("'furlong'", format='vounit')
ValueError: ''furlong'' did not parse as vounit unit: At col 0, Object with name 'furlong' already exists in given namespace (Unit("furlong")).
```
– in the new version analogously with `u.Unit('"furlong"', format='vounit')` in the first call.

mhvk:
Ah, that first `ValueError` is rather surprising, since it should just be consistent. And looking at the code, I now understand your suggestion of removing prefixes before looking whether something is already defined - if it is that simple to fix it, I think we should do it!!

p.s. Sorry for not reading what you wrote more carefully before

mhvk:
p.s. On `vo[table]`, etc.: I think it is a real worry, but let's deal with it separately.

dhomeier:
Admittedly so far this was only a vague idea that checking first against existing base unit names and then parsing the prefix could resolve this. If it actually is an inconsistency -if, say, `'furlong'` already exists, then it does hide the `furlong` (femto-`urlong`), and consequently the `urlong` and any of its other prefixed forms can no longer be defined.
What is less consistent is that the explicit definition of unknown units (by single quotes) does work with existing known unit names (as literally specified for a unit `'m'` in the standard. Effectively for these cases the quotes become part of the unknown unit's name, although it is still represented by two entries in `u.format.VOUnit._custom_units) = {'m': Unit("m"), "'m'": Unit("m"), ...}`. But here the former is effectively shadowed by the `'m': Unit("m")` in `u.format.VOUnit._units`. So my thoughts about solutions at this point are:

- Keep the existing behaviour and leave it to the user not to define conflicting units.

- Keep only the quoted name, i.e. remove the alternative name property and the unquoted form in in `_custom_units`. Then `'furlong'` could no longer be accessed as `furlong` at all, so that might break some existing code.

- Store the explicitly defined units with both name variants in their own namespace, like `_unrecognized_units` (and perhaps give them a specific attribute like `UnrecognizedUnit` or `ImmutableUnit` as well) and make the resolution order  `_units` -> `_custom_units` -> `_unrecognized_units` (so `'furlong'` would remain accessible as `furlong` as long as no `urlong`/`femto-urlong` has been declared).

One more thing I noticed is that unknown units still have an automatically assigned `.physical_type`, which is most likely not the type this special unit is meant to represent and rather adds to the confusion. E.g. after defining `'m'`, this happens:
```
> u.Unit("m", format='vounit').to("km")                                                                                                                                           
0.001

> u.Unit("'m'", format='vounit').to("km")
UnitConversionError: 'm' (length) and 'km' (length) are not convertible
```
– I think at least the 'm' should not be described as 'length' in the latter case...

pllim:
@dhomeier , unfortunately a merge commit crept in and it needs to be gotten rid of. This is because it might confused our release scripts.

Try a rebase on latest `master` and see if that works. See http://docs.astropy.org/en/latest/development/workflow/development_workflow.html#rebase-but-only-if-asked for more info.

dhomeier:
Thought I had taken care of that with the inline edits here, but apparently some conflicts in other files sneaked in.

pllim:
Ops, sorry @dhomeier , but you would have to rebase again to get rid of the change log conflict and pick up new CI updates from `master`.

@mhvk and @tomdonaldson -- Any chance we can wrap this up by Feature Freeze Friday? Otherwise, milestone needs moving.

Thanks!

dhomeier:
Uh, this had completely slipped off my radar...

dhomeier:
@mhvk I am actually happy to put it in as is, although I still think something should be done to avoid or at least reduce the problems of custom units "shadowing" standard units of the same name. Technically that is not central to this PR and could be dealt with in a new one.
But this one is not very pressing, I guess, so I'd be fine with pushing it to 4.1 as well.

pllim:
@dhomeier and @mhvk , any plans to wrap this up? I got lost in the discussions.

dhomeier:
> @dhomeier and @mhvk , any plans to wrap this up? I got lost in the discussions.

Since I haven't seen much further input to https://github.com/astropy/astropy/pull/8979#issuecomment-512840422, I would strongly vote for merging as soon as possible and keeping the points above for a new issue. I can check if I can fix `get_physical_type` for quoted units (so `Unit("'m'")` would become type `'unknown'` rather than `'length'`, but that's about all I could add at this point.

mhvk:
@dhomeier - I agree, we can postpone the points you had raised yourself to a new issue. It would be nice if you could fix the `physical_type` - it would be even nicer if that became something that would give a hint about what the problem was - as is, the text of your example is very unhelpful!

dhomeier:
> be nice if you could fix the `physical_type` - it would be even nicer if that became something that would give a hint about what the problem was - as is, the text of your example is very unhelpful!

You mean why it is identifying as a real physical unit? Well, I don't think I can tell for sure until I have actually fixed it, but the units return identical `._get_physical_type_id()` because they report the same `.name` properties (`'m'` in the above example) whether formed from a quoted string or not. More specifically it might be related to the alternative name property mentioned under #2 of the issues above, i.e.
```
>>> u.Unit("m", format='vounit').names
['m', 'meter']
>>> u.Unit("'m'", format='vounit').names
['m', "'m'"]
u.Unit("m", format='vounit')._get_physical_type_id()
(('m', 1),)
u.Unit("'m'", format='vounit')._get_physical_type_id()
(('m', 1),)
```

mhvk:
@dhomeier - darn, that is annoying. Maybe for now we could go the route of a warning (on unit creation) that a unit shadows an existing one and that therefore the physical type might well be wrong? I think eventually we'd need to move to your solution of a separate namespace, but we should not get the perfect be the enemy of at least sanitizing this mess somewhat!

dhomeier:
@mhvk I have `_get_physical_type_id()` now check for `aliases`/`long_names` to identify custom units by their quoted name form and made the `UnitConversionError` message more descriptive.

tomdonaldson:
@dhomeier Thank you for this work.  I apologize for arriving late to the discussion, and also for probably not fully grasping many of the points made here.

That said, I am a bit confused about the intent of the PR and of the issue #8978.  If I'm reading it correctly, it seems like this PR is not correcting any non-compliance with the IVOA standards, but is adding something to the syntax with the intent of reducing confusion about the standard.  Is that correct?

If so, I have some concerns with the approach.  The main issue to me is that by changing the syntax, the units parser no longer complies with the IVOA spec.  While it may be confusing that `"furlong"` parses as "femto-urlong", with `.bases[0]._represents` being `Unit("1e-15 urlong")`, that is what the spec says to do.  By changing that behavior, the parser will no longer be spec-compliant.

I suppose that the main use case of parsing units with format `"vounit"` will be in the parsing of VOTables.  For those cases it will be particularly important for the unit parsing to achieve what was intended by the spec.

Secondarily, I'm not sure that this approach really addresses the confusing aspects cited.  Even in following this discussion, the new syntax seems to have its own set of confusing aspects which don't seem to be documented.

To improve the usability of the original implementation, I wonder if just adding to the docs would help.  The test suite could also probably be improved (at least documented) to clarify the expected behaviors all the cases.

astropy/units/core.py
```diff
@@ -536,7 +536,9 @@ def _get_physical_type_id(self):
         useful as a dictionary key.
         """
         unit = self.decompose()
-        r = zip([x.name for x in unit.bases], unit.powers)
+        # custom units are stored under their quoted (alias) name
+        r = zip([x.name if f"'{x.name:}'" not in x.aliases else
```
mhvk, line 540: Total nitpick, but the trailing `:` are not necessary in the format strings.

astropy/units/format/vounit.py
```diff
@@ -91,7 +92,12 @@ def parse(cls, s, debug=False):
             raise core.UnitsError(
                 "'{}' contains multiple slashes, which is "
                 "disallowed by the VOUnit standard".format(s))
-        result = cls._do_parse(s, debug=debug)
+        if len(s) > 2 and s[0] == s[-1] == '"':
+            cls._custom_inline = True
```
mhvk, line 96: This seems rather fragile, to change the class while parsing a given unit. I would suggest instead to overwrite `_do_parse` and call `_parse_unit` with an extra keyword argument. (And the move this logic inside there? Or could it even be inside `_parse_unit`?

astropy/units/physical.py
```diff
@@ -60,7 +60,7 @@ def get_physical_type(unit):
     -------
     physical : str
         The name of the physical quantity, or unknown if not
-        known.
+        known (quoted custom units are always considered unknown).
```
mhvk, line 63: Maybe "(for vounit format, quoted ...)"

astropy/units/tests/test_format.py
```diff
@@ -419,27 +419,41 @@ def test_scaled_dimensionless():
...
+        u.Unit('ANGSTROM', format='cds')
+    assert str(e.value).count('Angstrom') == 5
 
-    with catch_warnings() as w:
+    with pytest.raises(ValueError,
+                       match=r'not supported by the VOUnit standard') as e:
         u.Unit('ANGSTROM', format='vounit')
+    assert str(e.value).count('nm') == 1
 
+    with catch_warnings() as w:
```
mhvk, line 439: There is also a `pytest.warns`, which works just like `pytest.raises`

astropy/units/tests/test_format.py
```diff
@@ -500,15 +516,51 @@ def test_vounit_custom():
...
+    assert u.physical.get_physical_type(x) == 'unknown'
+    with pytest.raises(u.UnitConversionError,
+                       match=r'\(custom unit ..m..\) and .cm. \(length\)'):
+        x.to('cm')
+
+
 def test_vounit_implicit_custom():
+    # We cannot overwrite already created custom units, so first make sure
+    # that the namespace is clean:
+    u.format.VOUnit._custom_units.clear()
```
mhvk, line 535: Not really necessary, but I think the proper way to write these types of test is to have a test class that has a `setup_class`/`teardown_class` method pair that ensures the directory stays in the same state.  Now, if someone runs `astropy.test()` their custom units disappear.

mhvk:
@tomdonaldson - good points. I'll let @dhomeier answer in more detail, but your comments do make clear that whatever we do, we should be sure to document it clearly.

mhvk:
I feel we're a bit frustratingly stuck with this PR, but as we didn't manage to get unstuck before the 4.1 deadline, sadly I have to move the milestone. @dhomeier, @tomdonaldson - might it make sense for the two of you to chat off-line about how best to proceed?

tomdonaldson:
> I feel we're a bit frustratingly stuck with this PR, but as we didn't manage to get unstuck before the 4.1 deadline, sadly I have to move the milestone. @dhomeier, @tomdonaldson - might it make sense for the two of you to chat off-line about how best to proceed?

Yes @mhvk , a discussion would be good.  With my current understanding, I would recommend closing the PR without merging, but I fully acknowledge that my current understanding may be missing important points.

dhomeier:
> Yes @mhvk , a discussion would be good. With my current understanding, I would recommend closing the PR without merging, but I fully acknowledge that my current understanding may be missing important points.

If nothing else, your comments and questions demonstrated that the main motivation of the PR is not even obvious enough for me to come up with a quick response. One initial trigger was that in #6302 there seemed consensus that custom units should not be _silently_ accepted, and that they should be identifiable as `UnrecognizedUnit`. Also there were (and possibly still are) a number of tests that quite obviously intended to check warnings on incorrect units like `Angstroms` but never caught `'vounit'` not raising anything, because `catch_warnings` was not checking the length of the warnings.
My interpretation of the standard was that the double quotes do not need to be considered part of the unit string, but of the way the string is passed to the reader, but of course that can be questioned. More importantly it probably depends on what downstream packages have adopted for their use, and if this would risk break anything.

Closing for now, part of this PR covers the **explicitly** defined custom units (with single quotes as per the standard), taking the position that those should not be identified as `physical_type` of something sounding similar and properly be interpreted as `_unrecognized_units` or some other suitable class. From the example discussed above, if someone puts a custom unit `m` in their table, and makes it clear that this unit has nothing to do with the standard metre by writing it as `'m'`, such a unit should not be read as `physical_type` `length`. This part perhaps should have better gone into its own PR, but if the double quote syntax is rejected, I still propose to extract the better treatment of single quote custom units (transferring it to a new PR, if necessary).

astropy/units/tests/test_format.py
dhomeier: Thanks for reminding me I had actually been a quite vocal proponent of using `pytest.warns` wherever a warning is really expected and required! ;-)
This time I had just lazily copied the entire template from another test...

astropy/units/tests/test_format.py
dhomeier: Good point; lately I am noticing a bunch of problems from the opposite side with tests (particularly from `io.asdf` it seems) adding "enabled" units that later fail to raise expected warnings or exceptions.
Will have to look into this and the class definition issue above later when it's settled how or if to proceed with the PR.

mhvk:
@dhomeier - If it is do-able, I think separating out the parts that ensure a custom unit never has a physical type would make a lot sense.

PR #8982, opened 2019-07-11
Fix uint conversion after slicing

### LINKED ISSUE #5490: Slicing FITS_rec objects produces unexpected dtypes.
Consider the following snippet:

```python
import numpy as np
from astropy.io import fits

def create_data():
    tbhdu = fits.BinTableHDU.from_columns([
        fits.Column(name='a', format='I', array=np.arange(2**16, dtype=np.int16)),
        fits.Column(name='b', format='I', bscale=1, bzero=2**15, array=np.arange(2**16, dtype=np.uint16))])
    tbhdu.writeto('test_scaled_slicing.fits')

def load_data(nonnegative=False):
    with fits.open('test_scaled_slicing.fits') as hdulist:
        data = hdulist[1].data
    if nonnegative:
        return data[data['a'] >= 0]
    else:
        return data

create_data()
t1 = load_data()
t2 = load_data(True)
assert t1['b'].dtype is np.dtype(np.uint16)
# assert t2['b'].dtype is np.dtype(np.float64)
assert t2['b'].dtype is np.dtype(np.uint16)
```

The second `assert` fails because the slicing the FITS_rec object does not trigger the mechanism that handles the scaling of the column 'b'.  In other words, some data access methods on FITS_rec trigger the scaling but some don't.

If however, I use a method that forces the scaling, then this problem goes away.  Here's an example:

```python
def load_data(nonnegative=False):
    with fits.open('test_scaled_slicing.fits') as hdulist:
        data = hdulist[1].data
    #
    # WORKAROUND
    #
    if any([c.bzero is not None for c in data.columns]):
        foo = [data[k].dtype for k in data.dtype.names]
    if nonnegative:
        return data[data['a'] >= 0]
    else:
        return data
```

That is, merely accessing the dtype of each column triggers the scaling.  Looking into the code itself, it looks like the `._converted` attribute needs to be filled in for slicing to work properly.

saimn:
Fix #5490.

The _uint attribute was not transfered to the sliced array, so the data
was later converted either correctly (if it was converted first before
slicing) or as float in the other case.

### DIFF
--- a/astropy/io/fits/fitsrec.py
+++ b/astropy/io/fits/fitsrec.py
@@ -497,6 +497,7 @@ class FITS_rec(np.recarray):
 
         # We got a view; change it back to our class, and add stuff
         out = out.view(type(self))
+        out._uint = self._uint
         out._coldefs = ColDefs(self._coldefs)
         arrays = []
         out._converted = {}
--- a/astropy/io/fits/tests/test_uint.py
+++ b/astropy/io/fits/tests/test_uint.py
@@ -4,6 +4,7 @@ import platform
 
 import pytest
 import numpy as np
+from numpy.testing import assert_array_equal
 
 from astropy.io import fits
 from . import FitsTestCase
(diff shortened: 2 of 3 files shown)


astropy/io/fits/fitsrec.py
```diff
@@ -497,6 +497,7 @@ def __getitem__(self, key):
 
         # We got a view; change it back to our class, and add stuff
         out = out.view(type(self))
+        out._uint = self._uint
```
mhvk: I vaguely recall adding some of the above code, but looking at it again I wonder a bit why... Normally, an `ndarray` subclass is supposed to do this in `__array_finalize__`, but here of course the view as a `recarray` and then the view back here circumvent that (and from a quick look, just doing `out.__arrray_finalize__(self)` would seem unlikely to do quite the right thing)

Anyway, I think the fix here is good, though perhaps have a quick look at `__array_finalize__` and see if anything else is missing (since `_uint` at least is updated there).

astropy/io/fits/tests/test_uint.py
```diff
@@ -112,3 +113,30 @@ def test_uint_columns(self, utype):
...
+        tbhdu.writeto(self.temp('test_scaled_slicing.fits'))
+
+        with fits.open(self.temp('test_scaled_slicing.fits')) as hdulist:
+            data = hdulist[1].data
+        assert_array_equal(data['b'], dataref)
+        sel = data['a'] >= 0
+        assert_array_equal(data[sel]['b'], dataref[sel])
+        assert data[sel]['b'].dtype == dataref[sel].dtype
+
+        with fits.open(self.temp('test_scaled_slicing.fits')) as hdulist:
```
mhvk: Are we opening the file twice to prevent being fooled by a cache or so?

astropy/io/fits/fitsrec.py
saimn: Yes, I should have mentioned that I requested your review because you changed a bit this code in the past (e.g. #6955), but the issue was already there before. The problem is that with the recarray view we loose the FITS_rec attributes, so they are no more available in `__array_finalize__` when viewing back to FITS_rec.
I'm not completely sure about the other attributes, I think that the missing ones can be recomputed from the coldefs.

astropy/io/fits/tests/test_uint.py
saimn: Yes somewhat, in the first case as `data['b']` is accessed the data is converted and cached, so before the fix `data[sel]` had the wrong value for `_uint` but was still able to print correct values because the cache is copied.

saimn:
Thanks for the review @mhvk !

mhvk:
And thanks for the fix, and for explaining - all makes sense. (Though I think we should be able to be more properly `ndarray` subclassy - but for another time...)

Issue #7661, opened 2018-07-17
Make @quantity_input pass over arguments with non-unit annotations


namurphy:
The `@astropy.units.quantity_input` decorator currently raises an exception if an argument has an annotation that cannot be converted to a unit.  This has the side effect that type hint annotations and annotations used by different decorators cannot be used in functions decorated by `@quantity_input`.

Here is a minimal example (using Astropy 3.0.2):
```Python
>>> import astropy.units as u

>>> @u.quantity_input
... def f(x: u.m, y: str) -> u.m:
...    return x

>>> f(1 * u.m, 'x')
TypeError: <class 'str'> can not be converted to a Unit
```

My suggestion would be for `@quantity_input` to skip arguments with annotations that are not units or cannot be converted to units.  However, if the annotation is a string that cannot be converted to a unit, then I would propose raising a `TypeError` (or maybe a `u.UnitsError` or a `u.UnitConversionError`) since there could be a typo in a string that could have been intended to be a unit.

Thanks!
-Nick

mhvk:
This seems like a good idea. Could be inside a `try/except` or perhaps one should explicitly omit trying the unit conversion for anything of the class that is used for type hints (if there is one). cc @Cadair.

namurphy:
The use case I had in mind was for compatibility with PlasmaPy's `@particle_input` decorator, which works on arguments that are annotated with our `Particle` class.  We aren't able to use both `@quantity_input` and `@particle_input` at the same time because `@quantity_input` will raise exceptions for the arguments annotated with `Particle`.  If `@quantity_input` were to ignore non-unit annotations, then we would be able to use `@quantity_input` a lot more frequently in the package.

Besides this issue, I really like `@quantity_input` and am really glad that it's in `astropy.units`!

Cadair:
When originally written we tried to duck type `quantity_input` to be unit implementation agnostic. This and other changes made since it was originally written have led to moving further away from this.

That said, I think this is a very good idea, and we should  implement it, but try not to make it too astropy units specific in the implementation.

maciekgroch:
Hey, 

I am dealing with exactly the same issue: one of the parameters of my function is not an astropy quantity because of which I am not able to use typing hints. Is there any progress on the issue?

I am using Python 3.7 and astropy 3.2.1

THanks for help.

likewei92:
I just ran into the same issue. Here's my attempt to fix it: https://github.com/astropy/astropy/pull/8984

[...]

PR #8984, opened 2019-07-11
Allowing non unit annotations in quantity_input decorator

likewei92:
(I copied the description from https://github.com/astropy/astropy/pull/7672, since that PR does the same thing but since has gone stale)

Closes https://github.com/astropy/astropy/issues/7661

Instead of raising an error, `@quantity_input` will skip any non-unit type annotations in function parameters. So, code such as below won't result in an error:
```python
>>> import astropy.units as u

>>> @u.quantity_input
... def f(x: u.m, y: str) -> u.m:
...    return x

>>> f(1 * u.m, 'x')
<Quantity 1. m>
# this would otherwise give
# TypeError: <class 'str'> can not be converted to a Unit
```

It will however, raise a ValueError in cases like these:
```python
>>> import astropy.units as u

>>> @u.quantity_input
... # consider the typo "kilogrammm" instead of "kilogram"
... def f(x: u.m, y: "kilogrammm") -> u.m:
...    return x

>>> f(1 * u.m, 2 * u.kg)
ValueError: Invalid unit or physical type 'kilogrammm'.
```

TLDR; @quantity_input now ignores any parameter in function definition which is non-unit type annotated, but will raise a ValueError if a string is passed and Unit(string) is unable to parse it (like above example).

EDIT: Close #7672

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -110,6 +110,8 @@ astropy.uncertainty
 astropy.units
 ^^^^^^^^^^^^^
 
+- Accept non-unit type annotations in @quantity_input. [#8984]
+
 - For numpy 1.17 and later, the new ``__array_function__`` protocol is used to
   ensure that all top-level numpy functions interact properly with
   ``Quantity``, preserving units also in operations like ``np.concatenate``.
--- a/astropy/units/decorators.py
+++ b/astropy/units/decorators.py
@@ -46,6 +46,9 @@ def _validate_arg_value(param_name, func_name, arg, targets, equivalencies):
     unit or physical type, ``target``.
     """
 
+    if len(targets) == 0:
+        return
+
     allowed_units = _get_allowed_units(targets)
 
     for allowed_unit in allowed_units:
@@ -94,7 +97,8 @@ class QuantityInput:
         the argument is not equivalent to the unit specified to the decorator
         or in the annotation.
         If the argument has no unit attribute, i.e. it is not a Quantity object, a
-        `ValueError` will be raised.
+        `ValueError` will be raised unless the argument is an annotation. This is to
+        allow non Quantity annotations to pass through.
 
         Where an equivalency is specified in the decorator, the function will be
         executed with that equivalency in force.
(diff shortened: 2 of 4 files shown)


likewei92:
@mhvk @Cadair could you help to review?

mhvk:
@likewei92 - I see you are still working on the PR. Generally, I would suggest to run the tests locally before pushing - every time you push a change, you start a new instance of all the tests, multiplied quite a number of time. This clogs up the system and wastes energy, which you might as well avoid.

Do ping us when you are happy and everything works!

ritiek:
@likewei92 Thanks for taking this over! Your approach is definitely better than mine. I'll close my stale PR.

likewei92:
@mhvk @Cadair I'm ready now :)

adrn:
Thanks for the PR, @likewei92! Another way of approaching this, and I think what I would prefer, would be to instead only pass through string or unit-subclass instances via `valid_targets`. i.e., to do some validation of the elements in `valid_targets` below the if-else clause at [L214](https://github.com/astropy/astropy/blob/master/astropy/units/decorators.py#L214). This would not catch your "kilograam" case, but then inside `_get_allowed_units()` you could instead wrap [L36](https://github.com/astropy/astropy/blob/master/astropy/units/decorators.py#L36) in a try/except that ignores the annotation. Does that make sense? I think this way will have logic that is easier to follow in the code.

likewei92:
@adrn I did as you suggested. COudl you help to take another look now?

likewei92:
pinging @mhvk @Cadair as well

likewei92:
Tests are passing now

likewei92:
Sure; I thought you would squash when merging. I squash all my changes to one commit now

likewei92:
I'm getting a CI timeout that seems to have nothing to do with my changes? Do you know how to rerun the pipeline?

likewei92:
@adrn I've squashed the commits and all the tests are passing now

adrn:
LGTM now - thanks @likewei92 !

PR #8993, opened 2019-07-14
http:// no no longer responds. but https:// does.

rtolesnikov:


mhvk:
OK, I heard back and they are indeed transitioning to https - though it wasn't meant to break yet. So, to future-proof ourselves, let's put this in. Thanks, @rtolesnikov!

PR #8854, opened 2019-06-15
Allow initialization of zero-size Times for all formats

mhvk:
Seeing astro-bot warn about #7868 again, I thought I would just take it over and make a PR that uses a bit less special-casing for zero-size (though still more than I'd like, but hard to avoid).

Milestoned this for 3.2.1 as I think it may be hard to backport.
 
Fixes #7866

PR #8998, opened 2019-07-15
Faster info via meta class

mhvk:
@taldcroft - I decided to try the metaclass approach to `DataInfo`, where for every attribute one automatically generates a descriptor that either gets/sets in `_attrs` or from the parent. It is a bit of magic in the meta-class, but not really that much more than there was before in `__getattr__` and `__setattr__`. Note that I had to move to `__slots__` in order to continue to disallow setting arbitrary attributes; not completely clear whether that is worth it. I have some other comments in-line.

Comparing this PR to current master:
```
from astropy.table import MaskedColumn
m = MaskedColumn([1, 2])
%timeit m.info
# 1000000 loops, best of 5: 290 ns per loop
# on current master: 1.24 µs
%timeit m.info.name
# 1000000 loops, best of 5: 659 ns per loop
# on current master: 2.12 µs

q = [1, 2] * u.m
%timeit q.info
# 1000000 loops, best of 5: 319 ns per loop
# on current master: 1.82 µs
%timeit q.info.name
# 1000000 loops, best of 5: 463 ns per loop 
# on current master: 2.46 µs
```

The last item is the most relevant for possibly moving everything to `info` - `463 ns` is still about 6 times slower than just `%timeit m.name`. But it is not obvious to me that with the general speed-up here, it would still be a limiting factor.

### DIFF
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -72,9 +72,11 @@ class SkyCoordInfo(MixinInfo):
         if 'distance' in attrs and np.all(obj.distance == 1.0):
             attrs.remove('distance')
 
-        self._represent_as_dict_attrs = attrs
-
-        out = super()._represent_as_dict()
+        out = {}
+        for attr in attrs:
+            value = getattr(obj, attr, None)
+            if value is not None:
+                out[attr] = value
 
         out['representation_type'] = obj.representation_type.get_name()
         out['frame'] = obj.frame.name
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -16,7 +16,7 @@ from astropy.time import Time
 from .distances import Distance
 from .angles import Angle
 from .baseframe import (BaseCoordinateFrame, frame_transform_graph,
-                        GenericFrame, _get_repr_cls)
+                        GenericFrame)
 from .builtin_frames import ICRS, SkyOffsetFrame
 from .representation import (SphericalRepresentation,
                              UnitSphericalRepresentation, SphericalDifferential)
@@ -72,11 +72,7 @@ class SkyCoordInfo(MixinInfo):
         if 'distance' in attrs and np.all(obj.distance == 1.0):
             attrs.remove('distance')
 
-        out = {}
-        for attr in attrs:
-            value = getattr(obj, attr, None)
-            if value is not None:
-                out[attr] = value
+        out = super()._represent_as_dict(attrs)
 
         out['representation_type'] = obj.representation_type.get_name()
         out['frame'] = obj.frame.name
--- a/astropy/table/column.py
+++ b/astropy/table/column.py
@@ -1092,21 +1092,16 @@ class MaskedColumnInfo(ColumnInfo):
         method = self.serialize_method[self._serialize_context]
 
         if method == 'data_mask':
-            # Note that adding to _represent_as_dict_attrs triggers later code which
-            # will add this to the '__serialized_columns__' meta YAML dict.
-
-            # Note also one driver here is a performance issue in #8443 where repr() of a
+            # Note: a driver here is a performance issue in #8443 where repr() of a
             # np.ma.MaskedArray value is up to 10 times slower than repr of a normal array
             # value.  So regardless of whether there are masked elements it is useful to
             # explicitly define this as a serialized column and use col.data.data (ndarray)
             # instead of letting it fall through to the "standard" serialization machinery.
             out['data'] = col.data.data
-            self._represent_as_dict_attrs += ('data',)
 
             if np.any(col.mask):
                 # Only if there are actually masked elements do we add the ``mask`` column
                 out['mask'] = col.mask
-                self._represent_as_dict_attrs += ('mask',)
 
         elif method is 'null_value':
             pass
--- a/astropy/table/serialize.py
+++ b/astropy/table/serialize.py
@@ -58,7 +58,6 @@ def _represent_mixin_as_column(col, name, new_cols, mixin_cols,
     ``MaskedColumn`` can also be serialized.
     """
     obj_attrs = col.info._represent_as_dict()
-    ordered_keys = col.info._represent_as_dict_attrs
 
     # If serialization is not required (see function docstring above)
     # or explicitly specified as excluded, then treat as a normal column.
@@ -83,8 +82,8 @@ def _represent_mixin_as_column(col, name, new_cols, mixin_cols,
         if nontrivial(col_attr):
             info[attr] = xform(col_attr) if xform else col_attr
 
-    data_attrs = [key for key in ordered_keys if key in obj_attrs and
-                  getattr(obj_attrs[key], 'shape', ())[:1] == col.shape[:1]]
+    data_attrs = [key for key, value in obj_attrs.items() if
+                  getattr(value, 'shape', ())[:1] == col.shape[:1]]
 
     for data_attr in data_attrs:
         data = obj_attrs[data_attr]
--- a/astropy/table/tests/test_info.py
+++ b/astropy/table/tests/test_info.py
@@ -248,14 +248,6 @@ def test_no_deprecation_warning():
         assert len(warns) == 0
 
 
-# No problem any more?
-# def test_lost_parent_error():
-#     c = table.Column([1, 2, 3], name='a')
-#     with pytest.raises(AttributeError) as err:
-#         c[:].info.name
-#     assert 'failed access "info" attribute' in str(err.value)
-
-
 def test_info_serialize_method():
     """
     Unit test of context manager to set info.serialize_method.  Normally just
--- a/astropy/table/tests/test_info.py
+++ b/astropy/table/tests/test_info.py
@@ -248,11 +248,12 @@ def test_no_deprecation_warning():
         assert len(warns) == 0
 
 
-def test_lost_parent_error():
-    c = table.Column([1, 2, 3], name='a')
-    with pytest.raises(AttributeError) as err:
-        c[:].info.name
-    assert 'failed access "info" attribute' in str(err.value)
+# No problem any more?
+# def test_lost_parent_error():
+#     c = table.Column([1, 2, 3], name='a')
+#     with pytest.raises(AttributeError) as err:
+#         c[:].info.name
+#     assert 'failed access "info" attribute' in str(err.value)
 
 
 def test_info_serialize_method():
--- a/astropy/table/tests/test_info.py
+++ b/astropy/table/tests/test_info.py
@@ -248,6 +248,13 @@ def test_no_deprecation_warning():
         assert len(warns) == 0
 
 
+def test_lost_parent_error():
+    c = table.Column([1, 2, 3], name='a')
+    with pytest.raises(AttributeError) as err:
+        c[:].info.name
+    assert 'failed access "info" attribute' in str(err.value)
+
+
 def test_info_serialize_method():
     """
     Unit test of context manager to set info.serialize_method.  Normally just
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -98,7 +98,6 @@ class TimeInfo(MixinInfo):
     required when the object is used as a mixin column within a table, but can
     be used as a general way to store meta information.
     """
-    attrs_from_parent = set(['unit'])  # unit is read-only and None
     attr_names = MixinInfo.attr_names | {'serialize_method'}
     _supports_indexing = True
 
--- a/astropy/utils/data_info.py
+++ b/astropy/utils/data_info.py
@@ -293,36 +293,25 @@ class DataInfo(metaclass=DataInfoMeta):
         # If bound to a data object instance then create the dict of attributes
         # which stores the info attribute values. Default of None for "unset"
         # except for dtype where the default is object.
-        self._parent_ref = None
         if bound:
             self._attrs = {attr: (None if attr != 'dtype' else np.dtype('O'))
                            for attr in self.attr_names}
 
     @property
     def _parent(self):
-        if self._parent_ref is None:
+        try:
+            return self._parent_ref()
+        except AttributeError:
             return None
-        else:
-            parent = self._parent_ref()
-            if parent is not None:
-                return parent
-
-            else:
-                raise AttributeError("""\
+        except TypeError:
+            raise AttributeError("""\
 failed access "info" attribute on a temporary object.
 
 It looks like you have done something like ``col[3:5].info``, i.e.
 you accessed ``info`` from a temporary slice object ``col[3:5]`` that
 only exists momentarily.  This has failed because the reference to
 that temporary object is now lost.  Instead force a permanent
-reference with ``c = col[3:5]`` followed by ``c.info``.""")
-
-    @_parent.setter
-    def _parent(self, value):
-        if value is None:
-            self._parent_ref = None
-        else:
-            self._parent_ref = weakref.ref(value)
+reference with ``c = col[3:5]`` followed by ``c.info``.""") from None
 
     def __get__(self, instance, owner_cls):
         if instance is None:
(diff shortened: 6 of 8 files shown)


astropy/coordinates/sky_coordinate.py
```diff
@@ -72,9 +72,11 @@ def _represent_as_dict(self):
         if 'distance' in attrs and np.all(obj.distance == 1.0):
             attrs.remove('distance')
 
-        self._represent_as_dict_attrs = attrs
```
mhvk: One is no longer allowed to change this - but it is also not necessary any more, so a change for the better I think.

astropy/table/serialize.py
```diff
@@ -58,7 +58,6 @@ def _represent_mixin_as_column(col, name, new_cols, mixin_cols,
     ``MaskedColumn`` can also be serialized.
     """
     obj_attrs = col.info._represent_as_dict()
-    ordered_keys = col.info._represent_as_dict_attrs
```
mhvk: I removed this thinking order was no longer an issue, since `dict` are guaranteed to be ordered - but I'm not sure this is true for `set` - so perhaps this is too much. If so, it'll need some thought.

astropy/table/tests/test_info.py
```diff
@@ -248,11 +248,12 @@ def test_no_deprecation_warning():
         assert len(warns) == 0
 
 
-def test_lost_parent_error():
-    c = table.Column([1, 2, 3], name='a')
-    with pytest.raises(AttributeError) as err:
-        c[:].info.name
-    assert 'failed access "info" attribute' in str(err.value)
+# No problem any more?
```
mhvk: I actually don't know why this test started failing...

astropy/time/core.py
```diff
@@ -98,7 +98,6 @@ class TimeInfo(MixinInfo):
     required when the object is used as a mixin column within a table, but can
     be used as a general way to store meta information.
     """
-    attrs_from_parent = set(['unit'])  # unit is read-only and None
```
mhvk: No need any more, since the info class sets the relevant property. So, a bit cleaner.

astropy/utils/data_info.py
```diff
@@ -231,46 +306,40 @@ class DataInfo:
...
+        # which stores the info attribute values. Default of None for "unset"
+        # except for dtype where the default is object.
         if bound:
-            self._attrs = dict((attr, None) for attr in self.attr_names)
+            self._attrs = {}
 
     @property
     def _parent(self):
-        if self._parent_ref is None:
+        try:
```
mhvk: Trying to speed this critical property up...

taldcroft:
@mhvk - this looks great from the description, but clearly needs a decent chunk of time to dig into and evaluate.  So it's on my radar, but probably not in the next few days...

mhvk:
No worries - but if you do have a small amount of time, it would be lovely to get #8854 out of the way...

taldcroft:
@mvhk - have you had a look at https://www.python.org/dev/peps/pep-0487/ ?  This is available as of Python 3.6.  Not clear if the `__slots__` stuff is possible, but it would be nice to avoid metaclasses if possible.

astropy/utils/data_info.py
```diff
@@ -184,7 +184,82 @@ def _get_data_attribute(dat, attr=None):
...
+Container for meta information like name, description, format.
+This is required when the object is used as a mixin column within
+a table, but can be used as a general way to store meta information.
+""")
+        return super().__new__(mcls, name, bases, dct)
+
+    def __init__(cls, name, bases, dct):
+        super().__init__(name, bases, dct)
+
+        # Define default getters/setters for attributes, if needed.
```
taldcroft: I just started looking at this and my immediate question is why it is not possible to use simple class attributes and rely on class-based inheritance.
```
class DataInfo(..):
    name = InfoAttribute()
    dtype = InfoAttribute()
   ...

class BaseColumnInfo(DataInfo):
    parent_table = InfoAttribute(copy=False)  # Attribute knows not to get copied

class ParentDtypeInfo(MixinInfo):
    dtype = ParentAttribute()
```
I ended up going crazy with descriptors for a class that stores most of its attributes in astropy Table meta.  :smile: 
https://github.com/sot/proseco/blob/ad6f861d0fd9eb27cc7034731ed758c9b875f84b/proseco/core.py#L511

mhvk:
(Taking this out of the inline comments as I think it goes to the heart of the matter.)

@taldcroft - The approach with just defining descriptors on the class is indeed a lot clearer! 

The main reason I went for the meta-class approach was to keep the API similar, with its `attr_names` and `attrs_from_parent` - which is why there are relatively few changes to `Info` subclasses.

The question then is whether we are happy to change it, which in turn depends on whether anybody outside of astropy relies on it. In principle, I guess the metaclass could help here, by doing the auto-generation while emitting a deprecation warning. But of course easier if we can do without it...

mhvk:
p.s. Have been meaning to look at the effects of PEP 487... Definitely useful for simple registration like Time fornats.

taldcroft:
> .. was to keep the API similar, with its attr_names and attrs_from_parent

Ah yes, I didn't immediately think of the API.  That's possibly thorny... So what about some radical thoughts just for the sake of conversation:

- What if we left the existing `DataInfo` hierarchy in place for external usage (but deprecate) and then made a brand new infrastructure that keeps most of the public API but is not constrained to implement in the same way.  For the sake of discussion all it `XDataInfo` or `XMixinInfo`.
- `attr_names` and `attrs_from_parents` can be done with `__set_name__` in the descriptors.
- Basically saying, `Info` is a honking good idea, let's make it better and faster with a reboot.  In practice it probably won't be radically different, but just a chance to take stock of what is there and make it nicer.
- I *think* that in the rest of astropy there are not too many places we would need to change to reflect this additional `XDataInfo` class hierarchy (but I haven't tried looking).
- Our Astropy classes would have e.g. `class QuantityInfoBase(XParentDtypeInfo)`, `class TimeInfo(XMixinInfo)`, etc.

astropy/utils/data_info.py
taldcroft: Another thing to put on the table is making this a direct ref instead of weak ref to the instance.  Making the weak ref for every access to `info` takes ~100 ns.  Since this ref is contained in an object stored on the instance, it seems like this is the straightforward reference cycle that GC should have no problem detecting.  But of course that may be simplistic and memory leaks are a bit annoying.

astropy/table/serialize.py
taldcroft: I don't recall seeing any statements about set order being preserved.

astropy/coordinates/sky_coordinate.py
```diff
@@ -72,9 +72,11 @@ def _represent_as_dict(self):
...
-
-        out = super()._represent_as_dict()
+        out = {}
```
taldcroft: Use `data_info._get_obj_attrs_map`.  (Which maybe needs to be made public?).

Or maybe `_represent_as_dict` could take an optional `attrs` to allow overriding `self._represent_as_dict_attrs`?

taldcroft:
Having given a decent first pass through this code and seen the timing, it looks great!  So it might be a reasonable strategy to get this in shape to merge and separately think about a bigger redesign to make the code more elegant / readable and maybe even faster.

astropy/table/serialize.py
mhvk: Actually, I think I managed to get myself confused: we call `_represent_as_dict()` here, which is these days guaranteed to be ordered (as it is done in order of `_represent_as_dict_attrs`).  So, that explains why no tests failed because of this!

astropy/utils/data_info.py
mhvk: Yes, I'd like that, but perhaps in the spirit of keeping things as similar as possible here, shall we leave it for later? I have this memory of confusing ourselves no end about memory leaks on `info` before... (Or was that for links to parent table??)

astropy/utils/data_info.py
```diff
@@ -184,7 +184,82 @@ def _get_data_attribute(dat, attr=None):
...
+class InfoAttribute:
+    def __init__(self, attr, default=None):
+        self.attr = attr
+        self.default = default
+
+    def __get__(self, instance, owner_cls):
+        if instance is None:
+            return self
+
+        return instance._attrs.get(self.attr, self.default)
```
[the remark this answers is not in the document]
mhvk: I in fact had that at some point! But I found it made almost no difference, while in this particular case it does feel substantially less clear, so I reverted it.

astropy/utils/data_info.py
```diff
@@ -184,7 +184,82 @@ def _get_data_attribute(dat, attr=None):
...
+
+        setattr(instance._parent, self.attr, value)
+
+
+class DataInfoMeta(type):
+    def __new__(mcls, name, bases, dct):
+        # Ensure that we do not gain a __dict__, which would mean
+        # arbitrary attributes could be set (as a by-product, save
+        # some memory).
+        dct.setdefault('__slots__', [])
```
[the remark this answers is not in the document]
mhvk: No, it has to be possible to define `__slots__` on the class (hence, the `setdefault`).

I'll remove the comment about the memory - I agree it is only confusiing.

Aside: the whole `__slots__` things is only there to disallow letting people stick random attributes on `info` (which otherwise requires overriding `__setattr__`, which makes every setting slow). Is that worth the complication?

astropy/utils/data_info.py
```diff
@@ -184,7 +184,82 @@ def _get_data_attribute(dat, attr=None):
...
+        # Set a default __doc__, mostly to avoid InheritDocstrings
```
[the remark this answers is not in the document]
mhvk: It copied `__doc__` to attributes - now that we do not use that anymore internally (#8881), this should indeed not be needed any more! 

... testing ...

Indeed! So, removed. Nice catch!

astropy/utils/data_info.py
```diff
@@ -184,7 +184,82 @@ def _get_data_attribute(dat, attr=None):
...
+a table, but can be used as a general way to store meta information.
+""")
+        return super().__new__(mcls, name, bases, dct)
+
+    def __init__(cls, name, bases, dct):
+        super().__init__(name, bases, dct)
+
+        # Define default getters/setters for attributes, if needed.
+        for attr in cls.attr_names:
+            if attr not in dct:
```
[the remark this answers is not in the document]
mhvk: Definitely. Since if we were to start from scratch, we'd do it as you suggested, we should check whether there is a path towards it. But I like the idea of finishing this up first.

astropy/coordinates/sky_coordinate.py
mhvk: Much better. Here, `represent_as_dict` itself is overridden, but, yes, a `super().represent_as_dict(attrs)` would work. I'm not quite sure it is worth changing, especially if we're considering a larger overhaul.

mhvk: Ah, darn, I now see why you suggested it: `_get_obj_attrs_map` is not on `DataInfo` but in the `data_info` module...

mhvk:
OK, I've implemented the suggested changes. I like the idea of merging this, and worrying about a simpler design separately. The one outstanding question, I think, is whether it is worth the hassle with `__slots__` to prevent people from sticking arbitrary attributes on `info`.

taldcroft:
Here are some timing results on my laptop with your branch and master.  I get qualitatively similar results.  This also explores the speedup of a different strategy (attrs are stored as object attributes, not dict elements).

```python
from weakref import ref as weakref
from astropy.utils.data_info import DataInfo
from astropy.table import MaskedColumn
```


```python
%astro
```

    astropy=4.0.dev25224



```python
DataInfo._attr_defaults  # Shows this is the faster-info-via-meta-class branch
```




    {'dtype': dtype('O')}




```python
class A:
    info = DataInfo()
```


```python
a = A()
```


```python
a.info.description = 'asdf'
```


```python
timeit a.info
```
    463 ns ± 3.27 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
MASTER:  2.5 µs ± 43.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)


```python
timeit a.info.description
```

    845 ns ± 13.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
 MASTER: 3.39 µs ± 64.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)


```python
a.__dict__
```




    {'info': dtype = object
     shape = --
     description = asdf
     class = A
     n_bad = 0}




```python
class SimpleDataInfo:
    """Simplest possible data info container"""
    __slots__ = ['name', 'description', 'dtype', 'meta', 'unit', '_parent_ref']
```


```python
class B:
    """Use SimpleDataInfo via a class property"""
    @property
    def info(self):
        try:
            info = self._info
        except AttributeError:
            info = self._info = SimpleDataInfo()
        finally:
            # info._parent_ref = self # weakref(self)
            return info            
```


```python
b = B()
```


```python
b.info.name = 'asdf'
```


```python
timeit b.info.name
```

    207 ns ± 6.02 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)



```python
class FDataInfo:
    """Descriptor that roughly matches the property above"""
    def __get__(self, instance, owner_cls):
        try:
            info = instance._info
        except AttributeError:
            if instance is None:
                # This is an unbound descriptor on the class
                self._parent_cls = owner_cls
                return self
            else:
                info = instance._info = SimpleDataInfo()
        finally:
            # info._parent_ref = weakref.ref(instance)
            return info
```


```python
class C:
    info = FDataInfo()
```


```python
c = C()
```


```python
c.info.description = 'hello'
```


```python
timeit c.info.description
```

    231 ns ± 4.75 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)



```python
m = MaskedColumn([1, 2])
m.info.description = 'asdf'
```


```python
timeit m.description
```

    50 ns ± 0.801 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)



```python
timeit m.info.description
```

    1.01 µs ± 10.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)



```python
junk = SimpleDataInfo()
m.junk = junk
```


```python
timeit m.junk
```

    58.1 ns ± 0.271 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)



```python
m.junk.name = 'name'
```


```python
timeit m.junk.name
```

    86.1 ns ± 0.644 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

taldcroft:
:+1: on explicitly preventing people from setting arbitrary attributes on `info`.  That matches the current release behavior and I think it prevents mistakes far more often than being a limitation.  Along those lines, I have often done things like `q.name = name` on a Quantity when I meant `q.info.name`.

The current implementation as slots is thus fine to me.  If we even move to a different implementation that does the same thing then that will also be fine.

astropy/table/column.py
```diff
@@ -1092,21 +1092,16 @@ def _represent_as_dict(self):
...
             # instead of letting it fall through to the "standard" serialization machinery.
             out['data'] = col.data.data
-            self._represent_as_dict_attrs += ('data',)
 
             if np.any(col.mask):
                 # Only if there are actually masked elements do we add the ``mask`` column
                 out['mask'] = col.mask
-                self._represent_as_dict_attrs += ('mask',)
 
         elif method is 'null_value':
```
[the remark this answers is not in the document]
mhvk: Also not sure, but coincidentally being fixed in #9225, so I'll leave it here.

astropy/utils/data_info.py
```diff
@@ -231,46 +298,40 @@ class DataInfo:
...
+        except AttributeError:
             return None
-        else:
-            parent = self._parent_ref()
-            if parent is not None:
-                return parent
-
-            else:
-                raise AttributeError("""\
+        except TypeError:
```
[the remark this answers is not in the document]
mhvk: Very weird - there must have been an intermediate state where it didn't happen, since the test was failing. In any case, I'll fix and reinstate the test!

mhvk:
OK, I reinstated the test and fixed the code so that the appropriate message is given again. The difference was that now there always is a `weakref` - it can just be dead, in which case `_parent_ref()` returns `None`.

PR #9004, opened 2019-07-16
Add example in narrative docs for the ``fixed`` parameter in models

hamogu:
I expeced an example in the docs for how to dfine a fixed parameter.
I understand why the current implementation is the way it is, but
it's not obvious that you need to pass in a dictionary where all
the fixed parameters are set to ``True``. I had expected a list with
parameter names or a tuple.
At the end of the day, all paraemtes that I pass are True, so a dict
where all values are True is not obvious.
I added a single line of code in the docs at the position where I expected
to find it.

### DIFF
--- a/docs/modeling/fitting.rst
+++ b/docs/modeling/fitting.rst
@@ -182,6 +182,13 @@ Fitters support constrained fitting.
           1.0 2.0 3.0
           1.0 2.0 3.0
 
+  The syntax to fix the same parameter ``c0`` using an argument to the model
+  instead of ``p1.c0.fixed = True`` would be::
+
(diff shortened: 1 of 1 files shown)


pllim:
I'll let Nadia set the milestone, as I don't know how far back this behavior applies. Also, is this still true after Perry's refactoring?

nden:
It has been like this for a very long time. Thanks for the clarification @hamogu!

pllim:
@nden , if you are okay with this, can you please approve?

nden:
@lim, I am holding off on merging PRs to modeling to see if #8769 can be merged first, if that's OK.

PR #9005, opened 2019-07-16
Fix units warnings

mhvk:
Further progress towards warnings-free testing! This time, no bugs in the code found, but did find some bugs in the test routines.

### DIFF
--- a/astropy/units/tests/test_format.py
+++ b/astropy/units/tests/test_format.py
@@ -489,7 +489,7 @@ def test_vounit_custom():
 
 
 def test_vounit_implicit_custom():
-    # Yikes, really...
+    # Yikes, this becomes "femto-urlong"...  But at least there's a warning.
     with catch_warnings() as w:
         x = u.Unit("furlong/week", format="vounit")
     assert x.bases[0]._represents.scale == 1e-15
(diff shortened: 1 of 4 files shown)


astropy/units/tests/test_format.py
```diff
@@ -429,9 +489,14 @@ def test_vounit_custom():
 
 
 def test_vounit_implicit_custom():
-    x = u.Unit("furlong/week", format="vounit")
+    # Yikes, really...
```
pllim: Why the "yikes"? Can you please clarify?

astropy/units/tests/test_units.py
```diff
@@ -160,8 +160,6 @@ def test_unknown_unit():
 
 
 def test_multiple_solidus():
-    assert u.Unit("m/s/kg").to_string() == u.m / u.s / u.kg
```
pllim: Is this removal on purpose? It sounds like the multiple calls were needed for a test named `test_multiple_solidus` but I am just guessing.

mhvk: It just duplicates the next test, but without a `catch_warnings`

astropy/units/tests/test_format.py
mhvk: Clarified:
```
    # Yikes, this becomes "femto-urlong"...  But at least there's a warning.
```

mhvk:
@pllim - thanks for the comments. Addressed. I did `skip ci` since the previous run worked (well, apart from a false-positive from coverage)

pllim:
Thanks, @mhvk !

PR #8857, opened 2019-06-16
Added property to BaseCoordinateFrame for cylindrical representation

ayshih:
This PR fills in the omission where `BaseCoordinateFrame` has the shorthand-for-representation properties `.cartesian`, `.spherical`, and `.sphericalcoslat`, but no analogous `.cylindrical`.  I copied the TODO message from the other properties.

PR #9008, opened 2019-07-17
Coordinate matching fails for unit-less distances

adrn:
This is a revival of #8029, now with a rebase and changelog entry.

The text from @eteq:

> 
> The Problem this PR is trying to solve is most easily shown in this snippet:
> ```python
> from astropy.coordinates import ICRS, CartesianRepresentation, match_coordinates_sky
> i1 = ICRS([[1], [2], [3]], representation_type=CartesianRepresentation)
> i2 = ICRS([[1], [2], [4, 5]], representation_type=CartesianRepresentation)
> i, sep, sep3d = match_coordinates_sky(i1, i2)
> ```
> yields ``astropy.units.core.UnitTypeError: Distance instances require units equivalent to 'm', so cannot set it to ''.``.  
> 
> The problem is: it seems counter-intuitive in that it fails only in `match_coordinates_sky` and not in the `ICRS` constructor or similar. The root cause is really `separation_3d`, which fails if the inputs are dimensionless because the "3d" part of coordinates matching uses `Distance`, which is supposed to be, well, a distance (and have appropriate units).  Given we are now allowing dimensionless distances in the frame classes, arguably we should allow it in `separation_3d`... But that's sort of an API change (albiet a "relaxing" one) - should we allow this "3d distance" to sometimes be dimensionless?
> 
> One could argue this is a bug... but I'm not sure it is, so I'm not sure if this is a bugfix, or new feature, or "shouldn't do", so have not yet flagged it bug, milestoned it, or changelogged it.
> 
> Also marking as a WIP because there's question above whether we really want to do this or not. The code itself I think is done, but the "discussion" part is not.

cc @mhvk @eteq

### DIFF
--- a/astropy/coordinates/tests/test_frames.py
+++ b/astropy/coordinates/tests/test_frames.py
@@ -509,7 +509,7 @@ def test_sep():
     i7 = ICRS(ra=1*u.deg, dec=2*u.deg, distance=3*u.one)
     i8 = ICRS(ra=1*u.deg, dec=2*u.deg, distance=4*u.one)
     sep3d = i7.separation_3d(i8)
-    assert_allclose(sep3d.to(u.one).value, [1])
+    assert_allclose(sep3d, 1*u.one)
 
     # but should fail with non-dimensionless
     with pytest.raises(ValueError):
--- a/astropy/coordinates/tests/test_frames.py
+++ b/astropy/coordinates/tests/test_frames.py
@@ -505,6 +505,16 @@ def test_sep():
     sep3d = i5.separation_3d(i6)
     assert_allclose(sep3d.to(u.kpc), np.array([2, 2])*u.kpc)
 
+    # 3d separations of dimensionless distances should still work
+    i7 = ICRS(ra=1*u.deg, dec=2*u.deg, distance=3*u.one)
+    i8 = ICRS(ra=1*u.deg, dec=2*u.deg, distance=4*u.one)
+    sep3d = i7.separation_3d(i8)
+    assert_allclose(sep3d.to(u.one).value, [1])
+
+    # but should fail with non-dimensionless
+    with pytest.raises(ValueError):
+        i7.separation_3d(i3)
+
 
 def test_time_inputs():
     """
--- a/astropy/coordinates/tests/test_matching.py
+++ b/astropy/coordinates/tests/test_matching.py
@@ -30,7 +30,7 @@ else:
     OLDER_SCIPY = True
 
 
-@pytest.mark.skipif('not HAS_SCIPY')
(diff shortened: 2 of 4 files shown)


astropy/coordinates/tests/test_matching.py
```diff
@@ -297,3 +297,14 @@ def test_match_catalog_empty():
     with pytest.raises(ValueError) as excinfo:
         sc1.match_to_catalog_3d(cat0)
     assert 'catalog' in str(excinfo.value)
+
```
mhvk: Need extra blank line

astropy/coordinates/tests/test_frames.py
```diff
@@ -505,6 +505,16 @@ def test_sep():
     sep3d = i5.separation_3d(i6)
     assert_allclose(sep3d.to(u.kpc), np.array([2, 2])*u.kpc)
 
+    # 3d separations of dimensionless distances should still work
+    i7 = ICRS(ra=1*u.deg, dec=2*u.deg, distance=3*u.one)
+    i8 = ICRS(ra=1*u.deg, dec=2*u.deg, distance=4*u.one)
+    sep3d = i7.separation_3d(i8)
+    assert_allclose(sep3d.to(u.one).value, [1])
```
mhvk: Here, `assert_allclose` is just`assert_quantity_allclose`, so you can do `assert_allclose(sep3d, 1*u.one)`.

astropy/coordinates/tests/test_matching.py
```diff
@@ -297,3 +297,14 @@ def test_match_catalog_empty():
...
+@pytest.mark.skipif(str('not HAS_SCIPY'))
```
mhvk:
```
@pytest.mark.skipif(not HAS_SCIPY or OLDER_SCIPY, reason="needs scipy>=???")
```

adrn: Yea, not sure why this was a string - I just took the commits from @eteq's branch

adrn:
> Beyond those, could you add a test for frames/coords without distances? (i.e., implicit distances of 1; that would seem he more likely usecase.)

@mhvk I think there are tests for this already in `test_matching_function()` (test_matching.py)?

mhvk:
Something odd about the test failure - perhaps an incorrect rebase after cylindrical was added? Also note the PEP8 failure

pllim:
`master` is currently broken by #8857. I think it is better to fix `master` first before proceeding here.

pllim:
@adrn , please rebase again and see if CI will turn green.

adrn:
Alrighty, let's see!

PR #9013, opened 2019-07-17
Add roundtrip test for user-defined unit in ecsv.

mhvk:
This already worked but was not tested. Based on the example in #8897.

@taldcroft - as you can see from the test, if a unit is not recognized, the whole parsing fails. I wonder if instead we should just let it be an `UnrecognizedUnit`? With the logic that one should be generous on input, strict on output.

If you agree, I can change it, either as part of this PR, or in a separate one.

### DIFF
--- a/astropy/io/ascii/tests/test_ecsv.py
+++ b/astropy/io/ascii/tests/test_ecsv.py
@@ -497,3 +497,45 @@ def test_round_trip_masked_table_serialize_mask(tmpdir):
         t[name].mask = False
         t2[name].mask = False
         assert np.all(t2[name] == t[name])
+
+
+@pytest.mark.skipif('not HAS_YAML')
+@pytest.mark.parametrize('table_cls', (Table, QTable))
+def test_round_trip_user_defined_unit(table_cls, tmpdir):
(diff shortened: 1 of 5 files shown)


mhvk:
p.s. Using `UnrecognizedUnit` is a bit closer to what we do for FITS.

taldcroft:
@mhvk - sure, `UnrecognizedUnit` sounds reasonable.

taldcroft:
I thought in #8897 that the issue came up with a `write` then `read` then `write` again sequence.  Does that need testing here for regression?

mhvk:
@taldcroft - OK, `.ecsv` (and all yaml header readers) now use `u.UnrecognizedUnit` as appropriate.  I also added a tip on how to get a real unit in the warning, as you suggested in #9015. Overall, this makes it much better!

Note that the test now is for both `Table` and `QTable`. The difference is that `Column` does not warn for an unrecognized unit.

mhvk:
Note: `mpldev` failure is unrelated; see #9031

taldcroft:
Apparently I can't merge on a phone with failing tests.  Go for it @mhvk.

PR #9015, opened 2019-07-18
Ensure custom units can round-trip in FITS

### LINKED ISSUE #8897: Writing quantity with user-defined unit to fits table breaks if unit is registered

mhvk:
Fixes #8897 

Currently, custom units sort-of round-trip through fits, but they do not compare equal, since the re-read version is an `UnrecognizedUnit`. This PR fixes it for astropy-native reading, by giving the serialized unit precedence over that stored in the FITS column. I'm not at all sure this is the best solution...

Like for `ecsv` in #9013, I'm also uncomfortable with reading failing if the unit is not locally enabled - it would seem better to warn and return an `UnrecognizedUnit`.

### DIFF
--- a/astropy/table/serialize.py
+++ b/astropy/table/serialize.py
@@ -74,13 +74,13 @@ def _represent_mixin_as_column(col, name, new_cols, mixin_cols,
     # - description: DO store
     # - meta: DO store
     info = {}
-    for attr, nontrivial, xform in (('unit', lambda x: x is not None and x != '', str),
-                                    ('format', lambda x: x is not None, None),
-                                    ('description', lambda x: x is not None, None),
-                                    ('meta', lambda x: x, None)):
+    for attr, nontrivial in (('unit', lambda x: x is not None and x != ''),
+                             ('format', lambda x: x is not None),
+                             ('description', lambda x: x is not None),
+                             ('meta', lambda x: x)):
         col_attr = getattr(col.info, attr)
         if nontrivial(col_attr):
-            info[attr] = xform(col_attr) if xform else col_attr
+            info[attr] = col_attr
 
     data_attrs = [key for key, value in obj_attrs.items() if
                   getattr(value, 'shape', ())[:1] == col.shape[:1]]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -382,6 +382,10 @@ astropy.table
 - Fix bug where adding a column consisting of a list of masked arrays was
   dropping the masks. [#9048]
 
+- ``Quantity`` columns with custom units can now round-trip via FITS tables,
+  as long as the custom unit is enabled during reading (otherwise, the unit
+  will become an ``UnrecognizedUnit``). [#9015]
+
 astropy.tests
 ^^^^^^^^^^^^^
 
--- a/astropy/io/fits/connect.py
+++ b/astropy/io/fits/connect.py
@@ -101,7 +101,7 @@ def _decode_mixins(tbl):
     # Use the `datatype` attribute info to update column attributes that are
     # NOT already handled via standard FITS column keys (name, dtype, unit).
     for col in info['datatype']:
-        for attr in ['description', 'meta']:
+        for attr in ['description', 'meta', 'unit']:
             if attr in col:
                 setattr(tbl[col['name']].info, attr, col[attr])
 
--- a/astropy/io/fits/connect.py
+++ b/astropy/io/fits/connect.py
@@ -101,7 +101,7 @@ def _decode_mixins(tbl):
     # Use the `datatype` attribute info to update column attributes that are
     # NOT already handled via standard FITS column keys (name, dtype, unit).
     for col in info['datatype']:
-        for attr in ['description', 'meta', 'unit']:
+        for attr in ['description', 'meta']:
             if attr in col:
                 setattr(tbl[col['name']].info, attr, col[attr])
 
--- a/astropy/io/fits/convenience.py
+++ b/astropy/io/fits/convenience.py
@@ -525,7 +525,10 @@ def table_to_hdu(table, character_as_bytes=False):
                     "the data or change the units.".format(col.name, str(scale)))
             except ValueError:
                 warnings.warn(
-                    "The unit '{}' could not be saved to FITS format".format(
+                    "The unit '{}' could not be saved in native FITS format "
+                    "and hence will be lost to non-astropy fits readers. "
+                    "Within astropy, it can roundtrip using QTable, but one "
+                    "has to ensure to enable to unit before reading.".format(
                         unit.to_string()), AstropyUserWarning)
             else:
                 # Try creating a Unit to issue a warning if the unit is not
--- a/astropy/io/fits/convenience.py
+++ b/astropy/io/fits/convenience.py
@@ -524,12 +524,26 @@ def table_to_hdu(table, character_as_bytes=False):
                     "is not recognized by the FITS standard. Either scale "
                     "the data or change the units.".format(col.name, str(scale)))
             except ValueError:
-                warnings.warn(
+                # Warn that the unit is lost, but let the details depend on
+                # whether the column was serialized (because it was a
+                # quantity), since then the unit can be recovered by astropy.
+                warning = (
                     "The unit '{}' could not be saved in native FITS format "
-                    "and hence will be lost to non-astropy fits readers. "
-                    "Within astropy, it can roundtrip using QTable, but one "
-                    "has to ensure to enable to unit before reading.".format(
-                        unit.to_string()), AstropyUserWarning)
+                    .format(unit.to_string()))
+                if any('SerializedColumn' in item and 'name: '+col.name in item
+                       for item in table.meta.get('comments', [])):
+                    warning += (
+                        "and hence will be lost to non-astropy fits readers. "
+                        "Within astropy, the unit can roundtrip using QTable, "
+                        "though one has to enable the unit before reading.")
+                else:
+                    warning += (
+                        "and cannot be recovered in reading. If pyyaml is "
+                        "installed, it can roundtrip within astropy by "
+                        "using QTable both to write and read back, "
+                        "though one has to enable the unit before reading.")
+                warnings.warn(warning, AstropyUserWarning)
+
             else:
                 # Try creating a Unit to issue a warning if the unit is not
                 # FITS compliant
--- a/astropy/io/fits/tests/test_connect.py
+++ b/astropy/io/fits/tests/test_connect.py
@@ -209,6 +209,24 @@ class TestSingleTable:
         t = Table.read(hdu)
         assert equal_data(t, self.data)
 
+    @pytest.mark.parametrize('table_type', (Table, QTable))
+    def test_write_drop_nonstandard_units(self, table_type, tmpdir):
+        # While we are generous on input (see above), we are strict on
+        # output, dropping units not recognized by the fits standard.
+        filename = str(tmpdir.join('test_nonstandard_units.fits'))
+        spam = u.def_unit('spam')
+        t = table_type()
+        t['a'] = [1., 2., 3.] * spam
+        with catch_warnings() as w:
+            t.write(filename)
+        assert len(w) == 1
+        assert 'spam' in str(w[0].message)
+        with fits.open(filename) as ff:
+            hdu = ff[1]
+            assert 'TUNIT1' not in hdu.header
+            t2 = table_type.read(hdu, astropy_native=False)
+        assert t2['a'].unit is None
+
     def test_memmap(self, tmpdir):
         filename = str(tmpdir.join('test_simple.fts'))
         t1 = Table(self.data)
--- a/astropy/io/fits/tests/test_connect.py
+++ b/astropy/io/fits/tests/test_connect.py
@@ -222,6 +222,12 @@ class TestSingleTable:
             t.write(filename)
         assert len(w) == 1
         assert 'spam' in str(w[0].message)
+        if table_type is Table or not HAS_YAML:
+            assert ('cannot be recovered in reading. '
+                    'If pyyaml is installed') in str(w[0].message)
+        else:
+            assert 'lost to non-astropy fits readers' in str(w[0].message)
+
         with fits.open(filename) as ff:
             hdu = ff[1]
             assert 'TUNIT1' not in hdu.header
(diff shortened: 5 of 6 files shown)


taldcroft:
:+1: in general on being more liberal with reading and never outright failing on units.  Even better if a nice warning message gives a good hint on the magic wrapping code to make it work for the user.  E.g.
```
Hint: to enable this unit use:
>>> import astropy.units as u
>>> u.add_enabled_units(u.def_unit('bandpass_sol_lum')) 
```

pllim:
Unfortunately, it failed in PYTHON=3.6 and NUMPY=1.13 job:
```
________________ TestSingleTable.test_with_custom_units_qtable _________________
self = <astropy.io.fits.tests.test_connect.TestSingleTable object at 0x7fa18166d748>
tmpdir = local('/tmp/pytest-of-travis/pytest-0/test_with_custom_units_qtable0')
    def test_with_custom_units_qtable(self, tmpdir):
        # Test only for QTable - Column cannot handle new units.
        filename = str(tmpdir.join('test_with_units.fits'))
        bandpass_sol_lum = u.def_unit('bandpass_sol_lum')
        t1 = QTable()
        t1['l'] = np.ones(5) * bandpass_sol_lum
        with catch_warnings() as w:
            t1.write(filename, overwrite=True)
        assert len(w) == 1
>       assert "'bandpass_sol_lum' did not parse" in str(w[0].message)
E       assert "'bandpass_sol_lum' did not parse" in "The unit 'bandpass_sol_lum' could not be saved to FITS format"
E        +  where "The unit 'bandpass_sol_lum' could not be saved to FITS format" = str(AstropyUserWarning("The unit 'bandpass_sol_lum' could not be saved to FITS format",))
E        +    where AstropyUserWarning("The unit 'bandpass_sol_lum' could not be saved to FITS format",) = <warnings.WarningMessage object at 0x7fa187a01390>.message
astropy/io/fits/tests/test_connect.py:126: AssertionError
```

*"I should not have touched FITS," said a wise person once.*

mhvk:
Rebased on #9013 to turn parsing error into warning, an adjusted the test to be less strict about the specific warning (which is somewhat mysterious).

mhvk:
@taldcroft - on to FITS...

As you can see, I currently "solve" the issue by letting units stored in the yaml header take precedence over those of the fits `TUNIT` header. Not sure if that is best. While looking into this I also noticed a strange inconsistency between `Table` and `QTable` writing (which I do not touch, and is independent of whether a unit is enabled; see below): `Table/Column` just drops the new unit, while `QTable/Quantity` write it to `TUNIT1` (and of course serialize). I think for `TUNIT` the behaviour should be consistent, and arguably the unit should be dropped for `Quantity` as well (which would make then using the serialised unit more logical). What do you think?

```
from astropy.table import Table, QTable
from astropy.io import fits
unit = u.def_unit('bandpass_sol_lum')
q = [1, 2, 3] * unit
QTable([q]).write('qt.fits')
# WARNING: UnitsWarning: 'bandpass_sol_lum' did not parse as fits unit: At col 0, Unit 'bandpass_sol_lum' not supported by the FITS standard.  If this is meant to be a custom unit, define it with 'u.def_unit'. To have it recognized inside a file reader or other code, enable it with 'u.add_enabled_units'. For details, see http://docs.astropy.org/en/latest/units/combining_and_defining.html [astropy.units.core]
Table([q]).write('t.fits')
# WARNING: The unit 'bandpass_sol_lum' could not be saved to FITS format [astropy.io.fits.convenience]

fits.open('t.fits')[1].header

XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                    8 / length of dimension 1                          
NAXIS2  =                    3 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
TFIELDS =                    1 / number of table fields                         
TTYPE1  = 'col0    '                                                            
TFORM1  = 'D       '                                                            

fits.open('qt.fits')[1].header

XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                    8 / length of dimension 1                          
NAXIS2  =                    3 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
TFIELDS =                    1 / number of table fields                         
TTYPE1  = 'col0    '                                                            
TFORM1  = 'D       '                                                            
TUNIT1  = 'bandpass_sol_lum'                                                    
COMMENT --BEGIN-ASTROPY-SERIALIZED-COLUMNS--                                    
COMMENT datatype:                                                               
COMMENT - {name: col0, unit: bandpass_sol_lum, datatype: float64}               
COMMENT meta:                                                                   
COMMENT   __serialized_columns__:                                               
COMMENT     col0:                                                               
COMMENT       __class__: astropy.units.quantity.Quantity                        
COMMENT       unit: !astropy.units.Unit {unit: bandpass_sol_lum}                
COMMENT       value: !astropy.table.SerializedColumn {name: col0}               
COMMENT --END-ASTROPY-SERIALIZED-COLUMNS--                                      
```

mhvk:
Actually, turns out there is a small difference whether or not a unit is enabled, that the write operation gives exactly the same warning for both cases,
```
# WARNING: The unit 'bandpass_sol_lum' could not be saved to FITS format [astropy.io.fits.convenience]
```

taldcroft:
How about this behavior (for both Table and QTable) for writing a unit not allowed by the standard:
- Drop unit entirely if it is not enabled.  Amend the current warning *"The unit 'bandpass_sol_lum' could not be saved to FITS format"* to add text about how to actually write out the unit by enabling it.
- Put unit into serialized metadata if it is enabled.  Warn that a unit which is not allowed by the FITS standard is being stored in the table in astropy-specific metadata to allow round-trip of the table.

In this way it is always possible to write the desired units to FITS, but it is necessary to opt in and then suffer a warning.  This is being somewhat strict on output, but allowing the astropy extension of FITS as a more arbitrary container for serializing objects.

mhvk:
I'm definitely OK with never writing an invalid FITS unit to `TUNIT`. My sense, though, is that it is OK to serialize without enabling it, but that the warning should tell that one can only read it back in astropy, and that even there one has to pre-define and enable the unit.

But I think it is not quite possible to do this for `Column` unless we also serialize it. Or is that what you were suggesting?

taldcroft:
> My sense, though, is that it is OK to serialize without enabling it, but that the warning should tell that one can only read it back in astropy, and that even there one has to pre-define and enable the unit.

I'm OK with that.

> But I think it is not quite possible to do this for Column unless we also serialize it. Or is that what you were suggesting?

I would say to not spend effort making Column with units work better.  One way to slowly deprecate Column with units (in favor of using proper Quantity with QTable) is by not fixing unit issues with Column.

mhvk:
@taldcroft - I think I managed to get the right behaviour for unit storing now, and attempted to clarify the warning, but am not sure that part is complete. It is slightly tricky since at the location where the warnings are raised, the code doesn't really know whether a `Quantity` or a `Column` is being dealt with.

astropy/table/serialize.py
```diff
@@ -75,13 +75,13 @@ def _represent_mixin_as_column(col, name, new_cols, mixin_cols,
...
-                                    ('description', lambda x: x is not None, None),
-                                    ('meta', lambda x: x, None)):
+    for attr, nontrivial in (('unit', lambda x: x is not None and x != ''),
+                             ('format', lambda x: x is not None),
+                             ('description', lambda x: x is not None),
+                             ('meta', lambda x: x)):
         col_attr = getattr(col.info, attr)
         if nontrivial(col_attr):
-            info[attr] = xform(col_attr) if xform else col_attr
+            info[attr] = col_attr
```
mhvk: In the end, the main thing was not to convert the unit to a string prematurely - this then later changed a user-defined unit into `UnrecognizedUnit` (in the creation of a `Column(..., **info)`, which was undesired. And it is not needed for the `yaml` storing.

taldcroft:
>  is slightly tricky since at the location where the warnings are raised, the code doesn't really know whether a Quantity or a Column is being dealt with.

Indeed it is not quite right with `Column`.  The warning says the unit is recoverable by reading with Quantity with the unit enabled, but instead it is dropped entirely:
```
In [35]: t = Table()
In [36]: t['l'] = np.ones(3) * unit
In [37]: t.write(filename, overwrite=True)
WARNING: The unit 'bandpass_sol_lum' could not be saved in native FITS format and hence will be lost to non-astropy fits readers. Within astropy, it can roundtrip using QTable, but one has to ensure to enable to unit before reading. [astropy.io.fits.convenience]

In [38]: with fits.open(filename) as ff:
    ...:     print('\n'.join(f'{k}: {v}' for k, v in ff[1].header.items()))
    ...: 
XTENSION: BINTABLE
BITPIX: 8
NAXIS: 2
NAXIS1: 8
NAXIS2: 3
PCOUNT: 0
GCOUNT: 1
TFIELDS: 1
TTYPE1: l
TFORM1: D
```
Compare:
```
In [39]: t = QTable()

In [40]: t['l'] = np.ones(3) * unit

In [41]: t.write(filename, overwrite=True)
WARNING: The unit 'bandpass_sol_lum' could not be saved in native FITS format and hence will be lost to non-astropy fits readers. Within astropy, it can roundtrip using QTable, but one has to ensure to enable to unit before reading. [astropy.io.fits.convenience]

In [42]: with fits.open(filename) as ff:
    ...:     print('\n'.join(f'{k}: {v}' for k, v in ff[1].header.items()))
    ...: 
XTENSION: BINTABLE
BITPIX: 8
NAXIS: 2
NAXIS1: 8
NAXIS2: 3
PCOUNT: 0
GCOUNT: 1
TFIELDS: 1
TTYPE1: l
TFORM1: D
COMMENT: --BEGIN-ASTROPY-SERIALIZED-COLUMNS--
COMMENT: datatype:
COMMENT: - {name: l, unit: bandpass_sol_lum, datatype: float64}
COMMENT: meta:
COMMENT:   __serialized_columns__:
COMMENT:     l:
COMMENT:       __class__: astropy.units.quantity.Quantity
COMMENT:       unit: !astropy.units.Unit {unit: bandpass_sol_lum}
COMMENT:       value: !astropy.table.SerializedColumn {name: l}
COMMENT: --END-ASTROPY-SERIALIZED-COLUMNS--
```

taldcroft:
> is slightly tricky since at the location where the warnings are raised, the code doesn't really know whether a Quantity or a Column is being dealt with.

I'm a bit confused by that statement.  At the point when that particular warning is issued, isn't `table[col.name]` the original table column in question?  I put in a debug statement there and it certainly *looks* like the table column, so you could check the type there.  If `Column` subclass then just warn that the non-compliant unit has been dropped entirely and that they should use `QTable` for the original table instead of `Table` if they want to round trip within astropy.

mhvk:
In `convenience`, any `QTable` quantities are already converted to `Column`. But one can still tell that it was a `Quantity` because it is stored in `table.meta['comments']` - so I look there now (see last commit).

mhvk:
@taldcroft - tests now pass. Is the change in warning message OK?

Issue #9003, opened 2019-07-16
Cannot read some non-standard FITS files


aplavin:
I'm trying to read a fits file and get its header and contents, but no matter what I do `astropy` gives me errors:

```python
with fits.open(Path('J0721+7120_C_2014_02_18_pet_vis.fits').open('rb')) as f:
    f['PRIMARY'].header['RA']

---------------------------------------------------------------------------
VerifyError                               Traceback (most recent call last)
<ipython-input-44-e7947a571936> in <module>
      1 with fits.open(Path('J0721+7120_C_2014_02_18_pet_vis.fits').open('rb')) as f:
----> 2     f['PRIMARY'].header['RA']

~/anaconda3/lib/python3.7/site-packages/astropy/io/fits/header.py in __getitem__(self, key)
    152             return card.rawvalue
    153 
--> 154         value = card.value
    155         if value == UNDEFINED:
    156             return None

~/anaconda3/lib/python3.7/site-packages/astropy/io/fits/card.py in value(self)
    283             value = self._value
    284         elif self._valuestring is not None or self._image:
--> 285             value = self._value = self._parse_value()
    286         else:
    287             if self._keyword == '':

~/anaconda3/lib/python3.7/site-packages/astropy/io/fits/card.py in _parse_value(self)
    765         if m is None:
    766             raise VerifyError("Unparsable card ({}), fix it first with "
--> 767                               ".verify('fix').".format(self.keyword))
    768 
    769         if m.group('bool') is not None:

VerifyError: Unparsable card (RA), fix it first with .verify('fix').
```

```python
with fits.open(Path('J0721+7120_C_2014_02_18_pet_vis.fits').open('rb')) as f:
    f.verify('fix')

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-43-ed0088ff4893> in <module>
      1 with fits.open(Path('J0721+7120_C_2014_02_18_pet_vis.fits').open('rb')) as f:
----> 2     f.verify('fix')

~/anaconda3/lib/python3.7/site-packages/astropy/io/fits/verify.py in verify(self, option)
     71             return
     72 
---> 73         errs = self._verify(opt)
     74 
     75         # Break the verify option into separate options related to reporting of

~/anaconda3/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py in _verify(self, option)
   1191 
   1192         # the first (0th) element must be a primary HDU
[...]

PR #9021, opened 2019-07-18
Remove continuation slash from FITS test

pllim:
Direct follow up of #9003 , as suggested by @mhvk . Given that I am touching FITS, I better let the full CI run, just in case... :crossed_fingers:


saimn:
Thanks @pllim :), given the change I think that there is no need to wait for the CI builds ;).

PR #9023, opened 2019-07-19
Fixed breaking of master by #8857

ayshih:
The merging of #8857 broke master because it hadn't been updated to handle the changes in pytest 5 (see #8949).  This PR fixes the issue.


pllim:
Thanks, @ayshih . Unfortunately, you were just a little too late as there is already a fix by @adrn  in #9022. Thanks anyway. Much appreciated!

PR #9039, opened 2019-07-23
Fix issue with data path that caused crash on Windows

astrofrog:
Fixes https://github.com/astropy/astropy/issues/8903

Not sure how I missed this before but I think this would happen on all platforms (It's just that on Windows long paths cause errors). I think these changes aren't in v2.0.x so no need to backport to LTS though. To reproduce the issue, just run ``python setup.py build`` several times repeatedly and noticed the nested build folders.

@MSeifert04 - feel free to try this out!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -456,6 +456,10 @@ astropy.wcs
 Other Changes and Additions
 ---------------------------
 
+- Fixed a bug that caused files outside of the astropy module directory to be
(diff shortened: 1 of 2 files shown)


MSeifert04:
Nice. Thank you @astrofrog for the fix and @pllim for the detective work 👍 

Works fine on my windows machine!

PR #9041, opened 2019-07-25
Accepting mask when creating new column from a list of masked arrays

### LINKED ISSUE #8977: Mask discarded when creating new column from a list of masked arrays
Populating a new column in a masked table using a list of masked arrays results in the mask being discarded:

```pycon
>>> import numpy as np
>>> t = Table([{'a': np.arange(3)} for _ in range(10)], masked=True)
>>> t['b'] = [np.ma.masked_all(3) for _ in range(10)]
>>> t
<Table masked=True length=10>
a [3]                     b [3]                   
int64                    float64                  
------ -------------------------------------------
0 .. 2                               0.0 .. 1e-323
0 .. 2                               0.0 .. 1e-323
0 .. 2                               0.0 .. 1e-323
0 .. 2                               0.0 .. 1e-323
0 .. 2                               0.0 .. 1e-323
0 .. 2                               0.0 .. 1e-323
0 .. 2                               0.0 .. 1e-323
0 .. 2            3.105036184601418e+231 .. 1e-323
0 .. 2 3.105036184601418e+231 .. 2.2363587243e-314
0 .. 2  3.105036184601418e+231 .. 2.144207845e-314
```

On the other hand, if we cast the list to a masked array, the list is preserved:

```pycon
>>> t = Table([{'a': np.arange(3)} for _ in range(10)], masked=True)
>>> t['b'] = np.ma.asarray([np.ma.masked_all(3) for _ in range(10)])
>>> t
<Table masked=True length=10>
a [3]   b [3]  
int64  float64 
------ --------
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
```

The mask should be preserved in both cases.

himanshupathak21061998:
Try to Solve: #8977 
Now  MaskedColumn can be created using a list of maksed_array.

```
In [3]: t['b'] =[np.ma.masked_all(3) for _ in range(10)]

In [4]: t
Out[4]: 
<Table masked=True length=10>
a [3]   b [3]  
int64  float64 
------ --------
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
0 .. 2 -- .. --
```

EDIT: Fix #8977

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -1782,6 +1782,10 @@ class Table:
 
         # If value doesn't have a dtype and won't be added as a mixin then
         # convert to a numpy array.
+        # See #8977 Using Masked array for working.
+        if type(col) == type([]) and any(isinstance(x, np.ma.MaskedArray) for x in col):
+            col = np.ma.asarray(col)
+
         if not hasattr(col, 'dtype') and not self._is_mixin_for_table(col):
             col = np.asarray(col)
 
--- a/astropy/table/tests/test_table.py
+++ b/astropy/table/tests/test_table.py
@@ -2403,3 +2403,13 @@ def test_data_to_col_convert_strategy():
     t['b'] = np.int64(2)  # Failed previously
     assert np.all(t['a'] == [1, 1])
     assert np.all(t['b'] == [2, 2])
+
+
+def test_mask_column_from_list():
+    """Test for #8977 Initialising masked column from a list
+    """
+    t = table.Table([(0, 1)], names='a', masked=True)
+    data = np.ma.masked_greater(np.arange(3), 1)
+    t['b'] = [data, data]
+    assert np.all(t['b'].mask == np.array([[False, False, True],
+                                          [False, False, True]]))


pllim:
Thanks! I milestoned this to 4.0 since the issue itself was milestoned so. I'll let @taldcroft review this and decide if a change log is necessary.

cc @lpsinger

himanshupathak21061998:
I think, 4.0 is right. Maybe changelog update is not required for this one. you're welcome @pllim

taldcroft:
@himanshupathak21061998 - thanks for taking a crack at this.  Your solution addresses the problem in #8977, but I realized that it goes deeper because initializing a table from the list of masked arrays is also failing.  I had to do a more core change in #9048 to fix both problems at once.  I would have made suggestions to you but I really had no idea how to do this without doing it because the changes are slightly subtle.

himanshupathak21061998:
Yes, #9048 is a better solution for this. I closed this PR.

PR #9049, opened 2019-07-26
Fix unresolved link

hamogu:
Sphinx does not resolve the link as given in the docs currently.
The easiest solution is spell out te entire path of the function
in question because I think `~astrop.io.fits.open` which will display just
"open" in the rendered document is too confusing.

[skip CI]


pllim:
p.p.s. In the future, for docs changes, we would still want to run the doc build job, so please use `[docs only]` directive or a combo of `[skip appveyor] [skip travis]`. For this time, it ran anyway, so no harm done.

PR #9054, opened 2019-07-26
MNT: Flip primary and mirror locations for data

pllim:
This flips the primary and mirror locations for Astropy data. cc @jhunkeler and @rendinam

I don't think change log is required?


astrofrog:
It might be better to simply flip the data domain to point to GitHub?

pllim:
If you end up changing the domain thingy, then feel free to close without merge!

astrofrog:
For the record, data.astropy.org now points at the gh-pages mirror

PR #9087, opened 2019-08-02
Converting remainder of multi-line test match expressions to raw strings

dhomeier:
Follow-up to #8997 completing the change of `pytest.raise` and `pytest.warns` `match` arguments to raw in multi-line strings. Should cover every instance of `match=` found in tests.


dhomeier:
Looks like a simple escape of the dot (unlike `?`?) is accepted by the CI, as in
`r"Keyword 'DP1\.AXIS\.3' not found"`; that would leave probably ~10 strings with unescaped special characters.

pllim:
@dhomeier , is there something else you want to add to this PR, or is it ready for merge? Thanks!

dhomeier:
Good to go from my side; looks like settling on a style policy how to deal with wildcard expressions etc. can take a bit more discussion, so that's it for now.

PR #9098, opened 2019-08-06
Fix backward-compatibility for ax.coords in WCSAxes

astrofrog:
In https://github.com/astropy/astropy/pull/8885, the behavior of ``ax.coords`` changed from returning all coordinates in the specified WCS to just the coordinates that were remaining after slicing the WCS, so this restores the original behavior.

Coordinates that will never be visible now have ``CoordinateHelper.coord_index`` set to ``None``, and I think I'm ok with breaking that backward-compatibility since I don't think anyone will likely be using that attribute. One open question is whether coordinates that will never be shown (because they are sliced out and uncorrelated) should be distinguishable in some other way from ones that are visible, e.g. by a different class from ``CoordinateHelper``? Maybe it's fine as-is (in the PR)?

This also:

* Fixes a bug introduced in #8885 that caused ``get_xlabel``/``get_ylabel`` to not work
* Implements a nice ``__repr__`` for ``ax.coords``
* Improves the error checking related to the ``slices`` argument and the dimensionality of the WCS

Note that I can't label this **Affects-dev** otherwise the bot complains about the presence of the changelog entries for the things that are new here and not just fixing #8885.

@Cadair - can you check whether this fixes SunPy? This fixes the tests for glue-core.

### DIFF
--- a/astropy/visualization/wcsaxes/wcsapi.py
+++ b/astropy/visualization/wcsaxes/wcsapi.py
@@ -20,7 +20,16 @@ IDENTITY.wcs.crpix = [1., 1.]
 IDENTITY.wcs.cdelt = [1., 1.]
 
 
-def transform_coord_meta_from_wcs(wcs, frame_class, aslice=None):
+def transform_coord_meta_from_wcs(wcs, frame_class, slices=None):
+
+    if wcs.pixel_n_dim > 2:
+        if slices is None:
+            raise ValueError("WCS has more than 2 pixel dimensions, so "
+                             "'slices' should be set")
+        elif len(slices) != wcs.pixel_n_dim:
+            raise ValueError("'slices' should have as many elements as WCS "
+                             "has pixel dimensions (should be {})"
+                             .format(wcs.pixel_n_dim))
 
     is_fits_wcs = isinstance(wcs, WCS)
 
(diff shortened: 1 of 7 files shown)


astropy/visualization/wcsaxes/wcsapi.py
```diff
@@ -100,6 +100,22 @@ def transform_coord_meta_from_wcs(wcs, frame_class, aslice=None):
...
     coord_meta['default_ticks_position'] = [''] * wcs.world_n_dim
 
+    invert_xy = False
+    if slices is not None:
+        wcs_slice = list(slices)
+        wcs_slice[wcs_slice.index("x")] = slice(None)
+        wcs_slice[wcs_slice.index("y")] = slice(None)
+        wcs = SlicedLowLevelWCS(wcs, wcs_slice[::-1])
+        invert_xy = slices.index('x') > slices.index('y')
+        world_keep = wcs._world_keep
```
astrofrog: @Cadair - this is a little ugly but I can't think of another way to do it. This is probably ok pragmatically since we are within the same overall package, but I wonder if we might want to somehow expose this information in a public API (maybe in a separate PR).

Cadair: yeah makes sense on both fronts.

PR #9109, opened 2019-08-09
Raise NotImplementedError in inverse

nden:
Modeling expects a `NotImplementedError` to be raised when an analytical inverse is not available. This tells the `CompoundModel` constructor not to attempt to create an analytical inverse for a compound model. The recently added `Tabular1D.inverse` raises a `ValueError` which breaks the model composition in some cases. Fixed here.


perrygreenfield:
Approved

pllim:
Tsk tsk, CI is broken and unrelated. Merging since approved. Thanks, everyone!

PR #9118, opened 2019-08-12
Allow for equal sized arrays to be added in nddata.utils.add_array

Onoddil:
This pull request makes a very minor change to `add_array` in `nddata.utils`, allowing for the possibility of `array_small` and `array_large` to be the same size. Instead of explicitly needing the larger array to be larger, now the smaller array cannot be smaller.

I am unsure of the reasoning behind `>` versus `>=` previously, but cannot see any reason why this logical flip is not allowed; please let me know if there is some reason `array_large` must always be larger than `array_small`.

This PR fixes astropy/photutils#818.

I don't think I am able to add labels, but I thought this change might be minor enough to get away with a `no-changelog-entry-needed` flag -- again, please let me know if it needs a changelog entry (and if I should add another commit to add one myself).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -460,6 +460,10 @@ bounding box. [#8799]
 astropy.nddata
 ^^^^^^^^^^^^^^
 
+- Fix to ``add_array``, which now accepts ``array_small`` having dimensions
+  equal to ``array_large``, instead of only allowing smaller sizes of
+  arrays. [#9118]
+
 astropy.samp
 ^^^^^^^^^^^^
 
--- a/astropy/nddata/tests/test_utils.py
+++ b/astropy/nddata/tests/test_utils.py
@@ -314,6 +314,21 @@ def test_add_array_even_shape():
     assert np.all(added_array == large_test_array_ref)
 
 
+def test_add_array_equal_shape():
+    """
+    Test add_array_2D utility function.
+
+    Test by adding an array of ones out of an array of zeros.
+    """
+    large_test_array = np.zeros((11, 11))
+    small_test_array = np.ones((11, 11))
+    large_test_array_ref = large_test_array.copy()
+    large_test_array_ref += small_test_array
+
+    added_array = add_array(large_test_array, small_test_array, (5, 5))
+    assert np.all(added_array == large_test_array_ref)
+
+
 @pytest.mark.parametrize(('position', 'subpixel_index'),
                          zip(test_positions, test_position_indices))
 def test_subpixel_indices(position, subpixel_index):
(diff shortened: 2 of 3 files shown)


pllim:
Thanks! I'll let the subpackage maintainer milestone this. When milestone is set, you will need to add a change log under "bug fix" for the appropriate milestone.

eteq:
One minor thought: perhaps while doing this you should add a sentence somewhere in the doctring that makes it clear that this is now the "intended" behavior?

(I'm not approving/reviewing this because I think it needs a sub-package maintainer to agree for sure it's a bug and not an API change, though.)

larrybradley:
I don't see why this behavior should be excluded.  Bug or API change, it definitely needs a changelog entry.  I'm inclined to call it a bug, although it seems like the original intent was that the large array be strictly larger.  The updated docstring should explicitly state that the "smaller" array can be the same size as the "larger" array.

This function has a long and nomadic history, and it wasn't easy to track down the original author.  It was originally added as `add_array_2D` in `photutils` by @adonath as part of his 2013 GSOC project on PSF/PRFs.  It went from there to the (now defunct) `imageutils` package, finally to the `astropy.nddata` subpackage.

@adonath - any comments?

Onoddil:
Thanks @larrybradley, I'll add a changelog as a bug fix for now, until told otherwise (either it being an API change, or the non-approval of the PR for whatever reason).

Seems slightly ironic that it came from there in the first place, as it's `photutils`'s `subtract_psf` that the issue arises in currently. The use case being times where you only really care about extracting a single source from an image, so you cutout the star in question (probably as a `EPSFStar` `NDData` object) and want to fit it, at which point you might as well fit the entire cutout (as you've deliberately cut out what you want to fit), but `subtract_psf` falls over due to needing a strictly larger cutout.

I'll update the docstring to have a sentence explicitly starting `array_small` can be equal to `array_large` as well while I'm doing the changelog.

Onoddil:
Hi @larrybradley I have updated the PR as requested. I have kept the change as a bug fix, and added a sentence explicitly saying equal size arrays are acceptable in the docstring. However I realised a milestone was never set, so the astropy-bot checks fail; I’ve put the changelog in `3.2.2` for now, but let me know if I should move it to `4.0` instead.

larrybradley:
Thanks, @Onoddil

adonath:
@Onoddil and @larrybradley Sorry, I was on vacation and could not respond earlier. I just wanted to confirm, that there was probably no good reason for the "strictly smaller" check. It was just the obvious choice for the use-case I had and seemed less error-prone, when I implemented this helper function. So 👍 for the changes in this PR...

PR #9142, opened 2019-08-19
DOC: Add modeling ref tags back

pllim:
This is a direct follow-up #9078 . I think I found where all the old tags should go, **except** this one (which seems to be listed under TODO, so hopefully you can add this remaining one back in when you get to it):

```rst
.. _modeling-getting-started-masked-data:

Fitting masked data
-------------------
```

cc @karllark

### DIFF
--- a/docs/modeling/basic-fitting.rst
+++ b/docs/modeling/basic-fitting.rst
@@ -3,6 +3,8 @@
 Basic Fitting
 =============
 
+.. _modeling-getting-started-1d-fitting:
+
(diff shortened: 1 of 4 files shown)


pllim:
I don't understand the `html-docs` failure...
```
astropy/visualization/wcsaxes/core.py:docstring of astropy.visualization.wcsaxes.WCSAxes:93:
Block quote ends without a blank line; unexpected unindent.
```

PR #9148, opened 2019-08-21
Replace (deprecated) optparse with argparse in fits scripts

MSeifert04:
The [`optparse` module](https://docs.python.org/library/optparse.html) has been deprecated for a while, so I thought it might be useful to transition these scripts to use the `argparse` module.

The functionality of the scripts should not differ (the automated tests are probably not exhaustive...), however the help text for the `fitsdiff` will now look slightly different. Not sure how relevant that is.

Also I'm not sure if that needs a changelog.

### DIFF
--- a/astropy/io/fits/scripts/fitscheck.py
+++ b/astropy/io/fits/scripts/fitscheck.py
@@ -40,10 +40,9 @@ Example uses of fitscheck:
 """
 
 
+import argparse
 import logging
-import optparse
 import sys
-import textwrap
 
 from astropy.tests.helper import catch_warnings
 from astropy.io import fits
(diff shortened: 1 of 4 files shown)


astropy/io/fits/scripts/fitscheck.py
```diff
@@ -40,72 +40,77 @@
...
-import textwrap
 
 from astropy.tests.helper import catch_warnings
 from astropy.io import fits
 
 
 log = logging.getLogger('fitscheck')
 
+_DESCRIPTION = """
+.e.g. fitscheck example.fits
```
pllim: What was the leading period for? Seems like bad grammar. Did I miss something?
```suggestion
e.g. fitscheck example.fits
```

MSeifert04: Hm... 

maybe to make it more symmetric? 🤔

Next time I'll be more careful when copy&pasting.

MSeifert04: I'll remove that once I know if I should add a changelog, okay?

pllim: I am leaning on no need for change log but I'll let Simon chime in. Sure, we can wait. And I don't think it is your fault, as the leading period was already there before but I just never noticed it until the diff highlighted it. :sweat_smile:

saimn:
Thanks for taking care of this @MSeifert04, optparse will probably never be removed but it is a good idea to make the scripts more consistent and use only one method to parse arguments.
About the changelog entry, it is probably a good idea to have one as the help strings will change a bit, so it allows to track when the change happen.

astropy/io/fits/scripts/fitscheck.py
saimn: I would completely remove this sentence, it is useless as it is redundant with the help formatted by optparse/argparse. This is for example the current one with optparse:
```
Usage: fitscheck [options] <.fits files...>

        .e.g. fitscheck example.fits

        Verifies and optionally re-writes the CHECKSUM and DATASUM keywords
        for a .fits file.
        Optionally detects and fixes FITS standard compliance problems.
```

MSeifert04: The output (especially the usage) with this PR is somewhat different:

```
$ fitscheck -h
usage: fitscheck [-h] [-k {standard,remove,none}] [-w] [-f] [-c] [-i] [-v]
                 file [file ...]

.e.g. fitscheck example.fits

Verifies and optionally re-writes the CHECKSUM and DATASUM keywords
for a .fits file.
Optionally detects and fixes FITS standard compliance problems.

positional arguments:
  file                  .fits files to process.

optional arguments:
  -h, --help            show this help message and exit
  -k {standard,remove,none}, --checksum {standard,remove,none}
                        Choose FITS checksum mode or none. Defaults standard.
  -w, --write           Write out file checksums and/or FITS compliance fixes.
  -f, --force           Do file update even if original checksum was bad.
  -c, --compliance      Do FITS compliance checking; fix if possible.
  -i, --ignore-missing  Ignore missing checksums.
  -v, --verbose         Generate extra output.
```

MSeifert04:
I added the changelog.

pllim:
@saimn , please approve if you are happy with this now. Thank you both!

saimn:
Looks good, thanks Michael!

PR #9150, opened 2019-08-21
More usage of Py_CLEAR in WCS C-extension

MSeifert04:
1.) Correct code like this:

```
PyObject *tmp;
tmp = instance->attribute;
instance->attribute = NULL;
Py_XDECREF(tmp);
```

Can be replaced by a shorter `Py_CLEAR(instance->attribute)`.

2.) Incorrect code that used

```
Py_XDECREF(instance->attribute);
instance->attribute = NULL;
```

Can lead to a segmentation fault because the deleted attribute could try to access the deleted attribute "instance->attribute" during the `__del__` invocation.

In this case the `Py_CLEAR(instance->attribute)` is not only shorter but also corrects for the rare case that there are such `__del__` implementations.

### DIFF
--- a/astropy/wcs/src/astropy_wcs.c
+++ b/astropy/wcs/src/astropy_wcs.c
@@ -52,31 +52,12 @@ static int
 Wcs_clear(
     Wcs* self) {
 
-  PyObject* tmp;
-
-  tmp = self->py_det2im[0];
-  self->py_det2im[0] = NULL;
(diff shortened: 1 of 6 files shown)


MSeifert04:
Sorry for the confusion, somehow I made the branch on origin instead of my fork 😓, see #9152 for the new PR.

pllim:
Ops...! :grimacing:

PR #9151, opened 2019-08-21
Fix possible buffer overflow with cunit or ctype

MSeifert04:
If the negative index is too large to wrap around then the `index` can be negative and thus the `array` is accessed out-of-bounds, leading to a buffer overflow.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -506,6 +506,9 @@ astropy.wcs
 
 - Fixed a crash while loading a WCS from headers containing duplicate SIP keywords. [#8893]
 
+- Fixed a possible buffer overflow when using too large negative indices for
+  ``cunit`` or ``ctype`` [#9151]
+
 
 Other Changes and Additions
 ---------------------------
--- a/astropy/wcs/src/str_list_proxy.c
+++ b/astropy/wcs/src/str_list_proxy.c
@@ -113,7 +113,7 @@ PyStrListProxy_getitem(
     PyStrListProxy* self,
     Py_ssize_t index) {
 
-  if (index >= self->size) {
+  if (index >= self->size || index < 0) {
     PyErr_SetString(PyExc_IndexError, "index out of range");
     return NULL;
   }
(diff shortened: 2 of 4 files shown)


pllim:
Interesting... So negative indexing doesn't work like it does in pure Python?

p.s. To save you a few clicks in the future, the bot now autolabels the subpackage (e.g., "wcs") from crawling the directories of changed files, thanks to @adrn and @astrofrog . :wink:

MSeifert04:
> Interesting... So negative indexing doesn't work like it does in pure Python?

Somewhat, for a size 2 collection -1 and -2 return valid indices 1 and 0. But for example -3 will become -1 in the method. It's all rather weird with the [`sq_ass_item`](https://docs.python.org/3/c-api/typeobj.html#c.PySequenceMethods.sq_ass_item).

astrofrog:
@MSeifert04 - can you add a test for the getter too? (at the moment only setting is tested?)

MSeifert04:
Of course, I added tests for the getters as well.

CHANGES.rst
```diff
@@ -506,6 +506,9 @@ astropy.wcs
 
 - Fixed a crash while loading a WCS from headers containing duplicate SIP keywords. [#8893]
 
+- Fixed a possible buffer overflow when using too large negative indices for
+  ``cunit`` or ``ctype`` [#9151]
```
astrofrog: Any negative indices, right? (not just too large ones?)

MSeifert04: No, negative indices `< -size` are allowed and work.

The `sq_ass_item` gets the index that is already wrapped around so the methods where I fixed that get a positive index as long as the user-provided index is within `-size <= index < size` (see [`PySequenceMethods.sq_ass_item`](https://docs.python.org/3/c-api/typeobj.html#c.PySequenceMethods.sq_ass_item)).

astrofrog: Ok perfect, thanks!

astrofrog:
This is pretty straightforward, so I'll go ahead and merge.

MSeifert04:
Thanks @astrofrog

PR #9152, opened 2019-08-21
Use Py_CLEAR in wcs C extension

MSeifert04:
Closes #9150

1.) Correct code like this:

```
PyObject *tmp;
tmp = instance->attribute;
instance->attribute = NULL;
Py_XDECREF(tmp);
```

Can be replaced by a shorter `Py_CLEAR(instance->attribute)`.

2.) Incorrect code that used

```
Py_XDECREF(instance->attribute);
instance->attribute = NULL;
```

Can lead to a segmentation fault because the deleted attribute could try to access the deleted attribute "instance->attribute" during the `__del__` invocation.

In this case the `Py_CLEAR(instance->attribute)` is not only shorter but also corrects for the rare case that there are such `__del__` implementations.

### DIFF
--- a/astropy/wcs/src/astropy_wcs.c
+++ b/astropy/wcs/src/astropy_wcs.c
@@ -52,31 +52,12 @@ static int
 Wcs_clear(
     Wcs* self) {
 
-  PyObject* tmp;
-
-  tmp = self->py_det2im[0];
-  self->py_det2im[0] = NULL;
-  Py_XDECREF(tmp);
-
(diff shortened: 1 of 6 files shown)


pllim:
Seems identical to #9150. Why was the other one closed? Just curious.

MSeifert04:
Because I couldn't figure out how to use my forks branch for the PR instead of the branch I accidentally created on the astropy repo. So I closed the other and opened this one. 😓

PR #9154, opened 2019-08-21
Fixed WCS.wcs.cunit equality

MSeifert04:
When the first item was the same the previous approach considered them equal and didn't check the following items.

Also removed the comparison of `unit_class` since that's just the `astropy.units.Unit` class and shouldn't be relevant for equality. Maybe that reference should not be kept in the instances but rather as static reference but that's something that can be addressed in the future and is not appropriate to change in a bug-fix commit.

### DIFF
--- a/astropy/wcs/src/unit_list_proxy.c
+++ b/astropy/wcs/src/unit_list_proxy.c
@@ -192,6 +192,8 @@ PyUnitListProxy_richcmp(
 	PyObject *a,
 	PyObject *b,
 	int op){
+  PyUnitListProxy *lhs, *rhs;
+  Py_ssize_t idx;
   assert(a != NULL && b != NULL);
   if (!PyObject_TypeCheck(a, &PyUnitListProxyType) ||
       !PyObject_TypeCheck(b, &PyUnitListProxyType)) {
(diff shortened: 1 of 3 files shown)


MSeifert04:
Just as an aside: I found this bug because of a compiler generated warning that looked very suspicious:

```
astropy/wcs/src/unit_list_proxy.c:210:34: warning: incompatible pointer types
      passing 'char (*)[72]' to parameter of type 'const char *'
      [-Wincompatible-pointer-types]
  equal = equal == 1 && !strncmp(lhs->array, rhs->array, ARRAYSIZE) && l...
                                 ^~~~~~~~~~
```

And I somehow overlooked this in #8480.

astropy/wcs/src/unit_list_proxy.c
```diff
@@ -201,18 +200,36 @@ PyUnitListProxy_richcmp(
   if (op != Py_EQ && op != Py_NE) {
     Py_RETURN_NOTIMPLEMENTED;
   }
-  lhs = (PyUnitListProxy *)a;
-  rhs = (PyUnitListProxy *)b;
-  int equal = PyObject_RichCompareBool(lhs->unit_class, rhs->unit_class, Py_EQ);
-  if (equal == -1) {
-    return NULL;  // Exception will be set because the rich-compare failed
+
+  {
```
pllim: That is this `{` tied to? Seems to be orphaned. Am I reading this right?

MSeifert04: Yep, it's orphaned. It's a so called "block scope" (it creates a new scope for variables). I used it here to separate the code dealing with the pre-condition checks and the actual comparison.

If that's confusing I can remove it again (and put the declarations at the top of the function).

pllim: If it is not strictly necessary, I would vote to remove it. Or at least state the intention of the block. Thanks!

MSeifert04: Okay 👍

MSeifert04:
Cc @astrofrog (original issue)

MSeifert04:
On my phone currently but i think the patch adding the eq was added in 3.2 so no need to backport?

MSeifert04:
Thank you for the review! :+1:

pllim:
#8480 was indeed milestoned for 3.2. Thanks!

nden:
@pllim I wish you could give me a chance to review before my merging WCS PRs.

pllim:
@nden , I deeply apologize and I take full responsibility!

MSeifert04:
@nden In case you have some comments or questions regarding the changes I'll still answer or address them. Just let me know :)

nden:
@MSeifert04 I'm sure you will :) . I just want to be aware of what is happening in the package without feeling I'm in a race. Thanks for the contributions!

PR #9155, opened 2019-08-22
Collate module index by module name ignoring `astropy.` suffix

lpsinger:
For example, `astropy.visualization` will go in the "V" section, not "A".


pllim:
The patch is simple enough that I don't think there is a reason *not* to backport all the way to LTS. But @bsipocz is free to move the milestone (again) if she disagrees.

This is uncontroversial enough, so merging. Thank you!

PR #9163, opened 2019-08-24
Copy editing: modeling sub-package

lglattly:
I had started copy editing the modeling sub-package but then was informed that it is currently being reworked, so I'm pushing the changes but will hold off on any more copy editing for now!

### DIFF
--- a/docs/modeling/algorithms.rst
+++ b/docs/modeling/algorithms.rst
@@ -2,28 +2,28 @@
 Algorithms
 **********
 
-Univariate polynomial evaluation
+Univariate Polynomial Evaluation
 ================================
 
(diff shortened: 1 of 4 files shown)


nden:
@lglattly The modeling documentation is going through major revisions. Sorry for not communicating this earlier. I don't think this PR will work. I can try to apply the changes here to the final version of the docs when we are done. Or we can close this PR and let you start from scratch. Let me know what you prefer.

lglattly:
@nden Probably closing this PR and starting from scratch makes more sense, as I'll want to go through all of the revisions from the start and I hadn't gotten all that far in this sub-package anyway. Sorry for the miscommunication and thanks for letting me know!

PR #9166, opened 2019-08-24
Fix potential segfault because of incorrect refcounting

MSeifert04:
This fixes a potential segfault (although a bit improbably because `WCSBase` is "well hidden") from code like this:

```
from astropy import wcs
import numpy as np
sip = wcs.Sip(np.ones((5, 5)), np.ones((5, 5)), np.ones((5, 5)), np.ones((5, 5)), np.array([1, 2]))

w = wcs.WCS()
wcs.WCSBase.__init__(w, sip, (None, None), None, (None, None))
try:
    wcs.WCSBase.__init__(w, sip, (None, None), 1, (None, None))
except:
    pass

try:
    wcs.WCSBase.__init__(w, sip, (None, None), 1, (None, None))
except:
    pass
```

I didn't add that as test-case because it's highly artificial and likely not useful in the future.

The `Py_CLEAR` instead of `Py_XDECREF` is also preventing highly unlikely bugs in case the "decrefed" instance has a `__del__` implementation that makes weird stuff (see [the inline documentation of `Py_CLEAR`](https://github.com/python/cpython/blob/v3.7.4/Include/object.h#L808-L841) for a more detailed explanation.).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -519,6 +519,8 @@ astropy.wcs
 - Fixed a possible buffer overflow when using too large negative indices for
   ``cunit`` or ``ctype`` [#9151]
 
+- Fixed reference counting in ``WCSBase.__init__`` [#9166]
+
 
 Other Changes and Additions
 ---------------------------
--- a/astropy/wcs/src/astropy_wcs.c
+++ b/astropy/wcs/src/astropy_wcs.c
@@ -123,8 +123,9 @@ Wcs_init(
         return -1;
       }
 
-      Py_XDECREF(self->py_det2im[i]);
+      Py_CLEAR(self->py_det2im[i]);
       self->py_det2im[i] = py_det2im[i];
+      Py_INCREF(py_det2im[i]);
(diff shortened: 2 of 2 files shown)


MSeifert04:
Not sure why the mpldev job failed.

MSeifert04:
Thank you @pllim

MSeifert04:
@nden I have to admit that I haven't timed it. 

However in my experience `Py_CLEAR` typically adds negligable overhead compared to `Py_XDECREF` in CPython - especially since we need to fetch the argument to set it afterwards (see also [the CPython implementation](https://github.com/python/cpython/blob/v3.7.4/Include/object.h#L842-L849). 

In theory there's also a better macro for these cases: [`Py_XSETREF`](https://github.com/python/cpython/blob/v3.7.4/Include/object.h#L893-L900) that would avoid the "set to NULL" `Py_CLEAR` does. However the definition has been moved to `cpython/object.h` and would require a conditional include...

PR #9167, opened 2019-08-25
Update Matplotlib dev CircleCI configuration for visualization tests

astrofrog:
This should help fix the Matplotlib dev builds on v3.2.x though I also need to update some reference images.


astrofrog:
This fixes the MPL dev build (the 32-bit and HTML builds are failing but those are different issues) so this is ready for review/merging.

PR #9170, opened 2019-08-26
Refactoring: Use Py_VISIT macro in the traverse methods of WCS C-extension types

MSeifert04:
The `Py_VISIT` macro is identical to the code before. It just makes the code more succinct.

### DIFF
--- a/astropy/wcs/src/str_list_proxy.c
+++ b/astropy/wcs/src/str_list_proxy.c
@@ -50,15 +50,7 @@ PyStrListProxy_traverse(
     visitproc visit,
     void *arg) {
 
-  int vret;
-
-  if (self->pyobject) {
-    vret = visit(self->pyobject, arg);
-    if (vret != 0) {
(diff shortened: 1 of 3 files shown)


MSeifert04:
Puh, i never given the tp_traverse that much attention. It's just something the gc sometimes needs to get all live objects.

MSeifert04:
I will have a looks later.

MSeifert04:
Does someone have a FITS file whose header has a non-zero `WCS.wcs.tab` field? 😕 😓

MSeifert04:
Just to make sure: The `tp_traverse` slot is quite trivial and the implementation is straight from the [Python docs](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse) - does this really need to be tested? And if yes, does it need to happen in this PR?

MSeifert04:
Thank you @nden

Issue #9132, opened 2019-08-15
Write a better starting example in the modeling documentation


nden:
The example should show model evaluation and fitting.

PR #9177, opened 2019-08-28
Update docstring to Table

hamogu:
Table now has individual table arrays, so I'm just updating the docstring to avoid confusion.

[skip ci]

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -323,12 +323,11 @@ class TableReadWrite:
 class Table:
     """A class to represent tables of heterogeneous data.
 
-    `~astropy.table.Table` provides a class for heterogeneous tabular data,
-    making use of a `numpy` structured array internally to store the data
-    values.  A key enhancement provided by the `~astropy.table.Table` class is
-    the ability to easily modify the structure of the table by adding or
-    removing columns, or adding new rows of data.  In addition table and column
-    metadata are fully supported.
+    `~astropy.table.Table` provides a class for heterogeneous tabular data.
+    A key enhancement provided by the `~astropy.table.Table` class over 
+    e.g. a `numpy` structured array is the ability to easily modify the 
+    structure of the table by adding or removing columns, or adding new
+    rows of data.  In addition table and column metadata are fully supported.
 
     `~astropy.table.Table` differs from `~astropy.nddata.NDData` by the
     assumption that the input data consists of columns of homogeneous data,


pllim:
Thanks! Should at least get the html-docs job to run with a `[skip travis]` directive. :)

taldcroft:
@hamogu - good catch, I didn't notice that bit of ancient history.  Per the advice from @pllim please update the commit message and force push to get the html-docs job run.  Otherwise LGTM.

hamogu:
You guys are getting unreasonably anal. :-)
The commit changes five lines of plain text and contains the exact same Sphinx markup it did before, just with some plain text moved around.

Fine. I'll do it eventually, but I may forget for a while because I'm on my phone right now, where I can edit in the github interface, but not clone, gitfu, and force push.

hamogu:
(I should say y my comment is only half joking. I can deal work that, but I think that this is exactly why people don't open a quick pr to fix minor issues unless they are cute developers already.)

taldcroft:
@hamogu - fair point.  I just can't tell you how many times I've been burned by sphinx so I no longer even try to guess with my brain whether a change will work, I let a computer tell me.  So as reviewers it's just easier to stick to the rules and ask for a quick rebase (from someone that we know can do it).

BTW, in the course of the conversation I learned something new and very useful, namely the `[skip travis]` directive!

hamogu:
@taldcroft is it not a little scary that we have all those options that even you don't know about?

If you agree to this change, can you approve the review so it can be merged when the tests pass?

pllim:
For future reference, when in doubt, just let the CI run. We can cancel it if we have to. Thanks for your patience! I'll let Tom A have the honor of approval/merge.

taldcroft:
> @taldcroft is it not a little scary that we have all those options that even you don't know about?

The infrastructure baffles me, but I am not fearful since I know there are infrastructure geniuses lurking to help out!

pllim:
I admit the flags are buried but they are no secret:

https://github.com/astropy/ci-helpers/blob/e1bddf10a07e88f571ea7e382051b5cab6fc4ee4/travis/setup_conda.sh#L29-L30

😸

taldcroft:
Hrrm, now I'm slightly confused.  For a change that involves only the narrative docs, which one to use?  For a change that includes only code doc string updates, which one?

pllim:
`[skip travis]` should be good. `[docs only]` was created when doc build was tested in Travis, but now it is in CircleCI.

karllark:
skipping travis here meant that two flake8 errors were introduced that caused failed travis builds.  For example, see PR #9132.

hamogu:
@karllark: Sorry about that. I did not even know that flake8 errors now cause a mandatory failure. (And I think that is too strict, because it raises the bar for contributions too high, but I don't make the rules.)

What I learn from that is that I will never skip ci in any form again - I'll leave those options to the expert developers who know what they are doing.

hamogu:
(And next time I will also think twice if it's really worth the effort to fix a minor inconsistency in the docs or if I just ignore it and move on.)

karllark:
@hamogu: Yep.   Lots of moving parts in astropy!  Don't take this as a reason not to put in a PR, no matter how small.  I would much prefer having such problems to not having correct docs!

nden:
I second @karllark's comment. This is a minor issue and easy to fix.

PR #9183, opened 2019-08-29
fixed typo: "source" -> "conda"

atakan:
"source" would not work in this case, I guess it is a typo.


pllim:
This is shown as merged but it is not really merged. It just shares the same commit with #9184 that was merged.

Issue #8676, opened 2019-05-08
Even when a cached downloaded file exists, urlopen() is called


dstndstn:
I'm tracking down a problem using a cached coordinates/sites.json on a read-only filesystem.

An issue I am finding is this:
in astropy/utils/data.py,
get_pkg_data_fileobj
ends up calling
get_readable_fileobj
with a URL, which calls
download_file()

Now, *even if the file is already in the cache*, on data.py:1005 it ends up calling
            with urllib.request.urlopen(conf.dataurl, timeout=timeout) as remote:
                _dataurls_to_alias[conf.dataurl] = [conf.dataurl, remote.geturl()]
to find out if the URL is an alias.

Now, okay, this only gets called once per session, but *really*?  Please can you avoid doing this somehow?  Pre-fill known aliases for data.astropy.org, maybe?  ie, initialize:

_dataurls_to_alias = {'http://data.astropy.org/': ['http://data.astropy.org/', 'https://astropy.stsci.edu/data/']}

pllim:
Did you pass in a `cache=True` option?

dstndstn:
Yes, it gets passed in from get_pkg_data_contents.  And it also doesn't matter, this code fragment gets called regardless of the cache setting.

https://github.com/astropy/astropy/blob/master/astropy/utils/data.py#L1001

pllim:
Ah, okay, yes. That is not related to the `cache`. The way it is designed currently is because the URL can actually be configured by user (if they want to) and we have no idea if user provided URL even have alias or not. And if we start to hardcode all the known aliases that might be used by users, where do we stop? Once per session is not that high of an overhead, is it? What is your use case that this is a blocker?

dstndstn:
so.... you don't think there should be a way to pre-download a file so that astropy will work offline?

dstndstn:
Well.... you could make the simple case simple :) and pre-populate the _dataurls_to_alias with the single alias of the default URL.

OR, have a config option to not resolve aliases.  This would allow us to pre-download the file to the cache, and then the user does not need internet access to use it.

aarchiba:
If I understand the code correctly, the failure of the alias lookup doesn't cause a problem, it just gets the URL listed as not having an alias and the function proceeds fine (possibly after a timeout, which you could set small if the Internet is never accessible on the machine). So this one call poses no particular problem to offline use of astropy.

dstndstn:
[...]

Issue #6982, opened 2017-12-13
download_file() bug with cached results from a mirror


adrn:
This is something I stumbled on when I was having connection issues with data.astropy.org but not the mirror: when using `download_file()`, my download times were very long because it tries to connect several times before giving up and switching to the mirror. When it then finally decides to use the mirror, it works, downloads the file, and caches it. But because the file gets cached using the *mirror* url, if I run `download_file()` again with the same exact filename, it won't read it from the cache because the first URL it tries is with data.astropy.org, so then the same timeout issues happen again.

This isn't a serious issue, because I suspect the failure mode is pretty rare: this will only happen if connections to data.astropy.org timeout. If the connection is explicitly blocked, it should quickly fall back to the mirror. But it should probably first check the cache for all possible URLs before doing any internet connection.

cc @saimn @pllim

pllim:
I might have a fix. PR forthcoming after I figured out how to write a test for this.

PR #9187, opened 2019-08-30
initialize data url alias for data.astropy.org

dstndstn:
To address #8676

### DIFF
--- a/astropy/utils/data.py
+++ b/astropy/utils/data.py
@@ -36,8 +36,9 @@ __all__ = [
     'check_free_space_in_dir', 'download_file',
     'download_files_in_parallel', 'is_url_in_cache', 'get_cached_urls']
 
-_dataurls_to_alias = {}
-
+_dataurls_to_alias = {
+    'http://data.astropy.org/': ['http://data.astropy.org/',
+                                 'https://astropy.stsci.edu/data/']}
 
 class Conf(_config.ConfigNamespace):
     """


aarchiba:
If I understand the code's current logic, the point of the alias mechanism is to treat URLs that are requested from `https://astropy.stsci.edu/data/` be stored under the same key as if they had been requested from `http://data.astropy.org/` . No other URL rewriting happens. This seems pointless, at best it reduces some data duplication. It doesn't help with mirroring. What is the user experience you want from the alias mechanism?

More to the point might be an ability for download_file to accept a list of mirrors, perhaps configured in the conffile. Then when `http://data.astropy.org/thing` is requested, it would know internally that if there were some problem with `data.astropy.org` it could also request `http://astropy.stsci.edu/data/thing` and store the result under `http://data.astropy.org/thing`. 

Perhaps I'll try adding that to PR #9182 . It's not really the same topic but I can't easily make it a separate pull request.

aarchiba:
Actually, because this is a much easier decision to merge, why not just strip out the alias mechanism entirely?

dstndstn:
Thanks for your reply, @aarchiba .  Please feel free to close this PR if you're willing to address the issue in #9182!  I totally agree with your questioning of the desired user experience of the aliasing mechanism (and suggestion of removing it).

pllim:
The alias was added in #6987 to address #6982 . FYI.

pllim:
Superseded by #9182. Thanks anyway!

PR #9202, opened 2019-09-05
Documenting imshow_norm return

wtgee:
Adding in the missing `returns` for the `imshow_norm` docstring.


larrybradley:
@wtgee Many thanks for your first `astropy` contribution!  The docs render fine for me locally.  I also took this opportunity to clean up the `imshow_norm` docstring a bit and to add an example.

PR #9203, opened 2019-09-05
Configuration option for the encoding of log files

ayshih:
Python's built-in `logging` package – and hence `AstropyLogger` – uses the platform-preferred encoding by default when writing to a file, so on Windows it uses CP-1252 instead of UTF-8.  That means that sending certain Unicode characters to a log file causes errors on Windows.  My ~suggested fix is simple: explicitly set the encoding to UTF-8, irrespective of platform.~

~Questions I have: Are there Windows users out there relying on CP-1252 for logging to a file?  How bad is it to append to an existing log file that had a different encoding?~

Per comments, this PR now implements a configuration option for specifying the encoding of log files, with the default behavior of using the platform-preferred encoding.

### DIFF
--- a/astropy/logger.py
+++ b/astropy/logger.py
@@ -399,7 +399,7 @@ class AstropyLogger(Logger):
                 # your code here
         '''
 
-        fh = logging.FileHandler(filename)
+        fh = logging.FileHandler(filename, encoding="UTF-8")
         if filter_level is not None:
             fh.setLevel(filter_level)
         if filter_origin is not None:
@@ -495,7 +495,7 @@ class AstropyLogger(Logger):
                 else:
                     log_file_path = os.path.expanduser(log_file_path)
 
-                fh = logging.FileHandler(log_file_path)
+                fh = logging.FileHandler(log_file_path, encoding="UTF-8")
             except OSError as e:
                 warnings.warn(
                     'log file {!r} could not be opened for writing: '
--- a/astropy/logger.py
+++ b/astropy/logger.py
@@ -67,7 +67,7 @@ class Conf(_config.ConfigNamespace):
         "file.")
     log_file_path = _config.ConfigItem(
         '',
-        "The file to log messages to. When ``''``, "
+        "The file to log messages to.  If empty string is given, "
         "it defaults to a file ``'astropy.log'`` in "
         "the astropy config directory.")
     log_file_level = _config.ConfigItem(
@@ -80,8 +80,8 @@ class Conf(_config.ConfigNamespace):
         "Format for log file entries.")
     log_file_encoding = _config.ConfigItem(
         '',
-        "The encoding (e.g., UTF-8) to use for the log file.  If ``''``, it "
-        "defaults to the platform-preferred encoding.")
+        "The encoding (e.g., UTF-8) to use for the log file.  If empty string "
+        "is given, it defaults to the platform-preferred encoding.")
 
 
 conf = Conf()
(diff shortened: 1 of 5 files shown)


MSeifert04:
Puh, it's definitely not a good idea to append to a file in a different encoding ... 

I would be more comfortable if the encoding would be a parameter or a config-value. It's a real shame that it defaults to cp-1252 encoding on windows 😢

pllim:
Making it configurable is actually pretty straight-forward, as there is already a `Conf` class in that same file. Good idea, @mseifert04!

ayshih:
The encoding can now be specified through a configuration option

astropy/logger.py
```diff
@@ -78,6 +78,10 @@ class Conf(_config.ConfigNamespace):
         "%(asctime)r, "
         "%(origin)r, %(levelname)r, %(message)r",
         "Format for log file entries.")
+    log_file_encoding = _config.ConfigItem(
+        '',
+        "The encoding (e.g., UTF-8) to use for the log file.  If ``''``, it "
```
pllim:
```suggestion
        "The encoding (e.g., UTF-8) to use for the log file.  If empty string is given, it "
```

astropy/logger.py
```diff
@@ -398,8 +402,8 @@ def log_to_file(self, filename, filter_level=None, filter_origin=None):
             with logger.log_to_file('myfile.log'):
                 # your code here
         '''
-
-        fh = logging.FileHandler(filename)
+        encoding = conf.log_file_encoding if conf.log_file_encoding != "" else None
```
pllim: I think it would work simply like this?
```suggestion
        encoding = conf.log_file_encoding if conf.log_file_encoding else None
```

astropy/logger.py
```diff
@@ -495,7 +499,8 @@ def _set_defaults(self):
                 else:
                     log_file_path = os.path.expanduser(log_file_path)
 
-                fh = logging.FileHandler(log_file_path)
+                encoding = conf.log_file_encoding if conf.log_file_encoding != "" else None
```
pllim:
```suggestion
                encoding = conf.log_file_encoding if conf.log_file_encoding else None
```

docs/logging.rst
```diff
@@ -135,16 +135,21 @@ of the Astropy configuration file::
     log_exceptions = False
 
     # Whether to always log messages to a log file
-    log_to_file = True
+    log_to_file = False
```
pllim: Changing this does not seem necessary. I think this example is to show a non-default setting.
```suggestion
    log_to_file = True
```

docs/logging.rst
```diff
@@ -135,16 +135,21 @@ of the Astropy configuration file::
...
 
-    # The file to log messages to
-    log_file_path = '~/.astropy/astropy.log'
+    # The file to log messages to. When '', it defaults to a file `astropy.log` in
+    # the astropy config directory.
+    log_file_path = ''
```
pllim: Changing this does not seem necessary. I think this example is to show a non-default setting.

docs/logging.rst
```diff
@@ -135,16 +135,21 @@ of the Astropy configuration file::
...
 
     # Threshold for logging messages to log_file_path
     log_file_level = 'INFO'
 
     # Format for log file entries
-    log_file_format = '%(asctime)s, %(origin)s, %(levelname)s, %(message)s'
+    log_file_format = '%(asctime)r, %(origin)r, %(levelname)r, %(message)r'
```
pllim: Changing this does not seem necessary. I think this example is to show a non-default setting.

docs/logging.rst
```diff
@@ -135,16 +135,21 @@ of the Astropy configuration file::
...
+
+    # The encoding (e.g., UTF-8) to use for the log file.  If '', it
```
pllim:
```suggestion
    # The encoding (e.g., UTF-8) to use for the log file.  If empty string is given, it
```

PR #9207, opened 2019-09-06
Moved APE-14 str/repr to standalone function

astrofrog:
As discussed in https://github.com/astropy/astropy/pull/9156, this moves the str/repr for APE-14-compliant WCS out of the BaseLowLevelWCS class and into a standalone function, and then only opts in to it for ``SlicedLowLevelWCS``.


astrofrog:
@nden - thanks! Fixed, so feel free to merge when CI passes.

PR #9209, opened 2019-09-06
Efficient pixel to pixel transformation function

astrofrog:
This PR adds a function to ``astropy.wcs.wcsapi`` called ``efficient_pixel_to_pixel`` which given two WCSes ``wcs_in`` and ``wcs_out`` does effectively:

```python
def efficient_pixel_to_pixel(wcs_in, wcs_out, *inputs)
    return wcs_out.world_to_pixel(*wcs_in.pixel_to_world(inputs))
```

except that it uses broadcasting to make things as efficient as possible. Here's an example with two WCSes that have correlated celestial axes, and an independent (uncorrelated) spectral axis:

```python
In [1]: import numpy as np                                                                                                                             

In [2]: x, y, z = np.arange(50), np.arange(100), np.arange(200)                                                                                        

In [3]: X, Y, Z = np.meshgrid(x, y, z, copy=False)                                                                                                     

In [4]: from astropy.wcs import WCS                                                                                                                    

In [5]: wcs1 = WCS('/Users/tom/Data/PPV/l1448_13co_small.fits')                                                                                        

In [6]: wcs2 = WCS('/Users/tom/Data/PPV/l1448_13co_tan.fits')                                                                                          

In [7]: %timeit _ = wcs2.world_to_pixel(*wcs1.pixel_to_world(X, Y, Z))                                                                                 
1.28 s ± 57.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: from astropy.wcs.wcsapi import efficient_pixel_to_pixel                                                                                        

In [9]: %timeit _ = efficient_pixel_to_pixel(wcs1, wcs2, X, Y, Z)                                                                                      
44.5 ms ± 586 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

(a 32x speedup in this case)

This code originally started off in glue, and I realized I could use it in reproject as well, so since it's very general code that just requires APE 14-compliant WCSes, I figured it could make sense to include in astropy core.

In future we could also consider adding simpler ``efficient_pixel_to_world`` and ``efficient_world_to_pixel`` functions that could re-use some of the machinery here (@Cadair expressed interest in this).

Just to be clear, only ``efficient_pixel_to_pixel`` is being added to the public API, though we could choose to expose some of the helper functions in future.

(@MSeifert04 - I'm tagging you for review since you've made a bunch of WCS-related PRs recently!)


astropy/wcs/wcsapi/tests/test_utils.py
```diff
@@ -23,3 +32,287 @@ def test_invalid():
...
+
+    matrix = pixel_to_pixel_correlation_matrix(wcs_in, wcs_out)
+    assert_equal(matrix, [[1, 1, 0], [0, 0, 1], [1, 1, 0]])
+
+
+def test_pixel_to_pixel_correlation_matrix_spectral_cube_correlated():
+
+    # We need the following fix when celestial axes are correlated with
+    # non-celestial axes: https://github.com/astropy/astropy/pull/8420
+    pytest.importorskip("astropy", minversion="3.1.2")
```
bsipocz: these should be cleaned up from here 😄

bsipocz:
I suppose the name `efficient_pixel_to_pixel` could be iterated on a bit as it doesn't really say when  not  to use it (why  would anyone want to use a not efficient way when there is an efficient function to do it?)

astrofrog:
Thanks for all the feedback! Really the main thing that determines whether this is efficient or not is not so much the size but whether any of the arrays are views that can be 'unbroadcasted'. I'm hesitant to call this function ``pixel_to_pixel`` because I ideally want to add ``pixel_to_world`` and ``world_to_pixel`` equivalents which are not methods on the WCS but can operate on a WCS, and I want to somehow indicate that these are not the same as the ``pixel_to_world`` and ``world_to_pixel`` methods that exist on the WCS. One option would be ``unbroadcasted_pixel_to_pixel`` but that's quite a mouthful!

mhvk:
I think it needs broadcasted coordinates *and* a WCS matrix that doesn't mix all the coordinates, right?  
Still, even so, the tests are relatively cheap (just look for zero strides and if any are present check the matrices) and that, therefore, we can simply create a general routine that always uses the fast route if it helps.

mhvk:
p.s. Overall point is that I don't think we need an "efficient" or "unbroadcasted" in the name - it should just be a general `pixel_in_one_to_pixel_in_other` routine for which the docstring clarifies that it makes some shortcuts if broadcasted pixels are present.

Aside: could one not get a similar speed-up by doing the checks inside the `pixel_to_world` and `world_to_pixel` routines? The only disadvantage would seem that the checks would be done twice.

astrofrog:
@mhvk - yes I see your point about just making it ``pixel_to_pixel`` and cover all cases

Unfortunately we don't always have control over the ``pixel_to_world`` or ``world_to_pixel`` implementation since this could be implemented directly by another project. We could optimize the method for our implementation but the idea here would be to have a generalized implementation that could be used with any other APE-14 WCS (e.g. GWCS or others)

mhvk:
OK, makes sense; I'm definitely in favour of adding a general `pixel_to_pixel` method that can take shortcuts; we can see later if some of those are also useful in our own wcs implementations.

astropy/wcs/wcsapi/tests/test_utils.py
astrofrog: Done!

astropy/utils/misc.py
```diff
@@ -1159,6 +1161,22 @@ def dtype_bytes_or_chars(dtype):
...
+
+    See https://stackoverflow.com/questions/40845769/un-broadcasting-numpy-arrays
+    for more details.
+    """
+
+    if array.ndim == 0:
+        return array
+
+    new_shape = np.where(np.array(array.strides) == 0, 1, array.shape)
+    return as_strided(array, shape=new_shape)
```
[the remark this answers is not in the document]
astrofrog: Nice! Done

bsipocz:
@astrofrog - appveyor has popped up here, too, I can't understand it yet, but anyway please rebase.

astrofrog:
@bsipocz - I removed efficient from the name, so i think this should hopefully now be ok?

astrofrog:
@mhvk - ``as_strided`` is now returning a deprecation warning that I don't understand. As our resident Numpy expert, would you be happy to take a quick look at this? :smile:

astropy/utils/misc.py
```diff
@@ -1159,6 +1161,27 @@ def dtype_bytes_or_chars(dtype):
...
+    original array that can be re-broadcasted back to the original array.
+
+    See https://stackoverflow.com/questions/40845769/un-broadcasting-numpy-arrays
+    for more details.
+    """
+
+    if array.ndim == 0:
+        return array
+
+    new_shape = np.where(np.array(array.strides) == 0, 1, array.shape)
```
mhvk: Hmm, it may be that the numpy version is one where there was a bug on interpreting writable; I certainly cannot reproduce it on latest numpy. But how about the following replacement?
```
array = array[tuple((slice(0, 1) if stride == 0 else slice(None)) for stride in array.strides)]
```
(I think this may even be faster...) I'll just push it up for speed...

CHANGES.rst
```diff
@@ -83,6 +83,7 @@ astropy.modeling
   and ``n_outputs`` are class attributes. Backwards compatible default values of
   ``inputs`` and ``outputs`` are generated. ``Model.inputs`` and ``Model.outputs``
   are now settable which allows renaming them on per user case. [#9298]
+- Significant reorganization of the documentation. [#9078, #9171]
```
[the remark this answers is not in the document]
bsipocz: hmm, that's docs related entry shouldn't really be in the changelog?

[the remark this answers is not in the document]
mhvk: Good point, I'll look into it.

mhvk:
@astrofrog - this now all looks good but I worry nobody will be able to find it. It might be good to put up an issue with a path forward: where it is going to be used internally, and how (if it is meant to be) it can be found externally.

astropy/wcs/wcsapi/utils.py
```diff
@@ -114,3 +116,188 @@ def wcs_info_str(wcs):
...
+        high-level shared APE 14 WCS API.
+    wcs_out : `~astropy.wcs.wcsapi.BaseHighLevelWCS`
+        A WCS object for the target dataset which complies with the
+        high-level shared APE 14 WCS API.
+    *inputs :
+        Scalars or arrays giving the pixel coordinates to transform.
+    """
+
+    # Shortcut for scalars
+    if np.isscalar(inputs[0]):
```
nden: Is this supposed to work for WCS with 1 input and 1 output, e.g. spectral only? If so, perhaps add also a check for the dimensionality.

astrofrog:
@mhvk - what about just exposing it via ``astropy.wcs.utils`` where there are a whole bunch of public-facing utilities?

astropy/wcs/wcsapi/tests/test_utils.py
```diff
@@ -56,3 +67,279 @@ def test_wcs_info_str():
...
+
+    # And these arrays should match the input
+    assert_allclose(X1, X3)
+    assert_allclose(Y1, Y3)
+    assert_allclose(Z1, Z3)
+
+
+def test_efficient_pixel_to_pixel_simple():
+
+    # Simple test to make sure that when WCS only returns one world coordinate
```
nden: Perhaps I misunderstand this comment. Did you mean one world axis in world_out? Because the test uses 2 axes and the case with one world axis should be tested.

mhvk:
@astrofrog - I think it is fine to keep it in `wcs.utils`, but currently the documentation only lists that module somewhere in the API pages, i.e., people are unlikely to find it.

astrofrog:
@mhvk - just to be clear, right now it's in ``astropy.wcs.wcsapi.utils`` which isn't even listed in the API docs. I think having it imported into ``astropy.wcs.utils`` would be better. I agree that ideally we could document it in the narrative docs but it's actually quite an advanced function that isn't *really* meant for the average user, but is more something that is needed for use in e.g. reproject and glue.

bsipocz:
> I agree that ideally we could document it in the narrative docs but it's actually quite an advanced function that isn't really meant for the average user, but is more something that is needed for use in e.g. reproject and glue.

But then those projects would find it in the API docs, don't they? (I mean not those two, but similar ones). Or is it just me who first reads the API docs only and only looks at the narrative if super badly stuck somewhere?

astrofrog:
> But then those projects would find it in the API docs, don't they? (I mean not those two, but similar ones). Or is it just me who first reads the API docs only and only looks at the narrative if super badly stuck somewhere?

Yes - right now it's not even in the API docs though, but I think exposing it in the API docs but not in the narrative docs would be sufficient.

astrofrog:
Actually I was mistaken, the function *is* already exposed in the API section of the shared WCS page:

https://44785-2081289-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/wcs/wcsapi.html#reference-api

Which I think is sufficient for now? At some point I think the whole WCS docs will need an overhaul anyway to show the APE 14 API as the default API, even for FITS-WCS, but I think we're not quite there yet.

astropy/wcs/wcsapi/utils.py
astrofrog: This is only for scalars - for the 1D case one might still pass in broadcasted arrays that can be transformed more efficiently. This is covered by the new test I just added.

astropy/wcs/wcsapi/tests/test_utils.py
astrofrog: Good catch, I think there was a test missing, so I've added one for the actual 1D case now (this test above was actually about correlated axes)

mhvk:
@bsipocz - it may just be you...  At least, I usually start with a "getting started" and then go for trial and error plus general discovery with tab-completion and checking help on methods (the exception might be class initializers). But I would definitely look if there were a table of contents that included "Utility functions".

Anyway, given that this is meant for power users, it certainly should be discoverable, and I think `wcs.utils` is a good place. I think we could advertise that simply in the "other information" section.

astrofrog:
Thinking about this more, I agree this belongs in astropy.wcs.utils, so I've moved it there. I think this should give sufficient visibility (it will be in the API docs of the main page). The WCS narrative docs have to be rethought anyway given APE 14, so once that happens we can think about whether to expose these kinds of functions in the narrative.

bsipocz:
So, how do we stand with this, ready to go?

astrofrog:
This is ready from my point of view. @mhvk?

Issue #8699, opened 2019-05-14
Invalid LaTeX format for arcsec2


petercamps:
astropy version 3.1.2

`print (astropy.units.Unit("arcsec2").to_string("latex"))`
`$\mathrm{{}^{\prime\prime}^{2}}$`

This is invalid LaTeX because the same base string carries two exponents. The problem arises because the LaTeX string for arcsec uses an exponent to raise the primes above the baseline.

A quick fix would be to always enclose the LaTeX string for arcsec in curly braces, but the resulting layout is not so nice (the primes and the exponent are on the same level). The better fix would be to include the LaTeX string for arcsec in regular parentheses if and only if it carries an exponent.

pllim:
> The better fix

@petercamps , thanks for reporting this. Would you be interested to open a PR with your suggested patch?

http://docs.astropy.org/en/latest/development/workflow/development_workflow.html

mhvk:
Hmm, I think with an exponent we should indeed not use the primes at all, but rather just `$\mathrm{arcsec}^{2}$. Arguably this should happen in composite units as well (say, `pix/arcsec`). A change is not entirely trivial - the actual formatting is done in `Latex.to_string` in `units/format/latex.py`.

Note that there is already a special treatment for `arcsec` in `si.py`, in that only the un-prefixed unit is shown using primes. The same should also be done for `arcmin` and `deg`. Right now:
```
In [12]: u.mdeg.to_string('latex')
Out[12]: '$\\mathrm{m{}^{\\circ}}$'
```
(And the same for unicode output)

Possibly the simplest would be for the format dict to `def_unit` to include an extra `latex_simple` (or so) entry with the value that gets used when the unit is shown on its own.

petercamps:
I agree with @mhvk 's suggestions. However, I can't find my way around the astropy code well enough to actually implement the change. Sorry.

mhvk:
@petercamps - no worries, this is not a trivial part of astropy, with quite a few interlocking classes.

pllim:
> The same should also be done for arcmin and deg

https://github.com/astropy/astropy/blob/eb20883bed9f815a0db177c4fd1474ab4077cfa1/astropy/units/si.py#L69-L75

seems to limit prefix for arcsec to "milli" and "micro". Do you mean that we should disable prefix for arcmin and deg **and** only limit them to "milli" and "micro" as well?

> include an extra latex_simple

[...]

Issue #5870, opened 2017-03-09
Units of table.write with output format as latex


battersbot:
Hi,

When using astropy.table to write in latex, the units of degree squared are represented as: 
$\mathrm{{}^{\circ}^{2}}$

However, this causes LaTeX to crash because it doesn't like the double superscript.

I think replacing that with deg$^2$ would solve this problem, and it seems to be a common way to represent square degrees.

Thanks for your help.  Code below.

Best,
  Cara

###
from astropy.table import Table

tbl = Table.read('column_and_temperature_and_luminosity_and_HCN_catalog.ipac', format='ascii.ipac')

name_map = {'_idx': 'Index',
            'area_ellipse': 'Ellipse Area',
           }

for key,val in name_map.items():
    tbl.rename_column(key, val)

for colname in list(tbl.columns):
    # TEMPORARY HACK
    tbl.rename_column(colname, colname.replace("_",'\_'))

tbl.write('dendrotable.tex', format='ascii.latex', overwrite=True)

## this fixes it temporarily:
with open('dendrotable.tex','r') as fh:
    lines = fh.readlines()
with open('dendrotable.tex','w') as fh:
    for line in lines:
        if '{}^{\circ}^{2}' in line:
            line = line.replace('{}^{\circ}^{2}', 'deg^2')
        fh.write(line)

PR #9218, opened 2019-09-09
Avoid double superscripts in Latex representations of units

barentsen:
### Problem

Certain units suffer from an invalid Latex representation when they are raised to an exponent.  The units affected are those which already contain a superscript instruction in their representation.  Raising such units to an exponent triggers an invalid double superscript instruction (ie. `x^{.}^{.}`).

This is a very low-impact bug, but it is nevertheless annoying because matplotlib will raise an exception when it encounters invalid Latex.

### Examples

Example notebook rendering failures:

Screenshot from 2019-09-09 14-26-28

Example matplotlib crash:

Screenshot from 2019-09-09 14-27-07

### Solution

This PR proposes a fix using a regular expression substitution.  (Happy to hear ideas for a more elegant fix?)


barentsen:
Looks like this issue was previously reported via #5870 and #8699.

dhomeier:
Not sure if that is more elegant, but maybe check for the existing exponent directly (`latex.py:35`)?
```
    @classmethod
    def _format_unit_list(cls, units):
        out = []
        for base, power in units:
            if power == 1:
                out.append(cls._get_unit_name(base))
            else:
                if cls._get_unit_name(base).find('^') < 0:
                    out.append('{0}^{{{1}}}'.format(
                        cls._get_unit_name(base),
                        utils.format_power(power)))
                else:
                    out.append('{0}^{{{1}}}}}'.format(
                        cls._get_unit_name(base).rstrip('}}'),
                        utils.format_power(power)))
        return r'\,'.join(out)
```

barentsen:
Thank you for the pointer @dhomeier.  I changed the PR according to your suggestion, which yielded a fix that is definitely less obscure.

I ended up fixing the issue by wrapping the unit in parentheses when required, i.e. this is the new behavior introduced by the PR:

Screenshot from 2019-09-10 08-38-42

**Is this what we want?**  I can imagine two alternative solutions:

### Alternative 1: add superscript inside existing superscript
 Screenshot from 2019-09-10 08-40-29

### Alternative 2: concatenate exponent to existing superscript
 Screenshot from 2019-09-10 08-41-18

Any opinions on which solution is the best?  I went for parentheses merely because it avoids having to use regular expression substitution.

mhvk:
Hmm, this is tricky. My general sense for angular units remains that expressed in https://github.com/astropy/astropy/issues/8699#issuecomment-492689476, that in any case where they are a composite unit (raised to a power, something like `arcsec/s`, etc.), they should be spelled out.

For electrons, the case is less clear, as something like `e-/ADU` makes sense... Parentheses for the powers (quite unlikely for electron...) seems like a good suggestion.

barentsen:
Side note: I stumbled upon this because of `u.electron**2`.  Kepler and TESS fluxes are in electron/second by default, so you end up with (electron/second)^2 when computing flux variance or power spectral densities.

dhomeier:
Aesthetically, I find parentheses around the angular unit symbols looking odd, but so are the powers to the second superscript level (alternative 1) in the typesetting above, and they would probably be quite hard to read. Alternative 2 creates ambiguity for `$e^{-2}$` with the inverse square of a hypothetical unit `e`. So I tend to agree with @mhvk about switching to full names for the angular unit names in this case.
The example with `(u.electron/u.s)**2` raises the question if parentheses should not simply be preserved in such cases, but I suspect writing it as `$(e^-/s)^2$` is against SI and IAU guidelines.

barentsen:
Later today I intend to have a go at changing this PR to use full names when units are raised to an exponent.  I understand from the comments above that we'll want this behavior for any unit that is raised to an exponent, not just the problematic ones!

mhvk:
Great!  Note that some of the machinery can likely be borrowed from the prefix units - we already get `mdeg` and not `m°`.

dhomeier:
My understanding was that this should apply for any units that are already represented by, or include a superscript in their short notation (guess that would be your "problematic" ones, effectively only deg/min/sec of arc and the special case of `e^-` I can think of - we don't have `^h`, `^m`, `^s` of RA, do we?).
I don't see a need why e.g. `u.m/u.s**2` should be turned into
`$\mathrm{\frac{meter}{second^{2}}}$`.

PR #9250, opened 2019-09-17
Fix broadcasting in SlicedLowLevelWCS

Cadair:
~Also some solar related tweaks to WCSAxes.~


astrofrog:
Looks good, but needs tests for the broadcasting part at least :)

Cadair:
I have removed the wcsaxes thing from this, I think that this should be a bugfix to SlicedLowLevelWCS now.

Cadair:
@bsipocz I think this should be a bugfix onto 3.2.

PR #9263, opened 2019-09-18
Cannot call numpy.random.rand with float as argument

MSeifert04:
Cannot call numpy.random.rand with float as argument. 
Correct import location for a few cosmology doctests.
Context managers cannot simply contain an inline comment.

### DIFF
--- a/astropy/cosmology/funcs.py
+++ b/astropy/cosmology/funcs.py
@@ -66,12 +66,13 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
     in a Planck13 cosmology, you could do the following:
 
     >>> import astropy.units as u
-    >>> from astropy.cosmology import Planck13, z_at_value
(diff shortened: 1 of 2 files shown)


astropy/cosmology/funcs.py
```diff
@@ -66,12 +66,13 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
     in a Planck13 cosmology, you could do the following:
 
     >>> import astropy.units as u
-    >>> from astropy.cosmology import Planck13, z_at_value
+    >>> from astropy.cosmology import z_at_value
+    >>> from astropy.cosmology.parameters import Planck13
```
MSeifert04: The import will work now, but the following tests try to use the `distmod` and `age` properties (which the dict doesn't have).

MSeifert04:
Sorry, something went totally wrong with this branch. It got somehow deconnected from the branch on my repo.

PR #9265, opened 2019-09-18
Fix EOF SytnaxError error in doctest examples

MSeifert04:


pllim:
Why not include this fix as part of #9264 ?

PR #9275, opened 2019-09-23
Add more solar frames to the wcsaxes known list

Cadair:
@astrofrog


pllim:
Let's hope no problem would *flare* up. 😏

PR #9285, opened 2019-09-25
Fix a small typo.

MSeifert04:


pllim:
I re-milestoned because this typo is also in 3.2.x but not 2.x.

PR #9298, opened 2019-09-27
Inputs on Model instances

nden:
@astrofrog @Cadair @perrygreenfield  Please review as this is a substantial change.

The current `Model` base class defines two class variables - `inputs` and `outputs` - both tuples of strings, the names of inputs and outputs of that model. There are use cases when it is useful from a user point of view to be able to change the names, especially of the inputs. An example is a WCS transform which takes ("x", "y", "spectral_order"), instead of (""x00", "x10", "x01"). The fact that these are class variables makes it difficult to do so.
Enabling change of names of inputs will allow better interface to gwcs using the `fix_inputs` operator.

`Model.inputs` and `Model.outputs` are used to derive `Model.n_inputs` and `Model.n_outputs`. While the first two are not intrinsically needed by a Model to function, the last two are essential. 

This PR changes the base class `Model` to make `n_inputs` and `n_outputs` class variables and `inputs` and `outputs` - instance variables. Care is taken to make the changes within modeling in such a way that they are backwards compatible, essentially providing default values which match the original values of `inputs` and `outputs`.

What is the effect on user defined models which already define `inputs` and `outputs` as class variables?

- The changes are transparent if they continue to use the models as before
   - evaluating them musing positional arguments
   - evaluating them using keyword arguments with the old inputs as keys
- They won't be able to evaluate the models with renamed inputs. For this to work they will need to change their models to redefine `n_inputs`, `n_outputs`, `inputs` and `outputs`.

It may be possible to raise a `DeprecationWarning` if it is detected that there are models which define `inputs` as class variables. Is this necessary?

I'd appreciate any comments.

 
TODO:
[ ] fix documentation
[ ] possibly add DeprecationWarnings

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -380,7 +380,7 @@ class _ModelMeta(abc.ABCMeta):
             elif n_inputs == 2:
                 inputs = ('x', 'y')
             else:
-                inputs = tuple(['x'+str(i) for i in range(members['n_inputs'])])
+                inputs = tuple(['x' + str(i) for i in range(members['n_inputs'])])
             # Don't create a custom __call__ for classes that already have one
             # explicitly defined (this includes the Model base class, and any
             # other classes that manually override __call__
@@ -735,8 +735,8 @@ class Model(metaclass=_ModelMeta):
             self._inputs = ("x", "y")
             self._outputs = ("z",)
         else:
-            self._inputs = tuple("x"+ str(idx) for idx in range(self.n_inputs))
-            self._outputs = tuple("x"+ str(idx) for idx in range(self.n_outputs))
+            self._inputs = tuple("x" + str(idx) for idx in range(self.n_inputs))
+            self._outputs = tuple("x" + str(idx) for idx in range(self.n_outputs))
 
     @property
     def inputs(self):
@@ -744,6 +744,8 @@ class Model(metaclass=_ModelMeta):
 
     @inputs.setter
     def inputs(self, val):
+        if len(val) != self.n_inputs:
+            raise ValueError(f"Expected {self.n_inputs} number of inputs, got {len(val)}.")
         self._inputs = val
         self._initialize_unit_support()
 
@@ -753,6 +755,8 @@ class Model(metaclass=_ModelMeta):
 
     @outputs.setter
     def outputs(self, val):
+        if len(val) != self.n_outputs:
+            raise ValueError(f"Expected {self.n_outputs} number of outputs, got {len(val)}.")
         self._outputs = val
 
     def _initialize_unit_support(self):
@@ -866,13 +870,60 @@ class Model(metaclass=_ModelMeta):
             else:
                 super().__setattr__(attr, value)
 
-    def __call__(self, *inputs, **kwargs):
+    def __call__(self, *args, **kwargs):
         """
         Evaluate this model using the given input(s) and the parameter values
         that were specified when the model was instantiated.
         """
+        def _keyword2positional(kwargs):
+            # these are the keys that are always present as keyword arguments
+            keys = ['model_set_axis', 'with_bounding_box', 'fill_value',
+                    'equivalencies', 'inputs_map']
 
-        return generic_call(self, *inputs, **kwargs)
+            new_inputs = {}
+            # kwargs contain the names of the new inputs + ``keys``
+            allkeys = list(kwargs.keys())
+            # Remove the names of the new inputs from kwargs and save them
+            # to a dict ``new_inputs``.
+            for key in allkeys:
+                if key not in keys:
+                    new_inputs[key] = kwargs[key]
+                    del kwargs[key]
+
+            return new_inputs
+
+        if not args:
+            # Inputs were passed as keyword (not positional) arguments.
+            # Because the signature of the ``__call__`` is defined at
+            # the class level, the name of the inputs cannot be changed at
+            # the instance level and the old names are always present in the
+            # signature of the method. In order to use the new names of the
+            # inputs, the old names are taken out of ``kwargs``, the input
+            # values are sorted in the order of self.inputs and passed as
+            # positional arguments to ``__call__``.
+
+            new_inputs = _keyword2positional(kwargs)
+            # Create positional arguments from the keyword arguments in ``new_inputs``.
+            new_args = []
+            for k in self.inputs:
+                new_args.append(new_inputs[k])
+        elif len(args) != self.n_inputs:
+            # Some inputs are passed as positional, others as keyword arguments.
+            args = list(args)
+
+            # Create positional arguments from the keyword arguments in ``new_inputs``.
+            new_inputs = _keyword2positional(kwargs)
+            new_args = []
+            for k in self.inputs:
+                if k in new_inputs:
+                    new_args.append(new_inputs[k])
+                else:
+                    new_args.append(args[0])
+                    del args[0]
+        else:
+            new_args = args
+        return generic_call(self, *new_args, **kwargs)
+
 
     # *** Properties ***
     @property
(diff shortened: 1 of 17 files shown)


astropy/modeling/tests/test_quantities_evaluation.py
```diff
@@ -107,40 +108,40 @@ def test_evaluate(self):
...
         assert_quantity_allclose(self.model(4 * u.rad, 2), 8 * u.rad)
         assert_quantity_allclose(self.model(4 * u.rad, 2 * u.s), 8 * u.rad * u.s)
 
         with pytest.raises(UnitsError) as exc:
             self.model(4 * u.s, 3)
-        assert exc.value.args[0] == ("MyTestModel: Units of input 'a', s (time), could not be "
+        assert exc.value.args[0] == ("MyTestModel: Units of input 'x', s (time), could not be "
                                      "converted to required input units of deg (angle)")
-
+        print('allow', self.model.input_units_allow_dimensionless)
```
pllim: Do we need this print?

nden: nope

astropy/modeling/core.py
```diff
@@ -725,19 +727,47 @@ def __init__(self, *args, meta=None, name=None, **kwargs):
...
+            self._inputs = tuple("x"+ str(idx) for idx in range(self.n_inputs))
+            self._outputs = tuple("x"+ str(idx) for idx in range(self.n_outputs))
+
+    @property
+    def inputs(self):
+        return self._inputs
+
+    @inputs.setter
+    def inputs(self, val):
+        self._inputs = val
```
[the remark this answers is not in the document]
perrygreenfield: Validating the inputs seems like a good idea here.

astropy/modeling/core.py
```diff
@@ -2212,6 +2221,36 @@ class FittableModel(Model):
...
+            # Remove the names of the new inputs from kwargs and save them
+            # to a dict ``new_inputs``.
+            for key in allkeys:
+                if key not in keys:
+                    new_inputs[key] = kwargs[key]
+                    del kwargs[key]
+            # Create positional arguments from the keyword arguments in ``new_inputs``.
+            args = []
+            for k in self.inputs:
+                args.append(new_inputs[k])
```
perrygreenfield: What about the case where someone uses both positional and keyword arguments? E.g., the original input names are x and y, and they rename them to b and c and then call f(1.2, b=7.1). It would seem that in this case the conversion of the keyword argument to positional won't happen or is that incorrect?

astropy/modeling/mappings.py
```diff
@@ -47,31 +47,34 @@ class Mapping(FittableModel):
...
         if n_inputs is None:
-            self._inputs = tuple('x' + str(idx)
-                                 for idx in range(max(mapping) + 1))
+            self.inputs = tuple('x' + str(idx)
+                                for idx in range(max(mapping) + 1))
+            self.outputs = tuple('x' + str(idx) for idx in range(len(mapping)))
         else:
-            self._inputs = tuple('x' + str(idx)
+            self.inputs = tuple('x' + str(idx)
                                  for idx in range(n_inputs))
```
perrygreenfield: Could this if/else block code be merged with the previous since the logical test is the same?

astropy/modeling/mappings.py
```diff
@@ -122,6 +125,8 @@ def inverse(self):
         inv = self.__class__(mapping)
         inv._inputs = self._outputs
         inv._outputs = self._inputs
+        inv._n_inputs = len(inv._inputs)
+        inv._n_outputs = len(inv._outputs)
         return inv
```
perrygreenfield: So this model doesn't support .N_inputs and .N_outputs? If not, should it?

astropy/modeling/polynomial.py
```diff
@@ -1391,8 +1395,9 @@ class SIP(Model):
     .. [1] `David Shupe, et al, ADASS, ASP Conference Series, Vol. 347, 2005 <http://adsabs.harvard.edu/abs/2005ASPC..347..491S>`_
     """
 
-    inputs = ('u', 'v')
-    outputs = ('x', 'y')
+    n_inputs = 2
+    n_outputs = 2
+
```
perrygreenfield: Does this change the default name of the inputs since now it uses the default names x, y?

astropy/modeling/mappings.py
nden: The two mapping models are special in that `n_inputs` and `n_outputs` are determined dynamically when the model is initialized. However, they need to be defined before `super` is called. `inputs` and `outputs` need to be defined after `super` is called so that they can overwrite any defaults.

astropy/modeling/mappings.py
nden: See comment above.

astropy/modeling/core.py
nden: You are right, this won't work. I will push a fix shortly.

astropy/modeling/polynomial.py
nden: Yes, fix coming.

astropy/modeling/core.py
```diff
@@ -866,13 +870,60 @@ def __setattr__(self, attr, value):
...
+
+            # Create positional arguments from the keyword arguments in ``new_inputs``.
+            new_inputs = _keyword2positional(kwargs)
+            new_args = []
+            for k in self.inputs:
+                if k in new_inputs:
+                    new_args.append(new_inputs[k])
+                else:
+                    new_args.append(args[0])
+                    del args[0]
```
perrygreenfield: Suppose there are 3 inputs x, y, z, and this call is made f(1, 2, x=3, z=-1), it looks like to me what will happen is that the new_args list will be [2,  1,  -1], which is not what I would expect (as it goes through new_inputs, it does find x, so that value goes on the list, doesn't find y, so the first value of args goes on the list, then the value for z goes n the list.) So two possible issues: should it be an error if a values is specified twice, or do we take the keyword value? Did I say two? OK, one I guess. The other part seems like an error in that it can get the two out of synch when a positional argument is also provided as a keyword argument. Perhaps  split up the new_inputs into two keyword lists: one that corresponds to the length of args and the other which is the rest. If one of the new keywords is in the first, its an error, and if not, just append to args. But I may be confused.

astropy/modeling/core.py
```diff
@@ -3903,6 +3915,7 @@ def _prepare_inputs_single_model(model, params, inputs, **kwargs):
 def _prepare_outputs_single_model(model, outputs, format_info):
     broadcasts = format_info[0]
     outputs = list(outputs)
+    print('broadcasts, info', broadcasts, outputs)
```
perrygreenfield: Is this print statement supposed to stay?

astropy/modeling/core.py
```diff
@@ -727,6 +720,19 @@ def __init__(self, *args, meta=None, name=None, **kwargs):
...
+        # ``inputs`` and ``outputs``.
+        self._inputs_deprecation()
+
+    def _inputs_deprecation(self):
+        if hasattr(self.__class__, 'n_inputs') and isinstance(self.__class__.n_inputs, property):
+            warnings.warn(
+            f"""Class {self.__class__.__name__} defines class attributes ``inputs``.
+            This has been deprecated in v4.0 and support will be removed in v4.1.
+            Starting with v4.0 classes must defnie a class attribute ``n_inputs``.
+            Please consult the documentaiton for details.
```
Cadair: typo

nden:
@perrygreenfield I think this is ready for a final review.

astropy/modeling/core.py
```diff
@@ -727,6 +720,19 @@ def __init__(self, *args, meta=None, name=None, **kwargs):
         self._initialize_slices()
         self._initialize_unit_support()
 
+        # Raise DeprecationWarning on classes with class attributes
+        # ``inputs`` and ``outputs``.
+        self._inputs_deprecation()
+
+    def _inputs_deprecation(self):
```
perrygreenfield: I'm easily confused :-) The following if statement checks for class level n_inputs and then if it exists says the class level inputs is deprecated. Naively I  would have thought the if statement would look for the existence of 'inputs'. Is this check implied elsewhere (e.g, the existence of n_inputs means inputs is defined at the class level, but I don't see that in the recent changes. (and as @Cadair points out defnie should be fixed)

nden: Both work, just different logic :-). I changed it and it's clearer.

Issue #8793, opened 2019-06-04
download_file and flaky server (was: Trouble opening IERS table)


Warren-Porter:
This causes NumPy to throw an error:
```
from astropy.utils import iers
dat = iers.IERS_Auto.open()
```
I am guessing the error is because an empty array is being generated from the IERS data and being sent to NumPy. Here is the error traceback and error message:
```
ERROR: ValueError: zero-size array to reduction operation minimum which has no identity [numpy.core.fromnumeric]
Error in sys.excepthook:
Traceback (most recent call last):
  File "/data/home004/warren.porter/anaconda3/envs/CrIS_geolocation_3/lib/python3.7/site-packages/astropy/logger.py", line 262, in _excepthook
    self._excepthook_orig(etype, value, traceback)
TypeError: 'NoneType' object is not callable

Original exception was:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/data/home004/warren.porter/anaconda3/envs/CrIS_geolocation_3/lib/python3.7/site-packages/astropy/utils/iers/iers.py", line 660, in open
    cls.iers_table = cls.read(file=filename)
  File "/data/home004/warren.porter/anaconda3/envs/CrIS_geolocation_3/lib/python3.7/site-packages/astropy/utils/iers/iers.py", line 507, in read
    table = cls._combine_a_b_columns(iers_a)
  File "/data/home004/warren.porter/anaconda3/envs/CrIS_geolocation_3/lib/python3.7/site-packages/astropy/utils/iers/iers.py", line 470, in _combine_a_b_columns
    table.meta['predictive_index'] = np.min(np.flatnonzero(is_predictive))
  File "/data/home004/warren.porter/anaconda3/envs/CrIS_geolocation_3/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 2442, in amin
    initial=initial)
  File "/data/home004/warren.porter/anaconda3/envs/CrIS_geolocation_3/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 83, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
ValueError: zero-size array to reduction operation minimum which has no identity

```
Any assistance would be greatly appreciated. Thank you!

pllim:
Works fine for me. Are you sure it wasn't because the remote service was unavailable? 

```python
>>> from astropy.utils import iers
>>> dat = iers.IERS_Auto.open()
[...]

PR #9302, opened 2019-09-29
Check downloaded files are the right length

aarchiba:
Current behaviour is for download_file to treat as a success any condition that doesn't actually raise an exception, but in #8793 users report that server problems are resulting in incomplete IERS files being downloaded and used, leading to cryptic errors. This PR checks the size on download, raising an exception if the server provides too many or too few bytes. The authoritative size also comes from the server, of course, but this should at worst never be triggered and at best catch the above-mentioned dropping downloads.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -570,6 +570,10 @@ astropy.utils
 - Fixed ``find_api_page`` access by using custom request headers and HTTPS
   when version is specified. [#9032]
 
+- Make ``download_file`` (and by extension ``get_readable_fileobj`` and others)
+  check the size of downloaded files against the size claimed by the server.
+  [#9302]
+
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


aarchiba:
So normally ContentTooShortError has a "content" attribute containing the downloaded data, in case the caller wants it even though it's too short (the temporary file is deleted). I wasn't sure whether to do that here - it's easy enough to just read in the downloaded data, but some of the things lying around in my cache are well over a hundred megabytes, and it seems dubious to create a hundred-megabyte string just to attach it to an exception people probably won't catch (the big ones are exactly the ones likely to be truncated too). I do know how big the contents actually are, I could set an upper limit, hopefully enough for people to inspect it and see the 404 contents page or whatever. Unless servers are misconfigured, though, I really expect short files to happen only with dropped connections, so there's not much information in them.

The partial file might be useful for resuming, I suppose, but implementing resuming seems out of scope for astropy (even my PR #9182 is not that ambitious).

pllim:
@aarchiba , the email you used for these commits does not seem to be tied to your GitHub account. It does not affect the PR but it might also not give you due credit. FYI.

pllim:
I restarted https://travis-ci.org/astropy/astropy/jobs/591261549 just to make sure the remote data test failure is transient. FYI.

PR #9315, opened 2019-10-04
Fix bug creating Table with rows as generator

### LINKED ISSUE #9290: Table returns inconsistent results when rows= argument is a generator
I tried to construct a Table with rows which come out of a generator (originally a call to `filter`), and for some reason it skips one result when you do this.

```python
In [12]: from astropy.table import Table                                                                                                                                                                                                                                                                       

In [13]: arow = lambda x: [random.random() for i in range(x)]                                                                                                                                                                                                                                                  

In [14]: it = (arow(3) for i in range(10))                                                                                                                                                                                                                                                                     

In [15]: li = [arow(3) for i in range(10)]                                                                                                                                                                                                                                                                     

In [16]: Table(rows=it)           
```
```                                                                                                                                                                                                                                                                             
Out[16]: 
<Table length=9>
        col0                col1                col2       
      float64             float64             float64      
------------------- ------------------- -------------------
...
```
```python
In [17]: Table(rows=li)                                                                                                                                                                                                                                                                                        
Out[17]: 
```
```
<Table length=10>
        col0                 col1                col2       
      float64              float64             float64      
-------------------- ------------------- -------------------
...
```

taldcroft:
Fixes #9290.

### DIFF
--- a/astropy/table/tests/test_init_table.py
+++ b/astropy/table/tests/test_init_table.py
@@ -6,7 +6,7 @@ from collections.abc import Mapping
 import pytest
 import numpy as np
 
-from astropy.table import Column, TableColumns
+from astropy.table import Column, TableColumns, Table
 
 
 class TestTableColumnsInit():
@@ -519,3 +519,10 @@ def test_init_from_row_OrderedDict(table_type):
 
     with pytest.raises(ValueError):
         table_type(rows=[OrderedDict([('b', 1)]), {'a': 10, 'b': 20}])
+
+
+def test_init_from_rows_as_generator():
+    rows = [(1 + ii, 2 + ii) for ii in range(2)]
+    t = Table(rows=rows)
+    assert np.all(t['col0'] == [1, 2])
+    assert np.all(t['col1'] == [2, 3])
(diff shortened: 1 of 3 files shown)


astropy/table/tests/test_init_table.py
```diff
@@ -519,3 +519,10 @@ def test_init_from_row_OrderedDict(table_type):
 
     with pytest.raises(ValueError):
         table_type(rows=[OrderedDict([('b', 1)]), {'a': 10, 'b': 20}])
+
+
+def test_init_from_rows_as_generator():
+    rows = [(1 + ii, 2 + ii) for ii in range(2)]
```
mhvk: I don't think this test is right: `rows` is a list here, not a generator...

pllim: Ops! Sorry if I merged too quickly!

pllim: @Cadair 's original use case does seemed fixed though. I'll fix the test since it is my bad to merge it too soon.

mhvk: Yes, the solution looks good, just the test is not...

pllim: See #9316

PR #9322, opened 2019-10-04
Add a classmethod to retrieve a Galactocentric frame with modern solar position/motion parameters

### LINKED ISSUE #9199: Update Galactocentric default parameters?
Thanks to [GRAVITY](https://ui.adsabs.harvard.edu/abs/2018A%26A...615L..15G/abstract), [Gaia](https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1417B/abstract), and [a combination of GRAVITY+VLBA](https://ui.adsabs.harvard.edu/abs/2018RNAAS...2d.210D/abstract), we now have new standard measurements of the solar motion, sun-galactic center distance, and solar height above the midplane. Given this, it's probably time to update the default Galactocentric frame parameters in astropy.coordinates. But is this a good idea? Updating the defaults will affect code downstream for anyone using the default frame. On one hand, maybe this is OK: if anyone wants to be certain of the frame they are using, we could recommend that they hard-set the values. On the other hand, this could lead to some confusion from users. So, what should we do: Change the values? Make a new `Galactocentric2019` frame? or?

Thoughts? @eteq @mhvk @Juanlu001

adrn:
As noted in #9199, thanks to Gaia, we now have new, much more precise measurements of the solar motion, sun-galactic center distance, and solar height above the midplane, which are all parameters of the `Galactocentric` coordinate frame. Unlike most other frames in `astropy.coordinates`, these values are not standardized in the same way (e.g., there is no official IAU reference frame). Still, the parameters that are in the `Galactocentric` frame are now somewhat out of date, and different from what many people now adopt: It would therefore be good to provide a way to retrieve a frame with newer values.

We discussed a few options for how to do this:
1. Just update the parameters and tell people that the frame could change. I decided that the disadvantages of this were too great to go through with it, even though it would be the simplest programmatically! Disadvantages: (a) users who update may have their analysis code change spontaneously, (2) this would require updating all of the doctest values every time the defaults are changed, (3) this doesn't provide any kind of history of the parameter values
2. Add a function or classmethod to retrieve the frame with a specified "version" of the parameters. This could be a nice way of doing it, and would fit in well conceptually with the "ScienceState". But this adds a lot of overhead in terms of "versioning" machinery.
3. Keep the defaults in the class initializer, but add a classmethod that allows retrieving the "latest" parameter values. This is what I ended up going with in this PR.

This fixes #9199

### DIFF
--- a/astropy/coordinates/builtin_frames/galactocentric.py
+++ b/astropy/coordinates/builtin_frames/galactocentric.py
@@ -81,6 +81,19 @@ doc_footer = """
             [( -9.6083819 ,  -9.40062188,  6.52056066),
              (-21.28302307,  18.76334013,  7.84693855)]>
 
+    This uses a somewhat out-of-date set of parameters for the solar position
+    and motion in a Galactocentric sense: If your code or analysis does not
+    depend sensitively on these values, and you are just interested in
+    retrieving the latest measurements of these quantities, you can instead use
+    the ``.get_latest()`` classmethod to retrieve a frame with newer parameter
+    values::
+
+        >>> c.transform_to(coord.Galactocentric.get_latest()) # doctest: +FLOAT_CMP
+        <Galactocentric Coordinate (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
+            (266.4051, -28.936175)>, galcen_distance=8.122 kpc, galcen_v_sun=(12.9, 245.6, 7.78) km / s, z_sun=20.8 pc, roll=0.0 deg): (x, y, z) in kpc
+            [( -9.43489286, -9.40062188, 6.51345359),
+            (-21.11044918, 18.76334013, 7.83175149)]>
+
     To specify a custom set of parameters, you have to include extra keyword
     arguments when initializing the Galactocentric frame object::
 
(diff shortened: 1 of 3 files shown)


astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -216,6 +225,43 @@ def get_roll0(cls):
...
+          Brunthaler (2004) proper motion of Sgr A*:
+          https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R/abstract
+          https://ui.adsabs.harvard.edu/abs/2018RNAAS...2..210D/abstract
+
+        See the docstring for the initializer of this
+        `~astropy.coordinates.Galacotcentric` class for information about the
+        frame and its default component (data) names.
+        """
+
+        kwargs['galcen_distance'] = kwargs.pop('galcen_distance', 8.122 * u.kpc)
```
mhvk: Clearer would be
```
kwargs.setdefault('galcen_distance', 8.122*u.kpc)
(etc)
```
(note: PEP8 avoids the spaces around operators for arguments)

mhvk:
@adrn - the code looks good and I can see the logic for not wanting to change the defaults (too often). On the other hand, I'm not so sure on having a class-specific method for getting a different version. Might an intermediate solution be to still define a new `ScienceState` in this module and use that? It could then have the standard options `None` and `'latest'`, but also allow the user to set a default for their whole session. One advantage of this mechanism is that it allows the user control even if `GalactoCentric` is used somewhere in a library.

I don't think it would be all that tricky to use the `ScienceState` (see #9244 for what I think is a more complicated example).

adrn:
Now that we're moving forward with #9346, I will close this!

PR #9329, opened 2019-10-06
Expose reading and setting of the ERFA leap second table

mhvk:
@eteq - this is step 1 to using/updating the leap second table. I went a bit beyond @jwoillez's implementation by also allowing one to *retrieve* the current leap second table, using a new `eraGetLeapSeconds`. With that, it seemed it would be better to also rename `eraUpdateLeapSeconds` to `eraSetLeapSeconds`.

One can now retrieve these as structured arrays using ~`_erfa.get_leap_second_table` and `_erfa.set_leap_second_table`.~ EDIT (name change): `erfa.get_leap_seconds` and `_erfa.set_leap_seconds`.

EDIT 2: I hid the C functions behind a new leap_second singleton, with `get`, `set`, and `update` methods.

### DIFF
--- a/astropy/_erfa/core.py.templ
+++ b/astropy/_erfa/core.py.templ
@@ -38,15 +38,12 @@ from astropy.utils.exceptions import ErfaError, ErfaWarning
 from astropy.utils.misc import check_broadcast
 
 from . import ufunc
-from .ufunc import (dt_eraASTROM, dt_eraLDBODY, dt_pv,
-                    dt_sign, dt_type, dt_ymdf, dt_hmsf, dt_dmsf)
 
 __all__ = ['ErfaError', 'ErfaWarning',
            {{ funcs|map(attribute='pyname')|surround("'","'")|join(", ") }},
            {{ constants|map(attribute='name')|surround("'","'")|join(", ") }},
            # TODO: delete the functions below when they can get auto-generated
-           'version', 'version_major', 'version_minor', 'version_micro', 'sofa_version',
-           'dt_eraASTROM', 'dt_eraLDBODY', 'dt_pv', 'dt_ymdf', 'dt_hmsf', 'dt_dmsf']
+           'version', 'version_major', 'version_minor', 'version_micro', 'sofa_version']
 
 
 # <---------------------------------Error-handling---------------------------->
--- a/astropy/_erfa/interface.py
+++ b/astropy/_erfa/interface.py
@@ -1,0 +1,177 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Interfaces to the ERFA library, in particular for leap seconds.
+"""
+from datetime import datetime, timedelta
+from warnings import warn
+
+import numpy as np
+
+from astropy.utils.decorators import classproperty
+from astropy.utils.exceptions import AstropyWarning
+
+from .ufunc import get_leap_seconds, set_leap_seconds, dt_eraLEAPSECOND
+
+
+class leap_seconds:
+    """Leap second management.
+
+    This singleton class allows access to ERFA's leap second table,
+    using the methods 'get', 'set', and 'update'.
+
+    One can also check expiration with 'expires' and 'expired'.
+
+    Note that usage of the class is similar to a ``ScienceState`` class,
+    but it cannot be used as a context manager.
+    """
+    _expires = None
+    """Explicit expiration date inferred from leap-second table."""
+    _expiration_days = 180
+    """Number of days beyond last leap second at which table expires."""
+
+    def __init__(self):
+        raise RuntimeError("This class is a singleton.  Do not instantiate.")
+
+    @classmethod
+    def get(cls):
+        """Get the current leap-second table used internally."""
+        return get_leap_seconds()
+
+    @classmethod
+    def validate(cls, table):
+        """Validate a leap-second table.
+
+        Raises
+        ------
+        ValueError
+            If the leap seconds in the table are not on the 1st of January or
+            July, or if the sorted TAI-UTC do not increase in increments of 1.
+        """
+        try:
+            day = table['day']
+        except Exception:
+            day = 1
+
+        # Take care of astropy Table.
+        if hasattr(table, '__array__'):
+            table = table.__array__()[list(dt_eraLEAPSECOND.names)]
+
+        table = np.array(table, dtype=dt_eraLEAPSECOND, copy=False,
+                         ndmin=1)
+
+        # Simple sanity checks.
+        if table.ndim > 1:
+            raise ValueError("can only pass in one-dimensional tables.")
+
+        if not np.all(((day == 1) &
+                       (table['month'] == 1) | (table['month'] == 7)) |
+                      (table['year'] < 1972)):
+            raise ValueError("leap seconds inferred that are not on "
+                             "1st of January or 1st of July.")
+
+        if np.any((table['year'][:-1] > 1970) &
+                  (np.diff(table['tai_utc']) != 1)):
+            raise ValueError("jump in TAI-UTC by something else than one.")
+
+        return table
+
+    @classmethod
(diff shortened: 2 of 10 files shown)


astropy/_erfa/core.py.templ
```diff
@@ -38,15 +38,12 @@ from astropy.utils.exceptions import ErfaError, ErfaWarning
 from astropy.utils.misc import check_broadcast
 
 from . import ufunc
-from .ufunc import (dt_eraASTROM, dt_eraLDBODY, dt_pv,
```
mhvk: Removed imports here since they are not used anyway, and make much more sense in the `_erfa` module.

astropy/_erfa/erfa_generator.py
```diff
@@ -388,7 +388,7 @@ def __init__(self, name, source_path, match_line=None):
             else:
                 filecontents = f.read()
 
-        pattern = fr"\n([^\n]+{name} ?\([^)]+\)).+?(/\*.+?\*/)"
+        pattern = fr"\n([^\n=]+{name} ?\([^)]+\)).+?(/\*.+?\*/)"
```
mhvk: This is because `dat.c` now has an `stat = eraDat`, which should not be recognized as the function definition.

mhvk:
@eteq - one possible issue with the whole procedure is that the ERFA update is, per definition, global, ie., it will affect anything else that has the library in memory (sub-process, etc.). Similarly, the module is not sub-process safe. Of course, this is an issue only if different subprocesses had different wishes for the leap-second table, which is arguably very unlikely. I guess in principle it might be better instead to offer a hook in `eraDat` that calls out to get the leap-second table (which then by default returns the built-in one, but can be overridden somehow).

Anyway, my own sense is to punt on this for now, and leave this for 5.0 or so.

mhvk:
@eteq - yet another issue: if we go for an astropy-only approach on this for 4.0, it will not work for any distribution that separately ships `liberfa`, since in those we cannot actually override the leap second table. Since I work on such a distribution myself, this is a problem..

Or would it work for those to compile just `dat.c` ourselves?

mhvk:
@eteq - would you be able to review? For now, I decided to forge ahead - we can worry about the distribution aspect later.

astropy/_erfa/interface.py
```diff
@@ -0,0 +1,177 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Interfaces to the ERFA library, in particular for leap seconds.
+"""
+from datetime import datetime, timedelta
```
mhvk: It would be nice to use `Time` but that leads to round-trip issues when adding auto-updates upon loading of `_erfa`.

astropy/_erfa/interface.py
```diff
@@ -0,0 +1,177 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Interfaces to the ERFA library, in particular for leap seconds.
```
eteq: I think it's better to make this *only* about leap seconeds.  I.e., something like:

```suggestion
Interfaces to controlling leap seconds in the ERFA library.
```
and then change the name of the module file from `interface.py` to `leap_seconds.py`

astropy/_erfa/erfa_generator.py
eteq: Gotta say: these regexes are pretty hairy.  To get the PR adding the new ERFA in I spent probably 45 min working out the one-or-two-character changes... (but this one looks good!)

astropy/_erfa/interface.py
eteq: None of this is actually *instantiated* at import time, right?  So can't we just use `time` imports *inside* the functions?  I think that's a more consistent approach.  

(If I missed an import-time requirement, then you can ignore this, though. It's not *that* bad since it's easy to transition between `Time` and `datetime` anyway)

mhvk:
Yes, I've been wondering about making the changes to `dat.c` more minimal, ideally such that we can use a system library and things still work. But it needs me understand global variables a bit better...

On `interfaces` vs `light_seconds` for the module - I had the latter originally, but the problem is that the singleton also is called `light_seconds`, and `from light_seconds import light_seconds` then hides the module. Or am I wrong?

mhvk:
p.s. The reason for a singleton *class* is that I started with `ScienceState` - which is a singleton class. If you think it is clearer, I can certainly make it a singleton module, i.e., `def get()`, etc. That would also solve the naming issue...

astropy/_erfa/erfa_generator.py
mhvk: Yes, hairy it is...

astropy/_erfa/interface.py
mhvk: I wish!! (and tried). The problem is that `Time` initiation needs `erfa` installed (even the plainest formats like 'iso' rely on it). More specifically, `Time.now()` needs to have the leap seconds up to date (in case 'now' is on a leap-second day), yet here I need to have `Time.now()` accessible to tell whether to update the leap-second table.

mhvk:
@eteq - I managed to get out most of the changes to `dat.c`, into a new `erfadatextra.c` (prefixed by `erfa` in analogy with `erfaversion.c` - I think it makes sense to avoid calling anything not from SOFA by a name that could imply it came from SOFA).

I looked a bit at the module replacement for the current `light_second` singleton class, but don't like that then I cannot have properties. My sense would be to leave it as is for now, since it is so analogous to `ScienceState` (and we use singleton classes also in `units.formats`).

mhvk:
@eteq - from the commits, it may not be so clear, but the total change in `dat.c` is now quite small: relative to the commit just before that by you of @jwoillez's changes,
```
git diff ddab2476d477fa4e92b058c4c72e70f05535f31b cextern/erfa/dat.c
 #include "erfa.h"
+#include "erfadatextra.h"
 
 int eraDat(int iy, int im, int id, double fd, double *deltat)
 /*
@@ -144,10 +145,7 @@ int eraDat(int iy, int im, int id, double fd, double *deltat)
    enum { NERA1 = (int) (sizeof drift / sizeof (double) / 2) };
 
 /* Dates and Delta(AT)s */
-   static const struct {
-      int iyear, month;
-      double delat;
-   } changes[] = {
+   static const eraLEAPSECOND _changes[] = {
       { 1960,  1,  1.4178180 },
       { 1961,  1,  1.4228180 },
       { 1961,  8,  1.3728180 },
@@ -193,7 +191,13 @@ int eraDat(int iy, int im, int id, double fd, double *deltat)
    };
 
 /* Number of Delta(AT) changes */
-   enum { NDAT = (int) (sizeof changes / sizeof changes[0]) };
+   enum { _NDAT = (int) (sizeof _changes / sizeof _changes[0]) };
+
+/* Get/initialise leap-second if needed */
+   int NDAT;
+   eraLEAPSECOND *changes;
+
+   NDAT = eraDatini(_changes, _NDAT, &changes);
 
 /* Miscellaneous local variables */
```

mhvk:
@eteq - this is still the simplest PR of the two - could you re-review?

mhvk:
I'm quite happy to change the name, but not to `leap_seconds`! How about `utils` or `helpers`? In a way, the actual interface is in C so this is just a helper module/routine/class.

eteq:
I like `helpers`.  If we do `utils` I'm 90% sure @adrn will just revert it all :wink:

pllim:
Why not *leap first* can make it `utils` anyway? 😏

eteq:
@mhvk - also if it would be helpful I can do some of the clean-up steps to finish some of this and/or #9365 off if you're running out of time today.

mhvk:
@eteq - thanks, may be needed. But right now pushing up a version with `helpers.py`...

mhvk:
coverage failure is a false positive.  Since approved before, merging...

PR #9346, opened 2019-10-09
Set the Galactocentric frame parameter defaults using a ScienceState

adrn:
Based on @mhvk's suggestion in #9322, I decided to have a go at implementing the `Galactocentric` frame attributes as being derived from a `ScienceState` subclass. It wasn't too bad in the end to get this to work, and I think this is a nicer solution than #9322, so this would replace that PR.

With this, it no longer makes sense to list citations/references to papers that contain the measurements of the Solar parameters, because the defaults can change. So here I added a `.get_references()` method that retrieves a dictionary of references (mapping parameter name to paper(s)) that is updated by and kept in sync with the parameters by the `ScienceState` instance. One remaining TODO on this, below.

This includes commits from #9345, which are necessary for the functionality in this PR.

TODO:
* [x] Changelog entry
* [x] Rebase once / if #9345 is merged
* [x] If a frame attribute is set by the user, e.g., `Galactocentric(galcen_distance=...)`, remove the reference for that frame attribute
* [x] Expose the `ScienceState` subclass at the `astropy.coordinates` level?
* [x] Documentation, especially on how to change the default parameter set
* [x] Update the link to the documentation of this new functionality in the warning raised (see TODO in code comment)

### DIFF
--- a/astropy/coordinates/builtin_frames/galactocentric.py
+++ b/astropy/coordinates/builtin_frames/galactocentric.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import warnings
+
 import numpy as np
 
 from astropy import units as u
 from astropy.utils.state import ScienceState
 from astropy.utils.decorators import format_doc
+from astropy.utils.exceptions import AstropyDeprecationWarning
 from astropy.coordinates.angles import Angle
 from astropy.coordinates.matrix_utilities import rotation_matrix, matrix_product, matrix_transpose
 from astropy.coordinates import representation as r
@@ -70,6 +73,7 @@ class galactocentric_frame_defaults(ScienceState):
     """
 
     # the default is to use the original definition of this frame
+    # TODO: change this to 'latest' in v4.1?
     _value = 'pre-v4.0'
     _references = None
 
@@ -303,13 +307,40 @@ class Galactocentric(BaseCoordinateFrame):
         self.frame_attribute_references = \
             galactocentric_frame_defaults._references.copy()
 
+        warn = False
         for k in default_params:
             # If a frame attribute is set by the user, remove its reference
             self.frame_attribute_references.pop(k, None)
 
+            # If a parameter is read from the defaults, we might want to warn
+            # the user that the defaults will change (see below)
+            if k not in kwargs:
+                warn = True
+
             # Keep the frame attribute if it is set by the user, otherwise use
             # the default value
             kwargs[k] = kwargs.get(k, default_params[k])
+
+        # If the frame defaults have not been updated with the ScienceState
+        # class, and the user uses any default parameter value, raise a
+        # deprecation warning to inform them that the defaults will change in
+        # the future:
+        if galactocentric_frame_defaults._value == 'pre-v4.0' and warn:
+            docs_link = ''  # TODO
+            warnings.warn('In v4.1 and later versions, the Galactocentric '
+                          'frame will adopt default parameters that may update '
+                          'with time. An updated default parameter set is '
+                          'already available through the '
+                          'astropy.coordinates.galactocentric_frame_defaults '
+                          'ScienceState object, as described in {} '
+                          'but the default is currently still set to '
+                          'the pre-v4.0 parameter defaults. The safest way to '
+                          'guard against changing default parameters in the '
+                          'future is to either (1) specify all Galactocentric '
+                          'frame attributes explicitly when using the frame, '
+                          'or (2) set the galactocentric_frame_defaults '
+                          'parameter set name explicitly.'.format(docs_link),
+                          AstropyDeprecationWarning)
 
         super().__init__(*args, **kwargs)
 
--- a/astropy/coordinates/builtin_frames/galactocentric.py
+++ b/astropy/coordinates/builtin_frames/galactocentric.py
@@ -260,13 +260,12 @@ doc_footer = """
             [(  86.2585249 ,  28.85773187,  2.75625475e-05),
              ( 289.77285255,  50.06290457,  8.59216010e+01)]>
 
+    .. testcleanup:: *
 
-    # Reset the state after doctests above.
-    # TODO: this needs to be audited and removed if we change the default
-    # defaults (yes) for v4.1. This is a hack to get around the fact that
-    # the doctests here actually change the *class* state::
-
-        >>> galactocentric_frame_defaults._value = 'pre-v4.0'
+        # TODO: this needs to be audited and removed if we change the default
+        # defaults (yes) for v4.1. This is a hack to get around the fact that
+        # the doctests here actually change the *class* state:
+        galactocentric_frame_defaults._value = 'pre-v4.0'
 
 """
 
--- a/astropy/coordinates/builtin_frames/galactocentric.py
+++ b/astropy/coordinates/builtin_frames/galactocentric.py
@@ -1,14 +1,11 @@
 # -*- coding: utf-8 -*-
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import warnings
-
 import numpy as np
 
 from astropy import units as u
 from astropy.utils.state import ScienceState
 from astropy.utils.decorators import format_doc
-from astropy.utils.exceptions import AstropyDeprecationWarning
 from astropy.coordinates.angles import Angle
 from astropy.coordinates.matrix_utilities import rotation_matrix, matrix_product, matrix_transpose
 from astropy.coordinates import representation as r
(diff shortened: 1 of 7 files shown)


astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -216,6 +276,12 @@ def get_roll0(cls):
         # API, so it's better for it to be accessable from Galactocentric
         return _ROLL0
 
+    def get_references(self):
+        if self._references is None:
+            raise ValueError("No references were provided or are available for "
```
mhvk: I think I'd make this a property `references` (or perhaps `frame_references`) instead, and then here raise `AttributeError`.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -30,6 +28,97 @@
 # prevent having to create new Angle objects every time `get_roll0` is called.
 _ROLL0 = Angle(58.5986320306*u.degree)
 
+
+class default_sun_galactocentric(ScienceState):
+    """The ScienceState instance that controls global setting of Galactocentric
+    default solar motion parameters
+    """
```
mhvk: Will want to elaborate a bit on this...  In particular, on how to initialize it.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -30,6 +28,97 @@
 # prevent having to create new Angle objects every time `get_roll0` is called.
 _ROLL0 = Angle(58.5986320306*u.degree)
 
+
+class default_sun_galactocentric(ScienceState):
```
mhvk: I think I'd give a bit more explicit name, like `galactocentric_frame_[definition|parameters]` or so, in analogy with things we already have, like `solar_system_ephemeris` and `earth_rotation_table`.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -118,9 +207,25 @@
...
+    This frame allows specifying the Sun-Galactic center distance, the height of
+    the Sun above the Galactic midplane, and the solar motion relative to the
+    Galactic center. However, as there is no modern standard definition of a
+    Galactocentric reference frame, it is important to pay attention to the
+    default values used in this class if precision is important in your code.
+    The default values of the parameters of this frame are taken from the
+    original definition of the frame in 2014. As such, the defaults are somewhat
+    out of date relative to recent measurements made possible by, e.g., Gaia.
+    The defaults can, however, be changed at runtime by setting the parameter
+    set name in ``astropy.coordinates.default_sun_galactocentric``. The default
```
mhvk: Should be a proper link (single back quotes)

eteq:
I also like this direction.  Two thoughts, though:

* "Expose the ScienceState subclass at the astropy.coordinates level?" - I'd say "yes" to that if the `?` meant you weren't sure, @adrn.  
* What are we thinking the long-term plan is? I asked a few Galactic astronomers about this today and they all agreed "we'll never notice this without getting some sort of warning or something". So we *could* say that the default is to do pre-4.0 but with a deprecation warning, and say in 4.1 and on you'll get "latest", but you can silence the deprecation by doing <link to the docs>...?

adrn:
@eteq on your second point, yes, I should have mentioned: I was going to suggest that we add a warning whenever the frame is used "naked" without specifying parameters. I like your idea of saying that in 4.1 on it will default to "latest".

mhvk:
On exposing the science state: is it an idea to actually define it *on* the `GalactoCentric` class? Or is this crazy?

bsipocz:
> So we could say that the default is to do pre-4.0 but with a deprecation warning, and say in 4.1 and on you'll get "latest", but you can silence the deprecation by doing ...?

how much downstream  uses this? I suppose they don't need to rely on the default but be specific?

astropy/coordinates/builtin_frames/galactocentric.py
adrn: I opted for `galactocentric_frame_defaults` to try to make it clear that these only control the default parameters, i.e. I don't want users to mistakenly think that this is the way to generically set or change the frame parameters.

adrn:
> On exposing the science state: is it an idea to actually define it on the GalactoCentric class? Or is this crazy?

Huh, I hadn't though of that! I wonder if that is feasible...

astropy/coordinates/builtin_frames/galactocentric.py
adrn: I did this, but I guess if we make it an attribute like that, we don't really need a `get_` method?

astropy/coordinates/builtin_frames/galactocentric.py
mhvk: Certainly fine (though note that `solarsystem_ephemeris` is similarly a default - it can be overridden in most routines where it is used)

adrn:
> how much downstream uses this? I suppose they don't need to rely on the default but be specific?

I am often sloppy in my own code and use the defaults. Even worse, I often update/specify 1 or 2 of the attributes but rely on the others to be set to the defaults. So I'm coming around to raising a warning if any of the defaults are left un-specified, if we are indeed thinking of changing the default behavior in v4.1 to use the 'latest' parameters.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -163,46 +305,60 @@ class Galactocentric(BaseCoordinateFrame):
...
+                          'astropy.coordinates.galactocentric_frame_defaults '
+                          'ScienceState object, as described in but the '
+                          'default is currently still set to the pre-v4.0 '
+                          'parameter defaults. The safest way to guard against '
+                          'changing default parameters in the future is to '
+                          'either (1) specify all Galactocentric frame '
+                          'attributes explicitly when using the frame, '
+                          'or (2) set the galactocentric_frame_defaults '
+                          'parameter set name explicitly. See {} for more '
+                          'information.'.format(docs_link),
```
Cadair: You should use an f-string.
```suggestion
                          'information.',
```

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -163,46 +305,60 @@ class Galactocentric(BaseCoordinateFrame):
...
+                          'already available through the '
+                          'astropy.coordinates.galactocentric_frame_defaults '
+                          'ScienceState object, as described in but the '
+                          'default is currently still set to the pre-v4.0 '
+                          'parameter defaults. The safest way to guard against '
+                          'changing default parameters in the future is to '
+                          'either (1) specify all Galactocentric frame '
+                          'attributes explicitly when using the frame, '
+                          'or (2) set the galactocentric_frame_defaults '
+                          'parameter set name explicitly. See {} for more '
```
Cadair:
```suggestion
                          f'parameter set name explicitly. See {docs_link} for more '
```

docs/coordinates/galactocentric.rst
```diff
@@ -1,11 +1,31 @@
...
-********************************************************
-Description of Galactocentric Coordinates Transformation
-********************************************************
+**************************************************
+Description of the Galactocentric Coordinate Frame
+**************************************************
+
+While many other frames implemented in `astropy.coordinates` are standardized in
+some way (e.g., defined by the IAU), there is no standard Milky Way
+Galactocentric reference frame. The `~astropy.coordinates.Galactocentric` frame
```
eteq: I think this is a good place to defince the term for those who stumble upon it.
```suggestion
reference frame with the center of the Milky Way as its origin. (This is
distinct from `~astropy.coordinates.Galactic` coordinates, which point
toward the Galactic Center but have their origin in the Solar System).
The `~astropy.coordinates.Galactocentric` frame
```

docs/coordinates/galactocentric.rst
```diff
@@ -118,3 +140,86 @@ The full transformation is then:
...
+.. _astropy-coordinates-galactocentric-defaults:
+
+Controlling the Default Frame Parameters
+========================================
+
+All of the frame-defining parameters of the
+`~astropy.coordinates.Galactocentric` frame are customizable and can be set by
+passing arguments in to the `~astropy.coordinates.Galactocentric` initializer.
+However, it is often convenient to use the frame without having to pass in every
+parameter. We have therefore tried to set reasonable default values for these
```
eteq:
```suggestion
parameter. Hence, the class comes with reasonable default values for these
```

astropy/coordinates/tests/test_frames.py
```diff
@@ -1139,3 +1139,63 @@ class NameChangeFrame(BaseCoordinateFrame):
...
+    galactocentric_frame_defaults._value = init_value
+
+
+def test_galactocentric_default_warning():
+    from astropy.coordinates import (Galactocentric,
+                                     galactocentric_frame_defaults)
+
+    # TODO: this needs to be audited and removed if we change the default
+    # defaults (yes) for v4.1. This is a hack to get around the fact that the
+    # doctests in galactocentric.py actually change the *class* state:
```
eteq: Still a bit of a hack, but maybe you can add a ``.. testsetup::`` clause in the docs that resets the science state to the default setting?  Then this wouldn't be necessary and more generally would prevent unintended consequences of the doctests...

adrn: Good point! I think we want `testcleanup` but yes, see latest commit!

adrn:
Docs build failed because sphinx didn't recognize the doctest directive `.. testcleanup::`. Hm...

32bit test failed for reasons I don't understand!

pllim:
I was just gonna say... Looking at https://github.com/sphinx-doc/sphinx/blob/1140e81df7ee9df1135ef9f35302ffd72d5bd764/tests/roots/test-ext-doctest/doctest.txt#L140-L143 , either that `*` is compulsory, or you needed a blank line after `.. testcleanup::`.

bsipocz:
One usually needs the blank lines after the directives.

pllim:
As for the test failure, any reason why under that set up, it would not emit a warning as expected?
```
        with pytest.warns(AstropyDeprecationWarning,
                          match="In v4.1 and later versions"):
>           Galactocentric()
E           Failed: DID NOT WARN. No warnings of type (<class 'astropy.utils.exceptions.AstropyDeprecationWarning'>,) was emitted. The list of emitted warnings is: [].
```

Also, `match` passes through regex, so you might want to turn it into a raw-string and maybe escape the `.`.

adrn:
RE: directive - ok, cool, will try
image


RE: 32bit - I saw the failure cause, but I can't reproduce locally. I'll try playing with the matching...

adrn:
RE: 32bit failure: I just noticed that the list of emitted warnings is empty, so it's not the regex matching. But the warning is definitely produced on my machine, and on travis...

pllim:
Well, then one of these conditions is false somehow in that job. You can try to SSH into that CI session and see?
```python
if galactocentric_frame_defaults._value == 'pre-v4.0' and warn:
```

pllim:
Seems like the 32-bit job also runs in parallel mode. Will that affect your test?

adrn:
Aha! It does seem to be an issue with parallelization...

bsipocz:
Thanks @adrn!

PR #9354, opened 2019-10-10
Use raise ... from to get tracebacks

aarchiba:
When an error occurs in format conversion, it is captured and re-raised
as a ValueError. This is necessary for format guessing but it can make
error reporting vague. Use raise ... from so that tracebacks go all the
way back to the line where the original exception was raised.

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -495,6 +495,8 @@ class Time(ShapedLikeNDArray):
         else:
             formats = [(format, self.FORMATS[format])]
 
+        assert formats
+        problems = {}
         for name, cls in formats:
             try:
                 return cls(val, val2, scale, precision, in_subfmt, out_subfmt)
@@ -504,14 +506,18 @@ class Time(ShapedLikeNDArray):
                 # If ``format`` specified then there is only one possibility, so raise
                 # immediately and include the upstream exception message to make it
                 # easier for user to see what is wrong.
-                if format is not None:
-                    raise ValueError(f'Input values did not match the format class {format}:' +
-                                     os.linesep +
-                                     f'{err.__class__.__name__}: {err}')
+                if len(formats)==1:
+                    raise ValueError(
+                        f'Input values did not match the format class {format}:'
+                        + os.linesep
+                        + f'{err.__class__.__name__}: {err}'
+                    ) from err
+                else:
+                    problems[name] = err
         else:
-            raise ValueError('Input values did not match any of the formats where '
-                             'the format keyword is optional {}'
-                             .format([name for name, cls in formats]))
+            raise ValueError(f'Input values did not match any of the formats '
+                             f'where the format keyword is optional: '
+                             f'{problems}') from problems[formats[0][0]]
 
     @classmethod
     def now(cls):
--- a/astropy/time/formats.py
+++ b/astropy/time/formats.py
@@ -86,6 +86,11 @@ class TimeFormatMeta(type):
         # it is not a user-accessible format and is only used for initialization into
         # a different format.
         if 'name' in members and cls.name != 'astropy_time':
+            # FIXME: check here that we're not introducing a collision with
+            # an existing method or attribute; problem is it could be either
+            # astropy.time.Time or astropy.time.TimeDelta, and at the point
+            # where this is run neither of those classes have necessarily been
+            # constructed yet.
             mcls._registry[cls.name] = cls
 
         if 'subfmts' in members:
--- a/astropy/time/tests/test_custom_formats.py
+++ b/astropy/time/tests/test_custom_formats.py
@@ -1,26 +1,93 @@
 import pytest
+from itertools import count
 
 from astropy.time import Time, TimeFormat
 
 
-def test_custom_time_format_exception():
-    class SpecificException(ValueError):
-        pass
+class SpecificException(ValueError):
+    pass
 
-    # Note there is no way to remove this after the test!
+
+@pytest.fixture
+def custom_format_name():
+    for i in count():
+        if not i:
+            custom = f"custom_format_name"
+        else:
+            custom = f"custom_format_name_{i}"
+        if custom not in Time.FORMATS:
+            break
+    yield custom
+    Time.FORMATS.pop(custom, None)
+
+
+def test_custom_time_format_set_jds_exception(custom_format_name):
     class Custom(TimeFormat):
-        name = "custom_time_format_test"
+        name = custom_format_name
 
         def set_jds(self, val, val2):
             raise SpecificException
 
-    # with pytest.raises(SpecificException):
-    #    Time(7., format="custom_time_format_test")
     try:
-        Time(7.0, format="custom_time_format_test")
-    except SpecificException:
-        # This is not what astropy does but it is fine too
-        pass
+        Time(7.0, format=custom_format_name)
     except ValueError as e:
-        assert (hasattr(e, "__cause__")
-                and isinstance(e.__cause__, SpecificException))
+        assert hasattr(e, "__cause__") and isinstance(e.__cause__, SpecificException)
+
+
+def test_custom_time_format_val_type_exception(custom_format_name):
+    class Custom(TimeFormat):
+        name = custom_format_name
+
+        def _check_val_type(self, val, val2):
+            raise SpecificException
+
+    try:
+        Time(7.0, format=custom_format_name)
+    except ValueError as e:
+        assert hasattr(e, "__cause__") and isinstance(e.__cause__, SpecificException)
+
+
+def test_custom_time_format_value_exception(custom_format_name):
+    class Custom(TimeFormat):
+        name = custom_format_name
+
+        def set_jds(self, val, val2):
+            self.jd1, self.jd2 = val, val2
+
+        @property
+        def value(self):
+            raise SpecificException
+
+    t = Time.now()
+    with pytest.raises(SpecificException):
+        getattr(t, custom_format_name)
+
+
+def test_custom_time_format_fine(custom_format_name):
+    class Custom(TimeFormat):
+        name = custom_format_name
+
+        def set_jds(self, val, val2):
+            self.jd1, self.jd2 = val, val2
+
+        @property
+        def value(self):
+            return self.jd1 + self.jd2
+
+    t = Time.now()
+    getattr(t, custom_format_name)
+
+
+def test_custom_time_format_forgot_property(custom_format_name):
+    class Custom(TimeFormat):
+        name = custom_format_name
+
+        def set_jds(self, val, val2):
+            self.jd1, self.jd2 = val, val2
+
+        def value(self):
+            return self.jd1, self.jd2
+
+    t = Time.now()
+    with pytest.raises(AttributeError):
+        getattr(t, custom_format_name)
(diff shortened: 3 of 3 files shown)


astropy/time/tests/test_custom_formats.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from astropy.time import Time, TimeFormat
+
+
+def test_custom_time_format_exception():
+    class SpecificException(ValueError):
+        pass
+
+    # Note there is no way to remove this after the test!
```
pllim: What do you mean? Can you `del Custom` in [module teardown](https://docs.pytest.org/en/latest/xunit_setup.html#module-level-setup-teardown) ?

astropy/time/tests/test_custom_formats.py
```diff
@@ -0,0 +1,26 @@
...
+    class Custom(TimeFormat):
+        name = "custom_time_format_test"
+
+        def set_jds(self, val, val2):
+            raise SpecificException
+
+    # with pytest.raises(SpecificException):
```
pllim: I am guessing `pytest.raises` didn't work? Please remove the commented code if it is not going to be used. Thanks!

pllim: Or you can try https://stackoverflow.com/questions/38403710/how-can-i-use-pytest-raises-with-multiple-exceptions

astropy/time/tests/test_custom_formats.py
mhvk: Agreed with @pllim, though I would make this a `class TestCustomTimeFormat`, so that you can have easy `setup` and `teardown` methods.  In `setup(self)', you can then define the format, and in `teardown` remove it, using
```
def setup(self):
    class Custom(...):
        ...

    self.Custom  = Custom

def self.teardown(self):
    Time.FORMATS.pop(self.Custom)
```
This has the advantage that we can (later) expand this class to test a custom class more thoroughly...

astropy/time/tests/test_custom_formats.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from astropy.time import Time, TimeFormat
+
+
+def test_custom_time_format_exception():
+    class SpecificException(ValueError):
```
mhvk: I'd take this outside - fine to define it locally in the test module.

astropy/time/tests/test_custom_formats.py
```diff
@@ -0,0 +1,26 @@
...
+        name = "custom_time_format_test"
+
+        def set_jds(self, val, val2):
+            raise SpecificException
+
+    # with pytest.raises(SpecificException):
+    #    Time(7., format="custom_time_format_test")
+    try:
+        Time(7.0, format="custom_time_format_test")
+    except SpecificException:
```
mhvk: I think we'd want to be warned that this exception was leaking out, so let's not allow it.

But, it *would* be good to check that the `Custom` class raises it, with a direct test
```
def test_custom_direct(self):
    with pytest.raises(SpecificException):
        self.Custom(... fill in *all* arguments...)
```

mhvk:
My sense is that since the type of exception does not change, this does not need a changelog entry (it is more like a docs change).

astropy/time/core.py
```diff
@@ -495,6 +495,8 @@ def _get_time_fmt(self, val, val2, format, scale,
         else:
             formats = [(format, self.FORMATS[format])]
 
+        assert formats
```
mhvk: @taldcroft - since we were just discussing a similar thing, this is the type of "assert that my assumptions hold" I meant in https://github.com/astropy/astropy/pull/9340#discussion_r333988907 - what if due to some bug `formats` is an empty list? Should definitely not happen, but who knows. The subsequent code would mean that what gets returned is `None`, which is not really what one wants!

Aside: not quite sure it is needed here, but it does show the intent.

astropy/time/tests/test_custom_formats.py
mhvk: Sorry, my example was somewhat wrong - edited.

aarchiba: It was the need to know about metaclass internals to deregister that I was referring to. And, more specifically, the need to make sure that different tests used different string names even though they were locally-defined classes. But de-registering them seems like a good trade-off.

Incidentally the register-when-class-created pattern seems to be the best available but it leads to some awkward situations where whether or not you imported a module affects the behaviour of code that uses no symbols from that module. In PINT my answer was to replace of "from astropy.time import Time" with "from pulsar_mjd import Time" so there's no reasonable way of constructing a Time object without first making available format="pulsar_mjd".

astropy/time/tests/test_custom_formats.py
aarchiba: I'm actually not sure what it would mean to instantiate a TimeFormat object directly? Or how to do it exactly. I'll figure it out. Or, well, the __cause__ check should be sufficient, particularly combined with the very simple code of CustomFormat.

astropy/time/tests/test_custom_formats.py
aarchiba: Yeah, sorry, this was a sort of work-in-progress thing, meant to make the test cope with whatever was decided was reasonable behaviour. Which really might be just reraising the original exception if there's only one format choice rather than wrapping it in a ValueError.

astropy/time/core.py
aarchiba: The distinction I use is, if the precondition is something the user should enforce, I raise a ValueError (or whatever) with helpful messages. If the precondition is obviously true as the result of immediately-visible astropy code, I don't check at all. But if the precondition is, as here, something astropy *ought* to enforce, but as a result of interactions of code and data structures that could easily be changed without thought to this function, then I use assert. Especially if the failure mode is hard to understand: failure to accept the input, signaled by an error raised by the attempt to index format[0] while constructing a different exception with a confusing message.

Essentially: ValueErrors (or TypeErrors or whatever) for errors caused by users or external circumstances; assertions for errors caused by people modifying astropy itself. Plus, especially given this rule, "assert" documents that the code will implode if ever there are no permitted formats left when one gets here.

astropy/time/tests/test_custom_formats.py
aarchiba: My thought was that I wanted this exception to be, well, specific to this function - no other code anywhere ought to be able to make an exception that would satisfy this test. There's no need to be quite so strict, though, and I suppose one ought to test generation of SpecificExceptions in all the various places CustomFormat might want code. (So far I think in val/val2 type validation, in construction, and in value extraction? The last poses no particular difficulty.)

aarchiba:
Traceback demo:
```
$ ipython
Type 'copyright', 'credits' or 'license' for more information
IPython 7.7.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from astropy.time import TimeFormat, Time                                                                      

In [2]: class Custom(TimeFormat): 
   ...:     name = "blerg" 
   ...:     def set_jds(self, val, val2): 
   ...:         raise ValueError("Oops!") 
   ...:                                                                                                                

In [3]: Time(7.0, format="blerg")                                                                                      
~/projects/astropy/astropy/astropy/time/core.py in _get_time_fmt(self, val, val2, format, scale, precision, in_subfmt, out_subfmt)
    501             try:
--> 502                 return cls(val, val2, scale, precision, in_subfmt, out_subfmt)
    503             except UnitConversionError:

~/projects/astropy/astropy/astropy/time/formats.py in __init__(self, val1, val2, scale, precision, in_subfmt, out_subfmt, from_jd)
    133             val1, val2 = self._check_val_type(val1, val2)
--> 134             self.set_jds(val1, val2)
    135 

<ipython-input-2-b906d5a77411> in set_jds(self, val, val2)
      3     def set_jds(self, val, val2):
----> 4         raise ValueError("Oops!")
      5 

ValueError: Oops!

The above exception was the direct cause of the following exception:

<ipython-input-3-99cbb1b875a2> in <module>
----> 1 Time(7.0, format="blerg")

~/projects/astropy/astropy/astropy/time/core.py in __init__(self, val, val2, format, scale, precision, in_subfmt, out_subfmt, location, copy)
    396         else:
    397             self._init_from_vals(val, val2, format, scale, copy,
--> 398                                  precision, in_subfmt, out_subfmt)
    399             self.SCALES = TIME_TYPES[self.scale]
    400 

~/projects/astropy/astropy/astropy/time/core.py in _init_from_vals(self, val, val2, format, scale, copy, precision, in_subfmt, out_subfmt)
    451         # Parse / convert input values into internal jd1, jd2 based on format
    452         self._time = self._get_time_fmt(val, val2, format, scale,
--> 453                                         precision, in_subfmt, out_subfmt)
    454         self._format = self._time.name
    455 

~/projects/astropy/astropy/astropy/time/core.py in _get_time_fmt(self, val, val2, format, scale, precision, in_subfmt, out_subfmt)
    512                         + os.linesep
    513                         + f'{err.__class__.__name__}: {err}'
--> 514                     ) from err
    515                 else:
    516                     problems[name] = err

ValueError: Input values did not match the format class blerg:
ValueError: Oops!

```

astropy/time/tests/test_custom_formats.py
aarchiba: I concluded two things:

- At least for these exception-raising tests, every test needs a custom time format that behaves differently. So by far the easiest way to do this is to construct a custom class in each test that needs one. This is kind of good anyway because custom formats shouldn't need a lot of boilerplate.
- Successfully installing and removing such a class is awkward and annoying, and the easiest place to put it was in a pytest fixture that returns a safe name and then deletes it after the test.

aarchiba:
Unfortunately if you create a custom format with a name that overlaps with a method or attribute of Time or TimeDelta you get hard-to-understand errors (for example "recursion depth exceeded" when you assert that something equals t.value). I can't see how to detect this upon construction of such a class, though. Aren't metaclasses fun?

mhvk:
Thanks for the traceback; looks very good!

mhvk:
@aarchiba - I think that it should not be difficult to check whether `name` conflicts with an existing attribute or method in the meta-class - but trickier might be the stuff done via `__getattr__`.

Though perhaps this is best done when we move the registration to `__init_subclass__`.

astropy/time/formats.py
```diff
@@ -86,6 +86,11 @@ def __new__(mcls, name, bases, members):
         # it is not a user-accessible format and is only used for initialization into
         # a different format.
         if 'name' in members and cls.name != 'astropy_time':
+            # FIXME: check here that we're not introducing a collision with
+            # an existing method or attribute; problem is it could be either
+            # astropy.time.Time or astropy.time.TimeDelta, and at the point
```
mhvk: Isn't the problem collisions with method names in `TimeFormat`?

mhvk: Ah, no, I now see the problem from your example with `sort`. Though, I am surprised, the name `sort` can only be looked up via `__getattr__` and that supposedly happens *after* looking for methods like `sort()`. I'm puzzled...

mhvk: And looking further I see that indeed `t.sort()` works fine - it is the `t.format = 'sort'` that screws one up. As you write, tricky to debug!!

PR #9359, opened 2019-10-12
Eliminate unnecessary 1+z factor from differential comoving volume

lpsinger:
### Description
Eliminate an unnecessary factor of 1 + z from the differential comoving volume using da = dm / (1 + z). This makes the code a little bit simpler and also makes it clearer that dVc = dm^2 d(dc).


pllim:
Thanks, @lpsinger !

PR #9368, opened 2019-10-15
Allow longdouble values through input validation

### LINKED ISSUE #9341: Impossible to input np.longdouble into custom time formats
I am trying to write a custom time format "mjd_long" which accepts and returns np.longdouble values upon conversion, to make nanosecond accuracy easier. Unfortunately during the preparation stage, before values are handed off to the custom code, they are passed through _make_array, which forcibly converts any non-approved types, including np.longdouble, into np.float64, losing accuracy in the process. As far as I can tell it is a simple matter of allowing higher-precision numerical dtypes (np.float96, np.float128, np.longdouble) through untouched. There is some delicacy about platforms like Windows/MSVC that do not have extended-precision types, where I think some of the type names are not defined. But as far as I can tell this is the only obstacle to writing a custom mjd_long type. (I think with some care plain mjd objects could be made to respect the extra precision of np.longdouble as well, but that's another issue.) 

```python
def _make_array(val, copy=False):
    """
    Take ``val`` and convert/reshape to an array.  If ``copy`` is `True`
    then copy input values.

    Returns
    -------
    val : ndarray
        Array version of ``val``.
    """
    val = np.array(val, copy=copy, subok=True)

    # Allow only float64, string or object arrays as input
    # (object is for datetime, maybe add more specific test later?)
    # This also ensures the right byteorder for float64 (closes #2942).
    if not (val.dtype == np.float64 or val.dtype.kind in 'OSUMaV'):
        val = np.asanyarray(val, dtype=np.float64)

    return val
```

aarchiba:
### Description

This pull request is to make it possible for third-party code to take higher-precision floating-point values as input, in order to more easily work with the extended precision available from astropy Time objects. This would be relevant to #9361 for example. 

Fixes #9341

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -2285,7 +2285,13 @@ def _make_array(val, copy=False):
     # Allow only float64, string or object arrays as input
     # (object is for datetime, maybe add more specific test later?)
     # This also ensures the right byteorder for float64 (closes #2942).
-    if not (val.dtype == np.float64 or val.dtype.kind in 'OSUMaV'):
+    if val.dtype == np.float64:
+        pass
+    elif val.dtype.kind in 'OSUMaV':
+        pass
+    elif val.dtype.kind == "f" and val.dtype.itemsize >= np.dtype(np.float64).itemsize:
+        pass
+    else:
         val = np.asanyarray(val, dtype=np.float64)
 
     return val
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -2285,11 +2285,9 @@ def _make_array(val, copy=False):
     # Allow only float64, string or object arrays as input
     # (object is for datetime, maybe add more specific test later?)
     # This also ensures the right byteorder for float64 (closes #2942).
-    if val.dtype == np.float64:
+    if val.dtype.kind == "f" and val.dtype.itemsize >= np.dtype(np.float64).itemsize:
         pass
     elif val.dtype.kind in 'OSUMaV':
-        pass
-    elif val.dtype.kind == "f" and val.dtype.itemsize >= np.dtype(np.float64).itemsize:
         pass
     else:
         val = np.asanyarray(val, dtype=np.float64)
--- a/astropy/time/tests/test_custom_formats.py
+++ b/astropy/time/tests/test_custom_formats.py
@@ -1,7 +1,13 @@
-import pytest
 from itertools import count
 
+import pytest
+
+import numpy as np
+
+import astropy.units as u
+from astropy._erfa import DJM0
 from astropy.time import Time, TimeFormat
+from astropy.time.utils import day_frac
 
 
 class SpecificException(ValueError):
@@ -76,6 +82,8 @@ def test_custom_time_format_fine(custom_format_name):
 
     t = Time.now()
     getattr(t, custom_format_name)
+    t2 = Time(7, 9, format=custom_format_name)
+    getattr(t2, custom_format_name)
 
 
 def test_custom_time_format_forgot_property(custom_format_name):
@@ -127,3 +135,56 @@ def test_custom_time_format_problematic_name():
 
     finally:
         Time.FORMATS.pop("sort", None)
+
+
+def test_mjd_longdouble_preserves_precision(custom_format_name):
+    class CustomMJD(TimeFormat):
+        name = custom_format_name
+
+        def _check_val_type(self, val, val2):
+            val = np.longdouble(val)
+            if val2 is not None:
+                raise ValueError("Only one value permitted")
+            return val, 0
+
+        def set_jds(self, val, val2):
+            mjd1 = np.float64(np.floor(val))
+            mjd2 = np.float64(val - mjd1)
+            self.jd1, self.jd2 = day_frac(mjd1 + DJM0, mjd2)
+
+        @property
+        def value(self):
+            mjd1, mjd2 = day_frac(self.jd1 - DJM0, self.jd2)
+            return np.longdouble(mjd1) + np.longdouble(mjd2)
+
+    m = 58000.0
+    t = Time(m, format=custom_format_name)
+    t2 = Time(m + 2 * m * np.finfo(np.longdouble).eps, format=custom_format_name)
+    assert t != t2
+    assert isinstance(getattr(t, custom_format_name), np.longdouble)
+    assert getattr(t, custom_format_name) != getattr(t2, custom_format_name)
+
+
+def test_mjd_unit_validation():
+    with pytest.raises(u.UnitConversionError):
+        Time(58000 * u.m, format="mjd")
+
+
+def test_mjd_unit_conversion():
+    assert Time(58000 * u.day, format="mjd") == Time(58000 * u.day, format="mjd")
+    assert Time(58000 * u.day, format="mjd") != Time(58000 * u.s, format="mjd")
+    assert Time(58000 * u.day, format="mjd") == Time(58000 * 86400 * u.s, format="mjd")
+
+
+@pytest.mark.parametrize("f", ["mjd", "unix", "cxcsec"])
+def test_existing_types_refuse_longdoubles(f):
+    t = np.longdouble(getattr(Time(58000, format="mjd"), f))
+    t2 = t + np.finfo(np.longdouble).eps * 2 * t
+    try:
+        tm = Time(np.longdouble(t), format=f)
+    except ValueError:
+        # Time processing makes it always ValueError not TypeError
+        return
+    else:
+        # accepts long doubles, better preserve accuracy!
+        assert Time(np.longdouble(t2), format=f) != tm
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -369,6 +369,9 @@ astropy.time
 - Time formats that do not use ``val2`` now raise ValueError instead of
   silently ignoring a provided value. [#9373]
 
+- Custom time formats can now accept floating-point types with extended
+  precision. Existing time formats raise exceptions rather than discarding
+  extended precision through conversion to ordinary floating-point. [#9368]
 
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
--- a/astropy/time/formats.py
+++ b/astropy/time/formats.py
@@ -184,9 +184,9 @@ class TimeFormat(metaclass=TimeFormatMeta):
     def _check_val_type(self, val1, val2):
         """Input value validation, typically overridden by derived classes"""
         # val1 cannot contain nan, but val2 can contain nan
-        ok1 = (val1.dtype == np.double and np.all(np.isfinite(val1)) or
+        ok1 = (val1.dtype.kind == 'f' and val1.dtype.itemsize == 8 and np.all(np.isfinite(val1)) or
                val1.size == 0)
-        ok2 = val2 is None or (val2.dtype == np.double and
+        ok2 = val2 is None or (val2.dtype.kind == 'f' and val2.dtype.itemsize == 8 and
                                not np.any(np.isinf(val2))) or val2.size == 0
         if not (ok1 and ok2):
             raise TypeError('Input values for {} class must be finite doubles'
@@ -619,9 +619,6 @@ class TimeDatetime(TimeUnique):
         if not all(isinstance(val, datetime.datetime) for val in val1.flat):
             raise TypeError('Input values for {} class must be '
                             'datetime objects'.format(self.name))
-        if val2 is not None:
-            raise ValueError(
-                f'{self.name} objects do not accept a val2 but you provided {val2}')
         return val1, None
 
     def set_jds(self, val1, val2):
@@ -901,9 +898,6 @@ class TimeString(TimeUnique):
         if val1.dtype.kind not in ('S', 'U') and val1.size:
             raise TypeError('Input values for {} class must be strings'
                             .format(self.name))
-        if val2 is not None:
-            raise ValueError(
-                f'{self.name} objects do not accept a val2 but you provided {val2}')
         return val1, None
 
     def parse_string(self, timestr, subfmts):
@@ -1135,9 +1129,6 @@ class TimeDatetime64(TimeISOT):
                                 'datetime64 objects'.format(self.name))
             else:
                 val1 = np.array([], 'datetime64[D]')
-        if val2 is not None:
-            raise ValueError(
-                f'{self.name} objects do not accept a val2 but you provided {val2}')
 
         return val1, None
 
@@ -1408,9 +1399,6 @@ class TimeDeltaDatetime(TimeDeltaFormat, TimeUnique):
         if not all(isinstance(val, datetime.timedelta) for val in val1.flat):
             raise TypeError('Input values for {} class must be '
                             'datetime.timedelta objects'.format(self.name))
-        if val2 is not None:
-            raise ValueError(
-                f'{self.name} objects do not accept a val2 but you provided {val2}')
         return val1, None
 
     def set_jds(self, val1, val2):
--- a/astropy/time/formats.py
+++ b/astropy/time/formats.py
@@ -619,6 +619,9 @@ class TimeDatetime(TimeUnique):
         if not all(isinstance(val, datetime.datetime) for val in val1.flat):
             raise TypeError('Input values for {} class must be '
                             'datetime objects'.format(self.name))
+        if val2 is not None:
+            raise ValueError(
+                f'{self.name} objects do not accept a val2 but you provided {val2}')
         return val1, None
 
     def set_jds(self, val1, val2):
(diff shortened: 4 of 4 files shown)


astropy/time/core.py
```diff
@@ -2285,7 +2285,13 @@ def _make_array(val, copy=False):
     # Allow only float64, string or object arrays as input
     # (object is for datetime, maybe add more specific test later?)
     # This also ensures the right byteorder for float64 (closes #2942).
-    if not (val.dtype == np.float64 or val.dtype.kind in 'OSUMaV'):
+    if val.dtype == np.float64:
+        pass
+    elif val.dtype.kind in 'OSUMaV':
+        pass
+    elif val.dtype.kind == "f" and val.dtype.itemsize > np.dtype(np.float64).itemsize:
```
mhvk: My sense would be just not to force at all for this case. If someone implemented a custom high-precision float dtype, then good for them!

astropy/time/core.py
```diff
@@ -2270,7 +2270,7 @@ class ScaleValueError(Exception):
     pass
 
 
-def _make_array(val, copy=False):
+def _make_array(val, copy=False, restrict_types=True):
```
mhvk: This new argument does not seem to be used at all!

astropy/time/tests/test_custom_formats.py
```diff
@@ -127,3 +135,48 @@ def value(self):
...
+def test_mjd_unit_conversion():
+    assert Time(58000 * u.day, format="mjd") == Time(58000 * u.day, format="mjd")
+    assert Time(58000 * u.day, format="mjd") != Time(58000 * u.s, format="mjd")
+    assert Time(58000 * u.day, format="mjd") == Time(58000 * 86400 * u.s, format="mjd")
+
+
+@pytest.mark.parametrize("f", ["mjd", "unix", "cxcsec"])
+def test_existing_types_refuse_longdoubles(f):
+    with pytest.raises(ValueError):
+        Time(np.longdouble(58000), format=f)
```
mhvk: Not sure we'd want these tests to fail!

aarchiba:
> Thanks! Some smaller comments, but also a general one: ideally, one could just put in higher precision floats and it would "just work". I _think_ it may be possible to do so by changing `day_frac` to special-case higher-precision - it may be possible to do so by adjusting `two_sum` with some calls that would be no-op for regular float (NOTE: NOT TESTED!)
> 
> ```
> def two_sum(a, b):
>     x = np.asanyarray(a + b, float)
>     eb = x - a  # bvirtual in Shewchuk
>     ea = x - eb  # avirtual in Shewchuk
>     eb = b - eb  # broundoff in Shewchuk
>     ea = a - ea  # aroundoff in Shewchuk
>     return x, (ea + eb).astype(float, copy=False)
> ```
> 
> It may be safer to just pre-condition the arguments, a bit like done in `quantity_day_frac`, doing a careful `divmod(val[12], 1)` before adding up the values and residuals.

This would absolutely call for testing with hypothesis, once we figured out exactly what the criterion was for successful computation. 

Well, okay, I think two_sum should do exactly what it does regardless of type it is passed: compute the sum to maximum precision available with the types. But making day_frac operate appropriately would take some doing, especially if we want it to behave well when also passed Quantities. And definitely testing.

astropy/time/core.py
aarchiba: Sorry, it's been a long day.

astropy/time/tests/test_custom_formats.py
aarchiba: Ah. Well. Hmm. I want precision loss to require explicit conversion, and none of these types preserve long double precision (yet?).

astropy/time/core.py
aarchiba: Yeah, I'm not sure I understand why this function is doing any type coercion at all so I'm hesitant to remove it. Well okay the existing type checker code chokes on integers, so we need to at least up-convert those. And I'm worried about the endianness thing, but the answer there is more tests!

mhvk:
Yes, makes sense to have `two_sum` be type-agnostic. There is already a `quantity_day_frac` so there is precedent... But, yes, substantial testing needed...

astropy/time/tests/test_custom_formats.py
mhvk: I think all of them start with `day_frac` (in `TimeFromEpoch.set_jds`) so they might all support it if that routine is adjusted.

aarchiba: The tests now all permit either outright rejection of long doubles or acceptance provided that the precision is preserved.

astropy/time/core.py
mhvk: That coercion was definitely there for integers and `float32` (and failure on complex, I think).

While testing may not be 100% complete, it is meant to be good enough that one can experiment and trust that things are not too bad if all tests pass...

aarchiba:
> Yes, makes sense to have `two_sum` be type-agnostic. There is already a `quantity_day_frac` so there is precedent... But, yes, substantial testing needed...

Sorry for the repetition, but my experience with hypothesis in getting the kinks out of PINT's numerical conversion code was extremely positive. It nailed down some difficult-to-spot bugs - including finding leap-second-day problems before I had even told it about leap second days - and programming it was easy (and had that delightful pieces-clicking-together feeling I get with Haskell).

astropy/time/core.py
aarchiba: Bleh. Lots of existing code chokes if I remove the byte swapping for doubles, and it's not so easy to see whether the underlying problem (ERFA might fail with wrong-endian data?) remains. So I still normalize the byte order of float64s.

mhvk:
I'm all for trying hypothesis on `Time` - I think it is a much easier target than units.

astropy/time/core.py
mhvk: Hmm, if ERFA fails then I didn't set the ufuncs up right. Could you make a separate issue for that?

aarchiba:
Maybe they're off-topic for this PR but I also found myself poking around to confirm that there is no data validation on the jd1 and jd2 values set in Time objects, and there is no validation of types returned by .value - but there are strict and hard-to-explain requirements on them. This means that implementing custom time formats involves a lot of impenetrable error messages. I'd like to add some validation on both ends; ensure that a Time object's jd1 and jd2 are always arrays of float64s of the same shape no matter what a custom format tries to set them to, and ensure that .value always returns an array of that same shape of whatevers, or a singleton whatever, as appropriate; also that implementors of custom time formats get helpful error messages if their .value methods don't do this correctly. (It'd sort of be nice to be able to provide element-by-element conversion funcions that would automatically be broadcast; these would be too slow for actual numeric arrays but for object arrays they're not any slower and handling arrays of arbitrary shape is awkward; see https://github.com/nanograv/PINT/blob/a4b33eae5fce21a6010e8665bfeecf9caec8cf96/src/pint/pulsar_mjd.py#L501 for the mess.)

astropy/time/core.py
aarchiba: Looking at #2942 I think the problem might just have been that astopy.time.core required a particular endianness from floats; no actual ERFA error reports are shown. I'll try relaxing the input validation and see what it breaks. I suspect nothing.

aarchiba:
So uh.

Right now we quietly ignore it if someone passes a val2 when creating a Time object for a string format. That is, the user provides some information in val2, and astropy throws it away and doesn't tell them. I do not approve.

I added some ValueErrors in this case, only to find that astropy has a test for explicitly this behaviour! Astropy's tests *demand* that we silently throw away information explicitly provided by the user. What? Can I change that? Please?

mhvk:
@aarchiba - I agree ignoring `val2` rather than checking it is `None` as appropriate is bad, and have meant to address that myself. So, happy to change, but let's do that in a separate PR!

@taldcroft  and I also discussed that somewhat of a design mistake is to have `jd1` and `jd2` on the `TimeFormat` class - it would make more sense to just let it do the transformations to and from input. 

p.s. Your `mjds_to_str` would seem rather like `Time._shaped_like_input`...

aarchiba:
> @aarchiba - I agree ignoring `val2` rather than checking it is `None` as appropriate is bad, and have meant to address that myself. So, happy to change, but let's do that in a separate PR!

I'll pull it out. I have a bad habit of fixing minor problems as they fly past, which gets them fixed but makes my PRs messy.
 
> @taldcroft and I also discussed that somewhat of a design mistake is to have `jd1` and `jd2` on the `TimeFormat` class - it would make more sense to just let it do the transformations to and from input.

This does seem awkward. We are sort of stuck with it, though properties allow for some deviousness.

> p.s. Your `mjds_to_str` would seem rather like `Time._shaped_like_input`...

Ah, yes, probably. I've had to write similar code too many times now, to get functions to behave like ufuncs in terms of correctly maintaining array-ness (scalar->scalar, array->array including zero-dimensional) and shape. It's worse needing to have multiple input arrays that are broadcast together. I thought numpy.vectorize might get me this for free but at least for strings it didn't quite. Modern numpy.vectorize can let a vector-capable function act on an entire one-dimensional array, at least, which broadens its applicability.

mhvk:
On ufuncs, see https://github.com/numpy/numpy/issues/13105 - would definitely be good to have a solution...

aarchiba:
> On ufuncs, see [numpy/numpy#13105](https://github.com/numpy/numpy/issues/13105) - would definitely be good to have a solution...

I'm not too hopeful, I remember frustration and argument about numpy zero-dimensional arrays versus scalars on the numpy mailing lists more than a decade ago. But it sure would be nice.

aarchiba:
Validation moved to #9373

aarchiba:
> > On ufuncs, see [numpy/numpy#13105](https://github.com/numpy/numpy/issues/13105) - would definitely be good to have a solution...
> 
> I'm not too hopeful, I remember frustration and argument about numpy zero-dimensional arrays versus scalars on the numpy mailing lists more than a decade ago. But it sure would be nice.

One of my ancient-times fixes was to record the input array shape (or that it wasn't an array), reshape it into a 1d array, do the vectorized calculation, then reshape it into the same shape (and unpack if it wasn't an array) on output. We could do that, though as we allow Time objects to be indexed and reshaped it would be more complicated. We could at least record whether the input was an array or not so as to know whether to strip off the array boxing on output.

astropy/time/core.py
aarchiba: Indeed all tests still pass once I remove the endianness check during construction.

mhvk:
I think the shaping, etc., and possibly returning to `np.float64` instead of python float, could be done as part of `_shaped_like_input`. Though I'm not sure in the `Time` case it matters all that much whether it is a `np.float64` or a `float` - I can only see arguments for never returning a python type, just to ensure output always has a `shape`, etc.

mhvk:
@aarchiba - on `day_frac`: like `two_sum`, it currently works fine on any type of float - it may make sense to do the conversion at the end of it - in particular, just before where the second rounding is done:
```
day = day.astype(np.float64, copy=False)
frac = frac.astype(np.float64, copy=False)
excess = np.round(frac)
...
```

aarchiba:
> @aarchiba - on `day_frac`: like `two_sum`, it currently works fine on any type of float - it may make sense to do the conversion at the end of it - in particular, just before where the second rounding is done:

The day_frac with multiplication/division is not type-agnostic, because two_prod is not - it has the "split" function with a magic constant that would need to be adapted for each type (which I think is not too hard).

More locally, I discovered in writing #9375 that you can't run jd1,jd2 through day_frac upon storage because some tests check for (and presumably some users rely upon?) their not being copied.

astropy/time/tests/test_custom_formats.py
taldcroft: I'm not sure about precision loss requiring explicit conversion (else raising an unavoidable exception).  This will outright break existing code that happens to be using long doubles, and I'm almost always against breaking code without a strong reason.

In general I prefer to be a bit pragmatic about things like this, letting users be responsible for making decisions about data types, and when it matters then requiring that they have some cognizance of the implementation details.  The second paragraph of the Time documentation starts with "All time manipulations and arithmetic operations are done internally using two 64-bit floats to represent time."  I think that is a pretty good starting point for someone to infer that providing a higher-precision input might not be helping.

But as a compromise it might be fine to issue a warning in the case of potential precision loss.

astropy/time/tests/test_basic.py
```diff
@@ -600,10 +600,9 @@ def test_init_from_time_objects(self):
 class TestVal2:
     """Tests related to val2"""
 
-    def test_val2_ignored(self):
-        """Test that val2 is ignored for string input"""
-        t = Time('2001:001', 'ignored', scale='utc')
-        assert t.yday == '2001:001:00:00:00.000'
+    def test_unused_val2_ignored(self):
```
taldcroft: Out of scope for this PR, and the test docstring no longer matches the actual test (the test does not check for a message that lets the user know we won't use it).

astropy/time/tests/test_custom_formats.py
```diff
@@ -127,3 +135,56 @@ def value(self):
...
+
+    m = 58000.0
+    t = Time(m, format=custom_format_name)
+    t2 = Time(m + 2 * m * np.finfo(np.longdouble).eps, format=custom_format_name)
+    assert t != t2
+    assert isinstance(getattr(t, custom_format_name), np.longdouble)
+    assert getattr(t, custom_format_name) != getattr(t2, custom_format_name)
+
+
+def test_mjd_unit_validation():
```
taldcroft: These two unit conversion tests are unrelated to the code change, but presuming they are just to improve coverage then fine to leave them.

astropy/time/tests/test_basic.py
aarchiba: Sorry, I was sloppy about reverting this change.

astropy/time/tests/test_custom_formats.py
mhvk: Ideally we just preserve precision in all.... If that isn't currently the case, I'm in favour of warning and setting a separate issue (I think my #9361 will solve it regardless)

astrofrog: I'm a bit of an outsider to this conversation, but if there is disagreement on here it seems like something here a kwarg could be used to implement a 'strict' mode that fails if there is precision loss?

aarchiba: Long doubles are substantially lower, have to be explicitly selected by the user, and use 50-100% more memory. If someone passes a long double into an astropy function, it is because they want extended precision. 

The astropy Time docs say "All time manipulations and arithmetic operations are done internally using two 64-bit floats to represent time." This says to me, correctly, that the Time object was constructed to make it possible to operate with extended precision. The fact that the *internal* format is based on doubles is no indication at all about how it handles its input. In fact I personally was astonished and horrified to find out that astropy silently discarded my extended precision information. Fortunately I have learned to distrust library writers and test these things.

As another example, PINT has an MCMC based fitting procedure based on the emcee package. In emcee 3.0 a numpy.linalg function was used that raises an exception when presented with long doubles, so I filed a bug report and started pursuing fixes. What I found was that prior to 3.0, emcee was silently discarding extended precision: our MCMC fitter *never* worked, but it failed in a subtle way that we were not aware of. So *please* if you can't provide extended precision, raise an exception when handed an extended-precision data type.

An explicit option to enable strict checking is missing the point: the point is that we should not surprise the user with subtle degradation of their data by default. If users want non-strict checking they can just use np.float on their arguments. The question is not how to support both behaviours, the question is how to avoid surprising our users, and the issue is made more important by the subtlety of the effect that we would be surprising them with.

taldcroft: I still think that a Warning in these cases instead of an Exception seems sufficient.  Will that not prevent users from having an expectation of precision that is not realized?

mhvk: I quite strongly feel that we should simply accept higher precision and test that we do it right as much as possible. I don't think, though, that we necessarily have to provide a way to *output* in every format.

taldcroft: Agreed.  The warning/exception question is predicated on the current behavior of not necessarily handling higher precision properly.

aarchiba: Our code currently emits a blizzard of warnings even for perfectly normal operation, so in fact people grow used to ignoring warnings. This is seen with LaTeX as well - too many warnings and they are ignored.

What is the use case where users pass long doubles in to these functions and would prefer lost precision to exceptions? Recall that manual downconversion is as easy as applying float.

mhvk: If we are not fixing the problem here, then raising an exception seems reasonable.

taldcroft: I think of exceptions for issues that are *definitely* wrong or an error, and warnings for things that *might* lead to unexpected results.  I don't think that saying that some code or package issues a blizzard of warnings is a good argument for ignoring the entire Python warning system.

If I do `Time(np.longdouble(1), format='cxcsec')`, there is no reason that should fail to execute.  It is perfectly legal and gives the result that I expect.  But I do think it is reasonable to issue a warning that about precision which an informed user can take into consideration.

It may be that somebody is already using long-double for some reason that we don't know about, and they understand what is happening, and they don't want their code to suddenly break for no good reason with astropy 4.0.  I sometimes do things that might seem irrational or downright stupid to someone else, but we all have our reasons.

mhvk: Ah, convinced the other way again. Let's just agree to fix it!

aarchiba: I agree on this - let's get something like #9361 merged and then there won't be much question left what to do with longdoubles because they will work.

Issue #7957, opened 2018-10-23
Implement support for returning Time objects in the new WCS API


astrofrog:
The new WCS API in https://github.com/astropy/astropy/pull/7326 (from APE 14) currently returns ``Quantity`` objects for time axes. However, once WCSLIB is updated to 6.2 in astropy (see https://github.com/astropy/astropy/issues/7929), it will be possible to return ``Time`` objects, so this issue is a reminder to do it once WCSLIB is updated.

Issue #4261, opened 2015-10-21
Consider making EarthLocation an ITRS subclass


mhvk:
See https://github.com/astropy/astropy/pull/4254#issuecomment-150049640
Right now, it is a `Quantity` subclass, which may not make most sense. It doesn't particularly need `obstime` though.

eteq:
To clarify just a bit more on this: it's actually a *feature* that it doesn't have an `obstime` - that's the main thing that makes it different from `ITRS`.  There's value in having an object that is a latitude/longitude/altitude but *not* necessarily at a particular time (which ITRS does have). So conceptually I don't think it can be a *subclass* of `ITRS` because `ITRS` has to fit into the transformation graph, and this *cannot* (it's degenerate how to transform to any other frame if you don't specify a time).

But what might make sense is to have it be a new frame class on its own, not hooked into the frame transform graph at all.  You'd have to use the `get_itrs` function as it exists right now to actually get something that's transformable, but that's probably good because it forces the user to think about the time.

Yet another option is to have it be hooked into the frame transform graph via ITRS.  Then users could just do standard things like ``loc.transform_to(ITRS(obstime=mytime))``.  The problem is that then they can *also* do things like ``loc.transform_to(AltAz(obstime='J2010',...))``, and you might naturally *think* that this is the same as asking for the EarthLocation in J2010... but in fact, subtly, you actually get J2000 (the default for ITRS).  So that's why I like the above approach of *requiring* the user to explicitly convert to an ITRS at a particular time - that way they are forced to think about what they actually want.

mhvk:
I don't know that it is a feature to discount that continental plates move...  But of course it is true that for many applications we simply do not care. 

It seems that in the end `EarthLocation` would be an `ITRS` position that has zero velocity. But maybe that is best done when we have support for velocities (and proper motions!).

In the meantime, there are some small steps that could go towards unifying the classes: 
- 1. Internally, `EarthLocation` could use `CartesianRepresentation` instead of be based on a `Quantity` record array.
- 2. The geodetic representations could be implemented using variants of `SphericalRepresentation` (variants, since there's `height` over a reference ellipsoid instead of `radius`).

eteq:
[...]

PR #9376, opened 2019-10-16
Implement support for Time in APE 14 FITS WCS

astrofrog:
Currently, the APE 14 machinery (for FITS WCS) returns Quantity for time axes, with a warning that this will change in future and return ``Time`` objects. This PR now adjusts the FITS WCS APE 14 implementation to understand and return ``Time`` objects.

This brought up some limitations in the APE 14 API (again!). Basically, the issue is that unlike other coordinates, time coordinates in the WCS are actually relative not absolute times, and need to be converted to absolute times using other WCS metadata. So the low-level WCS values are relative, and the high-level WCS objects need to be absolute, but we can't currently just instantiate ``Time`` objects with relative time and use kwargs to add a reference time.

So there are several options:

* Modify Time so that we can instantiate it with offsets and have a e.g. ``reference_time`` kwargs that specifies how to construct the final absolute time. However this adds complication to the ``Time`` class that is probably not desirable, and doesn't solve the issue of going the other way (see below).

* Modify APE 14 to allow an optional distinction between the class used to represent high-level coordinates and the class constructor which could be an arbitrary callable object. This PR implements this approach and updates the APE 14 base classes to describe this.

Likewise, when going from world to pixel coordinates, we need to go from high level world -> low level world -> pixel and the first step also involves going from an absolute to a relative time, which can't be done by simple attribute access on ``Time``. Therefore, I also modified the APE to allow the WCS to specify a callable instead of a string for getting numerical values from a high-level coordinate object.

Before anyone suggests subclassing ``Time`` to provide this kind of functionality built-in, that won't work since we still need users to be able to specify vanilla ``Time`` objects to convert to pixel values :)

It's unfortunate that this requires updating the APE 14 API again (with additions, nothing breaking) but the implementation is actually pretty simple. @Cadair ran into the same issues when implementing support for time in GWCS.

Remaining to-dos:

* [x] Make sure we cover all cases for time CTYPEs
* [x] Fix and add tests for FITS WCS implementation
* [ ] <s>Consider adding/updating documentation accordingly</s>

Before I do these, I want to get feedback on the approach so far though. I'd really like to get this into 4.0 if possible.

Fixes https://github.com/astropy/astropy/issues/7957

### DIFF
--- a/astropy/wcs/wcsapi/fitswcs.py
+++ b/astropy/wcs/wcsapi/fitswcs.py
@@ -388,7 +388,7 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
                     def time_from_reference_and_offset(offset):
                         if isinstance(offset, Time):
                             return offset
-                        return reference_time + offset * u.s
+                        return reference_time + u.Quantity(offset, unit=u.s)
 
                     def offset_from_time_and_reference(time):
                         return (time - reference_time).to_value(u.s)
--- a/astropy/wcs/wcsapi/fitswcs.py
+++ b/astropy/wcs/wcsapi/fitswcs.py
@@ -63,7 +63,7 @@ CTYPE_TO_UCD1 = {
     'TDB': 'time',
     'LOCAL': 'time'
 
-    # UT() is handled separately in world_axis_physical_types
+    # UT() and TT() are handled separately in world_axis_physical_types
 
 }
 
@@ -186,7 +186,7 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
         types = []
         # TODO: need to support e.g. TT(TAI)
         for ctype in self.wcs.ctype:
-            if ctype.startswith('UT('):
+            if ctype.startswith(('UT(', 'TT(')):
                 types.append('time')
             else:
                 ctype_name = ctype.split('-')[0]
@@ -305,7 +305,8 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
         # Avoid circular imports by importing here
         from astropy.wcs.utils import wcs_to_celestial_frame
         from astropy.coordinates import SkyCoord, EarthLocation
-        from astropy.time import Time
+        from astropy.time.formats import FITS_DEPRECATED_SCALES
+        from astropy.time import Time, TimeDelta
 
         components = [None] * self.naxis
         classes = {}
@@ -348,41 +349,31 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
                     else:
                         name = 'time'
 
-                    # Initalize JD offsets
-                    mjd1 = 0.
-                    mjd2 = 0.
+                    # Initialize delta
+                    reference_time_delta = None
 
                     # Extract time scale
                     scale = self.wcs.ctype[i].lower()
 
+                    # Drop sub-scales
+                    if '(' in scale:
+                        pos = scale.index('(')
+                        scale, subscale = scale[:pos], scale[pos+1:-1]
+                        warnings.warn(f'Dropping unsupported sub-scale '
+                                      f'{subscale.upper()} from scale {scale.upper()}',
+                                      UserWarning)
+
                     # TODO: consider having GPS as a scale in Time
                     # For now GPS is not a scale, we approximate this by TAI - 19s
                     if scale == 'gps':
-                        mjd2 += 19. / 86400.  # 19 seconds
+                        reference_time_delta = TimeDelta(19, format='sec')
                         scale = 'tai'
 
-                    # TDT is a deprecated name for TT
-                    elif scale == 'tdt':
-                        scale = 'tt'
-
-                    # IAT is a deprecated name for TAI
-                    elif scale == 'iat':
-                        scale = 'tai'
-
-                    # GMT is used for dates before 1972-01-01 and is continuous with UTC
-                    elif scale == 'gmt':
-                        scale = 'utc'
-
-                    # ET is used for dates before 1984-01-01 and is continuous with TT
-                    elif scale == 'et':
-                        scale = 'tt'
+                    elif scale.upper() in FITS_DEPRECATED_SCALES:
+                        scale = FITS_DEPRECATED_SCALES[scale.upper()]
 
                     elif scale not in Time.SCALES:
-                        warnings.warn(f'Unrecognized time CTYPE={self.wcs.ctype[i]}, treating as Quantity', UserWarning)
-                        continue
-
-                    mjd1 += self.wcs.mjdref[0]
-                    mjd2 += self.wcs.mjdref[1]
+                        raise ValueError(f'Unrecognized time CTYPE={self.wcs.ctype[i]}')
 
                     # Determine location
                     trefpos = self.wcs.trefpos.lower()
@@ -391,8 +382,8 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
                         # Note that some headers use TOPOCENT instead of TOPOCENTER
                         if np.any(np.isnan(self.wcs.obsgeo[:3])):
                             warnings.warn('Missing or incomplete observer location '
-                                        'information, setting location in Time to None',
-                                        UserWarning)
+                                          'information, setting location in Time to None',
+                                          UserWarning)
                             location = None
                         else:
                             location = EarthLocation(*self.wcs.obsgeo[:3], unit=u.m)
@@ -404,15 +395,21 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
                                        "supported, setting location in Time to None", UserWarning)
                         location = None
 
-                    reference_time = Time(mjd1, mjd2, format='mjd', scale=scale, location=location)
+                    reference_time = Time(self.wcs.mjdref[0],
+                                          self.wcs.mjdref[1],
+                                          format='mjd', scale=scale,
+                                          location=location)
+
+                    if reference_time_delta is not None:
+                        reference_time = reference_time + reference_time_delta
 
                     def time_from_reference_and_offset(offset):
                         if isinstance(offset, Time):
                             return offset
-                        return reference_time + u.Quantity(offset, unit=u.s)
+                        return reference_time + TimeDelta(offset, format='sec')
 
                     def offset_from_time_and_reference(time):
-                        return (time - reference_time).to_value(u.s)
+                        return (time - reference_time).sec
 
                     classes[name] = (Time, (), {}, time_from_reference_and_offset)
                     components[i] = (name, 0, offset_from_time_and_reference)
--- a/astropy/wcs/wcsapi/fitswcs.py
+++ b/astropy/wcs/wcsapi/fitswcs.py
@@ -393,7 +393,7 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
                     def offset_from_time_and_reference(time):
                         return (time - reference_time).to_value(u.s)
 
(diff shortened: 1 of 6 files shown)


astropy/wcs/wcsapi/fitswcs.py
```diff
@@ -330,12 +332,72 @@ def _get_components_and_classes(self):
...
+
+                    mjd1 += self.wcs.mjdref[0]
+                    mjd2 += self.wcs.mjdref[1]
+
+                    reference_time = Time(mjd1, mjd2, format='mjd', scale=scale)
+
+                    def time_from_reference_and_offset(offset):
+                        if isinstance(offset, Time):
+                            return offset
+                        return reference_time + offset * u.s
```
Cadair:
```suggestion
                        return reference_time + u.Quantity(offset, unit=u.s)
```
(just in case it was a quantity, also I think it's slightly faster)

astropy/wcs/wcsapi/fitswcs.py
```diff
@@ -330,12 +332,72 @@ def _get_components_and_classes(self):
...
+
+                    def offset_from_time_and_reference(time):
+                        return (time - reference_time).to_value(u.s)
```
Cadair: Is WCS always in seconds?

astropy/wcs/wcsapi/low_level_api.py
```diff
@@ -178,6 +181,12 @@ def world_axis_object_classes(self):
         * The last tuple element must be a dictionary with the keyword
           arguments required to initialize the class.
 
+        For advanced use cases, the tuple can also contain four elements, where
```
Cadair: I like this, to make it an extra, because then the type information is preserved. Why don't we add it at the end though, otherwise you end up with having to do backwards indexing to make things reliable?

astrofrog: Yes that could work and simplify things

astropy/wcs/wcsapi/fitswcs.py
astrofrog: I think for FITS WCS it is. We do the same with celestial and degrees. This is because of the WCSLIB unit normalization.

astropy/wcs/wcsapi/low_level_api.py
astrofrog: Done

astrofrog:
This is now ready for review! I'd really like to get this into 4.0 if possible. A few notes:

* This doesn't currently deal with CTYPES such as UT(...) and TT(...) because these are not scales that are supported by Time. For now we still just return Quantities for those, though we could also return ``scale='utc'`` for ``UT(...)`` and ``scale='tt'`` for ``TT(...)`` and emit a warning?

* We don't support non-geocentric locations because I think that ``TIme`` doesn't? Is there any reason that we couldn't allow ``Time.location`` to be a 3d astropy coordinate?

* I ran into the issue of GPS not being a scale again - which means I need to treat it in a special way. Is there any reason we can't go ahead and add a GPS scale (with a different name to the format, e.g. ``gps_scale``).

@mhvk @taldcroft - I've tagged you in case you want to review this since it's time related, and also I could benefit from your input for the questions above!

astropy/wcs/wcsapi/fitswcs.py
```diff
@@ -330,12 +332,93 @@ def _get_components_and_classes(self):
...
+
+                    # Initalize JD offsets
+                    mjd1 = 0.
+                    mjd2 = 0.
+
+                    # Extract time scale
+                    scale = self.wcs.ctype[i].lower()
+
+                    # TODO: consider having GPS as a scale in Time
+                    # For now GPS is not a scale, we approximate this by TAI - 19s
```
mhvk: You can do this by also including `format`, so use `scale='tai'` and `format='gps'` - though that actually doesn't allow you to put `format='mjd'` below, so never mind...

astropy/wcs/wcsapi/fitswcs.py
```diff
@@ -330,12 +332,93 @@ def _get_components_and_classes(self):
...
+                    if scale == 'gps':
+                        mjd2 += 19. / 86400.  # 19 seconds
+                        scale = 'tai'
+
+                    # TDT is a deprecated name for TT
+                    elif scale == 'tdt':
```
mhvk: Maybe use [`astropy.time.formats.FITS_DEPRECATED_SCALES`](https://github.com/astropy/astropy/blob/master/astropy/time/formats.py#L40)? (for use, see `TimeFITS`)

astropy/wcs/wcsapi/fitswcs.py
```diff
@@ -330,12 +332,93 @@ def _get_components_and_classes(self):
...
+
+                    # ET is used for dates before 1984-01-01 and is continuous with TT
+                    elif scale == 'et':
+                        scale = 'tt'
+
+                    elif scale not in Time.SCALES:
+                        warnings.warn(f'Unrecognized time CTYPE={self.wcs.ctype[i]}, treating as Quantity', UserWarning)
+                        continue
+
+                    mjd1 += self.wcs.mjdref[0]
```
mhvk: My sense would be to instantiate a `TimeDelta` here and then add that to a `Time(ref1, ref2, scale=...)`. Though this depends on how wcs actually defines the time offsets for UTC - `TimeDelta` doesn't allow that one cannot use this on leap-second days.  See general comments.

astropy/wcs/wcsapi/fitswcs.py
```diff
@@ -330,12 +332,93 @@ def _get_components_and_classes(self):
...
+                        warnings.warn(f"Observation location '{trefpos}' is not "
+                                       "supported, setting location in Time to None", UserWarning)
+                        location = None
+
+                    reference_time = Time(mjd1, mjd2, format='mjd', scale=scale, location=location)
+
+                    def time_from_reference_and_offset(offset):
+                        if isinstance(offset, Time):
+                            return offset
+                        return reference_time + u.Quantity(offset, unit=u.s)
```
mhvk: I'd do `TimeDelta(offset, format='sec')` here.

astropy/wcs/wcsapi/fitswcs.py
```diff
@@ -330,12 +332,93 @@ def _get_components_and_classes(self):
...
+
+                    def offset_from_time_and_reference(time):
+                        return (time - reference_time).to_value(u.s)
```
mhvk: Just `(time - reference_time).sec` (i.e., use `TimeDelta.sec`)

astropy/wcs/wcsapi/high_level_api.py
```diff
@@ -130,7 +130,7 @@ def world_to_pixel(self, *world_objects):
         unique_match = True
         for w in world_objects:
             matches = []
-            for key, (klass, _, _) in classes.items():
+            for key, (klass, *_) in classes.items():
```
mhvk: Didn't know this was possible, clever!

astropy/wcs/wcsapi/tests/test_fitswcs.py
```diff
@@ -445,23 +447,278 @@ def test_time_cube():
...
+
+def test_time_1d_values():
+
+    # Check that Time objects are instantiated with the correct values,
+    # scales, and formats.
+
+    header = Header.fromstring(HEADER_TIME_1D, sep='\n')
+
+    # Main supported time scales
+    for scale in ('tai', 'tcb', 'tcg', 'tdb', 'tt', 'ut1', 'utc', 'local'):
```
mhvk: Not really important, but this test suite seems to cry out for a class with a `setup` method that initializes the header and `pytest.mark.parametrize` to go through the scales.

mhvk:
On your three questions:
- For the scales, are those FITS-supported scales? If so, can we re-use some of the `TimeFITS` machinery? Overall, I'd pick dropping the sub-scale with a warning over still returning `Quantity`.
- In principle, there is no reason not to support any `SkyCoord` for `Time.location` (really, `EarthLocation` should be a coordinate anyway - #4261 - just have to make sure we then disable things like `sidereal_time` with a suitable error message.
- Yes, `gps` would be better as a scale, not a format. But not that little work, since it is not in the erfa transformations, which means it needs quite a bit of special-casing.

astropy/wcs/wcsapi/high_level_api.py
```diff
@@ -147,7 +147,12 @@ def world_to_pixel(self, *world_objects):
 
         if unique_match:
 
-            for key, (klass, args, kwargs) in classes.items():
+            for key, (klass, args, kwargs, *rest) in classes.items():
```
Cadair: This means there is no validation that this tuple is the right length? Is that deliberate and if so why?

(Also this probably isn't the best place for that, we should write something that tests your compliance with APE 14 :thinking: :rabbit: :hole: )

astropy/wcs/wcsapi/tests/test_fitswcs.py
Cadair:
```python

@pytest.fixture(params=('tai', 'tcb', 'tcg', 'tdb', 'tt', 'ut1', 'utc', 'local'))
def time_header_with_scale(request):
    header = Header.fromstring(HEADER_TIME_1D, sep='\n')
    header['CTYPE1'] = request.param.upper()
    return header
```

astrofrog:
@mhvk - thanks for the review!

> Anyway probably too much hassle so close to 4.0, but it might make sense to mark your fourth parameter as experimental or so, so that in principle you could roll it back.

Yes I agree it would be nice to be able to do this properly without needing this fourth argument. I think it's a good idea to keep the option of the fourth element in the APE 14 API though as there might be other unforeseen cases where this could be needed.

> p.s. For the reference time from a FITS header, there seems to be duplication with what was done to make reading Time from tables possible (http://docs.astropy.org/en/latest/io/fits/usage/table.html#fits-tables-with-time-columns).

Yes I agree we could factor out some of the common code. In the interest of time I'd like to defer this to after feature freeze though since this is just code cleanup and not user-facing API, and time is very short this week.

astropy/wcs/wcsapi/fitswcs.py
astrofrog: Yeah I went through the same reasoning :) I'll mark this as resolved.

astropy/wcs/wcsapi/fitswcs.py
astrofrog: Yes good plan, can do that

astropy/wcs/wcsapi/high_level_api.py
astrofrog: Yes it's very handy!

astropy/wcs/wcsapi/high_level_api.py
astrofrog: we can validate this in the if statement below though, to make sure rest has length 0 or 1?

astropy/wcs/wcsapi/fitswcs.py
astrofrog: Ok I've changed this a bit along the lines you suggested (will push changes soon)

astropy/wcs/wcsapi/fitswcs.py
astrofrog: Done

astropy/wcs/wcsapi/fitswcs.py
astrofrog: Done

astropy/wcs/wcsapi/high_level_api.py
astrofrog: Added some validation

astropy/wcs/wcsapi/tests/test_fitswcs.py
astrofrog: Done!

astrofrog:
Provided tests pass, this should be good to go with the caveat that some duplication could be reduced in a follow up PR.

@mhvk @Cadair - does this seem ok to you?

Cadair:
Looks great!

PR #9378, opened 2019-10-16
fix incorrect markup on emacs setup page

DougBurke:
The emacs example at  https://docs.astropy.org/en/v3.2.2/development/codeguide_emacs.html#delete-trailing-white-spaces is not being rendered correctly. I assume this is because the `::` caused the code-block directive to be rendered as an element, rather than as a directive.

I have not run this through a sphinx build to check whether it fixes the problem or not.

### DIFF
--- a/docs/development/codeguide_emacs.rst
+++ b/docs/development/codeguide_emacs.rst
@@ -40,7 +40,7 @@ Delete trailing white spaces
 One can `delete trailing whitespace
 <https://www.emacswiki.org/emacs/DeletingWhitespace#toc3>`_ with ``M-x
 delete-trailing-whitespace``. To ensure this is done every time a python file
-is saved, use::
+is saved, use:
 
 .. code-block:: scheme
 


DougBurke:
Yay - my guess was right: https://46894-2081289-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/development/codeguide_emacs.html?highlight=emacs#delete-trailing-white-spaces

saimn:
The PR should have been opened on master, not stable ...
@bsipocz : what's the best way to fix things?

pllim:
@DougBurke , you need to try again with that patch against `master`. Please don't open PRs against anything but the `master` branch. Also if it is documentation fix, you can put `[docs only]` in your commit message to save some CI CPU cycles. Thank you.

DougBurke:
Just to say that

a) being told I've done something wrong when I am just trying to make a
simple fix
b) the first thing that happens when a PR like this is made is that you
instantly see failures since there's no associated changelog entry or
bugfix or whatever

are not particularly welcoming to the casual submitter. I understand
there's documentation about part b, but the visceral reaction I had to the
"red failure mark" was surprising.

Doug

DougBurke:
@pllim I just tried a 'git rebase upstream/master' (where upstream is astropy) but there were conflicts in files un-related to these changes. I think it'll be easier just to redo the change - since it's only a one-character deletion - as a new pr.

DougBurke:
I've created #9400 with what I hope is the correct parent for the change.

PR #9380, opened 2019-10-16
TST: constants pickle test clean-up

pllim:
### Description

This pull request is a direct follow-up of #9377 to remove unused parametrization in the test.

### DIFF
--- a/astropy/constants/tests/test_pickle.py
+++ b/astropy/constants/tests/test_pickle.py
@@ -10,11 +10,8 @@ originals = [const.Constant('h_fake', 'Not Planck',
                             system='si'),
              const.h,
              const.e.si]
-xfails = [False, False, False]
 
 
-@pytest.mark.parametrize(("original", "xfail"), zip(originals, xfails))
-def test_new_constant(pickle_protocol, original, xfail):
-    if xfail:
-        pytest.xfail()
+@pytest.mark.parametrize("original", originals)
+def test_new_constant(pickle_protocol, original):  # noqa
     check_pickling_recovery(original, pickle_protocol)


astropy/constants/tests/test_pickle.py
```diff
@@ -10,11 +10,8 @@
...
              const.e.si]
-xfails = [False, False, False]
 
 
-@pytest.mark.parametrize(("original", "xfail"), zip(originals, xfails))
-def test_new_constant(pickle_protocol, original, xfail):
-    if xfail:
-        pytest.xfail()
+@pytest.mark.parametrize("original", originals)
+def test_new_constant(pickle_protocol, original):  # noqa
```
mhvk: Out of curiosity, why the `noqa`? I thought test files were excluded anyway...

pllim: It is complaining about redefinition of unused "pickle_protocol". :woman_shrugging:

mhvk: It is actually weird that it is in the `test_new_constant` definition, since the code below should just take whatever was imported. Have you tried removing it? (from the function definition)

pllim: I am not familiar with such `pytest` fixture magic. If you say to remove it, I'll try.

pllim: Nope, looks like you need to pass it in; can't remove it -- https://travis-ci.org/astropy/astropy/jobs/598740888

I am dropping the second commit.

mhvk: OK, don't understand it really either, but probably it is some type of fixture as well. Will merge, since this cleans things up nicely anway.

PR #9390, opened 2019-10-17
Ignore reserved keywords when creating FITS header cards from Table meta

### LINKED ISSUE #9387: `table_to_hdu` warns about reserved keywords 'NAXIS', 'NAXIS1', but does not ignore them
### Description
When converting a table with existing keys with reserved keywords in `table.meta`, `table_to_hdu` duly warns
`Meta-data keyword NAXIS1 will be ignored since it conflicts with a FITS reserved keyword`
but does not in fact ignore or update them to the correct values.
### Expected behavior
E.g. on writing such a table, the header should be written with the correct dimensions, but instead
```
>>> spec = Table([np.arange(5.), np.ones(5)], names=['wave', 'flux'], meta={'NAXIS1': 42, 'EXPTIME': 80.0})
>>> spec.write('table.fits', format='fits', overwrite=True)
WARNING: Meta-data keyword NAXIS1 will be ignored since it conflicts with a FITS reserved keyword [astropy.io.fits.convenience]
>>> fits.open('table.fits')[1].header
XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                   42 / length of dimension 1                          
NAXIS2  =                    5 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
TFIELDS =                    2 / number of table fields                         
TTYPE1  = 'wave    '                                                            
TFORM1  = 'D       '                                                            
TTYPE2  = 'flux    '                                                            
TFORM2  = 'D       '                                                            
EXPTIME =                 80.0
```
Similarly, a conflicting `NAXIS = 1` key gives a warning, but in this case fails the final verification before writing the file:
```
astropy.io.fits.verify.VerifyError: 
Verification reported errors:
HDU 1:
    NAXISj keyword out of range ('NAXIS2' when NAXIS == 1)
    'PCOUNT' card at the wrong place (card 5).
    'GCOUNT' card at the wrong place (card 6).
    'NAXIS' card has invalid value '1'.
```
whereas `NAXIS2` is replaced with the correct table dimension; so this seems to be updated at a later point.

It appears that the responsible block in
https://github.com/astropy/astropy/blob/master/astropy/io/fits/convenience.py#L564
```
    for key, value in table.meta.items():
        if is_column_keyword(key.upper()) or key.upper() in REMOVE_KEYWORDS:
            warnings.warn(
                "Meta-data keyword {} will be ignored since it conflicts "
                "with a FITS reserved keyword".format(key), AstropyUserWarning)
```
may just be missing a `continue` after the warning...

### System Details
```
Darwin-16.7.0-x86_64-i386-64bit
Python 3.7.4 (default, Aug 14 2019, 00:44:22) 
Numpy 1.17.1
Scipy 1.3.0
astropy 3.2.2
```
same for `astropy 4.0.dev26100`

dhomeier:
Fixes #9387 - `table_to_hdu` now passes over cards with reserved keywords found in `table.meta` after warning about them. One word of code added, plus test and additional test for cards with incompatible data types (`array`, `dict`).

### DIFF
--- a/astropy/io/fits/tests/test_convenience.py
+++ b/astropy/io/fits/tests/test_convenience.py
@@ -15,6 +15,7 @@ from astropy.tests.helper import catch_warnings
 from astropy.utils.exceptions import AstropyUserWarning
 
 from . import FitsTestCase
+from ..connect import REMOVE_KEYWORDS
 
 
 class TestConvenience(FitsTestCase):
@@ -110,9 +111,7 @@ class TestConvenience(FitsTestCase):
         assert hdu.header.get('PCOUNT') == 0
         np.testing.assert_almost_equal(hdu.header.get('EXPTIME'), 3.21e1)
 
-    @pytest.mark.parametrize('card', ['XTENSION', 'BITPIX', 'PCOUNT', 'GCOUNT',
-                                      'NAXIS', 'NAXIS1', 'NAXIS2',
-                                      'TFIELDS', 'THEAP'])
+    @pytest.mark.parametrize('card', REMOVE_KEYWORDS)
     def test_table_to_hdu_warn_reserved(self, card):
         """
         Test warning for each keyword in ..connect.REMOVE_KEYWORDS, 1 by 1
--- a/astropy/io/fits/tests/test_convenience.py
+++ b/astropy/io/fits/tests/test_convenience.py
@@ -110,6 +110,34 @@ class TestConvenience(FitsTestCase):
         assert hdu.header.get('PCOUNT') == 0
         np.testing.assert_almost_equal(hdu.header.get('EXPTIME'), 3.21e1)
 
+    @pytest.mark.parametrize('card', ['XTENSION', 'BITPIX', 'PCOUNT', 'GCOUNT',
+                                      'NAXIS', 'NAXIS1', 'NAXIS2',
+                                      'TFIELDS', 'THEAP'])
+    def test_table_to_hdu_warn_reserved(self, card):
+        """
+        Test warning for each keyword in ..connect.REMOVE_KEYWORDS, 1 by 1
+        """
+        diagnose = 'be ignored since it conflicts with a FITS reserved keyword'
+        res_cards = {'XTENSION': 'BINTABLE', 'BITPIX': 8,
+                     'NAXIS': 2, 'NAXIS1': 12, 'NAXIS2': 3,
+                     'PCOUNT': 0, 'GCOUNT': 1, 'TFIELDS': 2, 'THEAP': None}
+        ins_cards = {'XTENSION': 'TABLE', 'BITPIX': 16,
+                     'NAXIS': 1, 'NAXIS1': 2, 'NAXIS2': 6,
+                     'PCOUNT': 2, 'GCOUNT': 2, 'TFIELDS': 4, 'THEAP': 36}
+
+        table = Table([[1.0, 2.0, 3.0], [2.3, 4.5, 6.7]],
+                      names=['wavelength', 'flux'], dtype=['f8', 'f4'])
+        table.meta['ORIGIN'] = 'Min.Silly Walks'
+        table.meta[card] = ins_cards[card]
+        assert table.meta.get(card) != res_cards[card]
+
+        with pytest.warns(AstropyUserWarning,
+                          match=f'Meta-data keyword {card} will {diagnose}'):
+            hdu = fits.table_to_hdu(table)
+
+        assert hdu.header.get(card) == res_cards[card]
+        assert hdu.header.get('ORIGIN') == 'Min.Silly Walks'
+
     def test_table_to_hdu_filter_incompatible(self):
         """
         Test removal of unsupported data types from header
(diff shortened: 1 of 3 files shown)


dhomeier:
@saimn any chance you find time for a review before Friday?

dhomeier:
Thanks @saimn, I've heard a bit about the unrest; be safe!

> I think it would be useful to test explicitly each warning, because currently in the tests it is not clear which keyword will cause a warning. If you have time to address this it would be great but otherwise we can merge as is.

This should hopefully enforce an update to the tests if `REMOVE_KEYWORDS` ever are to be extended.

Note that this adds some updates to f-strings in the warning and error messages; I manually tested the `UnitScaleError` with the new version, but am not sure if there are unit tests for all exceptions.

saimn:
Thanks @dhomeier !
For the next time it is better to keep changes like the f-string ones for another PR, to avoid mixing unrelated changes.

astropy/io/fits/tests/test_convenience.py
```diff
@@ -89,6 +91,68 @@ def test_table_to_hdu_convert_comment_convention(self):
...
+                     'PCOUNT': 0, 'GCOUNT': 1, 'TFIELDS': 2, 'THEAP': None}
+        ins_cards = {'XTENSION': 'TABLE', 'BITPIX': 16,
+                     'NAXIS': 1, 'NAXIS1': 2, 'NAXIS2': 6,
+                     'PCOUNT': 2, 'GCOUNT': 2, 'TFIELDS': 4, 'THEAP': 36}
+
+        table = Table([[1.0, 2.0, 3.0], [2.3, 4.5, 6.7]],
+                      names=['wavelength', 'flux'], dtype=['f8', 'f4'])
+        table.meta['ORIGIN'] = 'Min.Silly Walks'
+        table.meta[card] = ins_cards[card]
+        assert table.meta.get(card) != res_cards[card]
```
saimn: I don't understand what you're doing here...
What I had in mind is that when doing `fits.table_to_hdu(table)` you can get several (possibly different) warnings, and `pytest.warns` does not allow to check which warning is raised.

saimn: For instance:
```
In [2]:  
   ...:         table = Table([[1, 2, 3], ['a', 'b', 'c'], [2.3, 4.5, 6.7]], 
   ...:                       names=['a', 'b', 'c'], dtype=['i4', 'U1', 'f8']) 
   ...:         table.meta.update({'EXPTIME': 32.1, 'XTENSION': 'NEWTABLE', 
   ...:                            'NAXIS': 1, 'NAXIS1': 3, 'NAXIS2': 9, 
   ...:                            'PCOUNT': 42, 'OBSERVER': 'Adams'})          

In [3]: fits.table_to_hdu(table)                                                
WARNING: Meta-data keyword XTENSION will be ignored since it conflicts with a FITS reserved keyword [astropy.io.fits.convenience]
WARNING: Meta-data keyword NAXIS will be ignored since it conflicts with a FITS reserved keyword [astropy.io.fits.convenience]
WARNING: Meta-data keyword NAXIS1 will be ignored since it conflicts with a FITS reserved keyword [astropy.io.fits.convenience]
WARNING: Meta-data keyword NAXIS2 will be ignored since it conflicts with a FITS reserved keyword [astropy.io.fits.convenience]
WARNING: Meta-data keyword PCOUNT will be ignored since it conflicts with a FITS reserved keyword [astropy.io.fits.convenience]
```
Here I think it would be good to test that the warnings are the expected ones.

dhomeier: Now I am not sure what kind of test you had in mind...
`pytest.warns(..., match=f'Meta-data keyword {card} will {diagnose}')`
does check for the specific keyword, just one at a time. Or did you intend to check them all at once, e.g. using `catch_warnings() as w` or `pytest.warns() as w` and then go through all the lines in `w`? Not sure if the warning lines will have a predetermined order, so this could be a bit trickier, and I do not see much advantage in this variant, except for checking that multiple reserved keywords indeed raise a warning for each of them.

saimn: My point is to be sure that you get the expected number of warnings, 5 for the example above. Or said another way, that for each reserved keyword put in `table.meta` you get a warning.
So yes maybe with `catch_warnings() as w` (I don't know if this is possible with `pytest.warns` ?). And maybe using fewer reserved keywords in the test to simplify a bit ?

dhomeier: Testing the number is simple; and as the warnings do appear to pop up in the order of the keys (which are always ordered in a 3.6+ dict, as @pllim has let me know ;-), it is even possible to test for the specific keywords. Not sure if we can rely on the warnings to stay in order, though.
Since the second test now checks each reserved keywords, I guess we can do here with fewer - my main objective was to have the `NAXIS` keywords covered.

dhomeier: Oh, and this change assumes that no other warnings are emitted...

saimn: I think that the order should stay, indeed. And I actually missed that `res|ins_cards` contain all the keywords in `REMOVE_KEYWORDS`. So I think we can go with this, thanks @dhomeier !

dhomeier: Thanks for the review, @saimn !

PR #9396, opened 2019-10-17
MNT: Use return_units for Box1D

pllim:
### Description

This pull request is to utilize `return_units` in `Box1D` model.

I am trying to understand when to use `return_units` and when to hardcode unit conversion within `evaluate`. I am hoping that this PR would help me understand such things. From the documentation, I don't see why we cannot use `return_units` for `Box1D`. Did I miss something? The tests passed for me locally.

cc @karllark


karllark:
There is a case where return_units does not cover.  When the parameter has units, but the value passed to evaluate does not have units.  This is the case with fitting if I understand things correctly and why the evaluate function would need specific code.  I think.  Looking forward to hearing from others.

PR #9401, opened 2019-10-18
add numpy ufuncs as models

nden:
This PR adds many of the numpy ufuncs as models in astropy.modeling.

The schema is defined in spacetelescope/asdf-standard#210.

### DIFF
--- a/astropy/modeling/math_functions.py
+++ b/astropy/modeling/math_functions.py
@@ -42,10 +42,12 @@ def ufunc_model(name):
     nout = ufunc.nout
     if nin == 1:
         separable = True
+
         def evaluate(self, x):
             return self.func(x)
     else:
         separable = False
+
         def evaluate(self, x, y):
             return self.func(x, y)
 
--- a/astropy/modeling/math_functions.py
+++ b/astropy/modeling/math_functions.py
@@ -3,6 +3,7 @@ import warnings
 import numpy as np
 
 from astropy.modeling.core import Model
+from astropy.utils.exceptions import AstropyUserWarning
 
 
 trig_ufuncs = ["sin", "cos", "tan", "arcsin", "arccos", "arctan", "arctan2",
@@ -26,10 +27,6 @@ class _NPUfuncModel(Model):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
-    def evaluate(self, x):
-        warnings.warn("Models in math_functions are experimental.")
-        return self.func(x)
-
 
 def _make_class_name(name):
     """ Make a ufunc model class name from the name of the ufunc. """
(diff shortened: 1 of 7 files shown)


astropy/modeling/math_functions.py
```diff
@@ -0,0 +1,67 @@
...
+
+from astropy.modeling.core import Model
+
+
+trig_ufuncs = ["sin", "cos", "tan", "arcsin", "arccos", "arctan", "arctan2",
+               "hypot", "sinh", "cosh", "tanh", "arcsinh", "arccosh",
+               "arctanh", "deg2rad", "rad2deg"]
+
+
+math_ops = ["add", "subtract", "multiply", "divide", "logaddexp", "logaddexp2",
```
perrygreenfield: Should we expose add, subtract, multiply, divide, and power since they are already implicitly in modeling?

nden: These take two inputs and add, ..., them, e.g. two arrays.
Model.add adds the result of two Model.evaluate functions.
The former can be achieved with the latter but is way more handy to have it as a ufunc.

astropy/modeling/math_functions.py
```diff
@@ -0,0 +1,67 @@
...
+supported_ufuncs = trig_ufuncs + math_ops
+
+
+class _NPUfuncModel(Model):
+    _is_dynamic = True
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def evaluate(self, x):
```
perrygreenfield: Does this ever get called since the factory function below supplies a new definition? If not, should the warning go in the init instead?

nden: Ah, right, I moved the definition to the factory function and forgot to delete this one.

astropy/modeling/tests/test_math_func.py
```diff
@@ -0,0 +1,18 @@
...
+
+import numpy as np
+from numpy.testing import assert_allclose#, assert_quantity_allclose
+
+from .. import math_functions
+
+x = np.linspace(-20, 360, 100)
+y = np.linspace(-20, 360, 100)
+
+def test_math():
```
perrygreenfield: Perhaps for more persuasiveness, use ufunc= getattr(np, name) and test against that instead of model.func?

nden:
Thanks @perrygreenfield . I addressed the comments, now waiting for the tests to finish.

pllim:
Looks interesting. Do we need a "what's new" entry for this?

nden:
@pllim I'm not sure if should mention all new models in "what's new". Though we could mention this as it provides different type of models.

pllim:
Now that I look a little more, I also don't see any user-facing examples. I wouldn't have any idea on how to use them.

nden:
I am working on a documentation PR. As it is they are all dynamically generated and there's no space for an example in docstrings.

PR #9402, opened 2019-10-18
Add info on creating multi-ext files to "Getting started" page of io.fits

eigenbrot:
The concept of using ``ImageHDU`` objects to construct files with more than one extension is not really addressed on the Getting Started page. This PR attempts to add that information in a simple way.

Maybe this is beyond what should be on this page, but I'm not sure it's explicitly called out anywhere else.

### DIFF
--- a/docs/io/fits/index.rst
+++ b/docs/io/fits/index.rst
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
 This will write a single HDU to a FITS file without having to manually
 encapsulate it in an :class:`HDUList` object first.
 
+Creating a Multi-Extension Image File
+"""""""""""""""""""""""""""""""""""""
+
+In the previous example we created an :class:`HDUList` object with a single
+extension (a :class:`PrimaryHDU`). To create a file with multiple extensions we need to
+use :class:`ImageHDU`\s and append them to an :class:`HDUList`.
+
+First, let's create some more data::
+
+    >>> n2 = np.random.random((3, 3)) # A random 3x3 array
+    >>> n3 = np.random.random((100, 100)) # A random 100x100 array
+
+Note that the data shapes of the different extensions do not need to be the same. Next
+place the data into separate :class:`ImageHDU` objects::
+
+    >>> hdu2 = fits.ImageHDU(n2)
+    >>> hdu3 = fits.ImageHDU(n3)
+
+Now when we create the :class:`HDUList` we simply list all extensions we want to include::
+
+    >>> hdul = fits.HDUList([hdu, hdu2, hdu3])
+
+Because :class:`HDUList` acts like a :class:`list` we can also append an :class:`ImageHDU`
+to an already-existing :class:`HDUList`::
+
+    >>> hdul.append(hdu3)
+
+Multi-extension :class:`HDUList` are treated just like those with only a :class:`PrimaryHDU`,
+so to save the file use :func:`HDUList.writeto` as shown above.
+
+.. note::
+
+    The FITS standard enforces all files to have exactly one :class:`PrimaryHDU` that is the
+    first HDU present in the file. This standard is enforced during the call to
+    :func:`HDUList.writeto` and an error will be raised if it is not met. See the ``output_verify``
+    option in :func:`HDUList.writeto` for ways to fix or ignore these warnings.
 
 Creating a New Table File
 ^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/io/fits/index.rst
+++ b/docs/io/fits/index.rst
@@ -575,7 +575,7 @@ Next, we create a :class:`PrimaryHDU` object to encapsulate the data::
 
     >>> hdu = fits.PrimaryHDU(n)
 
-We then create a HDUList to contain the newly created primary HDU, and write to
+We then create an :class:`HDUList` to contain the newly created primary HDU, and write to
 a new file::
 
     >>> hdul = fits.HDUList([hdu])
--- a/docs/io/fits/index.rst
+++ b/docs/io/fits/index.rst
@@ -589,42 +589,6 @@ lines to accomplish the same behavior::
 This will write a single HDU to a FITS file without having to manually
 encapsulate it in an :class:`HDUList` object first.
 
-Creating an Image File with Multiple Extensions
-"""""""""""""""""""""""""""""""""""""""""""""""
-
-In the previous example we created an :class:`HDUList` object with a single
-extension (a :class:`PrimaryHDU`). To create a file with multiple extensions we need to
-use :class:`ImageHDU`\s and append them to an :class:`HDUList`.
-
-First, we create some more data::
-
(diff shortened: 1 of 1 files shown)


pllim:
Thanks, @eigenbrot ! For future commits, please add  a `[docs only]` directive in your commit message to save some CPU cycles on Travis CI, as this PR only touches the documentation.

eigenbrot:
Whoa, awesome. Thanks for the tip, @pllim!

docs/io/fits/index.rst
```diff
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
...
+Creating a Multi-Extension Image File
+"""""""""""""""""""""""""""""""""""""
+
+In the previous example we created an :class:`HDUList` object with a single
+extension (a :class:`PrimaryHDU`). To create a file with multiple extensions we need to
+use :class:`ImageHDU`\s and append them to an :class:`HDUList`.
+
+First, let's create some more data::
+
+    >>> n2 = np.random.random((3, 3)) # A random 3x3 array
```
MSeifert04: Since the actual data isn't important, I would prefer to use `np.zeros`, `np.ones`, `np.arange`, or `np.linspace`.

Generating lots of random numbers can be slow and unnecessarily slows down examples.

docs/io/fits/index.rst
```diff
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
...
+
+    >>> hdu2 = fits.ImageHDU(n2)
+    >>> hdu3 = fits.ImageHDU(n3)
+
+Now when we create the :class:`HDUList` we simply list all extensions we want to include::
+
+    >>> hdul = fits.HDUList([hdu, hdu2, hdu3])
+
+Because :class:`HDUList` acts like a :class:`list` we can also append an :class:`ImageHDU`
+to an already-existing :class:`HDUList`::
```
MSeifert04: I don't think the `-` between already and existing is needed, or is it?

docs/io/fits/index.rst
```diff
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
...
+
+Note that the data shapes of the different extensions do not need to be the same. Next
+place the data into separate :class:`ImageHDU` objects::
+
+    >>> hdu2 = fits.ImageHDU(n2)
+    >>> hdu3 = fits.ImageHDU(n3)
+
+Now when we create the :class:`HDUList` we simply list all extensions we want to include::
+
+    >>> hdul = fits.HDUList([hdu, hdu2, hdu3])
```
MSeifert04: Maybe just use `[hdu, hdu2]` because we append the third one in the next example.

MSeifert04:
Not really relevant for this PR but the index.rst is getting really long. Maybe we need to split it at some point...

pllim:
@MSeifert04 , feel free to open a new issue about splitting up the `index.rst` file but I agree that splitting it up is out of scope for this PR.

pllim:
If the latest round of edits are satisfactory, please approve. Thanks!

docs/io/fits/index.rst
```diff
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
...
 encapsulate it in an :class:`HDUList` object first.
 
+Creating an Image File with Multiple Extensions
+"""""""""""""""""""""""""""""""""""""""""""""""
+
+In the previous example we created an :class:`HDUList` object with a single
+extension (a :class:`PrimaryHDU`). To create a file with multiple extensions we need to
+use :class:`ImageHDU`\s and append them to an :class:`HDUList`.
+
+First, let's create some more data::
```
lglattly:
```suggestion
First, we create some more data::
```

docs/io/fits/index.rst
```diff
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
...
+
+    >>> n2 = np.ones((3, 3))
+    >>> n3 = np.ones((100, 100))
+
+Note that the data shapes of the different extensions do not need to be the same. Next
```
lglattly:
```suggestion
Note that the data shapes of the different extensions do not need to be the same. Next,
```

docs/io/fits/index.rst
```diff
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
...
+place the data into separate :class:`ImageHDU` objects::
+
+    >>> hdu2 = fits.ImageHDU(n2)
+    >>> hdu3 = fits.ImageHDU(n3)
+
+Now when we create the :class:`HDUList` we simply list all extensions we want to include::
```
lglattly:
```suggestion
Now when we create the :class:`HDUList` we list all extensions we want to include::
```

docs/io/fits/index.rst
```diff
@@ -589,6 +589,42 @@ lines to accomplish the same behavior::
...
+First, we create some more data::
+
+    >>> n2 = np.ones((3, 3))
+    >>> n3 = np.ones((100, 100))
+
+Note that the data shapes of the different extensions do not need to be the same. Next,
+place the data into separate :class:`ImageHDU` objects::
+
+    >>> hdu2 = fits.ImageHDU(n2)
+    >>> hdu3 = fits.ImageHDU(n3)
```
saimn: It would be useful to mention here the possibility to give a name to HDUs. Also the variable names starting at 2 look a bit weird, why not using a more real word like example ? Suggestion:
```
    >>> data_hdu = fits.ImageHDU(data=np.zeros(10), name='DATA')
    >>> var_hdu = fits.ImageHDU(data=np.ones(10), name='VAR')
    >>> hdul = fits.HDUList([data_hdu, var_hdu])
```
And then with 'DQ' for the third extension ?

eigenbrot: I'm going to respectfully push back on this. The additions in this PR are designed to follow naturally from the previous section that details creating a single-HDU file with a ``PrimaryHDU``. In other words, I view this new subsection as an extension of its parent section and as such I think the form should be the same.

The point here is that it's almost trivially easy to add another HDU with nothing more than a numpy array. I'd worry that throwing `data` and `name` keywords will make this look more complicated than the ``PrimaryHDU`` case. The finer details you mentioned are still only a click a way for those more curious than the basic example.

saimn: Actually after reading the whole section, I think that this should be a new subsection after the "Creating a New Table File" and at the same level. Because multi-extension files can contain table as well I would rename to "Creating a file with Multiple Extensions", and maybe include a table extension. Also I think that the variable names should make sense even when you don't read the previous sections.

eigenbrot: That's a good point, @saimn, the New Table section already has some of this information in it, and because an HDUList can have any type of HDU in it it makes sense to make this a separate section (not a subsection of the image section). I've moved the next section to be after the New Table section and consolidated the information.

saimn:
Thanks @eigenbrot for the nice improvement!

PR #9415, opened 2019-10-19
Minor formatting fixes in docs

astrofrog:
Just a few small issues I found today when reading over the docs. For the ephemeris one, the issue is that the lack of empty line means that the double :: shows up in the HTML.

Not worth backporting!


astrofrog:
I checked giles and the output looks good now

PR #9468, opened 2019-10-25
Expanded documentation for the radio_velocity_correction method to de…

### LINKED ISSUE #9269: Potential trap with the accuracy of velocity correction
Hello!

I came across this issue in my science work where I tried to use the Doppler correction model implemented in astropy (radial_velocity_correction method of SkyCoord class) instead of a build-in model in an unrelated software for radio-interferometry data reduction in order to improve the accuracy of measured radial velocities. I believe the current description of the radial_velocity_correction method may be misleading and can result (well, in fact resulted in my project) in an error well above the claimed 1 mm/s accuracy (assuming correct application procedure is followed to account for multiplicative nature of the correction).  Here is the description of what I think is going on:

1. Wright & Eastman (2014) treatment of the barycentric correction is expressed in terms of the redshift, rather than the radial velocity returned by the method in question (this is kind of makes sense as some corrections, e.g. gravitational, are not strictly speaking the velocity effects).
2. The radial_velocity_correction method returns z_b * c, which is only the first order approximation. One has to use c * ((z_b+1)^2 - 1) / ((z_b+1)^2 + 1) to avoid making an approximation and get true velocity. Failure to do so introduces a sine wave residual with half a year period and amplitude up to 0.75 m/s (i.e. a quadratic term). 
3. The application formula given in the example in the method description, 
rv = rv + vcorr + rv * vcorr / consts.c, is strictly speaking valid only in the first order approximation v = z * c (e.g. compare with equation (11) of Wright & Eastman). Otherwise, one has to deal with redshifts directly (as per eq. 11 of Wright & Eastman) and then use the full relativistic formula shown above to get the final velocity. My reading of Wright & Eastman paper is that they used z * c formula for simplicity when considered residual terms or physical sense, although it wasn't articulated very clearly.

So, the issue is related to definitions and is not necessarily a bug as such. Although personally I'd prefer no implicit approximations at the library level (e.g. the method returning z_b itself or the velocity calculated using full relativistic formula), I see that the current interface may have some advantages for other users. In addition, changing the output now would potentially introduce bugs. Perhaps, it is better to just add a comment/warning to the help of the radial_velocity_correction method and state clearly that it returns z_b of Wright & Eastman (2014) multiplied by the speed of light.

Cheers,
              Max

VoronkovMA:
…scribe intricasies related to different velocity conentions Closes #9269

### Description

This pull request is to address documentation of the radial_velocity_correction method of SkyCoord class. Basically, it is now mentioned explicitly that this method returns correction velocity following the optical convention. An example is given how to use the method if full relativistic velocities are needed (mainly for use with an external application as is common in radio astronomy). Also, there is now enough information to link the details to the math of the Wright and Eastman (2014) paper referred in the documentation. 

Fixes #9269

### DIFF
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -1,4 +1,3 @@
-
 import re
 import copy
 
@@ -1493,7 +1492,7 @@ class SkyCoord(ShapedLikeNDArray):
           >>> loc = EarthLocation('149d33m00.5s','-30d18m46.385s',236.87*u.m)
           >>> sc = SkyCoord(1*u.deg, 2*u.deg)
           >>> vcorr = sc.radial_velocity_correction(kind='barycentric', obstime=t, location=loc)  # doctest: +REMOTE_DATA
-          >>> rv = rv + vcorr + rv * vcorr / c
+          >>> rv = rv + vcorr + rv * vcorr / c  # doctest: +SKIP
 
         If your target is nearby and/or has finite proper motion you may need to account
         for terms arising from this. See Wright & Eastman (2014) for details.
@@ -1501,7 +1500,7 @@ class SkyCoord(ShapedLikeNDArray):
         Also note that this method returns the correction velocity in the so-called
         *optical convention*::
 
-          >>> vcorr = zb * c
+          >>> vcorr = zb * c  # doctest: +SKIP
 
         where ``zb`` is the barycentric correction redshift as defined in section 3
         of Wright & Eastman (2014). The application formula given above follows from their
@@ -1513,13 +1512,13 @@ class SkyCoord(ShapedLikeNDArray):
         it as the input to another software which performs the application), the
         following recipe can be used::
 
-          >>> zb = vcorr / c
-          >>> zb_plus_one_squared = (zb + 1) ** 2
-          >>> vcorr_rel = c * (zb_plus_one_squared - 1) / (zb_plus_one_squared + 1)
+          >>> zb = vcorr / c  # doctest: +REMOTE_DATA
+          >>> zb_plus_one_squared = (zb + 1) ** 2  # doctest: +REMOTE_DATA
+          >>> vcorr_rel = c * (zb_plus_one_squared - 1) / (zb_plus_one_squared + 1)  # doctest: +REMOTE_DATA
 
         or alternatively using just equivalencies::
 
-          >>> vcorr_rel = vcorr.to(u.Hz, u.doppler_optical(1*u.Hz)).to(vcorr.unit, u.doppler_relativistic(1*u.Hz))
+          >>> vcorr_rel = vcorr.to(u.Hz, u.doppler_optical(1*u.Hz)).to(vcorr.unit, u.doppler_relativistic(1*u.Hz))  # doctest: +REMOTE_DATA
 
         See also `~astropy.units.equivalencies.doppler_optical`,
         `~astropy.units.equivalencies.doppler_radio`, and
@@ -1532,9 +1531,10 @@ class SkyCoord(ShapedLikeNDArray):
         either directly or via ``with`` statement.  For example, to use the JPL
         ephemeris, do::
 
-          >>>  sc = SkyCoord(1*u.deg, 2*u.deg)
-          >>>  with coord.solar_system_ephemeris.set('jpl'):
-          ...      rv += sc.radial_velocity_correction(obstime=t, location=loc)
+          >>> from astropy.coordinates import solar_system_ephemeris
+          >>> sc = SkyCoord(1*u.deg, 2*u.deg)
+          >>> with solar_system_ephemeris.set('jpl'):  # doctest: +REMOTE_DATA
+          ...     rv += sc.radial_velocity_correction(obstime=t, location=loc)  # doctest: +SKIP
 
         """
         # has to be here to prevent circular imports
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -1492,7 +1492,7 @@ class SkyCoord(ShapedLikeNDArray):
           >>> t = Time(56370.5, format='mjd', scale='utc')
           >>> loc = EarthLocation('149d33m00.5s','-30d18m46.385s',236.87*u.m)
           >>> sc = SkyCoord(1*u.deg, 2*u.deg)
-          >>> vcorr = sc.radial_velocity_correction(kind='barycentric', obstime=t, location=loc)
+          >>> vcorr = sc.radial_velocity_correction(kind='barycentric', obstime=t, location=loc)  # doctest: +REMOTE_DATA
           >>> rv = rv + vcorr + rv * vcorr / c
 
         If your target is nearby and/or has finite proper motion you may need to account
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -1491,7 +1491,31 @@ class SkyCoord(ShapedLikeNDArray):
            rv = rv + vcorr + rv * vcorr / consts.c
 
         If your target is nearby and/or has finite proper motion you may need to account
-        for terms arising from this. See Wright & Eastmann (2014) for details.
+        for terms arising from this. See Wright & Eastman (2014) for details.
+
+        Also note that this method returns the correction velocity in the so called 
+        *optical convention*::
+   
+           vcorr = zb * consts.c  
+
+        where `zb` is the barycentric correction redshift as defined in section 3
+        of Wright & Eastman (2014). The application formula given above follows from their
+        equation (11) under assumption that the radial velocity `rv` has also been defined 
+        using the same optical convention. Note, this can be regarded as a matter of 
+        velocity definition and does not by itself imply any loss of accuracy, provided 
+        sufficient care has been taken during interpretation of the results. If one needs
+        the barycentric correction expressed as the full relativistic velocity (e.g. to provide
+        it as the input to another software which performs the application), the 
+        following recipe can be used::
+
+           zb = vcorr / consts.c
+           zb_plus_one_squared = (zb + 1) * (zb + 1)
+           vcorr_rel = consts.c * (zb_plus_one_squared - 1) / (zb_plus_one_squared + 1)
+
+        See also `~astropy.units.equivalencies.doppler_optical`, 
+        `~astropy.units.equivalencies.doppler_radio` and 
+        `~astropy.units.equivalencies.doppler_relativistic` for more information on
+        the velocity conventions. 
 
         The default is for this method to use the builtin ephemeris for
         computing the sun and earth location.  Other ephemerides can be chosen
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -1486,36 +1486,45 @@ class SkyCoord(ShapedLikeNDArray):
         care is needed in application. Strictly speaking, the barycentric correction is
         multiplicative and should be applied as::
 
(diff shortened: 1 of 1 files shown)


pllim:
Please fix failures caused by PEP 8 -- see CI log to find out exactly what to fix. Thank you!

Also for subsequent commits, please add `[docs only]` directly in your commit message.

astropy/coordinates/sky_coordinate.py
```diff
@@ -1485,7 +1485,31 @@ def radial_velocity_correction(self, kind='barycentric', obstime=None,
            rv = rv + vcorr + rv * vcorr / consts.c
 
         If your target is nearby and/or has finite proper motion you may need to account
-        for terms arising from this. See Wright & Eastmann (2014) for details.
+        for terms arising from this. See Wright & Eastman (2014) for details.
+
+        Also note that this method returns the correction velocity in the so called
```
lglattly:
```suggestion
        Also note that this method returns the correction velocity in the so-called 
```

astropy/coordinates/sky_coordinate.py
```diff
@@ -1485,7 +1485,31 @@ def radial_velocity_correction(self, kind='barycentric', obstime=None,
...
+        *optical convention*::
+   
+           vcorr = zb * consts.c  
+
+        where `zb` is the barycentric correction redshift as defined in section 3
+        of Wright & Eastman (2014). The application formula given above follows from their
+        equation (11) under assumption that the radial velocity `rv` has also been defined 
+        using the same optical convention. Note, this can be regarded as a matter of 
+        velocity definition and does not by itself imply any loss of accuracy, provided 
+        sufficient care has been taken during interpretation of the results. If one needs
```
lglattly:
```suggestion
        sufficient care has been taken during interpretation of the results. If you need
```

astropy/coordinates/sky_coordinate.py
```diff
@@ -1485,7 +1485,31 @@ def radial_velocity_correction(self, kind='barycentric', obstime=None,
...
+        the barycentric correction expressed as the full relativistic velocity (e.g. to provide
```
lglattly:
```suggestion
        the barycentric correction expressed as the full relativistic velocity (e.g., to provide
```

VoronkovMA:
Sorry, I've added the pull-request number instead of the ticket number to the end of the last commit message

VoronkovMA:
it looks like CI failed again. This time it didn't like `zb` which was treated as a reference. What should I use to highlight in the text a variable ised in the code/example? I thought I did it according to the documentation

astropy/coordinates/sky_coordinate.py
```diff
@@ -1485,7 +1485,31 @@ def radial_velocity_correction(self, kind='barycentric', obstime=None,
...
+        equation (11) under assumption that the radial velocity `rv` has also been defined
+        using the same optical convention. Note, this can be regarded as a matter of
+        velocity definition and does not by itself imply any loss of accuracy, provided
+        sufficient care has been taken during interpretation of the results. If you need
+        the barycentric correction expressed as the full relativistic velocity (e.g., to provide
+        it as the input to another software which performs the application), the
+        following recipe can be used::
+
+           zb = vcorr / consts.c
+           zb_plus_one_squared = (zb + 1) * (zb + 1)
```
pllim: Is there benefit using this over `(zb + 1) ** 2`?

astropy/coordinates/sky_coordinate.py
```diff
@@ -1485,7 +1485,31 @@ def radial_velocity_correction(self, kind='barycentric', obstime=None,
...
+           vcorr_rel = consts.c * (zb_plus_one_squared - 1) / (zb_plus_one_squared + 1)
+
+        See also `~astropy.units.equivalencies.doppler_optical`,
+        `~astropy.units.equivalencies.doppler_radio` and
```
pllim: Oxford comma! 😹 
```suggestion
        `~astropy.units.equivalencies.doppler_radio`, and
```

astropy/coordinates/sky_coordinate.py
VoronkovMA: in the c-world such an explicit multiplication would be faster. In python, honestly I don't know. Anyway, I am happy to change it if it adds to readability

pllim: In the Python world, it is very easy to find out. And you appear to be correct! 😄 
```python
In [1]: zb = 0.05

In [2]: %timeit  (zb + 1) * (zb + 1)
83.2 ns ± 0.291 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %timeit  (zb + 1) ** 2
144 ns ± 0.577 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

mhvk: Go for readability. Generally `zb` will be an array and numpy is smart enough to turn `**2` into `np.square` (likely even in-place).

Plus, the calculations before this are *far* slower.

bsipocz:
Not related to this PR itself, but it would be great to have these code snippets in  docstrings actually  run through doctest. cc @adrn

VoronkovMA:
I am very surprised about the CI failure given the content of my last commit. I can't make much sense of the error message either.

mhvk:
@VoronkovMA - the failures are certainly unrelated to your commit! I think a rebase would solve it, but am not sure it is really needed, since the documentation tests passed already. @bsipocz?

pllim:
Ops... it's been a while now, so a rebase now might make sense.

VoronkovMA:
Ok, I will look into this when I get back into office mid-January.

Max

VoronkovMA:
pushed after a rebase attempt

VoronkovMA:
I am confused about the CI failure following the recent rebase and what to do to clear them as my changes were documentation only. Also, following rebase (which was successful btw with no conflicts), I got modified astropy_helpers (new commits) according to 'git status'. As I haven't done anything with that directory explicitly I didn't bother adding/committing it. Was this the likely source of the problem?

VoronkovMA:
update: after removing and reinstalling sphinx-astropy 'git status' doesn't complain about modified astropy_helpers any more

pllim:
Please prepend `>>> ` to your code snippets so that `pytest-doctestplus` picks them up for testing. Thanks!

> modified astropy_helpers (new commits)

For future reference, a `git submodule update` would have gotten rid of that.

VoronkovMA:
It looks like CI still fails, but according to the log, the failure doesn't seem to be related to the part I touched.

astropy/coordinates/sky_coordinate.py
```diff
@@ -1486,22 +1486,50 @@ def radial_velocity_correction(self, kind='barycentric', obstime=None,
...
+        using the same optical convention. Note, this can be regarded as a matter of
+        velocity definition and does not by itself imply any loss of accuracy, provided
+        sufficient care has been taken during interpretation of the results. If you need
+        the barycentric correction expressed as the full relativistic velocity (e.g., to provide
+        it as the input to another software which performs the application), the
+        following recipe can be used::
+
+          >>> zb = vcorr / consts.c
+          >>> zb_plus_one_squared = (zb + 1) ** 2
+          >>> vcorr_rel = consts.c * (zb_plus_one_squared - 1) / (zb_plus_one_squared + 1)
```
eteq: There is a way to do this using *just* equivalencies:

```
>>> vcorr_rel = vcorr.to(u.Hz, u.doppler_optical(1*u.Hz)).to(vcorr.unit, u.doppler_relativistic(1*u.Hz))
```
I'd say best to switch to that *except* that it's sort of un-intuitive that it requires a choice of "rest" frequency (which doesn't matter in this case as long as you use the same one for both equivalencies).

Perhaps it's best to use what I have written there because it's more concise and units-complient, but then add "or more expressively:" and then what you have here?

VoronkovMA: thanks, @eteq. Your suggestions make sense (although for the second code snippet, I'd perhaps give the explicit formula first as this is what I've tested extensively as part of my science work and what helps relating the result to the notations used in Wright & Eastman's paper cited earlier. Your way of doing it via equivalencies can be given later as an alternative). 

I've added the imports at some later iteration when CI started to run these examples (they were just text before) and just did the minimum addition which worked (perhaps because consts were imported through some other import?). I don't have any problem with importing c explicitly. 

Do I need to rebase again before making these changes (given that it took a lot of time since the previous commit)?

bsipocz:
> Do I need to rebase again before making these changes (given that it took a lot of time since the previous commit)?

Rebase is not necessary, but is welcome, if you don't mind.  I wouldn't expect much changes in the modules this test requires, but a rebase would get rid of the unrelated CI failures.

So a rebase is not a blocker at all for merging, and I'm happy to do it before it goes into master.

VoronkovMA:
Made changes to reflect previous suggestions by @eteq (explicit import of constants + alternative example of velocity conversion via equivalencies). I'll do the rebase shortly (I tried it earlier in my tree but it seems to cause issues with astropy_helpers which I couldn't resolve. This made local testing/building docs impossible. So I'll rebase a bit later and do a separate commit).

VoronkovMA:
pushed changes after rebase. Interestingly, CI still fails. But the message suggests it's IERS-related exception and nothing to do with my doc-only patch.

pllim:
Hmm. Looks like a doctest that should have been marked as remote data. Definitely unrelated to this PR.

pllim:
Oh, wait...it is part of this PR... Hang on... suggestion incoming.

astropy/coordinates/sky_coordinate.py
```diff
@@ -1486,22 +1486,55 @@ def radial_velocity_correction(self, kind='barycentric', obstime=None,
...
-           sc = SkyCoord(1*u.deg, 2*u.deg)
-           vcorr = sc.rv_correction(kind='barycentric', obstime=t, location=loc)
-           rv = rv + vcorr + rv * vcorr / consts.c
+          >>> from astropy.time import Time
+          >>> from astropy.coordinates import SkyCoord, EarthLocation
+          >>> from astropy.constants import c
+          >>> t = Time(56370.5, format='mjd', scale='utc')
+          >>> loc = EarthLocation('149d33m00.5s','-30d18m46.385s',236.87*u.m)
+          >>> sc = SkyCoord(1*u.deg, 2*u.deg)
+          >>> vcorr = sc.radial_velocity_correction(kind='barycentric', obstime=t, location=loc)
```
pllim: Any code affected by this line also need to be marked as `# doctest: +REMOTE_DATA`
```suggestion
          >>> vcorr = sc.radial_velocity_correction(kind='barycentric', obstime=t, location=loc)  # doctest: +REMOTE_DATA
```

bsipocz:
The commit should make the CI happy. I also went ahead and squashed the review addressing commits into one.

I'm not super happy with all the ` # doctest` additions in docstrings, but currently we don't have a better way to handle this.

mhvk:
Yes, it would be nice if there were a `.. doctest-requires: remote-data`...

bsipocz:
@mhvk - see https://github.com/astropy/pytest-doctestplus/issues/2, maybe we will get to it if we ever have an infrastructure sprint day or similar.

eteq:
Boo squashing! I think this is a good example of where the multiple commits actually contained useful information because there's a narrative of how the documentation got built...

But I don't care enough to reconstructor it, so on the basis of do-ocracy @bsipocz wins.  For now :wink:

Merging!

bsipocz:
@eteq - thus I kept the two commits. I looked into the individual ones, but there wasn't really useful info keeping it more granular, to the level of adding >>> or fighting with ci in addition of addressing content review.

mhvk:
If we really want a useful history, we need to start insisting on useful commit messages; right now, most real history is in the github comments (with unknown reliability; certainly feels much slower than it used to).

Issue #9449, opened 2019-10-23
Issues with the default value for `clip` in `ImageNormalize`


ayshih:
The [documentation for `ImageNormalize`](https://docs.astropy.org/en/latest/api/astropy.visualization.mpl_normalize.ImageNormalize.html) is confusing because it provides conflicting information about the default value for `clip` (which is True).  The call string and the parameter list correctly note that the default is True, but the text below that, which looks to be copied from Matplotlib's `Normalize`, states that the default is False because it otherwise "silently defeats" other functionality and "is likely to lead to surprises".

I agree with the sentiment the default should be `clip=False`.  I'll note that this was changed in #7800 due to a different inconsistency in the claimed defaults, but there may have been insufficient discussion about what the default ought to be.

Even if the consensus is that the default should be `clip=True`, the documentation still needs to be made consistent.  I'm not clear how the Matplotlib text is being pulled into the documentation.

Issue #8165, opened 2018-11-20
norm's behavior with NaN and masked values


saimn:
It seems that recently the behavior of astropy's norms (or most the way it interacts with matplotib) changed, and now NaN values and masked values are not displayed with the NaN color (when `clip=True` which is the default) as it used to be. 

For instance with NaNs:
![image](https://user-images.githubusercontent.com/311639/48803610-b237dd80-ed12-11e8-9d5d-2f5f808f5d66.png)

And with masked values:
![image](https://user-images.githubusercontent.com/311639/48803813-499d3080-ed13-11e8-8325-e9e2e3de63af.png)

I don't have this behavior (the middle image) if I test on an old py2 env with Astropy 2.0.7 and matplotlib 2.2.2.
I didn't find a recent change in Astropy that could explain this, so I guess that something changed in matplotlib. But still I wonder why the mask is dropped here (@astrofrog - I think you introduced this) when `clip` is True:
https://github.com/astropy/astropy/blob/f7db70d21b1b9b10124314202a835a4485ebf695/astropy/visualization/mpl_normalize.py#L92-L97

saimn:
Maybe not clear above: removing the `mask = False` line mentioned above fixes the issue, though I don't know if this is the correct fix.

Knusper:
On a slightly related note, I find it quite confusing that the `clip` defaults in `matlplotlib.Normalize` to `False` (which is imho a reasonable default), wheras in astropy's `ImageNormalize` it defaults to `True`.
Moreover, the behavior of `clip` with respect to masked values should be documented.

pllim:
cc @larrybradley and @astrofrog

PR #9478, opened 2019-10-25
Change default clip to False for consistency with matplotlib

larrybradley:
Fixes the issue reported in #8165.  This also makes plotting masked arrays consistent.

Note this is a breaking change.  @astrofrog, please review.

fix: #8165
fix: #9449


saimn:
Oh, in #8165 I did not realize that the default value was changed in astropy (in  #7800). This explains the sudden behavior change!

So LGTM, and I guess that this closes #9449 as well ?

larrybradley:
Hmm, apparently this was my fault in #7800.  Sorry about that!

Yes, this also closes #9449.

PR #9489, opened 2019-10-27
Add hint for filtering multidim cols in to_pandas and io.ascii ECSV

taldcroft:
### Description

PR #9423 was not accepted by all as a way to deal with handling multidimensional columns in `to_pandas()` and `ascii.write`.  This takes a different approach and provides the 2-liner for users in this case, instead of embedding it as a new Table method.

Closes #9423


mhvk:
Remaining failure is a URL timeout problem, so ignoring it. Thanks, @taldcroft

PR #9491, opened 2019-10-27
Ensure we check leap seconds only when Time is actually used.

### LINKED ISSUE #9490: Ordering of leap second update can lead to import warnings
After #9365 was merged, @taldcroft [noticed](https://github.com/astropy/astropy/pull/9365#issuecomment-546683559) that "In a clean build of master right now I'm seeing this:"
```
In [1]: from astropy.io import ascii
WARNING: leap-second auto-update failed due to the following exception: ImportError("cannot import name 'convert_numpy'",) [astropy.time.core]
```
???

Clearly, import order is important in ways it shouldn't be.  I'll investigate but suggestions would be most welcome...

EDIT: reason it is tricky: many of the most common time formats need an up-to-date leap-second table (e.g., `Time.now()` uses `TimeDatetime`, which, like `TimeISO*` calls `erfa.dtf2d`, which needs leap second information. But the `LeapSecond` class subclasses `QTable`, so needs to import `astropy.table`. `Table` itself only imports `Time` locally, so is safe, but some parts of `io.ascii` import `Time` on top and therefore are not.

Possible solutions:
1. Make `LeapSeconds` a class independent of `Table` - which means also moving it out of `iers`. That means parsing the files outside of `io.ascii` and having very IERS-like configuration elsewhere; this seems a horrible duplication of work.
2. Ensure `Time` is imported only locally in `io`. But it seems rather annoying not to be able to count on being able to import such a fundamental class.
3. Ensure `erfa` is imported only locally in `Time`, `iers`, etc., and move the update back there. More than a little annoying, since used a *lot* in `time.formats`. Also, is a very basic package that should just be importable (and is not currently causing the problems).
4. Ensure the leap-second update is only done on first *use* of `Time`.

Option (4) may be the best. Should I do this with something in `Time.__new__`?

cc @taldcroft, @eteq

mhvk:
I went with only calling `update_leap_seconds` on first call of `Time`. Somewhat ugly but it does work. Better suggestions most welcome...

fixes #9490

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -95,6 +95,9 @@ SIDEREAL_TIME_MODELS = {
         'IAU1994': {'function': erfa.gst94, 'scales': ('ut1',)}}}
 
 
+_LEAP_SECONDS_CHECKED = False
+
+
 class TimeInfo(MixinInfo):
     """
     Container for meta information like name, description, format.  This is
@@ -373,6 +376,17 @@ class Time(ShapedLikeNDArray):
     def __new__(cls, val, val2=None, format=None, scale=None,
                 precision=None, in_subfmt=None, out_subfmt=None,
                 location=None, copy=False):
+
+        # Because of import problems, this can only be done on
+        # first call of Time.
+        global _LEAP_SECONDS_CHECKED
+        if not _LEAP_SECONDS_CHECKED:
+            # *Must* set to True first as update_leap_seconds uses Time.
+            # In principle, this may cause wrong leap seconds in
+            # update_leap_seconds itself, but since expiration is in
+            # units of days, that is fine.
+            _LEAP_SECONDS_CHECKED = True
+            update_leap_seconds()
 
         if isinstance(val, cls):
             self = val.replicate(format=format, copy=copy)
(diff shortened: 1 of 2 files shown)


astropy/time/core.py
```diff
@@ -374,6 +377,17 @@ def __new__(cls, val, val2=None, format=None, scale=None,
                 precision=None, in_subfmt=None, out_subfmt=None,
                 location=None, copy=False):
 
+        # Because of import problems, this can only be done on
+        # first call of Time.
+        global _LEAP_SECONDS_CHECKED
+        if not _LEAP_SECONDS_CHECKED:
```
taldcroft: Maybe you could save two lines of code and a `global` by holding this attribute in the `Time` class.  So
```
if not getattr(Time, '_LEAP_SECONDS_CHECKED', False):
    Time._LEAP_SECONDS_CHECKED = True
    update_leap_seconds()
```
But on the whole this solution seems fine.  I did confirm that it resolves the problems I found.

mhvk: I thought about putting it on the class, but ended up feeling it was slightly more elegant (less inelegant?) to have it as global state,. Also perhaps slightly faster?

PR #9496, opened 2019-10-28
Added Python 3.8 32-bit build

astrofrog:
While trying to build 32-bit wheels for Python 3.8 I saw a test failure so I think we should make sure we test this configuration here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,10 +32,10 @@ jobs:
           command: PYTHONHASHSEED=42 /opt/python/cp37-cp37m/bin/python setup.py test --parallel=4 -a "--durations=50"
       - run:
           name: Install dependencies for Python 3.8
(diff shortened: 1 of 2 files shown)


astrofrog:
@nden @perrygreenfield - I tagged you for review since I had to tweak a test to avoid precision issues. With (0,0) the output coordinates in the test sometimes came out as (360,0) so I thought it'd make more sense to go slightly off zero.

@bsipocz - this doesn't need backporting, I'm just patching this in the wheels infrastructure.

mhvk:
Would it make sense to remove a 32-bit test with an older python?

astrofrog:
@mhvk - we already removed Python 3.5 recently. In any case I think we should keep all the supported Pythons in the 32-bit build since this is often a source of problems for wheels and linux distributions otherwise. Note that this build is usually pretty fast because it also serves as our test for parallel builds.

mhvk:
OK, sounds good!

astrofrog:
I'll go ahead and merge this since @nden approved - thanks!

PR #9500, opened 2019-10-28
Fix wcsaxes when world_axis_object_classes contains 4 values

Cadair:
#9376 broke this method in wcsaxes.

@astrofrog


pllim:
Maybe add a regression test?

PR #9505, opened 2019-10-29
VO table no longer access char field as bytes

### LINKED ISSUE #8716: Extended Table unicode sandwich for varchar data access from votable
As reported by @tomdonaldson , currently, this is the behavior; i.e., variable-length columns in VO table return byte string data. And this behavior was not seen in Python 2 but only in Python 3:
```python
>>> filename = '.../astropy/io/votable/tests/data/gemini.xml'
>>> t = Table.read(filename)
>>> x = t['access_url'][0]
>>> x
b'http://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/...'
```

To obtain a plain text output (not byte string), one needs to do the following. However, converting byte string to unicode also takes up more memory:

```python
>>> x.decode('utf-8')
'http://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/...'
```

So, we wouldn't want to automatically convert everything to unicode, but rather do it on-the-fly only to requested data, by extending the "unicode sandwich" implemented in #5700.

Many thanks to @astrofrog for the advice and suggestion at IVOA 2019. cc @bmatthieu3  

**Appendix**

Here is the XML header of the data file in the example above:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<VOTABLE xmlns="http://www.ivoa.net/xml/VOTable/v1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2">
  <RESOURCE type="results">
    <TABLE>
      <FIELD name="ID" datatype="char" ucd="meta.id;meta.main" arraysize="*" />
      <FIELD name="access_url" datatype="char" ucd="meta.ref.url" arraysize="*" />
      <FIELD name="service_def" datatype="char" ucd="meta.ref" arraysize="*" />
      <FIELD name="error_message" datatype="char" ucd="meta.code.error" arraysize="*" />
      <FIELD name="semantics" datatype="char" ucd="meta.code" arraysize="*" />
      <FIELD name="description" datatype="char" ucd="meta.note" arraysize="*" />
      <FIELD name="content_type" datatype="char" ucd="meta.code.mime" arraysize="*" />
      <FIELD name="content_length" datatype="long" ucd="phys.size;meta.file" unit="byte" />
      <FIELD name="product_type" datatype="char" utype="caom:Artifact.productType" arraysize="*" />
      <FIELD name="file_uri" datatype="char" ID="fileURIRef" arraysize="*" />
```

pllim:
### Description

This pull request is to address the issue of VO table parsing `char` field as bytes unnecessarily.

This does change a subtle behavior, so it should only go into a major release. I suspect it would affect packages downstream, e.g., `astroquery.vo_conesearch` or whoever is parsing VO table using `Table.read()` and then decoding bytes to UTF-8 themselves.

So, the gist is that VO table parser was relying on bytes vs str to differentiate between `char` and `unicodeChar` fields. A quick glance at VO table standards did not seem to indicate that this is a rule from the standards. Since I can't ask Mike D anymore, I am wildly guessing that the behavior was something that was just convenient to do at the time in Python 2 but turned into a nuisance in Python 3.

**Note to reviewers:** Those who are familiar with the VO standards and use VO table a lot should really look at this carefully and think about the implications of merging this.

**TODO:**

- [x] Wait for in-depth discussions from VO table stakeholders.
- [x] Add change log.

Fixes #8716 

Supersedes #8745

### DIFF
--- a/astropy/io/votable/converters.py
+++ b/astropy/io/votable/converters.py
@@ -20,7 +20,8 @@ from astropy.utils.xml.writer import xml_escape_cdata
 
 # LOCAL
 from .exceptions import (vo_raise, vo_warn, warn_or_raise, W01,
-    W30, W31, W39, W46, W47, W49, W51, E01, E02, E03, E04, E05, E06)
+    W30, W31, W39, W46, W47, W49, W51, W55, E01, E02, E03, E04,
+    E05, E06, E24)
 
 
 __all__ = ['get_converter', 'Converter', 'table_column_to_votable_datatype']
@@ -298,6 +299,8 @@ class Char(Converter):
 
         Converter.__init__(self, field, config, pos)
 
+        self.field_name = field.name
+
         if field.arraysize is None:
             vo_warn(W47, (), config, pos)
             field.arraysize = '1'
@@ -314,7 +317,7 @@ class Char(Converter):
                 self.arraysize = int(field.arraysize)
             except ValueError:
                 vo_raise(E01, (field.arraysize, 'char', field.ID), config)
-            self.format = f'S{self.arraysize:d}'
+            self.format = f'U{self.arraysize:d}'
             self.binparse = self._binparse_fixed
             self.binoutput = self._binoutput_fixed
             self._struct_format = f">{self.arraysize:d}s"
@@ -325,6 +328,12 @@ class Char(Converter):
     def parse(self, value, config=None, pos=None):
         if self.arraysize != '*' and len(value) > self.arraysize:
             vo_warn(W46, ('char', self.arraysize), config, pos)
+
+        # Warn about non-ascii characters if warnings are enabled.
+        try:
+            value.encode('ascii')
+        except UnicodeEncodeError:
+            vo_warn(W55, (self.field_name, value), config, pos)
         return value, False
 
     def output(self, value, mask):
@@ -350,14 +359,20 @@ class Char(Converter):
         if mask or value is None or value == '':
             return _zero_int
         if isinstance(value, str):
-            value = value.encode('ascii')
+            try:
+                value = value.encode('ascii')
+            except ValueError:
+                vo_raise(E24, (value, self.field_name))
         return self._write_length(len(value)) + value
 
     def _binoutput_fixed(self, value, mask):
         if mask:
             value = _empty_bytes
         elif isinstance(value, str):
-            value = value.encode('ascii')
+            try:
+                value = value.encode('ascii')
+            except ValueError:
+                vo_raise(E24, (value, self.field_name))
         return struct_pack(self._struct_format, value)
 
 
@@ -1394,9 +1409,12 @@ def table_column_to_votable_datatype(column):
        A dict containing 'datatype' and 'arraysize' keys that can be
        set on a VOTable FIELD element.
     """
+    votable_string_dtype = None
+    if column.info.meta is not None:
+        votable_string_dtype = column.info.meta.get('_votable_string_dtype')
     if column.dtype.char == 'O':
-        if '_votable_string_dtype' in column.info.meta:
-            string_dtype = column.info.meta['_votable_string_dtype']
+        if votable_string_dtype is not None:
+            string_dtype = votable_string_dtype
             return {'datatype': string_dtype, 'arraysize': '*'}
         elif isinstance(column[0], np.ndarray):
             dtype, shape = _all_matching_dtype(column)
@@ -1411,4 +1429,10 @@ def table_column_to_votable_datatype(column):
         # All bets are off, do the most generic thing
         return {'datatype': 'unicodeChar', 'arraysize': '*'}
 
-    return numpy_to_votable_dtype(column.dtype, column.shape[1:])
+    # For fixed size string columns, datatype here will be unicodeChar,
+    # but honor the original FIELD datatype if present.
+    result = numpy_to_votable_dtype(column.dtype, column.shape[1:])
+    if result['datatype'] == 'unicodeChar' and votable_string_dtype == 'char':
+        result['datatype'] = 'char'
+
+    return result
--- a/astropy/io/votable/converters.py
+++ b/astropy/io/votable/converters.py
@@ -1430,7 +1430,7 @@ def table_column_to_votable_datatype(column):
     if column.dtype.char == 'O':
         if votable_string_dtype is not None:
             string_dtype = votable_string_dtype
-            return {'datatype': string_dtype, 'arraysize': '*'}
+            return {'datatype': votable_string_dtype, 'arraysize': '*'}
         elif isinstance(column[0], np.ndarray):
             dtype, shape = _all_matching_dtype(column)
             if dtype is not False:
--- a/astropy/io/votable/converters.py
+++ b/astropy/io/votable/converters.py
@@ -339,8 +339,23 @@ class Char(Converter):
     def output(self, value, mask):
         if mask:
             return ''
-        if not isinstance(value, str):
-            value = value.decode('ascii')
+
+        # The output methods for Char assume that value is either str or bytes.
+        # This method needs to return a str, but needs to warn if the str contains
+        # non-ASCII characters.
+        try:
+            if isinstance(value, str):
+                value.encode('ascii')
+            else:
+                # Check for non-ASCII chars in the bytes object.
+                value = value.decode('ascii')
+        except (ValueError, UnicodeEncodeError):
+            warn_or_raise(E24, UnicodeEncodeError, (value, self.field_name))
+        finally:
+            if isinstance(value, bytes):
+                # Convert the bytes to str regardless of non-ASCII chars.
+                value = value.decode('utf-8')
+
         return xml_escape_cdata(value)
 
     def _binparse_var(self, read):
--- a/astropy/io/votable/converters.py
+++ b/astropy/io/votable/converters.py
@@ -319,23 +319,13 @@ class Char(Converter):
             self.binoutput = self._binoutput_fixed
             self._struct_format = f">{self.arraysize:d}s"
 
-        if config.get('verify', 'ignore') == 'exception':
-            self.parse = self._ascii_parse
-        else:
-            self.parse = self._str_parse
-
     def supports_empty_values(self, config):
         return True
 
-    def _ascii_parse(self, value, config=None, pos=None):
+    def parse(self, value, config=None, pos=None):
         if self.arraysize != '*' and len(value) > self.arraysize:
             vo_warn(W46, ('char', self.arraysize), config, pos)
-        return value.encode('ascii'), False
-
-    def _str_parse(self, value, config=None, pos=None):
-        if self.arraysize != '*' and len(value) > self.arraysize:
-            vo_warn(W46, ('char', self.arraysize), config, pos)
-        return value.encode('utf-8'), False
+        return value, False
 
     def output(self, value, mask):
         if mask:
@@ -346,11 +336,12 @@ class Char(Converter):
 
     def _binparse_var(self, read):
         length = self._parse_length(read)
-        return read(length), False
+        return read(length).decode('ascii'), False
 
     def _binparse_fixed(self, read):
         s = struct_unpack(self._struct_format, read(self.arraysize))[0]
         end = s.find(_zero_byte)
+        s = s.decode('ascii')
         if end != -1:
             return s[:end], False
         return s, False
@@ -358,11 +349,15 @@ class Char(Converter):
     def _binoutput_var(self, value, mask):
         if mask or value is None or value == '':
             return _zero_int
+        if isinstance(value, str):
+            value = value.encode('ascii')
         return self._write_length(len(value)) + value
 
     def _binoutput_fixed(self, value, mask):
         if mask:
             value = _empty_bytes
+        elif isinstance(value, str):
+            value = value.encode('ascii')
         return struct_pack(self._struct_format, value)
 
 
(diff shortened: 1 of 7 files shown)


astropy/io/votable/converters.py
```diff
@@ -346,23 +336,28 @@ def output(self, value, mask):
 
     def _binparse_var(self, read):
         length = self._parse_length(read)
-        return read(length), False
+        return read(length).decode('ascii'), False
```
pllim: Note: This no longer choose between `'ascii'` and `'utf-8'` based on `'verify'` setting. This one is stricter, but I suspect it wouldn't be hard to add it back in, though the VO table standard did say `char` is ASCII. 🤷‍♀

tomdonaldson: With the updated commit, I changed this behavior to allow non-ASCII characters for FIELDs with `datatype="char"` in TABLEDATA serializations, but with a new warning message.  Whether all ASCII or not, values for those FIELDs are stored in the astropy Table as strings.

tomdonaldson:
This PR has been updated to ensure that it handles both fixed-length and variable length strings.  The upshot is that for `FIELD`s with `datatype="char"`, the values will always be stored in the astropy Table as `str` instead of `bytes` (e.g., `'myvalue'` instead of `b'myvalue'`).

That behavior was already the case for `FIELD`s with `datatype="unicodeChar"`, so the general idea was to make the `char` handling more like `unicodeChar`, but with some differences depending on the serialization.

For `TABLEDATA` serialization:
- `char` is read in just like `unicodeChar` except that if non-ASCII characters are present for `char`, a warning is issued.  This behavior is similar to many existing VOTable parsers which also allow non-ASCII characters, often without a warning.
- On write, an exception is raised if attempting to write non-ASCII characters to a `char` `FIELD`.  The stricter enforcement here is to prevent astropy from writing a VOTable that is not compliant with the spec.

For `BINARY` and `BINARY2`:
- The `FIELD` `datatype` determines the number of bytes per character, so `char` values are ASCII-encoded into `str`s,  and `unicodeChar` values are `UTF-16-BE` -encoded into `str`s.

For all serializations, the astropy Table column `dtype` is `O` for variable length strings and `U<length>` for fixed length strings.  This is consistent with the prior behavior except that fixed length `char` strings were `S<length>`.

msdemlei:
Without having looked at the patch as such – someone will have to poke me again for that –, I heartily support the intent of the PR.  VOTable char[] most definitely is supposed to be an ASCII string (as opposed to unsignedByte), and so it shouldn't have been turned into bytes in the first place.

I kind-of also support interpreting content as utf-8; I've spoken for this ages ago (http://mail.ivoa.net/pipermail/apps/2014-October/001010.html is a thread from 2014), and if astropy establishes a quasi-standard here, that may actually push this along.  It should be made clear somewhere that it's possible that behaviour with non-ASCII in char is actually undefined by the IVOA.

Of course, this will break some code that did something with strings from VOTable.  But bytes coming out of VOTable are so painful that to me that seems worth it.

Hence: informally, +1.

astropy/io/votable/converters.py
hugobuddel: It seems to me that @pllim's comment is still valid; but I'm not familiar with this code so maybe I'm misreading it. As in, putting utf-8 text in a char field and setting 'verify' to 'ignore' seems to raise an exception, but I've not actually tried this. I'm in favor of allowing utf-8, even if it is not in the official standard.

tomdonaldson: Hi @hugobuddel .  Thanks for the taking a look at this.

The context here a is a little misleading.  You are correct that in the `_binparse_var()` method shown above, non-ASCII characters will cause an exception.  This occurs for the `BINARY` and `BINARY2` serializations when any of the byte values are outside the range 0-127.  I think that behavior is probably OK since values outside that range may just as likely represent corrupt binary data as opposed to deliberate characters in the extended ASCII range (128-255), but I could go either way.  I should (and will) at least add a test case to demonstrate that as the expected behavior.

My comment above though was referring to the `parse()` method that is just above `_binparse_var()`.  `parse()` handles the TABLEDATA serialization (in which non-ASCII unicode might most easily sneak in).  In that case a `try` clause catches the decode exception and calls `vo_warn()` with a new warning message.  `vo_warn()` looks at the `verify` value to decide whether to show the message.

hugobuddel: Thanks for the clarification @tomdonaldson . Adding tests would indeed help, because that's where I looked to see whether my interpretation of the code was correct, but it is clear now.

I'm personally not using the BINARY/BINARY2 serializations (only FITS or TABLEDATA), so I'll let others decide. (At least, I'm not writing them; no idea whether I read them from other sources.)

astropy/io/votable/tests/converter_test.py
```diff
@@ -65,6 +66,42 @@ def test_unicode_mask():
     assert c.output("Foo", True) == ''
 
 
+def test_unicode_as_char():
+    config = {'verify': 'exception'}
+    #with catch_warnings(exceptions.W55) as w:
```
pllim: @tomdonaldson , if you have no plans to re-enable `catch_warnings` here, please remove the code.

astropy/io/votable/tests/converter_test.py
```diff
@@ -65,6 +66,42 @@ def test_unicode_mask():
...
+    field = tree.Field(
+        None, name='unicode_in_char', datatype='char',
+        arraysize='*', config=config)
+    c = converters.get_converter(field, config=config)
+
+    c.parse("XXX")  # ASCII succeeds
+
+    with pytest.raises(
+        exceptions.W55,
```
pllim: Easier to read if the continuation of `with` statement is at a different indentation level from the code inside the block.
```suggestion
            exceptions.W55,
```

astropy/io/votable/tests/converter_test.py
```diff
@@ -65,6 +66,42 @@ def test_unicode_mask():
...
+        match=r'.*FIELD \(unicode_in_char\) has datatype="char" but contains non-ASCII value.*'):
```
pllim:
```suggestion
            match=r'.*FIELD \(unicode_in_char\) has datatype="char" but contains non-ASCII value.*'):
```

astropy/io/votable/tests/converter_test.py
```diff
@@ -65,6 +66,42 @@ def test_unicode_mask():
...
+
+def test_unicode_as_char_binary():
+    config = {'verify': 'exception'}
+
+    field = tree.Field(
+        None, name='unicode_in_char', datatype='char',
+         arraysize='*', config=config)
+    c = converters.get_converter(field, config=config)
+    c._binoutput_var('abc', False)  # ASCII succeeds
+    with pytest.raises(exceptions.E24, match=r".*E24: Attempt to write non-ASCII value.*"):
```
pllim: I don't think the trailing `.*` is necessary but I guess it does not hurt.

astropy/io/votable/tests/vo_test.py
```diff
@@ -75,10 +75,10 @@ def _test_regression(tmpdir, _python_based=False, binary_mode=1):
 
     dtypes = [
         (('string test', 'string_test'), '|O8'),
-        (('fixed string test', 'string_test_2'), '|S10'),
+        (('fixed string test', 'string_test_2'), '<U10'),
```
pllim: Is this type change in the expected results still abide by the VOTable standards? (Asking as a person unfamiliar with those standards. 😬 )

astropy/table/tests/test_showtable.py
```diff
@@ -99,6 +99,8 @@ def test_votable(capsys):
     showtable.main([os.path.join(VOTABLE_ROOT, 'data/regression.xml'),
                     '--table-id', 'main_table', '--max-width', '50'])
     out, err = capsys.readouterr()
+    print('output:')
```
pllim: We do not need these, right?
```suggestion
```

astropy/table/tests/test_showtable.py
```diff
@@ -99,6 +99,8 @@ def test_votable(capsys):
...
+    print(out.splitlines())
```
pllim:
```suggestion
```

funbaker:
Reading the ascii datatype as unicode char is fine, as long there is a warning emitted. This is necessary for writing a service compliance checker with astropy/pyvo (This idea came up somewhere).
 
I'm missing a `vo_raise(E24)` when trying to write out non ascii values in the `output` method of the `char` datatype.

astropy/io/votable/tests/converter_test.py
tomdonaldson: @pllim Thank you for the review!  These are all good comments and I will make the suggested changes.

astropy/io/votable/tests/converter_test.py
tomdonaldson: This is interesting.  At some point I had convinced myself that was needed, but in looking at the docs now, it doesn't look like the leading `.*` is needed either.  (https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions)

I've removed the leading and trailing `.*` in 3 cases in that file and the tests still run well.

pllim:
@tomdonaldson , do you think this is ready for another round of review, or are you still waiting to add more code or hear back from VO stakeholders?

tomdonaldson:
> I'm missing a `vo_raise(E24)` when trying to write out non ascii values in the `output` method of the `char` datatype.

@funbaker Thanks for noticing that!  I've added code to the Char converter that warns if the value contains non-ASCII characters, and added corresponding test cases.  Perhaps that check can be a little cleaner with [isascii()](https://docs.python.org/3.7/library/stdtypes.html#str.isascii) once we require Python 3.7+. 

@pllim  With this last commit, this is ready for another round of review.  Perhaps @funbaker and/or @hugobuddel could also have another look?  It would be great to get this in for the upcoming 4.0.1 release.

To summarize, I believe we've changed the behavior to parse all `char` and `unicodeChar` VOTable FIELDs into `str` instead of `bytes` objects (allowing but warning on non-ASCII values for the `char` case).  We've also added checks on outputting non-ASCII values to `char` FIELDs (warnings for TABLEDATA serialization and errors for BINARY and BINARY2 serializations).  Outside of those changes, the previous behaviors, assumptions and test coverage should have been preserved.

pllim:
> Perhaps that check can be a little cleaner with isascii() once we require Python 3.7+

Note to self: Open a follow-up issue for this after this PR is merged.

astropy/io/votable/converters.py
```diff
@@ -1413,13 +1424,13 @@ def table_column_to_votable_datatype(column):
...
     if column.dtype.char == 'O':
-        if isinstance(column[0], bytes):
-            if _all_bytes(column[1:]):
-                return {'datatype': 'char', 'arraysize': '*'}
-        elif isinstance(column[0], str):
-            if _all_unicode(column[1:]):
-                return {'datatype': 'unicodeChar', 'arraysize': '*'}
+        if votable_string_dtype is not None:
+            string_dtype = votable_string_dtype
+            return {'datatype': string_dtype, 'arraysize': '*'}
```
pllim: Why is `string_dtype = votable_string_dtype` needed? Maybe we can do the following instead?
```suggestion
            return {'datatype': votable_string_dtype, 'arraysize': '*'}
```

astropy/io/votable/exceptions.py
```diff
@@ -1082,6 +1082,19 @@ class W54(VOTableSpecWarning):
     default_args = ('1.3',)
 
 
+class W55(VOTableSpecWarning):
+    """
+    When non-ASCII characters are detected when reading
+    a TABLEDATA value for a FIELD with datatype="char", we
```
pllim: Nitpick: API doc formatting.
```suggestion
    a TABLEDATA value for a FIELD with ``datatype="char"``, we
```

astropy/io/votable/exceptions.py
```diff
@@ -1448,6 +1461,18 @@ class E23(VOTableSpecWarning):
     default_args = ('x',)
 
 
+class E24(VOWarning, ValueError):
+    """
+    Non-ASCII unicode values should not be written when the FIELD datatype="char",
```
pllim: Nitpick: API doc formatting.
```suggestion
    Non-ASCII unicode values should not be written when the FIELD ``datatype="char"``,
```

astropy/io/votable/tests/converter_test.py
```diff
@@ -65,6 +66,56 @@ def test_unicode_mask():
...
+            exceptions.W55,
+            match=r'FIELD \(unicode_in_char\) has datatype="char" but contains non-ASCII value'):
+        c.parse("zła")  # non-ASCII
+
+    # Test output.
+    c.output('XYZ', False)  # ASCII str succeeds
+    c.output(b'XYZ', False)  # ASCII bytes succeeds
+    value = 'zła'
+    value_bytes = value.encode('utf-8')
+    with pytest.raises(
```
pllim: I wonder if these two `pytest.raises` should be in a loop, but given `n=2`, maybe not? I'll leave this for you to decide.

astropy/io/votable/tests/converter_test.py
```diff
@@ -65,6 +66,56 @@ def test_unicode_mask():
...
+            exceptions.E24,
+            match=r'E24: Attempt to write non-ASCII value'):
+        c.output(value, False)  # non-ASCII str raises
+    with pytest.raises(
+            exceptions.E24,
+            match=r'E24: Attempt to write non-ASCII value'):
+        c.output(value_bytes, False)  # non-ASCII bytes raises
+
```
pllim: Nitpick: Extra blank line.
```suggestion
```

astropy/io/votable/tests/converter_test.py
tomdonaldson: Good question.  Since the second value to test is derived from the first, I think I'll leave this as is for now.

pllim:
Now I gotta clean up `astroquery.vo_conesearch`... :sweat_smile:

pllim:
@bsipocz and @tomdonaldson , as discussed on Slack, since this change is causing some downstream breakage (e.g., `astroquery`), I am remilestoning to 4.1 instead of 4.0.1. It is my understanding that there is no action needed on my part in terms of consolidating the change log. This PR needs to be backed out of the 4.0.x branch too. I apologize for any inconvenience caused.

PR #9539, opened 2019-11-06
No longer import display plugin from test runner

astrofrog:
This will get rid of the deprecation warning reported in https://github.com/astropy/package-template/issues/435 - I think this may also be the cause of the warning about the plugin being registered twice - let's see in the CI logs.


pllim:
To close the loop, I ran tests with this patch and warnings turned back on, and the warning is indeed gone.

PR #9541, opened 2019-11-06
Adding pandas Int64 support [fix #9280]

### LINKED ISSUE #9280: Consider using IntegerArrays in to_pandas() for masked int columns
Currently, to_pandas converts any masked columns automatically to float (even if they are integers). This is actually a pretty big problem for integer identifiers (for example, GAIA source ids). If the integer is too large, the double-precision float actually gets approximated in the last few digits, completely changing the identifier. As this can happen with something as common as GAIA data, it's a big problem. What I assume most people would do is convert `to_pandas`, and then convert the data-type back to `int`, but this will now produce a different number.

More recent versions of pandas now support nullable integers (https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html), using the new `IntegerArrays` (with data type `pd.Int64Dtype()`). With that now supported, it would make sense for the astropy.table.to_pandas to turn masked int columns into this new data type, rather than float. pandas `IntegerArrays` are apparently still "experimental" but at the same time I think it's worth doing anyway since turning integers to floats can have pretty severe side-effects in astronomical data, which can take a long time to debug.

At the very least, if the above isn't implemented yet, I think it would make sense for astropy to produce a warning when it's converting integers to floats in this way, especially if the integers are large enough that the floats will only approximate them.

bmorris3:
### Description

This pull request is to address issue #9280. Adding support for pandas nullable integers in the `Table.to_pandas` function, as suggested by @mar-ses. We're new to contributing to `Table`, so we're happy to iterate if generating a pandas table from a combination of masked arrays and a pandas `Series` doesn't jive with the Table style.

Fixes #9280.

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3526,6 +3526,14 @@ class Table:
 
         for name, column, data, mask, unit in zip(names, columns, datas, masks, units):
 
+            if column.dtype.kind in ['u', 'i'] and np.any(mask):
+                # Special-case support for pandas nullable int
+                np_dtype = str(column.dtype).lower()
+                data = np.zeros(shape=column.shape, dtype=np_dtype)
+                data[~mask] = column[~mask]
+                out[name] = MaskedColumn(data=data, name=name, mask=mask, unit=unit, copy=False)
+                continue
+
             if data.dtype.kind == 'O':
                 # If all elements of an object array are string-like or np.nan
                 # then coerce back to a native numpy str/unicode array.
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3425,7 +3425,8 @@ class Table:
             else:
                 out[name] = column
 
-            if out[name].dtype.byteorder not in ('=', '|'):
+            if (hasattr(out[name].dtype, 'byteorder') and
+                out[name].dtype.byteorder not in ('=', '|')):
                 out[name] = out[name].byteswap().newbyteorder()
 
         kwargs = {'index': out.pop(index)} if index else {}
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3417,7 +3417,9 @@ class Table:
         for name, column in tbl.columns.items():
             if isinstance(column, MaskedColumn) and np.any(column.mask):
                 if column.dtype.kind in ['i', 'u']:
-                    out[name] = Series(column, dtype='Int64')
+                    # str(type) must be int<n> or uint<n>, convert to
+                    pd_dtype = str(column.dtype).replace('i', 'I').replace('u', 'U')
+                    out[name] = Series(column, dtype=pd_dtype)
                 elif column.dtype.kind in ['f', 'c']:
                     out[name] = column.filled(np.nan)
                 else:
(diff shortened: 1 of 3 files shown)


pllim:
Thanks!

Test failure seems real, but change log can wait until the relevant section is available.

I'll let subpackage maintainer decides if a new test is necessary.

taldcroft:
@bmorris3 - my sincere apologies for the late reply.  This slipped off my radar and I just noticed it again today.

This looks like a great direction and we should go forward on it.  Some issues that need to be addressed are:
- The current failing test.  The current behavior (in master) is that an int MaskedColumn will round-trip through pandas and come back as a masked float.  So what appears to be happening in your branch is that it comes back as an Object `'O'` dtype.  The desired behavior would be perfect round trip where it comes back as an int MaskedColumn.
- There should be a check on the pandas version to ensure that the installed version is >= 0.24 so that the Int64 type for `Series` is supported.  You'll need to take a little care in testing with regards to the Pandas version since I don't think we have a strict requirements on pandas version for testing.
- A slight sticky wicket with regards to back-compatibility.  We likely need a new keyword argument in `to_pandas` like `use_nullable_int`, which would default to `None` implying use it if possible (pandas version >= 0.24).  `False` would mean never use it and `True` would mean always use it.

astropy/table/table.py
```diff
@@ -3283,18 +3283,16 @@ def _encode_mixins(tbl):
         for name, column in tbl.columns.items():
             if isinstance(column, MaskedColumn) and np.any(column.mask):
                 if column.dtype.kind in ['i', 'u']:
-                    out[name] = column.astype(float).filled(np.nan)
-                    warnings.warn(
-                        "converted column '{}' from integer to float".format(
-                            name), TableReplaceWarning, stacklevel=3)
+                    out[name] = Series(column, dtype='Int64')
```
taldcroft: The dtype should match the input dtype, e.g. `np.uint8` should go to the same (`pd.UInt8Dtype`) and so on.

astropy/table/table.py
```diff
@@ -3283,18 +3283,16 @@ def _encode_mixins(tbl):
...
                 elif column.dtype.kind in ['f', 'c']:
                     out[name] = column.filled(np.nan)
                 else:
                     out[name] = column.astype(object).filled(np.nan)
             else:
                 out[name] = column
 
-            if out[name].dtype.byteorder not in ('=', '|'):
+            if (hasattr(out[name].dtype, 'byteorder') and
+                out[name].dtype.byteorder not in ('=', '|')):
```
taldcroft: This line should be indented 4 characters for PEP8.

astrofrog:
@bmorris3 - would you have time to rebase this and implement @taldcroft's comments to get this working?

taldcroft:
@astrofrog @bmorris3 - I took a quick crack at fixing this.  It looks like `from_pandas()` needs work as well as it is round-tripping back to astropy as an object column. Will see if this is an easy fix or not.

taldcroft:
Passing tests locally!  @astrofrog please have a look and think about what this might break.

astropy/table/table.py
```diff
@@ -3417,18 +3425,26 @@ def _encode_mixins(tbl):
...
+                    out[name] = Series(column, dtype=pd_dtype)
+
+                    # If pandas is older than 0.24 the type may have turned to float
+                    if column.dtype.kind != out[name].dtype.kind:
+                        warnings.warn(
+                            f"converted column '{name}' from {column.dtype} to {out[name].dtype}",
+                            TableReplaceWarning, stacklevel=3)
                 elif column.dtype.kind in ['f', 'c']:
-                    out[name] = column.filled(np.nan)
+                    out[name] = column
```
taldcroft, line 3440: It turns out this intermediate nan-filled copy is not needed, pandas deals with a `MaskedArray` subclass just fine.

taldcroft:
@astrofrog -this needs a quick review, or we can just push to 4.2.

bmorris3:
Thanks for carrying this over the finish line, @taldcroft!

Issue #9551, opened 2019-11-06
TST: CircleCI mpldev job failing


pllim:
### Description

CircleCI `mpldev` job is failing on `master` and PRs.

### Expected behavior

No failure.

### Actual behavior

```
Reading state information... Done

The following additional packages will be installed:
  g++-5 libstdc++-5-dev
Suggested packages:
  g++-multilib g++-5-multilib gcc-5-doc libstdc++6-5-dbg libstdc++-5-doc
The following NEW packages will be installed:
  g++ g++-5 libstdc++-5-dev
0 upgraded, 3 newly installed, 0 to remove and 11 not upgraded.
Need to get 9738 kB of archives.
After this operation, 39.1 MB of additional disk space will be used.
Err:1 http://archive.ubuntu.com/ubuntu xenial-updates/main amd64 libstdc++-5-dev amd64 5.4.0-6ubuntu1~16.04.11
  404  Not Found [IP: 91.189.88.31 80]
Err:2 http://archive.ubuntu.com/ubuntu xenial-updates/main amd64 g++-5 amd64 5.4.0-6ubuntu1~16.04.11
  404  Not Found [IP: 91.189.88.31 80]
Get:3 http://archive.ubuntu.com/ubuntu xenial/main amd64 g++ amd64 4:5.3.1-1ubuntu1 [1504 B]
Fetched 1504 B in 0s (2852 B/s)
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/g/gcc-5/libstdc++-5-dev_5.4.0-6ubuntu1~16.04.11_amd64.deb  404  Not Found [IP: 91.189.88.31 80]

E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/g/gcc-5/g++-5_5.4.0-6ubuntu1~16.04.11_amd64.deb  404  Not Found [IP: 91.189.88.31 80]

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Exited with code 100
```

pllim:
Should get this fixed before RC goes out at least...

PR #9552, opened 2019-11-06
MNT: Handle coordinates test warnings

pllim:
### Description

This pull request is to address warnings emitted by tests when `setup.cfg` is set to stop ignoring `pytest` warnings and I run `python setup.py test -P coordinates --remote-data`.

As part of #7928

### DIFF
--- a/astropy/coordinates/builtin_frames/galactocentric.py
+++ b/astropy/coordinates/builtin_frames/galactocentric.py
@@ -215,20 +215,22 @@ doc_footer = """
 
     Examples
     --------
+    .. testsetup::
+
+        >>> from astropy.coordinates import galactocentric_frame_defaults
+        >>> _ = galactocentric_frame_defaults.set('pre-v4.0')
+
     To transform to the Galactocentric frame with the default
     frame attributes, pass the uninstantiated class name to the
     ``transform_to()`` method of a coordinate frame or
     `~astropy.coordinates.SkyCoord` object::
 
-        >>> import warnings
         >>> import astropy.units as u
         >>> import astropy.coordinates as coord
         >>> c = coord.ICRS(ra=[158.3122, 24.5] * u.degree,
         ...                dec=[-17.3, 81.52] * u.degree,
         ...                distance=[11.5, 24.12] * u.kpc)
-        >>> with warnings.catch_warnings():
-        ...     warnings.simplefilter('ignore')  # Ignore deprecation warning
-        ...     c.transform_to(coord.Galactocentric) # doctest: +FLOAT_CMP
+        >>> c.transform_to(coord.Galactocentric) # doctest: +FLOAT_CMP
         <Galactocentric Coordinate (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
             ( 266.4051, -28.936175)>, galcen_distance=8.3 kpc, galcen_v_sun=( 11.1,  232.24,  7.25) km / s, z_sun=27.0 pc, roll=0.0 deg): (x, y, z) in kpc
             [( -9.6083819 ,  -9.40062188,  6.52056066),
@@ -237,9 +239,7 @@ doc_footer = """
     To specify a custom set of parameters, you have to include extra keyword
     arguments when initializing the Galactocentric frame object::
 
-        >>> with warnings.catch_warnings():
-        ...     warnings.simplefilter('ignore')  # Ignore deprecation warning
-        ...     c.transform_to(coord.Galactocentric(galcen_distance=8.1*u.kpc)) # doctest: +FLOAT_CMP
+        >>> c.transform_to(coord.Galactocentric(galcen_distance=8.1*u.kpc)) # doctest: +FLOAT_CMP
         <Galactocentric Coordinate (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
             ( 266.4051, -28.936175)>, galcen_distance=8.1 kpc, galcen_v_sun=( 11.1,  232.24,  7.25) km / s, z_sun=27.0 pc, roll=0.0 deg): (x, y, z) in kpc
             [( -9.40785924,  -9.40062188,  6.52066574),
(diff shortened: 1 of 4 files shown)


astropy/coordinates/tests/test_frames_with_velocity.py
```diff
@@ -1,36 +1,37 @@
...
 
 
 def test_api():
     # transform observed Barycentric velocities to full-space Galactocentric
-    gc_frame = Galactocentric()
-    icrs = ICRS(ra=151.*u.deg, dec=-16*u.deg, distance=101*u.pc,
-                pm_ra_cosdec=21*u.mas/u.yr, pm_dec=-71*u.mas/u.yr,
-                radial_velocity=71*u.km/u.s)
-    icrs.transform_to(gc_frame)
+    with galactocentric_frame_defaults.set('pre-v4.0'):
```
pllim: Is this the correct value?

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -220,12 +220,15 @@ def validate(cls, value):
...
+        >>> import warnings
         >>> import astropy.units as u
         >>> import astropy.coordinates as coord
         >>> c = coord.ICRS(ra=[158.3122, 24.5] * u.degree,
         ...                dec=[-17.3, 81.52] * u.degree,
         ...                distance=[11.5, 24.12] * u.kpc)
-        >>> c.transform_to(coord.Galactocentric) # doctest: +FLOAT_CMP
+        >>> with warnings.catch_warnings():
+        ...     warnings.simplefilter('ignore')  # Ignore deprecation warning
+        ...     c.transform_to(coord.Galactocentric) # doctest: +FLOAT_CMP
```
astrofrog: I'm not a fan of needing to catch warnings in the doctests since this complicates the user examples. In fact if users run these examples we want them to see the warnings not ignore them. So I'm wondering if it would be better to include in the list of warnings we whitelist for tests in https://github.com/astropy/astropy/blob/master/astropy/tests/helper.py - alternatively we should set the value of ``galactocentric_frame_defaults`` at the start of the example. Same comment for the other changes in this file.

astropy/coordinates/tests/test_frames_with_velocity.py
astrofrog: I'm going to guess that it might make more sense to use 'latest' here to make sure this example always works with the latest version (since we're not checking output values here, just that it runs).

astropy/coordinates/tests/test_frames_with_velocity.py
```diff
@@ -178,15 +179,15 @@ def test_frame_affinetransform(kwargs, expect_success):
     an AffineTransform fails without full-space data, but this just checks that
     things work as expected at the frame level as well.
     """
+    with galactocentric_frame_defaults.set('pre-v4.0'):
```
astrofrog: Same here, maybe just use 'latest'

docs/coordinates/galactocentric.rst
```diff
@@ -179,8 +179,11 @@ globally set the frame defaults at runtime. For example, the default parameter
 values can be seen by initializing the `~astropy.coordinates.Galactocentric`
 frame with no arguments::
 
+    >>> import warnings
     >>> from astropy.coordinates import Galactocentric
-    >>> Galactocentric()
+    >>> with warnings.catch_warnings():
+    ...     warnings.simplefilter('ignore')  # Ignore deprecation warning
+    ...     Galactocentric()
```
astrofrog: Can't we let the warning be shown and include it in the expected output?

astropy/coordinates/builtin_frames/galactocentric.py
pllim: I vaguely remember @adrn had a tough time setting the science state in doctest but I don't know if this particular doctest was affected by it.

docs/coordinates/galactocentric.rst
pllim: Will that silence `pytest` warnings?

astrofrog: That's a good question! If it doesn't, what about including the warning here anyway and then using doctest SKIP?

pllim: Skipping code is better than two lines to catch warning? I am not convinced.

astropy/coordinates/builtin_frames/galactocentric.py
adrn: I agree with @astrofrog about catching warnings in the doctest! The fact that this issues a warning is actually "correct behavior", so maybe we want to just add words here to explain that it is expected?

docs/coordinates/galactocentric.rst
adrn: How about we skip this particular test, and add a block near the bottom of the file with a short example of how to catch the warning, and do this down there?

astropy/coordinates/builtin_frames/galactocentric.py
pllim: My goal is for `pytest` to not show that warning in the "warnings" when tests are done. I am not sure how explaining it in the doc will make `pytest` understand.

astropy/coordinates/tests/test_frames_with_velocity.py
pllim: Done!

astropy/coordinates/builtin_frames/galactocentric.py
pllim: Okay, let's see if `testsetup` is magical enough.

pllim:
@astrofrog and @adrn , I removed warning catching from doc. Thanks for the reviews!

astropy/coordinates/builtin_frames/galactocentric.py
astrofrog: @pllim - I guess that fundamentally what we present to the users is what matters the most, so we shouldn't obfuscate the examples just to avoid warnings in the test suite. I agree this example is tricky though, but we need to work around it in a way that doesn't complicate the user examples or they will copy that code and never see the warning they are supposed to see :)

pllim: Yeah, I get it. I think the `testsetup` should be acceptable to us all?

pllim:
> whether it makes sense to use 'pre-v4.0' or whether we should use v4.0 or latest for the examples

I did try using `'latest'` first but then all the doctest output numbers changed and it was more than what I want to do.

astrofrog:
Yes I guess we could leave pre-v4.0 for now since that is just making the current assumption explicit, then changing it in a separate PR?

pllim:
test failure is unrelated -- #9551

adrn:
Thanks @pllim !

PR #9522, opened 2019-11-02
fix bug with compound model inverse serialization

nden:
### Description

This pull request is to address a problem with serialization of compound models when only one of them has a custom inverse. In this case the inverse was not written to file.

```
m1 = Polynomial1D(1) & Polynomial1D(1)
m2 = Polynomial1D(1) & Polynomial1D(1)
m1.inverse = m2
m = Shift(1) & Shift(2) | m1
fa = AsdfFile()
fa.tree['model'] = m
fa.write_to('model.asdf')
f = asdf.open('model.asdf').tree['model'].inverse
>           raise NotImplementedError("Inverse function not provided")
E           NotImplementedError: Inverse function not provided

```

PR #9558, opened 2019-11-07
updating individual parameters of compound models

nden:
### Description

This pull request is to address a problem with updating parameters of compound models, for example:

```
g = Gaussian1D()
c = Const1D(2)
m = g+c
m.amplitude_1 = 100
m
 _<CompoundModel(amplitude_0=1., mean_0=0., stddev_0=1., amplitude_1=2.)>_
m.amplitude_1                                                                                                                  
_Parameter('amplitude', value=2.0)_
```
It also adds tests addressing #9522.

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -2560,7 +2560,7 @@ class CompoundModel(Model):
             self.linear = False
         self.eqcons = []
         self.ineqcons = []
-        self.map_parameters()
+        self._map_parameters()
 
     def __len__(self):
         return self._n_models
@@ -2620,7 +2620,6 @@ class CompoundModel(Model):
         # names.
 
         if 'equivalencies' in kw:
-            #self.map_parameters()
             # Restructure to be useful for the individual model lookup
             kw['inputs_map'] = [(value[0], (value[1], key)) for
                                 key, value in self.inputs_map().items()]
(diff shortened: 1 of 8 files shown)


perrygreenfield:
Any effect on performance by always doing map_parameters at __init__?

astropy/modeling/tests/test_compound.py
```diff
@@ -633,3 +640,34 @@ def test_bounding_box_with_units():
     t = Tabular1D(points, lt)
 
     assert(t(1 * u.pix, with_bounding_box=True) == 1. * u.AA)
+
+
+@pytest.mark.parametrize('poly', [Chebyshev2D(1, 2), Polynomial2D(2), Legendre2D(1, 2)])
+def test_compound_with_polynomials_2d(poly):
+    """
+    Tests that polynomials are scaled when used in compound models.
```
perrygreenfield: Shouldn't that say offset instead of scaled? (also for the following test)

astropy/modeling/core.py
```diff
@@ -2561,6 +2560,7 @@ def __init__(self, op, left, right, name=None, inverse=None):
             self.linear = False
         self.eqcons = []
         self.ineqcons = []
+        self.map_parameters()
```
perrygreenfield: Since this obviates the need for a user to ever all this method, make it private by renaming to _map_parameters?

nden:
Regarding performance - evaluation is obviously the same. Initialization is a bit slower. For 7 models it's 15 % slower.

perrygreenfield:
It probably would be good to check the NIRSPEC WCS models given how many components they have.

nden:
It's practically the same (but slightly less, about 10%) for a Nirspec file.

PR #9563, opened 2019-11-07
Address new warnings in time.

mhvk:
Mostly the new `Stardate` format, which probably should use TAI instead of UTC (see https://github.com/astropy/astropy/pull/7928#issuecomment-551305274). cc @SolarDrew. (Or should it be TDB? I guess really a galactic barycentric time might be more logical...)


pllim:
@astrofrog , I think you are the first one to write a test for, uh, "undocumented features".

Thanks for catching this, @mhvk !

PR #9567, opened 2019-11-08
Fix more warnings in test suite

astrofrog:
This fixes some of the remaining warnings in ``io``, ``stats``, and ``timeseries``

@mhvk - take a look at the updated test for astropy.time's argsort - you'll notice that I had to xfail the utc case, and I'm not sure how to fix that one. Any ideas?

⚠️ note that there is a reasonably significant change here, which is the removal of the bayesian blocks warning about ``p0``, as discussed in https://github.com/astropy/astropy/issues/9568 and https://github.com/astropy/astropy/issues/4927

### DIFF
--- a/astropy/io/ascii/tests/test_cds_header_from_readme.py
+++ b/astropy/io/ascii/tests/test_cds_header_from_readme.py
@@ -1,7 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import numpy as np
 from astropy.io import ascii
-from .common import (assert_equal, assert_almost_equal, has_isnan,
+from .common import (assert_equal, assert_almost_equal,
                      setup_function, teardown_function)
 
 
@@ -138,14 +139,12 @@ def test_header_from_readme():
          1.958,
          1.416,
          0.949]
-    if has_isnan:
-        from .common import isnan
-        for i, val in enumerate(table.field('Q')):
-            if isnan(val):
-                # text value for a missing value in that table
-                assert Q[i] == -9.999
-            else:
-                assert val == Q[i]
+    for i, val in enumerate(table.field('Q')):
+        if val is np.ma.masked:
+            # text value for a missing value in that table
+            assert Q[i] == -9.999
+        else:
+            assert val == Q[i]
 
 
 def test_cds_units():
--- a/astropy/stats/bayesian_blocks.py
+++ b/astropy/stats/bayesian_blocks.py
@@ -418,14 +418,6 @@ class Events(FitnessFunc):
         gamma})`.
         If ``ncp_prior`` is specified, ``gamma`` and ``p0`` is ignored.
     """
-    def __init__(self, p0=0.05, gamma=None, ncp_prior=None):
-        if p0 is not None and gamma is None and ncp_prior is None:
-            warnings.warn('p0 does not seem to accurately represent the false '
-                          'positive rate for event data. It is highly '
-                          'recommended that you run random trials on signal-'
-                          'free noise to calibrate ncp_prior to achieve a '
-                          'desired false positive rate.', AstropyUserWarning)
-        super().__init__(p0, gamma, ncp_prior)
 
     def fitness(self, N_k, T_k):
         # eq. 19 from Scargle 2012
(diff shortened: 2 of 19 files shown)


astropy/io/ascii/tests/test_cds_header_from_readme.py
```diff
@@ -138,14 +139,12 @@ def test_header_from_readme():
...
-    if has_isnan:
-        from .common import isnan
-        for i, val in enumerate(table.field('Q')):
-            if isnan(val):
-                # text value for a missing value in that table
-                assert Q[i] == -9.999
-            else:
-                assert val == Q[i]
+    for i, val in enumerate(table.field('Q')):
+        if val is np.ma.masked:
```
astrofrog: @taldcroft - is this an ok change?

astropy/stats/bayesian_blocks.py
```diff
@@ -105,13 +105,13 @@ def bayesian_blocks(t, x=None, sigma=None,
...
     >>> t = np.random.normal(size=100)
-    >>> edges = bayesian_blocks(t, fitness='events', p0=0.01)
+    >>> edges = bayesian_blocks(t, fitness='events', gamma=1)
 
     Event data with repeats:
 
     >>> t = np.random.normal(size=100)
     >>> t[80:] = t[:20]
-    >>> edges = bayesian_blocks(t, fitness='events', p0=0.01)
+    >>> edges = bayesian_blocks(t, fitness='events', gamma=1)
```
astrofrog: @crawfordsm @larrybradley - I'm not sure what the best way is to avoid the warning about p0 - this works but I'm not sure if it's a sensible value of gamma.

astropy/stats/info_theory.py
```diff
@@ -364,7 +364,7 @@ def akaike_info_criterion_lsq(ssr, n_params, n_samples):
     >>> # Fit with three Gaussians
     >>> g3_init = (models.Gaussian1D(.1, 0, 0.1)
     ...            + models.Gaussian1D(.1, 0.2, 0.15)
-    ...            + models.Gaussian1D(2., .4, 0.1))
+    ...            + models.Gaussian1D(2.4, .4, 0.1))
```
astrofrog: The fit was failing before

pllim:
If this is a competition on how fast both of us can fix warnings, you totally won. LOL!

xref #7928

astropy/visualization/tests/test_histogram.py
```diff
@@ -49,6 +49,7 @@ def test_hist_specify_ax(rseed=0):
 
 
 @pytest.mark.skipif('not HAS_PLT')
+@pytest.mark.filterwarnings('ignore:p0 does not seem to accurately represent')
```
pllim: Should we add a note to remove this filter when #9568 is fixed?

astrofrog: I'll add a note in #9568

astropy/stats/bayesian_blocks.py
larrybradley: Is this necessary now that the `p0` warning was removed?

astropy/visualization/tests/test_histogram.py
astrofrog: Warning now removed, so resolving this.

astropy/stats/bayesian_blocks.py
astrofrog: Yes this is fixed now so I've reverted.

pllim:
Re: reasonably significant change -- Does that mean we need a change log now?

astrofrog:
@pllim - I added one, since that warning was very annoying so I consider this an improvement in user experience! :)

pllim:
@astrofrog , are you waiting for anyone else in particular to approve?

astrofrog:
I'd like @taldcroft to check the table-related changes before we merge. There was a question for @mhvk too but I can actually just open an issue for that since it doesn't have to be solved here.

mhvk:
@astrofrog - this looks all OK to me; if there is anything specific I should look at, let me know, or indeed open another issue.

astropy/io/ascii/tests/test_cds_header_from_readme.py
taldcroft: Looks OK.

pllim:
Thank you, everyone!

PR #9569, opened 2019-11-08
Handle utils test warnings

pllim:
### Description

This pull request is to address test warnings emitted by `utils`. @aarchiba should have a look as she added most of these tests in #9182.

This is a part of #7928

### DIFF
--- a/astropy/table/tests/test_operations.py
+++ b/astropy/table/tests/test_operations.py
@@ -1012,7 +1012,7 @@ class TestVStack():
                     'vstack unavailable' in str(err.value))
 
 
-class TestCStack():
+class TestDStack():
 
     def _setup(self, t_cls=Table):
         self.t1 = t_cls.read([' a   b',
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -1001,6 +1001,7 @@ def test_is_url_in_cache_local(temp_cache, valid_urls, invalid_urls):
     assert not is_url_in_cache(nonexistent)
 
 
+@pytest.mark.filterwarnings('ignore:Not clearing data cache - cache inaccessible due to FileNotFoundError')
 def test_check_download_cache(tmpdir, temp_cache, valid_urls, invalid_urls):
     testurl, testurl_contents = next(valid_urls)
     testurl2, testurl2_contents = next(valid_urls)
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -12,7 +12,6 @@ import hashlib
 import pathlib
 import platform
 import tempfile
-import warnings
 import itertools
 import contextlib
 import urllib.error
@@ -1001,12 +1000,12 @@ def test_is_url_in_cache_local(temp_cache, valid_urls, invalid_urls):
     assert not is_url_in_cache(nonexistent)
 
 
-@pytest.mark.filterwarnings('ignore:Not clearing data cache - cache inaccessible due to FileNotFoundError')
 def test_check_download_cache(tmpdir, temp_cache, valid_urls, invalid_urls):
     testurl, testurl_contents = next(valid_urls)
     testurl2, testurl2_contents = next(valid_urls)
 
     zip_file_name = os.path.join(tmpdir, "the.zip")
+    # TODO: Ensure clear_download_cache not emit warning at all (PR 9569).
     clear_download_cache()
     check_download_cache()
 
--- a/astropy/utils/tests/test_timer.py
+++ b/astropy/utils/tests/test_timer.py
@@ -16,7 +16,6 @@ import pytest
 import numpy as np
 
 # LOCAL
-from astropy.utils.timer import RunTimePredictor
 from astropy.modeling.fitting import ModelsError
 
 
@@ -36,6 +35,7 @@ def func_to_time(x):
 @pytest.mark.filterwarnings("ignore")
 def test_timer():
     """Test function timer."""
+    from astropy.utils.timer import RunTimePredictor
     p = RunTimePredictor(func_to_time)
 
     # --- These must run before data points are introduced. ---


aarchiba:
I can't speak for the timer stuff but the rest should probably be turned into explicit checks that appropriate warnings are raised by specific lines. If the cache is readonly, for example, we want to confirm that operations that would normally write to it raise a relevant warning, and operations that only require reading don't warn.

astropy/utils/tests/test_data.py
```diff
@@ -1260,6 +1262,8 @@ def test_download_cache_update_doesnt_damage_cache(temp_cache, valid_urls):
     download_file(u, cache="update")
 
 
+@pytest.mark.filterwarnings(
+    'ignore:Remote data cache could not be accessed')
```
aarchiba: It's weird that this is needed; I thought all such warnings would be caught and suppressed by the `pytest.warns` context managers. I guess I missed some?

aarchiba:
I can go through and more carefully test warning generation if you like, but I'm not sure how to do that without tangling up this PR.

pllim:
> I can go through and more carefully test warning generation if you like, but I'm not sure how to do that without tangling up this PR.

@aarchiba , either a separate standalone PR or a PR-to-my-PR would be acceptable. Up to you. If you choose the former, I can deal with the conflicts later.

astropy/table/tests/test_operations.py
```diff
@@ -1012,7 +1012,7 @@ def test_mixin_functionality(self, mixin_cols):
                     'vstack unavailable' in str(err.value))
 
 
-class TestCStack():
+class TestDStack():
```
pllim: @taldcroft , this okay?

astropy/utils/tests/test_data.py
```diff
@@ -1001,6 +1001,7 @@ def test_is_url_in_cache_local(temp_cache, valid_urls, invalid_urls):
     assert not is_url_in_cache(nonexistent)
 
 
+@pytest.mark.filterwarnings('ignore:Not clearing data cache - cache inaccessible due to FileNotFoundError')
```
pllim: @aarchiba , seems like #9575 didn't cache this one. Is handling it like this okay?

aarchiba: I get no warnings on my machine, at least, though it's not tested in CI. No reason this should emit a warning - clear_download_cache should work fine (no warnings) on a nonexistent cache, and it shouldn't be running on a nonexistent cache here anyway. But if you can produce a warning, let me know and I'll try to track it down?

astropy/table/tests/test_operations.py
taldcroft: Yes, thanks.

pllim:
@aarchiba , here is what I get when running the test on my machine locally with `python setup.py test -P utils` and turning warning into error:
```
platform linux -- Python 3.7.3, pytest-5.0.1, py-1.8.0, pluggy-0.12.0

Running tests with Astropy version 4.1.dev26889.
Running tests in astropy/utils docs/utils.

Date: 2019-11-10T16:41:26

Platform: Linux-3.10.0-1062.1.2.el7.x86_64-x86_64-with-redhat-7.7-Maipo

Executable: .../miniconda/envs/py37/bin/python

Full Python Version: 
3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0]

byteorder: little

Package versions: 
Numpy: 1.17.3
Scipy: 1.3.1
Matplotlib: 3.1.1
h5py: not available
Pandas: 0.24.2
Cython: 0.29.13
Scikit-image: not available
asdf: 2.4.0.dev63+gb67cb02
astropy-helpers: 3.2rc1

rootdir: /tmp/astropy-test-5dwkmwkz/lib/python3.7/site-packages, inifile: setup.cfg
plugins: doctestplus-0.4.0, remotedata-0.3.1, openfiles-0.3.2, arraydiff-0.3,
astropy-header-0.1.1, forked-1.1.3, xdist-1.30.0,
ci-watson-0.0.0.dev76+g4b73d3b5, asdf-2.4.0.dev63+gb6
```
```
=== FAILURES ===
___ test_check_download_cache ___

hashorurl = None, pkgname = 'astropy'

    def clear_download_cache(hashorurl=None, pkgname='astropy'):
        # ...
                else:  # it's a path
                    filepath = os.path.join(dldir, hashorurl)
                    if not _is_inside(filepath, dldir):
                        # Should this be ValueError? IOError?
                        raise RuntimeError(
                            f"attempted to use clear_download_cache on the path "
                            f"{filepath} outside the data cache directory {dldir}")
                    if os.path.exists(filepath):
                        # Convert to list because we'll be modifying it as we go
                        for k, v in list(url2hash.items()):
                            if v == filepath:
                                del url2hash[k]
                        os.unlink(filepath)
>                       return

astropy/utils/data.py:1439: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <contextlib._GeneratorContextManager object at 0x7f26b40fc080>, type = None, value = None, traceback = None

    def __exit__(self, type, value, traceback):
        if type is None:
            try:
>               next(self.gen)

.../miniconda/envs/py37/lib/python3.7/contextlib.py:119: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

pkgname = 'astropy', write = True

    @contextlib.contextmanager
    def _cache(pkgname, write=False):
        # ...
        if write:
            with _cache_lock(pkgname, need_write=True), \
                    shelve.open(urlmapfn, flag="c") as url2hash:
>               yield dldir, url2hash

astropy/utils/data.py:1626: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <shelve.DbfilenameShelf object at 0x7f26b40fc3c8>, type = None, value = None, traceback = None

    def __exit__(self, type, value, traceback):
>       self.close()

.../miniconda/envs/py37/lib/python3.7/shelve.py:138: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <shelve.DbfilenameShelf object at 0x7f26b40fc3c8>

    def close(self):
        if self.dict is None:
            return
        try:
>           self.sync()

.../miniconda/envs/py37/lib/python3.7/shelve.py:144: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <shelve.DbfilenameShelf object at 0x7f26b40fc3c8>

    def sync(self):
        if self.writeback and self.cache:
            self.writeback = False
            for key, entry in self.cache.items():
                self[key] = entry
            self.writeback = True
            self.cache = {}
        if hasattr(self.dict, 'sync'):
>           self.dict.sync()

.../miniconda/envs/py37/lib/python3.7/shelve.py:172: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <dbm.dumb._Database object at 0x7f26b40fc240>

    def _commit(self):
        # CAUTION:  It's vital that _commit() succeed, and _commit() can
        # be called from __del__().  Therefore we must never reference a
        # global in this routine.
        if self._index is None or not self._modified:
            return  # nothing to do
    
        try:
            self._os.unlink(self._bakfile)
        except OSError:
            pass
    
        try:
            self._os.rename(self._dirfile, self._bakfile)
        except OSError:
            pass
    
>       with self._io.open(self._dirfile, 'w', encoding="Latin-1") as f:
E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-lim/pytest-75/test_check_download_cache0/astropy/download/py3/urlmap.dir'

.../miniconda/envs/py37/lib/python3.7/dbm/dumb.py:135: FileNotFoundError

tmpdir = local('/tmp/pytest-of-lim/pytest-75/test_check_download_cache0'), temp_cache = None, valid_urls = <generator object valid_urls.<locals>._valid_urls at 0x7f26b4b4b138>
invalid_urls = <generator object invalid_urls.<locals>._invalid_urls at 0x7f26b411c0c0>

    def test_check_download_cache(tmpdir, temp_cache, valid_urls, invalid_urls):
        testurl, testurl_contents = next(valid_urls)
        testurl2, testurl2_contents = next(valid_urls)
    
        zip_file_name = os.path.join(tmpdir, "the.zip")
>       clear_download_cache()

astropy/utils/tests/test_data.py:1009: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

hashorurl = None, pkgname = 'astropy'

    def clear_download_cache(hashorurl=None, pkgname='astropy'):
        # ...
        except OSError as e:
            msg = 'Not clearing data cache - cache inaccessible due to '
            estr = '' if len(e.args) < 1 else (': ' + str(e))
>           warn(CacheMissingWarning(msg + e.__class__.__name__ + estr))
E           astropy.utils.data.CacheMissingWarning:
E           Not clearing data cache - cache inaccessible due to FileNotFoundError:
E           [Errno 2] No such file or directory:
E '/tmp/pytest-of-lim/pytest-75/test_check_download_cache0/astropy/download/py3/urlmap.dir'

astropy/utils/data.py:1446: CacheMissingWarning
```

astropy/utils/tests/test_data.py
```diff
@@ -1006,7 +1005,9 @@ def test_check_download_cache(tmpdir, temp_cache, valid_urls, invalid_urls):
     testurl2, testurl2_contents = next(valid_urls)
 
     zip_file_name = os.path.join(tmpdir, "the.zip")
-    clear_download_cache()
+    with pytest.warns(CacheMissingWarning, match='Not clearing data cache - '
```
pllim: @aarchiba , I made it much more explicit now. So if indeed it is just my machine acting weird, the jobs would fail in CI.

pllim:
This is weird, seems like that warning is not emitted in CircleCI 32-bit job and Travis CI Python 3.8 job, but the other Travis CI jobs are seemed to catch it. (The other CircleCI jobs are irrelevant here.)

aarchiba:
There is some difference based on which dbm backend is used by shelve - gdbm versus dbm versus dumb.py. It might be possible to force all machines everywhere to use dumb.py, which would resolve certain problems - for example if your astropy installation creates a gdbm cache index but then you try to run astropy in conda, which doesn't provide gdbm, astropy will choke. This would have the side effect of making the shelve behaviour much more uniform (but also slower - though I'm not sure speed is much of a concern for the cache). I can also poke around and try to ensure the warning is not emitted when it's not necessary (the case here) but it'll be tricky unless I can arrange to control the database backend (I have ideas). I guess I should create an issue about database format.

astropy/utils/tests/test_data.py
aarchiba: This is still likely to be trouble - it will fail on any machine (like mine) that doesn't emit the warning. There's a filterwarnings context manager that would succeed in ignoring the warning but I think it's better to prevent the warning being emitted - it is fine if clear_download_cache() is called when no cache exists, and we shouldn't be emitting warnings.

Wait. I think what's going on is maybe that the shelve is opened, then the on-disk file is deleted, then the shelve is closed, and this backend doesn't like that. Which is not unreasonable. But I can catch that exception and let it go without a warning, because this only happens with a no-argument clear_download_cache().

pllim:
Thanks for the insight, @aarchiba ! Maybe I'll just not handling this one warning for now in this PR.

If necessary, to get #7928 in for 4.0, I can put in the blind `pytest.mark.filterwarnings` again on it at the last minute.

pllim:
I cancelled the rest of CircleCI tests, as they are unnecessary and the queue is too long.

PR #9582, opened 2019-11-11
Fix bug in the frame attributes references for Galactocentric frame

adrn:
### Description
 
While working on astrofrog/astropy#91 to write a what's new entry for the changes to `Galactocentric`, I noticed that the dictionary of references (i.e. scientific papers that the parameter values come from) was not being properly handled. This adjusts the logic to fix how references are stored, and adds a test.

### DIFF
--- a/astropy/coordinates/tests/test_frames.py
+++ b/astropy/coordinates/tests/test_frames.py
@@ -1141,14 +1141,19 @@ def test_component_names_repr():
     assert repr(frame).count("JUSTONCE") == 1
 
 
+@pytest.fixture
 def reset_galactocentric_defaults():
     # TODO: this can be removed, along with the "warning" test below, once we
     # switch the default to 'latest' in v4.1
     from astropy.coordinates import galactocentric_frame_defaults
+
+    # Resets before each test, and after (the yield is pytest magic)
+    galactocentric_frame_defaults._value = 'pre-v4.0'
+    yield
     galactocentric_frame_defaults._value = 'pre-v4.0'
 
 
-def test_galactocentric_defaults():
+def test_galactocentric_defaults(reset_galactocentric_defaults):
     from astropy.coordinates import (Galactocentric,
                                      galactocentric_frame_defaults,
                                      BaseCoordinateFrame,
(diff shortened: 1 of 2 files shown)


pllim:
In the spirit of #7928, please run these new tests without ignoring warnings and see if they introduce any new warnings. If so, please handle the warnings accordingly. Thanks!

adrn:
I copied the setup.cfg from #7928 and ran `python setup.py test -t astropy/coordinates` locally -- not seeing any new warnings

adrn:
@pllim Now I did - still no warnings!

pllim:
Thanks for checking, @adrn!

astropy/coordinates/tests/test_frames.py
```diff
@@ -1204,3 +1204,44 @@ def test_galactocentric_default_warning():
...
+        for k in galcen_custom.get_frame_attr_names():
+            if k == 'roll':  # no reference for this parameter
+                continue
+
+            if k == 'z_sun':
+                assert k not in galcen_custom.frame_attribute_references
+            else:
+                assert k in galcen_custom.frame_attribute_references
+
+    reset_galactocentric_defaults()
```
eteq: shouldn't this be in a `finally` block (and then the whole thing in a ``try``)?

Or alternatively, you could use the fixtures finalization magic of pytest by doing something like:
```
@pytest.fixture
def reset_galactocentric_defaults():
    yield 
    ... do what the current reset_galactocentric_defaults does...
```
and then have the pytest call for this function be:
```
def test_galactocentric_references(reset_galactocentric_defaults)
```

astropy/coordinates/tests/test_frames.py
```diff
@@ -1178,16 +1183,14 @@ def test_galactocentric_defaults():
...
 
-def test_galactocentric_default_warning():
+def test_galactocentric_default_warning(reset_galactocentric_defaults):
     from astropy.coordinates import (Galactocentric,
                                      galactocentric_frame_defaults)
 
     # Note: this is necessary because the doctests may alter the state of the
     # class globally and simultaneously when the tests are run in parallel
-    reset_galactocentric_defaults()
+    # reset_galactocentric_defaults()
```
mhvk: Oops, missed in review...

pllim: So is this line supposed to be removed?

mhvk: Indeed, no reason to just comment it out, it is now done in `setup` and `teardown`. Indeed, the comment above can go too...

pllim: OK, I'll fold it into #7928.

PR #9585, opened 2019-11-11
Improve test runner error when pytest dependency is missing

adrn:
This just adds a bit more information to the `RuntimeError` that is thrown when running tests with the Astropy test runner and a pytest dependency is missing.

As noted by @astrofrog, it might be even better to require a minimum version of `pytest-astropy`, and check for that version when executing the test runner

### DIFF
--- a/astropy/tests/runner.py
+++ b/astropy/tests/runner.py
@@ -176,7 +176,6 @@ class TestRunnerBase:
         "have a previous version installed, e.g., "
         "'pip install pytest-astropy --upgrade' or the equivalent with conda).")
 
-
     @classmethod
     def _has_test_dependencies(cls):  # pragma: no cover
         # Using the test runner will not work without these dependencies, but
--- a/astropy/tests/runner.py
+++ b/astropy/tests/runner.py
@@ -170,7 +170,11 @@ class TestRunnerBase:
         """
 
     _required_dependencies = ['pytest', 'pytest_remotedata', 'pytest_doctestplus', 'pytest_astropy_header']
-    _missing_dependancy_error = "Test dependencies are missing. You should install the 'pytest-astropy' package."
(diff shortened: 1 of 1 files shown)


astropy/tests/runner.py
```diff
@@ -170,7 +170,12 @@ def _generate_args(self, **kwargs):
         """
 
     _required_dependencies = ['pytest', 'pytest_remotedata', 'pytest_doctestplus', 'pytest_astropy_header']
-    _missing_dependancy_error = "Test dependencies are missing. You should install the 'pytest-astropy' package."
+    _missing_dependancy_error = (
+        "Test dependencies are missing: {module}. You should install the "
```
pllim: Need f-string here? Maybe not... My string-fu is weak, but just in case.
```suggestion
        f"Test dependencies are missing: {module}. You should install the "
```

astropy/tests/runner.py
```diff
@@ -170,7 +170,12 @@ def _generate_args(self, **kwargs):
...
+        "'pytest-astropy' package (you may need to update the package if you "
+        "have a previous version installed, e.g., "
+        "`pip install pytest-astropy --upgrade` or the equivalent with conda).")
```
pllim: This is not Sphinx, yo.
```suggestion
        "'pip install pytest-astropy --upgrade' or the equivalent with conda).")
```

adrn: Oops, brain fart.

astropy/tests/runner.py
adrn: Nope! I use `.format()` later on.

pllim: Ah, okay. Sorry for the noise!

adrn:
Ack, I should have `[ci skip]`'d that last commit...

PR #9589, opened 2019-11-11
Handle FITS doc warnings

pllim:
### Description

This pull request is to address the remaining warnings from FITS doctest. People do not seem to like explicit warning handling in the doctest snippet, so I opted for `# doctest: +SKIP` instead unless I can fix it.

This is part of #7928

### DIFF
--- a/docs/io/fits/usage/headers.rst
+++ b/docs/io/fits/usage/headers.rst
@@ -305,10 +305,11 @@ the example below is flagged by such verification. More about verification in
 
 ::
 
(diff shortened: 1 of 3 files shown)


saimn:
I wish there was a better way to test warnings in doctests, maybe with a plugin to redirect stderr to stdout, but ok let's go with this to help with #7928 :)

pllim:
astropy/pytest-doctestplus#84 would be what we need but it was just merged, so maybe we need to come back in the future and use the new `+IGNORE_WARNINGS` instead when a new `pytest-doctestplus` released is made.

PR #9593, opened 2019-11-11
Handle test warning in test_precision.py

pllim:
### Description

This pull request is to address test warning from a test added in #9577 . cc @aarchiba 

This is part of #7928

### DIFF
--- a/astropy/time/tests/test_precision.py
+++ b/astropy/time/tests/test_precision.py
@@ -10,6 +10,7 @@ import astropy._erfa as erfa
 from astropy.time import Time, TimeDelta
 from astropy.time.utils import day_frac, two_sum
 from astropy.utils import iers
+from astropy.utils.exceptions import ErfaWarning
 
 allclose_jd = functools.partial(np.allclose, rtol=2. ** -52, atol=0)
 allclose_jd2 = functools.partial(np.allclose, rtol=2. ** -52,
@@ -178,7 +179,10 @@ def test_conversion_preserves_jd1_jd2_invariant():
     scale2 = 'tcb'
     jd1, jd2 = 0., 0.
     t = Time(jd1, jd2, scale=scale1, format="jd")
-    t2 = getattr(t, scale2)
+    with pytest.warns(
+            ErfaWarning,
+            match=r'ERFA function "taiutc" yielded 1 of "dubious year'):
+        t2 = getattr(t, scale2)
     assert t2.jd1 % 1 == 0
     assert abs(t2.jd2) <= 0.5
     assert abs(t2.jd2) < 0.5 or t2.jd1 % 2 == 0


astropy/time/tests/test_precision.py
```diff
@@ -178,7 +179,10 @@ def test_conversion_preserves_jd1_jd2_invariant():
     scale2 = 'tcb'
     jd1, jd2 = 0., 0.
     t = Time(jd1, jd2, scale=scale1, format="jd")
-    t2 = getattr(t, scale2)
+    with pytest.warns(
+            ErfaWarning,
+            match=r'ERFA function "taiutc" yielded 1 of "dubious year'):
+        t2 = getattr(t, scale2)
```
astrofrog: I wonder if it would make sense to instead pick dates that don't generate a warning? (I defer to @aarchiba or @mhvk though)

aarchiba: It's never been obvious to me which dates made ERFA generate a warning, apart from anything too far from the present. Unfortunately it's not trivial to find cases where this fails, and hypothesis tends to aim for zero as a "simple" example. The test might even not fail in this way for reasonable dates? I think I had it set up to prefer dates close to now over these.

taldcroft: There is a specific and fixed epoch after which ERFA / SOFA will issue the dubious year warning.  Empirically anything after 1960-01-01 00:00:00 is OK and anything before is not.  So if it is possible to avoid the warning then that seems like a good thing.

mhvk: Yes, it is basically when UTC is properly defined, which I guess is after 1960 and some number of years from the present (for which one can hope no leap second will occur - it is in `extern/erfa/dat.c`)

pllim: Can anyone give me the magic numbers then? 🙏

astrofrog: @mhvk - see this is is what I meant in https://github.com/astropy/astropy/issues/9587#issuecomment-552538314 - it would be nice to have warnings about UTC that were more informative about the range of valid years :)

astrofrog: > The test might even not fail in this way for reasonable dates?

If that's true then maybe it makes sense to keep the warning catching as-is?

astrofrog: @aarchiba - do you think we should just leave it as-is with the warning catching?

aarchiba: I'd say yeah, leave the warning catching, we know it'll generate that warning and it's okay (unless we clean up ERFA itself warning-wise). The bug it's testing for is orthogonal to this.

aarchiba: I guess the alternative might be to suppress but not test for the warning. That depends whether there are plans to change ERFA's warning generation.

aarchiba: > Yes, it is basically when UTC is properly defined, which I guess is after 1960 and some number of years from the present (for which one can hope no leap second will occur - it is in `extern/erfa/dat.c`)

It would definitely be good for the warning to contain this information; I can also adjust `reasonable_jd` in the hypothesis testing to ensure that it spans the whole range that does not make ERFA grumpy in this way.

pllim: Hmm, but as far as this PR is concerned, is there a consensus on how to proceed?

astrofrog: I've opened an issue specifically regarding improving the warning message here: https://github.com/astropy/astropy/issues/9603

For the hypothesis tests, restricting the range of ``reasonable_jd`` makes sense but only for scales that actually produce warning - we should make sure we test a wide range of dates for scales that don't have leap seconds.

@aarchiba - for this PR, I guess the question is, does the bug this is testing for happen also in the ``tai`` scale? If so we could just set ``scale='tai'`` and avoid the warning altogether?

astrofrog:
Since this is the last warning in the test suite, I'll merge this PR shortly if there are no objections so that we can flip the switch and make warnings into error to prevent more warnings from creeping back in.

PR #9594, opened 2019-11-11
Test basic operation with all backends

aarchiba:
### Description

The cache mechanism depends on python's `shelve` module, which can use one of three backends (`dumb`, `gdbm` and `ndbm`). These sometimes have different failure modes, so create a test that tries all backends available to a given interpreter.

The hope was that this would trigger a warning (cache problems at the end of `clear_download_cache()`) so that I could be sure that I had banished it, but in the event that doesn't seem to happen on Linux. Nevertheless I think I fixed that problem.

Relevant to #7928

### DIFF
--- a/astropy/utils/data.py
+++ b/astropy/utils/data.py
@@ -5,6 +5,7 @@
 import atexit
 import contextlib
 import dbm
+import errno
 import fnmatch
 import hashlib
 import os
@@ -1414,11 +1415,13 @@ def clear_download_cache(hashorurl=None, pkgname='astropy'):
 
     """
 
+    zapped_cache = False
     try:
         with _cache(pkgname, write=True) as (dldir, url2hash):
             if hashorurl is None:
                 if os.path.exists(dldir):
                     shutil.rmtree(dldir)
+                    zapped_cache = True
             elif _is_url(hashorurl):
                 try:
                     filepath = url2hash.pop(hashorurl)
@@ -1449,10 +1452,16 @@ def clear_download_cache(hashorurl=None, pkgname='astropy'):
                 # Clearing download cache just makes sure that the file or url
                 # is no longer in the cache regardless of starting condition.
     except OSError as e:
-        msg = 'Not clearing data cache - cache inaccessible due to '
-        estr = '' if len(e.args) < 1 else (': ' + str(e))
-        warn(CacheMissingWarning(msg + e.__class__.__name__ + estr))
-        return
+        if zapped_cache and e.errno == errno.ENOENT:
+            # We just deleted the directory and, on Windows (?) the "dumb"
+            # backend tried to write itself out to a nonexistent directory.
+            # It's fine for this to fail.
+            return
+        else:
+            msg = 'Not clearing data cache - cache inaccessible due to '
+            estr = '' if len(e.args) < 1 else (': ' + str(e))
+            warn(CacheMissingWarning(msg + e.__class__.__name__ + estr))
+            return
 
 
 def _get_download_cache_locs(pkgname='astropy'):
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -3,6 +3,7 @@
 
 import io
 import os
+import dbm
 import sys
 import stat
 import base64
@@ -12,6 +13,7 @@ import hashlib
 import pathlib
 import platform
 import tempfile
+import importlib
 import itertools
 import contextlib
 import urllib.error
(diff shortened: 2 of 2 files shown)


pllim:
I installed this patch and applied the `conftest.py` and `setup.cfg` settings on top of it, then I ran the tests locally with `python setup.py test -P utils --remote-data`, but unfortunately there is one failure, as attached:

[test_log.txt](https://github.com/astropy/astropy/files/3833299/test_log.txt)

aarchiba:
Okay, I was able to trigger the warning - it happens with `dbm.dumb` when a previously nonexistent cache is cleared. Fixed now. (Actually all I needed wase ENOENT instead of EEXIST but now I'm able to trigger the behaviour.)

astropy/utils/tests/test_data.py
```diff
@@ -1755,3 +1779,64 @@ def test_download_parallel_respects_pkgname(temp_cache, valid_urls):
...
+def test_lock_behaviour_if_directory_disappears(temp_cache, b):
+    dldir, urlmapfn = _get_download_cache_locs()
+    try:
+        dbm = importlib.import_module(b)
+    except ImportError:
+        pytest.skip(f"module {b} not available")
+    if b == "dbm.dumb":
+        c = pytest.raises(FileNotFoundError)
+    else:
+        @contextlib.contextmanager
```
pllim: @astrofrog added this in `master`, should you should probably use it instead of rolling your own; FYI:

https://github.com/astropy/astropy/blob/dce08d5f6236d14a3b3cee7cb715f5864df48ac9/astropy/utils/compat/context.py#L8

aarchiba: Thanks!

astropy/utils/data.py
```diff
@@ -1637,7 +1647,7 @@ def _cache(pkgname, write=False):
             with _cache_lock(pkgname), shelve.open(urlmapfn, flag="r") as url2hash:
                 # Copy so we can release the lock.
                 d = ReadOnlyDict(url2hash.items())
-        except dbm.error:
+        except dbm.error as e:
```
pllim: `e` is not used?

astropy/utils/data.py
```diff
@@ -1414,11 +1415,14 @@ def clear_download_cache(hashorurl=None, pkgname='astropy'):
 
     """
 
+    zapped_cache = False
     try:
         with _cache(pkgname, write=True) as (dldir, url2hash):
             if hashorurl is None:
+                #import pdb; pdb.set_trace()
```
pllim: Remove leftover from debug session?

astropy/utils/tests/test_data.py
```diff
@@ -186,6 +189,26 @@ def no_mkdir(p):
...
+
+
+@contextlib.contextmanager
+def shelve_backend(name):
+    """Ensure that shelve has access only to one backend."""
+    defaultmod = dbm._defaultmod
+    modules = dbm._modules
+    try:
+        try:
+                dbm._defaultmod = __import__(name, fromlist=['open'])
```
pllim: Please fix indentation.

astropy/utils/tests/test_data.py
```diff
@@ -1755,3 +1780,61 @@ def test_download_parallel_respects_pkgname(temp_cache, valid_urls):
                                pkgname=a)
     assert not get_cached_urls()
     assert len(get_cached_urls(pkgname=a)) == FEW
+
+
+@pytest.mark.parametrize("b", _shelve_possible_backends)
+def test_cache_with_different_shelve_backends(temp_cache, b, valid_urls):
```
pllim: Just to be consistent with the way "b" is defined in the other test function above.
```suggestion
def test_cache_with_different_shelve_backends(b, temp_cache, valid_urls):
```

astropy/utils/tests/test_data.py
```diff
@@ -1755,3 +1780,61 @@ def test_download_parallel_respects_pkgname(temp_cache, valid_urls):
...
+
+        for u, c in uc:
+            assert get_file_contents(
+                download_file(u, cache=True, sources=[])) == c
+
+        clear_download_cache()
+
+
+@pytest.mark.parametrize("b", _shelve_possible_backends)
+def test_lock_behaviour_if_directory_disappears(temp_cache, b):
```
pllim:
```suggestion
def test_lock_behaviour_if_directory_disappears(b, temp_cache):
```

pllim:
Update: Warning is gone on my machine with this patch! 🎉

aarchiba:
Sorry about the debugging detritus.

aarchiba:
Glad I could help! And it's good to have some tests that explore all the backends.

PR #9595, opened 2019-11-12
Fix unclosed file warning with nonstandard HDU.

saimn:
Fix another warning with `io.fits` :
```
astropy/io/fits/tests/test_nonstandard.py::TestNonstandardHdus::test_create_fitshdu_from_filename
  /home/simon/dev/astropy/astropy/io/fits/hdu/nonstandard.py:55: ResourceWarning: unclosed file <_io.FileIO name='/tmp/simon/fits-test-ulxk6ga8/test.fits' mode='rb' closefd=True>
    return cls.fromhdulist(HDUList.fromfile(filename), compress=compress)
```

### DIFF
--- a/astropy/io/fits/hdu/nonstandard.py
+++ b/astropy/io/fits/hdu/nonstandard.py
@@ -52,7 +52,10 @@ class FitsHDU(NonstandardExtHDU):
             Gzip compress the FITS file
         """
 
-        return cls.fromhdulist(HDUList.fromfile(filename), compress=compress)
+        hdulist = HDUList.fromfile(filename)
+        fitshdu = cls.fromhdulist(hdulist, compress=compress)
+        hdulist.close()
+        return fitshdu
 
     @classmethod
     def fromhdulist(cls, hdulist, compress=False):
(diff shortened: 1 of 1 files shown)


astropy/io/fits/hdu/nonstandard.py
```diff
@@ -52,7 +52,10 @@ def fromfile(cls, filename, compress=False):
             Gzip compress the FITS file
         """
 
-        return cls.fromhdulist(HDUList.fromfile(filename), compress=compress)
+        hdulist = HDUList.fromfile(filename)
```
MSeifert04: Just curious, why not use the context manager here?

pllim: Does `fromfile` have a context manager form? I have ever only used context manager with `open`.

saimn: I didn't think about it, thanks @MSeifert04.
@pllim , yes the context manager is on `HDUList`, and `fits.open` is just a shortcut for `HDUList.fromfile`.

saimn:
Thanks for the review :)

PR #9601, opened 2019-11-12
fix problem with fitting with tied constraints

nden:
### Description

This pull request is to address a bug in fitting with tied constraints. 
I can't come up with a simple failing test but found the problem while writing up an example.
The example is coming in a separate documentation PR and can serve as a test.

### DIFF
--- a/astropy/modeling/fitting.py
+++ b/astropy/modeling/fitting.py
@@ -1336,6 +1336,9 @@ def _fitter_to_model_params(model, fps):
         parameters[slice_] = values
         offset += size
 
+    # Update model parameters before calling ``tied`` constraints.
+    model._array_to_parameters()
+
     # This has to be done in a separate loop due to how tied parameters are
     # currently evaluated (the fitted parameters need to actually be *set* on
     # the model first, for use in evaluating the "tied" expression--it might be
@@ -1345,8 +1348,10 @@ def _fitter_to_model_params(model, fps):
             if model.tied[name]:
                 value = model.tied[name](model)
                 slice_ = param_metrics[name]['slice']
-                parameters[slice_] = value
-    model._array_to_parameters()
+                model.parameters[slice_] = value
+                # To handle multiple tied constraints, model parameters
+                # need to be updated after each iterration.
+                model._array_to_parameters()
 
 
 def _model_to_fit_params(model):


astropy/modeling/fitting.py
```diff
@@ -1345,7 +1345,7 @@ def _fitter_to_model_params(model, fps):
             if model.tied[name]:
                 value = model.tied[name](model)
                 slice_ = param_metrics[name]['slice']
-                parameters[slice_] = value
+                model.parameters[slice_] = value
```
perrygreenfield: Just for clarification. I'm puzzling over why these two versions result in different behavior. Normally a slice is a view (model.parameters returns a slice), so assignment to a slice of a slice should be a view (for a 1d array with contiguous slices). The only case that seems to result in different behavior is if _slice is relative to model._parameters array, not the slice of it which is returned. Is that what is breaking here?

nden:
`model.parameters` and `parameters=model.parameters` are different objects.
```
g = Gaussian1D()
pars = g.parameters
pars                                                                                                                           
Out: array([1., 0., 1.])

pars[0]=100                                                                                                                    
pars                                                                                                                           
Out: array([100.,   0.,   1.])

g.parameters                                                                                                                   
Out: array([1., 0., 1.])
```

pllim:
Does L1336 also need the same patch then?

nden:
@pllim Good question. Changing it breaks tests. I think the issue is present with `tied` constraints only because the way they are set up now they require the model to be evaluated to compute the value while the rest of the constraints don't need the model (and the parameter array can be updated independently before passing it to the fitter for the next iteration.

perrygreenfield:
@nden The example you give only behaves that way because g._array_to_parameters() wasn't called. (Forcing the array to be back propagated to the parameter instances., try doing that and you will see that g.parameters does show the new value) But that call is in the fitting segment right after the the slice assignments. So I'm still puzzled.

nden:
I think I understand now why `_array_to_parameters` did not work the first time I tried it.
`tied` constraints need the updated model. So `_array_to_parameters` needs to be called just before dealing with `tied` constraints to make sure any parameter updates due to other type of constraints are propagated to the model. For the same reasons it needs to be called after each iteration to deal with multiple `tied` constraints. Does this make sense now? 
I have not looked into why using `model.parameters` worked better.

nden:
@perrygreenfield Please look over the code again, I updated it to use `_array_to_parameters`.

Issue #3320, opened 2015-01-21
hdf5 table writer append functionality is absent from implementation


aphearin:
When trying to append to an existing hdf5 file using table.write(), the method returns an error:
IOError: Table table already exists
So I tried additionally including a flag for overwrite=True, but this only overwrites the entire existing file, rather than append to it.

astrofrog:
@aphearin - can you show the exact call to ``Table.write`` you are using? You should be using the ``append=True`` keyword rather than ``overwrite=True``. Did you try this?

aphearin:
Yes, I have been using append=True explicitly, although I thought the docs indicated that this is the default setting. Here is a code distillation of the problem.

Npts=1e6
x = np.random.random(Npts)
tx = Table({'x':x})
y = np.random.random(Npts)
ty = Table({'y':y})
tx.write('dummydata.hdf5',path='table',append=True)
ty.write('dummydata.hdf5',path='table',append=True)

mhvk:
@aphearin - the problem is that the `path` for the two tables should be different -- this is essentially a name of the file in the hdf5 data container.

Would you have a suggestion for improving the documentation so that this would be clear?

aphearin:
Sure, @mhvk , I'll submit a pull request later this morning

aphearin:
Actually, @mhvk , I'm a little confused about the functionality. What I want is to be able to load the *entire* table into memory with a call to Table.read(), not just one of the subsets of data that I appended. If I have to specify different paths, then astropy will not read in the concatenated data. How would I accomplish this using astropy?

mhvk:
The problem would be that the two tables might well have different lengths, or be incompatible in other ways, so they cannot necessary be assigned to a single table upon reading. I think the only solution would be to merge the tables (hstack or vstack) before writing to the hdf file. (But perhaps I am misunderstanding what you would like to do.)

aphearin:
If astropy intends to provide tools to efficiently handle management of very large datasets, I think this is important functionality. Imagine you have a many Gb ASCII data that you want to reduce and store in binary. Well then the natural way to accomplish this would be to read in chunks of ASCII, turn the chunk into a Table, make your cuts, and then append to an hdf5 file storing the reduced data. This way, at no point does the user ever need to load the entire table of reduced data into memory. Is this behavior not possible with the current code?

astrofrog:
[...]

PR #9602, opened 2019-11-12
table (hdf5): pass additional kwargs through to create_dataset() in write_table_hdf5

adrn:
While thinking about #3320, and looking at the HDF5 table writer, I realized that there is no way to use `Table.write()` to pass through additional arguments to `h5py.File.create_dataset()` (some of its kwargs are added as kwargs to `write_table_hdf5()` itself, like `compression`, but there are many more optional kwargs in `h5py.File.create_dataset()`). This PR adds support for passing through additional kwargs, like `maxshape` or `chunks` (both of which are relevant for the discussion in #3320). 

Needs a changelog if this is something we would like to add...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,6 +78,9 @@ astropy.table
 - Added new ``Table`` method ``.round()``, which rounds numeric columns to the
   specified number of decimals. [#9862]
 
+- The HDF5 writer, ``write_table_hdf5()``, now allows passing through
+  additional keyword arguments to the ``h5py.Group.create_dataset()``. [#9602]
+
 astropy.tests
 ^^^^^^^^^^^^^
 
--- a/astropy/io/misc/hdf5.py
+++ b/astropy/io/misc/hdf5.py
@@ -253,7 +253,8 @@ def write_table_hdf5(table, output, path=None, compression=False,
         If ``append=True`` and ``overwrite=True`` then only the dataset will be
         replaced; the file/group will not be overwritten.
     **create_dataset_kwargs
-        Additional keyword arguments are passed to `h5py.File.create_dataset`.
+        Additional keyword arguments are passed to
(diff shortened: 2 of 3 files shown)


dhomeier:
Appears like the doc build might need some kind of reference to the intersphinx mapping for h5py on `h5py.File.create_dataset`.
Looks good otherwise; tested here on Python 3.7 + 3.8.

One suggestion for cleaning up the `create_dataset` call further for the work on #3320 is to always use the call with `compression` argument:
```
    if compression:
        if compression is True:
            compression = 'gzip'
    else:
        compression = None

    # Check whether table already exists
    if name in output_group:
        if append and overwrite:
            # Delete only the dataset itself
            del output_group[name]
        else:
            raise OSError(f"Table {path} already exists")

    # Write the table to the file
    dset = output_group.create_dataset(name, data=table.as_array(),
                                           compression=compression,
                                           **create_dataset_kwargs)
```
– actually even the `compression is True` conditional is not needed, as it translates to `gzip` in `h5py` as well; but maybe it's better to keep this explicit in case the `h5py` behaviour changes in the future.
That way you might more easily replace the `OSError` case (or add an `elif extend:`) with a call for appending the table to the existing `output_group[name]`.

taldcroft:
@adrn - just looking through open table PR's and saw this.  This seems like a fine idea and not too controversial -- I will approve once passing tests.  As @dhomeier mentioned it looks like just need to double-backtick the `create_dataset` reference below, and add a changelog entry.

```
/home/circleci/project/build/lib.linux-x86_64-3.6/astropy/io/misc/hdf5.py:docstring of astropy.io.misc.hdf5.write_table_hdf5:36:py:obj reference target not found: h5py.File.create_dataset
Sphinx Documentation subprocess failed with return code 2
```

adrn:
@taldcroft thanks for the reminder! I added a changelog entry and fixed the backtick issue

PR #9621, opened 2019-11-18
Avoid warnings in documentation build

astrofrog:
This fixes a few residual warnings in the docs - these weren't being caught before because of https://github.com/matplotlib/matplotlib/issues/15715

### DIFF
--- a/astropy/modeling/blackbody.py
+++ b/astropy/modeling/blackbody.py
@@ -67,6 +67,10 @@ class BlackBody1D(Fittable1DModel):
     .. plot::
         :include-source:
 
+        # Ignore warnings about deprecation of BlackBody1D
+        import warnings
+        warnings.simplefilter('ignore')
+
         import numpy as np
         import matplotlib.pyplot as plt
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -225,6 +225,12 @@ try:
         'abort_on_example_error': True
     }
 
+    # Filter out backend-related warnings as described in
+    # https://github.com/sphinx-gallery/sphinx-gallery/pull/564
+    warnings.filterwarnings("ignore", category=UserWarning,
+                            message='Matplotlib is currently using agg, which is a'
+                                    ' non-GUI backend, so cannot show the figure.')
+
 except ImportError:
     def setup(app):
         msg = ('The sphinx_gallery extension is not installed, so the '
--- a/astropy/modeling/physical_models.py
+++ b/astropy/modeling/physical_models.py
@@ -237,7 +237,6 @@ class Drude1D(Fittable1DModel):
         ax.set_xlabel('x')
         ax.set_ylabel('F(x)')
 
-        ax.legend(loc='best')
         plt.show()
     """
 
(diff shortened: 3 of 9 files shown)


astrofrog:
Ah wait there are some more that appear in CI here:

```
/home/circleci/project/venv/lib/python3.7/site-packages/numpy/lib/nanfunctions.py:1115: RuntimeWarning: All-NaN slice encountered
  overwrite_input=overwrite_input)
WARNING: Input data contains invalid values (NaNs or infs), which were automatically clipped. [astropy.stats.sigma_clipping]
WARNING: Input data contains invalid values (NaNs or infs), which were automatically clipped. [astropy.stats.sigma_clipping]
/home/circleci/project/build/lib.linux-x86_64-3.7/astropy/timeseries/downsample.py:28: RuntimeWarning: Mean of empty slice
  result.append(function(array[indices[i]:indices[i+1]]))
/home/circleci/project/build/lib.linux-x86_64-3.7/astropy/timeseries/downsample.py:29: RuntimeWarning: Mean of empty slice
  result.append(function(array[indices[-1]:]))
WARNING: ErfaWarning: ERFA function "utctai" yielded 1 of "dubious year (Note 3)" [astropy._erfa.core]
No handles with labels found to put in legend.
<string>:1: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to each axes instance.
```

astrofrog:
@pllim - ok that should be all of them now I think! We should check the html-docs build once it completes.

astropy/modeling/blackbody.py
```diff
@@ -67,6 +67,10 @@ class BlackBody1D(Fittable1DModel):
     .. plot::
         :include-source:
 
+        # Ignore warnings about deprecation of BlackBody1D
+        import warnings
+        warnings.simplefilter('ignore')
```
mhvk: Can we not fix the underlying cause? (The imporf of `FLAM`.)

pllim: I wouldn't spend too much time on this module, as it is deprecated and will be removed in the future.

docs/conf.py
```diff
@@ -225,6 +225,12 @@
         'abort_on_example_error': True
     }
 
+    # Filter out backend-related warnings as described in
+    # https://github.com/sphinx-gallery/sphinx-gallery/pull/564
+    warnings.filterwarnings("ignore", category=UserWarning,
+                            message='Matplotlib is currently using agg, which is a'
```
pllim: I am just wondering if we should relax the regex to match all the backends it might complain about, not just `agg`.

pllim:
As far as I can see, the build log looks pretty clean! 🎉 https://circleci.com/gh/astropy/astropy/53985

astropy/modeling/blackbody.py
mhvk: Ah, sorry, I didn't see this was also in `blackbody.py` - yes, definitely no problem!

docs/conf.py
astrofrog: This is the recommended warning catching in the sphinx-gallery docs, so I just thought I'd go with that.

PR #9632, opened 2019-11-19
Fix HighLevelWCSWrapper to take into account return type of pixel_to_world_values

Cadair:
I managed to construct a situation where we were putting a len 1 tuple inside a len 1 tuple, which then proceeded to break gWCS.

(affects dev bugfix I think)

tests etc tomorrow.

### DIFF
--- a/astropy/wcs/wcsapi/high_level_api.py
+++ b/astropy/wcs/wcsapi/high_level_api.py
@@ -3,6 +3,7 @@ from collections import defaultdict, OrderedDict
 
 import numpy as np
 
+from astropy.utils import isiterable
 from .utils import deserialize_class
 
 __all__ = ['BaseHighLevelWCS', 'HighLevelWCSMixin']
@@ -214,7 +215,7 @@ class HighLevelWCSMixin(BaseHighLevelWCS):
         # Compute the world coordinate values
         world = self.low_level_wcs.pixel_to_world_values(*pixel_arrays)
 
-        if self.world_n_dim == 1:
+        if self.world_n_dim == 1 and not (isiterable(world) and len(world) == 1):
             world = (world,)
 
         # Cache the classes and components since this may be expensive


pllim:
@Cadair , will you be able to wrap this up soon? If not, I propose to push the milestone to 4.0.1.

Cadair:
I actually wonder if this is rectified by #9677 and https://github.com/spacetelescope/gwcs/pull/261 I will try and investigate.

nden:
The way I am reading the [spec](https://github.com/astropy/astropy/blob/6005d3a7c426df499a5ded08f68f8b9a927e7960/astropy/wcs/wcsapi/high_level_api.py#L47), passing a float should return a float, passing an array of floats to a 1D transform should return an array. Am I misunderstanding it?

Assuming no misunderstanding, currently `astropy.wcs.pixel_to_world_values` returns an array when passed a float.

```
aw = awcs.WCS(naxis=1)
aw.pixel_to_world_values(1)                                                                                                                                     
Out[44]: array(2.)

aw.pixel_to_world(1)                                                                                                                                            
Out[45]: <Quantity 2.>
```

The equivalent gwcs object returns
```
w.pixel_to_world_values(1)                                                                                                                                      
Out[47]: 10.0

w.pixel_to_world(1)                                                                                                                                             
Out[48]: <Quantity 10. Hz>

```
This is related also to what the output of other methods should be, like `pixel_bounds`.
Both astropy and gwcs are self-consistent but they return different types. I think the spec linked above needs to be clarified.

nden:
And the definition for the low_level_api is [here](https://github.com/astropy/astropy/blob/6005d3a7c426df499a5ded08f68f8b9a927e7960/astropy/wcs/wcsapi/low_level_api.py#L62). In any case a failing example will help to understand the issue.

Cadair:
@astrofrog and I were discussing this in the context of #9677 

The reason for this is that we use some `np.broadcast_to` trickery in the FITS implementation of APE14 which leads to numpy converting things to scalar arrays. If we think this is worth correcting (I don't really have an opinion on if it matters) then we should be able to special case it on the exit of the relevant methods in the APE14 implementation.

pllim:
Is this resolved during the Astropy Coordination Meeting 2019? What is the path forward here?

Cadair:
There are still discrepancies in the return types of low level WCSes, which I am working on in #9677 but this was actually an issue with gWCS.

Issue #7883, opened 2018-10-11
wcs _TAB option: example please


knordsieck:
I have been unable to get the _TAB transformation class ("table lookup distortions") to work.  I need it because of quite nonlinear spectral distortions I need to manage in objective prism images.  Could someone point me to a successful use of this class?  An actual fits file with the correct header keywords and table extension would be very valuable.

I am currently using astropy 2.0.2.  I s this a problem?

Thanks

pllim:
Hello. Do you not have a FITS file already? If not, what format is your data in? What instrument did you use to get the data? Usually, an instrument comes with its own calibration software.

Also, unless you are stuck with Python 2, we recommend using the latest release of 3.x series.

knordsieck:
I’m in the process of commissioning a new mode on the Robert Stobie Spectrograph of the Southern African Large Telescope (I did the spectrograph optical design).  Right now I am writing the calibration/ analysis software for this mode.  I need to define a fits file data product that deals with the non-linear prism dispersion in this mode, and it seems that the wcs -TAB option is most appropriate.  So I am trying to put data that has already been processed from raw detector fits files into reduced fits files that incorporate wavelength calibration of the spectra.  All my attempts to do this result in garbage.

I thought I could find dumb mistakes or version problems by comparing with a –TAB that does work, hence the title. Failing this, I show a failed example here.  Below is the info() and header for a test fits file I made.

The current SALT pipeline uses python 2.7.3.  I would definitely like to avoid updating the whole pipeline!

>>> hdul.info()
Filename: testfits.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     167   ()
  1  SCI           1 ImageHDU        24   (1601, 514, 2)   float32
  2  VAR           1 ImageHDU         9   (1601, 514, 2)   float32
  3  BPM           1 ImageHDU         9   (1601, 514, 2)   uint8
  4  TMAP          1 ImageHDU         9   (1601, 514, 2)   uint8
  5  TGT           1 BinTableHDU     43   108R x 16C   [19A, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D]
  6  WCS-table     1 BinTableHDU     13   12R x 2C   [K, D]

>>> print repr(hdul['SCI'].header)
XTENSION= 'IMAGE   '           / Image extension
BITPIX  =                  -32 / array data type
NAXIS   =                    3 / number of array dimensions
[...]

Issue #5462, opened 2016-11-04
Lookup-table WCS for irregular spectral sampling doesn't look up


josePhoenix:
Bearing in mind this may be a) not implemented or b) user error, I seem to be having difficulties expressing the `-TAB` coordinate type. I have a (4096, 4096, 19) (in FITS axes order) datacube with 19 planes representing simulated observations through different filters.

I want to assign a bogus WCS so that I can visualize angular dimensions on each plane in ds9 or similar. It occurred to me that the third dimension actually is wavelength, and I *think* I can represent it in FITS WCS using a lookup table. However, when I try to do `mywcs.all_pix2world` I get unexpected behavior that might even be uninitialized memory (i.e. calling the same function twice gives two different results for the wavelength axis).

I call `mywcs.all_pix2world([[2048, 2048, i],], 0)` where `i` is in [0, 19). The first two dimensions come out as I expect (that is, `0, 0`) but the last is usually 0 and sometimes a very small float value (like 1e-322):

```python
>>> mywcs.all_pix2world([[2048, 2048, 2],], 0)
array([[  5.55555555e-005,   5.55555556e-005,   1.58101007e-322]])
```

Here's my WCS header:

```
WCSAXES =                    3 / Number of coordinate axes                      
CRPIX1  =               2048.0 / Pixel coordinate of reference point            
CRPIX2  =               2048.0 / Pixel coordinate of reference point            
CRPIX3  =                  0.0 / Pixel coordinate of reference point            
CDELT1  =  5.5555555555556E-05 / [deg] Coordinate increment at reference point  
CDELT2  =  5.5555555555556E-05 / [deg] Coordinate increment at reference point  
CDELT3  =                  1.0 / [m] Coordinate increment at reference point    
CUNIT1  = 'deg'                / Units of coordinate increment and value        
CUNIT2  = 'deg'                / Units of coordinate increment and value        
CUNIT3  = 'm'                  / Units of coordinate increment and value        
CTYPE1  = 'RA---TAN'           / Right ascension, gnomonic projection           
CTYPE2  = 'DEC--TAN'           / Declination, gnomonic projection               
CTYPE3  = 'WAVE-TAB'           / Vacuum wavelength                              
CRVAL1  =                  0.0 / [deg] Coordinate value at reference point      
CRVAL2  =                  0.0 / [deg] Coordinate value at reference point      
CRVAL3  =                  0.0 / [m] Coordinate value at reference point        
[...]

Issue #2362, opened 2014-04-23
WCS with lookuptables (tabular)


keflavich:
It appears that lookup table WCS cannot be created with `astropy.wcs` at present.  

http://astropy.readthedocs.org/en/latest/api/astropy.wcs.Tabprm.html#astropy.wcs.Tabprm says a `Tabprm` table cannot be directly instantiated.

This is a feature request or a request for documentation if the functionality already exists.  I'll help with implementation, but I need some direction.  Does anything need to be modified in the C wrapper?  Or is it just the python frontend that needs to be made?  

Does `astropy.wcs` presently know how to write a tabular WCS, or do we need to implement that too?

cdeil:
In Gammapy having a way to construct a lookup table WCS from Python would also be very useful.

Currently we have to handle our spectral log energy axis ourselves, without using `astropy.wcs` ... see [SpectralCube](https://gammapy.readthedocs.org/en/latest/api/gammapy.data.SpectralCube.html) and [LogEnergyAxis](https://gammapy.readthedocs.org/en/latest/api/gammapy.spectrum.LogEnergyAxis.html).

parejkoj:
This is likely related to #7883 and I think is still an open issue? @nden ?

mcara:
In PR #9641 I implemented support for the `-TAB` algorithm in FITS WCS defined in WCS Paper III.

mcara:
To those (still) interested: https://github.com/astropy/astropy/pull/9641 has been merged and `WCS-TAB` should now be supported in `astropy.wcs`. Please give it a try.

pllim:
Closing as solved. Thanks, @mcara !

PR #9641, opened 2019-11-21
Implement support for the -TAB algorithm in WCS

mcara:
This pull request is to implement support for the `-TAB` algorithm in FITS WCS as described in [WCS Paper III](https://www.atnf.csiro.au/people/mcalabre/WCS/scs.pdf).

This PR partially addresses #2362, #5462, #7883 in the sense that `astropy.wcs` now will be able to support `-TAB` in WCS if already defined in a HDUList object. However this PR does not provide "write" support for `-TAB`: the files must be created using external scripts.

### DIFF
--- a/astropy/wcs/include/astropy_wcs/wcslib_wrap.h
+++ b/astropy/wcs/include/astropy_wcs/wcslib_wrap.h
@@ -1,6 +1,5 @@
 /*
  Author: Michael Droettboom
-         mdroe@stsci.edu
 */
 
 #ifndef __WCSLIB_WRAP_H__
@@ -22,5 +21,9 @@ PyWcsprm_find_all_wcs(
     PyObject* self,
     PyObject* args,
     PyObject* kwds);
+
+int _update_wtbarr_from_hdulist(PyObject *hdulist, struct wtbarr *wtb);
+
+void _set_wtbarr_callback(PyObject* callback);
 
 #endif
--- a/astropy/wcs/src/astropy_wcs.c
+++ b/astropy/wcs/src/astropy_wcs.c
@@ -19,6 +19,9 @@
 #include <stdlib.h>
 #include <time.h>
 
+#include <tab.h>
+#include <wtbarr.h>
+
 /***************************************************************************
  * Wcs type
  ***************************************************************************/
@@ -26,6 +29,22 @@
 static PyTypeObject WcsType;
 
 static int _setup_wcs_type(PyObject* m);
+
+
+PyObject* PyWcsprm_set_wtbarr_fitsio_callback(PyObject *dummy, PyObject *args) {
+    PyObject *callback;
+
+    if (PyArg_ParseTuple(args, "O:set_wtbarr_fitsio_callback", &callback)) {
+        if (!PyCallable_Check(callback)) {
+            PyErr_SetString(PyExc_TypeError, "parameter must be callable");
+            return NULL;
+        }
+        _set_wtbarr_callback(callback);
+
+        Py_RETURN_NONE;
+    }
+    return NULL;
+}
 
 
 /***************************************************************************
@@ -724,6 +743,7 @@ static PyMethodDef Wcs_methods[] = {
 static PyMethodDef module_methods[] = {
   {"_sanity_check", (PyCFunction)_sanity_check, METH_NOARGS, ""},
   {"find_all_wcs", (PyCFunction)PyWcsprm_find_all_wcs, METH_VARARGS|METH_KEYWORDS, doc_find_all_wcs},
+  {"set_wtbarr_fitsio_callback", (PyCFunction)PyWcsprm_set_wtbarr_fitsio_callback, METH_VARARGS, NULL},
   {NULL}  /* Sentinel */
 };
 
@@ -837,7 +857,7 @@ PyInit__wcs(void)
       _setup_unit_list_proxy_type(m)||
       _setup_wcsprm_type(m)         ||
       _setup_tabprm_type(m)         ||
-      /* _setup_wtbarr_type(m)         || */
+      _setup_wtbarr_type(m)         ||
       _setup_distortion_type(m)     ||
       _setup_sip_type(m)            ||
       _setup_wcs_type(m)          ||
--- a/astropy/wcs/src/wcslib_tabprm_wrap.c
+++ b/astropy/wcs/src/wcslib_tabprm_wrap.c
@@ -94,6 +94,7 @@ PyTabprm*
 PyTabprm_cnew(PyObject* wcsprm, struct tabprm* x) {
   PyTabprm* self;
   self = (PyTabprm*)(&PyTabprmType)->tp_alloc(&PyTabprmType, 0);
+  if (self == NULL) return NULL;
   self->x = x;
   Py_INCREF(wcsprm);
   self->owner = wcsprm;
@@ -124,8 +125,7 @@ PyTabprm_set(
     return NULL;
   }
 
-  Py_INCREF(Py_None);
-  return Py_None;
+  Py_RETURN_NONE;
 }
 
 /*@null@*/ static PyObject*
@@ -139,13 +139,10 @@ PyTabprm_print_contents(
   /* This is not thread-safe, but since we're holding onto the GIL,
      we can assume we won't have thread conflicts */
   wcsprintf_set(NULL);
-
   tabprt(self->x);
-
   printf("%s", wcsprintf_buf());
-
-  Py_INCREF(Py_None);
-  return Py_None;
+  fflush(stdout);
+  Py_RETURN_NONE;
 }
 
 /*@null@*/ static PyObject*
@@ -413,7 +410,7 @@ PyTypeObject PyTabprmType = {
   0,                            /*tp_getattr*/
   0,                            /*tp_setattr*/
   0,                            /*tp_compare*/
-  (reprfunc)PyTabprm___str__,   /*tp_repr*/
+  0,                            /*tp_repr*/
   0,                            /*tp_as_number*/
   0,                            /*tp_as_sequence*/
   0,                            /*tp_as_mapping*/
--- a/astropy/wcs/src/wcslib_wrap.c
+++ b/astropy/wcs/src/wcslib_wrap.c
@@ -19,6 +19,8 @@
 #include <wcsprintf.h>
 #include <wcsunits.h>
 #include <tab.h>
+#include <wtbarr.h>
+#include <stdio.h>
 
 #include "astropy_wcs/isnan.h"
 #include "astropy_wcs/distortion.h"
@@ -157,6 +159,133 @@ convert_rejections_to_warnings() {
   return status;
 }
 
+
+/***************************************************************************
+ * wtbarr-related global variables and functions                           *
+ ***************************************************************************/
+
+static PyObject *get_wtbarr_data = NULL;
+
+
+void _set_wtbarr_callback(PyObject* callback) {
+  Py_XINCREF(callback);         /* Add a reference to new callback */
+  Py_XDECREF(get_wtbarr_data);  /* Dispose of previous callback */
+  get_wtbarr_data = callback;   /* Remember new callback */
+}
+
+
+int _update_wtbarr_from_hdulist(PyObject *hdulist, struct wtbarr *wtb) {
+  PyArrayObject *arrayp=NULL;
+  PyObject *result=NULL;
+  int i, naxis, nelem, naxes[NPY_MAXDIMS];
+  npy_intp *npy_naxes;
+  npy_double *appayp_data;
+
+  if (hdulist == NULL || hdulist == Py_None) {
+    PyErr_SetString(PyExc_ValueError,
+                    "HDUList is required to retrieve -TAB coordinates "
+                    "and/or indices.");
+    return 0;
+  }
+
+  if (wtb->ndim < 1) {
+    PyErr_SetString(PyExc_ValueError, "Number of dimensions should be positive.");
+    return 0;
+  }
+
+  result = PyObject_CallFunction(get_wtbarr_data, "(OsiiCsli)", hdulist,
+      wtb->extnam, wtb->extver, wtb->extlev, wtb->kind, wtb->ttype, wtb->row,
+      wtb->ndim);
+
+  if (result == NULL) return 0;
+
+  arrayp = (PyArrayObject *)PyArray_FromAny(result,
+      PyArray_DescrFromType(NPY_DOUBLE), 0, 0, NPY_ARRAY_CARRAY, NULL);
+
+  Py_DECREF(result);
+
+  if (arrayp == NULL) {
+    PyErr_SetString(PyExc_TypeError, "Unable to convert wtbarr callback "
+                    "result to a numpy.ndarray.");
+    return 0;
+  }
+
+  if (!PyArray_Check(arrayp)) {
+    PyErr_SetString(PyExc_TypeError,
+                    "wtbarr callback must return a numpy.ndarray type "
+                    "coordinate or index array.");
+    Py_DECREF(arrayp);
+    return 0;
+  }
+
+  naxis = PyArray_NDIM(arrayp);
+
+  if (naxis == 0) {
+    PyErr_SetString(PyExc_ValueError, "-TAB coordinate or index arrays "
+                    "cannot be 0-dimensional.");
+    Py_DECREF(arrayp);
+    return 0;
+  }
+
+  npy_naxes = PyArray_DIMS(arrayp);
+  for (i = 0; i < naxis; i++) {
+    naxes[i] = (int) npy_naxes[i];
+  }
+
+  if (naxis != wtb->ndim) {
+    if (wtb->kind == 'c' && wtb->ndim == 2 && naxis == 1) {
+      /* Allow TDIMn to be omitted for degenerate coordinate arrays. */
+      naxis = 2;
+      naxes[1] = 1;
+    } else {
+      PyErr_Format(PyExc_ValueError,
+          "An array with an unexpected number of axes was "
+          "received from the callback. Expected %d but got %d.",
+          wtb->ndim, (int) naxis);
+      Py_DECREF(arrayp);
+      return 0;
+    }
+  }
+
+  if (wtb->kind == 'c') {
+    /* Coordinate array; calculate the array size. */
+    nelem = naxes[naxis-1];
+    for (i = 0; i < naxis-1; i++) {
+      *(wtb->dimlen + i) = naxes[naxis-2-i];
+      nelem *= naxes[i];
+    }
+  } else {
+    /* Index vector; check length. */
+    if ((nelem = naxes[naxis-1]) != *(wtb->dimlen)) {
+      /* N.B. coordinate array precedes the index vectors. */
+      PyErr_Format(PyExc_ValueError,
+          "An index array with an unexpected number of dimensions was "
+          "received from the callback. Expected %d but got %d.",
+          *(wtb->dimlen), (int) nelem);
+      Py_DECREF(arrayp);
+      return 0;
+    }
+  }
+
+  /* Allocate memory for the array. */
+  if (!((*wtb->arrayp) = calloc((size_t)nelem, sizeof(double)))) {
+    PyErr_SetString(PyExc_MemoryError, "Out of memory: can't allocate "
+                    "coordinate or index array.");
+    Py_DECREF(arrayp);
+    return 0;
+  }
+
+  /* Read the array from the table. */
+  appayp_data = (npy_double*)PyArray_DATA(arrayp);
+  for (i = 0; i < nelem; i++) {
+    (*wtb->arrayp)[i] = (double)appayp_data[i];
+  }
+
+  Py_DECREF(arrayp);
+  return 1;
+}
+
+
 /***************************************************************************
  * PyWcsprm methods
  */
@@ -203,6 +332,7 @@ PyWcsprm_init(
 
   int            status;
   PyObject*      header_obj    = NULL;
+  PyObject*      hdulist       = NULL;
   char *         header        = NULL;
   Py_ssize_t     header_length = 0;
   Py_ssize_t     nkeyrec       = 0;
@@ -219,14 +349,14 @@ PyWcsprm_init(
   int            nreject       = 0;
   int            nwcs          = 0;
   struct wcsprm* wcs           = NULL;
-  int            i             = 0;
+  int            i, j;
   const char*    keywords[]    = {"header", "key", "relax", "naxis", "keysel",
-                                  "colsel", "warnings", NULL};
+                                  "colsel", "warnings", "hdulist", NULL};
 
   if (!PyArg_ParseTupleAndKeywords(
-          args, kwds, "|OsOiiOi:WCSBase.__init__",
+          args, kwds, "|OsOiiOiO:WCSBase.__init__",
           (char **)keywords, &header_obj, &key, &relax_obj, &naxis, &keysel,
-          &colsel, &warnings)) {
+          &colsel, &warnings, &hdulist)) {
     return -1;
   }
 
@@ -257,7 +387,6 @@ PyWcsprm_init(
       return -1;
     }
 
-    note_change(self);
     self->x.flag = -1;
     status = wcsini(1, naxis, &self->x);
 
@@ -442,6 +571,16 @@ PyWcsprm_init(
       return -1;
     }
 
+    if (self->x.ntab) {
+      wcstab(&self->x);
+      for (j = 0; j < self->x.nwtb; j++) {
+        if (!_update_wtbarr_from_hdulist(hdulist, &(self->x.wtb[j]))) {
+          wcsfree(&self->x);
+          return -1;
+        }
+      }
+    }
+
     note_change(self);
     wcsprm_c2python(&self->x);
     wcsvfree(&nwcs, &wcs);
@@ -485,8 +624,9 @@ PyWcsprm_bounds_check(
 PyWcsprm_copy(
     PyWcsprm* self) {
 
-  PyWcsprm*      copy      = NULL;
-  int            status;
+  PyWcsprm*     copy = NULL;
+  int           status, nelem, i, j, ndim;
+  struct wtbarr *wtb, *wtb0;
 
   copy = PyWcsprm_cnew();
   if (copy == NULL) {
@@ -504,6 +644,39 @@ PyWcsprm_copy(
       Py_XDECREF(copy);
       return NULL;
     }
+
+    if (self->x.ntab) {
+      wcstab(&copy->x);
+
+      for (j = 0; j < copy->x.nwtb; j++) {
+        wtb0 = self->x.wtb + j;
+        wtb = copy->x.wtb + j;
+        for (i = 0; i < wtb0->ndim - 1; i++) {
+          wtb->dimlen[i] = wtb0->dimlen[i];
+        }
+        /* Allocate memory for the array. */
+        if (wtb->kind == 'c') {
+          nelem = ndim = wtb->ndim - 1;
+          for (i = 0; i < ndim; i++) {
+            nelem *= wtb->dimlen[i];
+          }
+        } else {
+          nelem = *(wtb->dimlen);
+        }
+
+        if (!((*wtb->arrayp) = calloc((size_t)nelem, sizeof(double)))) {
+          PyErr_SetString(PyExc_MemoryError, "Out of memory: can't allocate "
+                                             "coordinate or index array.");
+          Py_DECREF(copy);
+          return NULL;
+        }
+
+        for (i = 0; i < nelem; i++) {
+          (*wtb->arrayp)[i] = (*wtb0->arrayp)[i];
+        }
+      }
+    }
+
     wcsprm_c2python(&copy->x);
     return (PyObject*)copy;
   } else {
@@ -3773,38 +3946,30 @@ PyWcsprm_set_velref(
   return set_int("velref", value, &self->x.velref);
 }
 
-/* static PyObject* */
-/* PyWcsprm_get_wtb( */
-/*     PyWcsprm* self, */
-/*     /\*@unused@*\/ void* closure) { */
 
-/*   PyObject* result; */
-/*   PyObject* subresult; */
-/*   int i, nwtb; */
+static PyObject* PyWcsprm_get_wtb(PyWcsprm* self, void* closure) {
+  PyObject* list;
+  PyObject* elem;
+  int i, nwtb;
 
-/*   nwtb = self->x.nwtb; */
+  nwtb = self->x.nwtb;
 
-/*   result = PyList_New(nwtb); */
-/*   if (result == NULL) { */
-/*     return NULL; */
-/*   } */
+  list = PyList_New(nwtb);
+  if (list == NULL) return NULL;
 
-/*   for (i = 0; i < nwtb; ++i) { */
-/*     subresult = (PyObject *)PyWtbarr_cnew((PyObject *)self, &(self->x.wtb[i])); */
-/*     if (subresult == NULL) { */
-/*       Py_DECREF(result); */
-/*       return NULL; */
-/*     } */
+  for (i = 0; i < nwtb; ++i) {
+    elem = (PyObject *)PyWtbarr_cnew((PyObject *)self, &(self->x.wtb[i]));
+    if (elem == NULL) {
+      Py_DECREF(list);
+      return NULL;
+    }
 
-/*     if (PyList_SetItem(result, i, subresult) == -1) { */
-/*       Py_DECREF(subresult); */
-/*       Py_DECREF(result); */
-/*       return NULL; */
-/*     } */
-/*   } */
+    PyList_SET_ITEM(list, i, elem);
+  }
 
-/*   return result; */
-/* } */
+  return list;
+}
+
 
 static PyObject*
 PyWcsprm_get_zsource(
@@ -3904,7 +4069,7 @@ static PyGetSetDef PyWcsprm_getset[] = {
   {"velosys", (getter)PyWcsprm_get_velosys, (setter)PyWcsprm_set_velosys, (char *)doc_velosys},
   {"velref", (getter)PyWcsprm_get_velref, (setter)PyWcsprm_set_velref, (char *)doc_velref},
   {"xposure", (getter)PyWcsprm_get_xposure, (setter)PyWcsprm_set_xposure, (char *)doc_xposure},
-  /* {"wtb", (getter)PyWcsprm_get_wtb, NULL, (char *)doc_tab}, */
+  {"wtb", (getter)PyWcsprm_get_wtb, NULL, (char *) NULL},
   {"zsource", (getter)PyWcsprm_get_zsource, (setter)PyWcsprm_set_zsource, (char *)doc_zsource},
   {NULL}
 };
--- a/astropy/wcs/src/wcslib_wtbarr_wrap.c
+++ b/astropy/wcs/src/wcslib_wtbarr_wrap.c
@@ -1,0 +1,239 @@
+#define NO_IMPORT_ARRAY
+
+#include "astropy_wcs/wcslib_wtbarr_wrap.h"
+
+#include <wcs.h>
+#include <wcsprintf.h>
+#include <tab.h>
+#include <wtbarr.h>
+
+/*
+ It gets to be really tedious to type long docstrings in ANSI C syntax
+ (since multi-line strings literals are not valid).  Therefore, the
+ docstrings are written in doc/docstrings.py, which are then converted
+ by setup.py into docstrings.h, which we include here.
+*/
+#include "astropy_wcs/docstrings.h"
+
+
+/***************************************************************************
+ * PyWtbarr methods                                                        *
+ ***************************************************************************/
+
+static PyObject*
+PyWtbarr_new(PyTypeObject* type, PyObject* args, PyObject* kwds) {
+  PyWtbarr* self;
+  self = (PyWtbarr*)type->tp_alloc(type, 0);
+  return (PyObject*)self;
+}
+
+
+static int
+PyWtbarr_traverse(PyWtbarr* self, visitproc visit, void *arg) {
+  Py_VISIT(self->owner);
+  return 0;
+}
+
+
+static int
+PyWtbarr_clear(PyWtbarr* self) {
+  Py_CLEAR(self->owner);
+  return 0;
+}
+
+
+static void PyWtbarr_dealloc(PyWtbarr* self) {
+  PyWtbarr_clear(self);
+  Py_TYPE(self)->tp_free((PyObject*)self);
+}
+
+
+PyWtbarr* PyWtbarr_cnew(PyObject* wcsprm, struct wtbarr* x) {
+  PyWtbarr* self;
+  self = (PyWtbarr*)(&PyWtbarrType)->tp_alloc(&PyWtbarrType, 0);
+  if (self == NULL) return NULL;
+  self->x = x;
+  Py_INCREF(wcsprm);
+  self->owner = wcsprm;
+  return self;
+}
+
+
+static void wtbarrprt(const struct wtbarr *wtb) {
+  int i, nd, ndim;
+
+  if (wtb == 0x0) return;
+
+  wcsprintf("     i: %d\n", wtb->i);
+  wcsprintf("     m: %d\n", wtb->m);
+  wcsprintf("  kind: %c\n", wtb->kind);
+  wcsprintf("extnam: %s\n", wtb->extnam);
+  wcsprintf("extver: %d\n", wtb->extver);
+  wcsprintf("extlev: %d\n", wtb->extlev);
+  wcsprintf(" ttype: %s\n", wtb->ttype);
+  wcsprintf("   row: %ld\n", wtb->row);
+  wcsprintf("  ndim: %d\n", wtb->ndim);
+  wcsprintf("dimlen: %p\n", (void *)wtb->dimlen);
+
+  ndim = wtb->ndim - (int)(wtb->kind == 'c');
+  nd = 1 + (int) log10(ndim ? ndim : 1);
+  for (i = 0; i < ndim; i++) {
+    wcsprintf("        %*d:   %d\n", nd, i, wtb->dimlen[i]);
+  }
+  wcsprintf("arrayp: %p\n", (void *)wtb->arrayp);
+
+  return;
+}
+
+
+static PyObject* PyWtbarr_print_contents(PyWtbarr* self) {
+  /* This is not thread-safe, but since we're holding onto the GIL,
+     we can assume we won't have thread conflicts */
+  wcsprintf_set(NULL);
+  wtbarrprt(self->x);
+  printf("%s", wcsprintf_buf());
+  fflush(stdout);
+  Py_RETURN_NONE;
+}
+
+
+static PyObject* PyWtbarr___str__(PyWtbarr* self) {
+  /* This is not thread-safe, but since we're holding onto the GIL,
+     we can assume we won't have thread conflicts */
+  wcsprintf_set(NULL);
+  wtbarrprt(self->x);
+  return PyUnicode_FromString(wcsprintf_buf());
+}
+
+
+/***************************************************************************
+ * Member getters/setters (properties)
+ */
+
+
+static PyObject* PyWtbarr_get_i(PyWtbarr* self, void* closure) {
+  return get_int("i", self->x->i);
+}
+
+
+static PyObject* PyWtbarr_get_m(PyWtbarr* self, void* closure) {
+  return get_int("m", self->x->m);
+}
+
+
+static PyObject* PyWtbarr_get_extver(PyWtbarr* self, void* closure) {
+  return get_int("extver", self->x->extver);
+}
+
+
+static PyObject* PyWtbarr_get_extlev(PyWtbarr* self, void* closure) {
+  return get_int("extlev", self->x->extlev);
+}
+
+
+static PyObject* PyWtbarr_get_ndim(PyWtbarr* self, void* closure) {
+  return get_int("ndim", self->x->ndim);
+}
+
+
+static PyObject* PyWtbarr_get_row(PyWtbarr* self, void* closure) {
+  return get_int("row", self->x->row);
+}
+
+
+static PyObject* PyWtbarr_get_extnam(PyWtbarr* self, void* closure) {
+  if (is_null(self->x->extnam)) return NULL;
+  return get_string("extnam", self->x->extnam);
+}
+
+
+static PyObject* PyWtbarr_get_ttype(PyWtbarr* self, void* closure) {
+  if (is_null(self->x->ttype)) return NULL;
+  return get_string("ttype", self->x->ttype);
+}
+
+
+static PyObject* PyWtbarr_get_kind(PyWtbarr* self, void* closure) {
+  return PyUnicode_FromFormat("%c", self->x->kind);
+}
+
+
+/***************************************************************************
+ * PyWtbarr definition structures
+ */
+
+static PyGetSetDef PyWtbarr_getset[] = {
+  {"i", (getter)PyWtbarr_get_i, NULL, (char *) NULL},
+  {"m", (getter)PyWtbarr_get_m, NULL, (char *) NULL},
+  {"kind", (getter)PyWtbarr_get_kind, NULL, (char *) NULL},
+  {"extnam", (getter)PyWtbarr_get_extnam, NULL, (char *) NULL},
+  {"extver", (getter)PyWtbarr_get_extver, NULL, (char *) NULL},
+  {"extlev", (getter)PyWtbarr_get_extlev, NULL, (char *) NULL},
+  {"ttype", (getter)PyWtbarr_get_ttype, NULL, (char *) NULL},
+  {"row", (getter)PyWtbarr_get_row, NULL, (char *) NULL},
+  {"ndim", (getter)PyWtbarr_get_ndim, NULL, (char *) NULL},
+/*  {"dimlen", (getter)PyWtbarr_get_dimlen, NULL, (char *) NULL}, */
+/*  {"arrayp", (getter)PyWtbarr_get_arrayp, NULL, (char *) NULL}, */
+  {NULL}
+};
+
+
+static PyMethodDef PyWtbarr_methods[] = {
+  {"print_contents", (PyCFunction)PyWtbarr_print_contents, METH_NOARGS, NULL},
+  {NULL}
+};
+
+PyTypeObject PyWtbarrType = {
+  PyVarObject_HEAD_INIT(NULL, 0)
+  "astropy.wcs.Wtbarr",         /*tp_name*/
+  sizeof(PyWtbarr),             /*tp_basicsize*/
+  0,                            /*tp_itemsize*/
+  (destructor)PyWtbarr_dealloc, /*tp_dealloc*/
+  0,                            /*tp_print*/
+  0,                            /*tp_getattr*/
+  0,                            /*tp_setattr*/
+  0,                            /*tp_compare*/
+  0,                            /*tp_repr*/
+  0,                            /*tp_as_number*/
+  0,                            /*tp_as_sequence*/
+  0,                            /*tp_as_mapping*/
+  0,                            /*tp_hash */
+  0,                            /*tp_call*/
+  (reprfunc)PyWtbarr___str__,   /*tp_str*/
+  0,                            /*tp_getattro*/
+  0,                            /*tp_setattro*/
+  0,                            /*tp_as_buffer*/
+  Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE, /*tp_flags*/
+  0,                            /* tp_doc */
+  0,                            /* tp_traverse */
+  0,                            /* tp_clear */
+  0,                            /* tp_richcompare */
+  0,                            /* tp_weaklistoffset */
+  0,                            /* tp_iter */
+  0,                            /* tp_iternext */
+  PyWtbarr_methods,             /* tp_methods */
+  0,                            /* tp_members */
+  PyWtbarr_getset,              /* tp_getset */
+  0,                            /* tp_base */
+  0,                            /* tp_dict */
+  0,                            /* tp_descr_get */
+  0,                            /* tp_descr_set */
+  0,                            /* tp_dictoffset */
+  0,                            /* tp_init */
+  0,                            /* tp_alloc */
+  0,                            /* tp_new */
+};
+
+
+int
+_setup_wtbarr_type(PyObject* m) {
+  if (PyType_Ready(&PyWtbarrType) < 0) {
+    return -1;
+  }
+
+  Py_INCREF(&PyWtbarrType);
+
+  PyModule_AddObject(m, "Wtbarr", (PyObject *)&PyWtbarrType);
+
+  return 0;
+}
--- a/astropy/wcs/tests/helper.py
+++ b/astropy/wcs/tests/helper.py
@@ -1,0 +1,112 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import pytest
+
+import numpy as np
+
+from astropy import wcs
+from astropy.io import fits
+
+
+class SimModelTAB:
+    def __init__(self, nx=150, ny=200, crpix=[1, 1], crval = [1, 1],
+                 cdelt = [1, 1], pc = {'PC1_1': 1, 'PC2_2': 1}):
+        """  set essential parameters of the model (coord transformations)  """
+        assert nx > 2 and ny > 1  # a limitation of this particular simulation
+        self.nx = nx
+        self.ny = ny
+        self.crpix = crpix
+        self.crval = crval
+        self.cdelt = cdelt
+        self.pc = pc
+
+    def fwd_eval(self, xy):
+        xb = 1 + self.nx // 3
+        px = np.array([1, xb, xb, self.nx + 1])
+        py = np.array([1, self.ny + 1])
+
+        xi = self.crval[0] + self.cdelt[0] * (px - self.crpix[0])
+        yi = self.crval[1] + self.cdelt[1] * (py - self.crpix[1])
+
+        cx = np.array([0.0, 0.26, 0.8, 1.0])
+        cy = np.array([-0.5, 0.5])
+
+        xy = np.atleast_2d(xy)
+        x = xy[:, 0]
+        y = xy[:, 1]
+
+        mbad = (x < px[0]) | (y < py[0]) | (x > px[-1]) | (y > py[-1])
+        mgood = np.logical_not(mbad)
+
+        i = 2 * (x > xb).astype(np.int)
+
+        psix = self.crval[0] + self.cdelt[0] * (x - self.crpix[0])
+        psiy = self.crval[1] + self.cdelt[1] * (y - self.crpix[1])
+
+        cfx = (psix - xi[i]) / (xi[i + 1] - xi[i])
+        cfy = (psiy - yi[0]) / (yi[1] - yi[0])
+
+        ra = cx[i] + cfx * (cx[i + 1] - cx[i])
+        dec = cy[0] + cfy * (cy[1] - cy[0])
+
+        return np.dstack([ra, dec])[0]
+
+    @property
+    def hdulist(self):
+        """ Simulates 2D data with a _spatial_ WCS that uses the ``-TAB``
+        algorithm with indexing.
+        """
+        # coordinate array (some "arbitrary" numbers with a "jump" along x axis):
+        x = np.array([[0.0, 0.26, 0.8, 1.0], [0.0, 0.26, 0.8, 1.0]])
+        y = np.array([[-0.5, -0.5, -0.5, -0.5], [0.5, 0.5, 0.5, 0.5]])
+        c = np.dstack([x, y])
+
+        # index arrays (skip PC matrix for simplicity - assume it is an
+        # identity matrix):
+        xb = 1 + self.nx // 3
+        px = np.array([1, xb, xb, self.nx + 1])
+        py = np.array([1, self.ny + 1])
+        xi = self.crval[0] + self.cdelt[0] * (px - self.crpix[0])
+        yi = self.crval[1] + self.cdelt[1] * (py - self.crpix[1])
+
+        # structured array (data) for binary table HDU:
+        arr = np.array(
+            [(c, xi, yi)],
+            dtype=[
+                ('wavelength', np.float64, c.shape),
+                ('xi', np.double, (xi.size,)),
+                ('yi', np.double, (yi.size,))
+            ]
+        )
+
+        # create binary table HDU:
+        bt = fits.BinTableHDU(arr);
+        bt.header['EXTNAME'] = 'WCS-TABLE'
+
+        # create primary header:
+        image_data = np.ones((self.ny, self.nx), dtype=np.float32)
+        pu = fits.PrimaryHDU(image_data)
+        pu.header['ctype1'] = 'RA---TAB'
+        pu.header['ctype2'] = 'DEC--TAB'
+        pu.header['naxis1'] = self.nx
+        pu.header['naxis2'] = self.ny
+        pu.header['PS1_0'] = 'WCS-TABLE'
+        pu.header['PS2_0'] = 'WCS-TABLE'
+        pu.header['PS1_1'] = 'wavelength'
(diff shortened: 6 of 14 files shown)


timj:
This is great. LSST recently started writing files with -TAB for a small spectrograph we have. I can donate a sample file if you like?  Also, wcsLib itself generates a rather complex -TAB example file. Have you tested with that file? 

My test files are at: https://jira.lsstcorp.org/browse/DM-21797

mcara:
@timj I got your files and I was able to create `astropy.wcs.WCS objects` without issues for both of them. The numbers appear "reasonable" to me though I really have no idea whether the coordinates that I am getting are what you expect to see.

nden:
@timj Could someone on the LSST team test this PR as well?

nden:
@MSeifert04 Since you've looked at the C wrappers before do you think you can review this one?

MSeifert04:
@mcara Is it feasible to undo the purely style-related changes? It makes the PR quite complicated to review and a bit of information is lost (the `/*@null@*/`, etc.).

astropy/wcs/include/astropy_wcs/wcslib_wrap.h
```diff
@@ -1,6 +1,5 @@
 /*
- Author: Michael Droettboom
```
MSeifert04: We really should remove these "authors" comments. But at least we shouldn't modify them. The code is BSD-licensed and authored by "astropy" not by individual contributors.

astropy/wcs/src/astropy_wcs.c
```diff
@@ -28,6 +31,24 @@ static PyTypeObject WcsType;
...
+
+    if (PyArg_ParseTuple(args, "O:set_wtbarr_fitsio_callback", &callback)) {
+        if (!PyCallable_Check(callback)) {
+            PyErr_SetString(PyExc_TypeError, "parameter must be callable");
+            return NULL;
+        }
+        _set_wtbarr_callback(callback);
+
+        /* Boilerplate to return "None" */
+        Py_INCREF(Py_None);
```
MSeifert04: There's the [`Py_RETURN_NONE`](https://docs.python.org/3/c-api/none.html#c.Py_RETURN_NONE) macro. Also the comment above isn't necessary.

MSeifert04:
This sounds like a great addition. Thank you @mcara . I'll do a review, but I'm not sure if I will be able to do the review today/tomorrow.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,149 @@ convert_rejections_to_warnings() {
...
+    naxes[i] = (int) npy_naxes[i];
+  }
+
+  if (naxis != wtb->ndim) {
+    if (wtb->kind == 'c' && wtb->ndim == 2 && naxis == 1) {
+      /* Allow TDIMn to be omitted for degenerate coordinate arrays. */
+      naxis = 2;
+      naxes[1] = 1;
+    } else {
+      sprintf(err_msg,
```
MSeifert04: I would prefer [`PyErr_Format`](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Format) over `sprintf` and `PyErr_SetString`.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -3773,38 +3907,32 @@ PyWcsprm_set_velref(
...
+  if (list == NULL)
+    return NULL;
 
-/*   result = PyList_New(nwtb); */
-/*   if (result == NULL) { */
-/*     return NULL; */
-/*   } */
+  for (i = 0; i < nwtb; ++i) {
+    elem = (PyObject *)PyWtbarr_cnew((PyObject *)self, &(self->x.wtb[i]));
+    if (elem == NULL) return NULL;
```
MSeifert04: You need to `Py_DECREF(list)` in this case.

astropy/wcs/src/wcslib_wtbarr_wrap.c
```diff
@@ -0,0 +1,238 @@
...
+static void PyWtbarr_dealloc(PyWtbarr* self) {
+  PyWtbarr_clear(self);
+  Py_TYPE(self)->tp_free((PyObject*)self);
+}
+
+
+PyWtbarr* PyWtbarr_cnew(PyObject* wcsprm, struct wtbarr* x) {
+  PyWtbarr* self;
+  self = (PyWtbarr*)(&PyWtbarrType)->tp_alloc(&PyWtbarrType, 0);
+  self->x = x;
```
MSeifert04: In case the `tp_alloc` fails this will de-reference a NULL pointer, or?

astropy/wcs/src/wcslib_wtbarr_wrap.c
```diff
@@ -0,0 +1,238 @@
+/*
+ Author: Mihai Cara
```
MSeifert04: We generally don't add authorship on a per file basis. Would you mind removing this?

astropy/wcs/src/wcslib_wtbarr_wrap.c
```diff
@@ -0,0 +1,238 @@
...
+
+
+static PyObject* PyWtbarr_print_contents(PyWtbarr* self) {
+  /* This is not thread-safe, but since we're holding onto the GIL,
+     we can assume we won't have thread conflicts */
+  wcsprintf_set(NULL);
+  wtbarrprt(self->x);
+  printf("%s", wcsprintf_buf());
+
+  Py_INCREF(Py_None);
```
MSeifert04: I would prefer [`Py_RETURN_NONE`](https://docs.python.org/3/c-api/none.html#c.Py_RETURN_NONE) here.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,149 @@ convert_rejections_to_warnings() {
...
+  if (wtb->ndim < 1) {
+    PyErr_SetString(PyExc_ValueError, "Number of dimensions should be positive.");
+    return 0;
+  }
+
+  arglist = Py_BuildValue("(OsiiCsii)", hdulist, wtb->extnam, wtb->extver,
+                          wtb->extlev, wtb->kind, wtb->ttype, wtb->row,
+                          wtb->ndim);
+
+  result = PyObject_CallObject(get_wtbarr_data, arglist);
```
MSeifert04: This needs a check if `result != NULL`.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,149 @@ convert_rejections_to_warnings() {
...
+                    "and/or indices.");
+    return 0;
+  }
+
+  if (wtb->ndim < 1) {
+    PyErr_SetString(PyExc_ValueError, "Number of dimensions should be positive.");
+    return 0;
+  }
+
+  arglist = Py_BuildValue("(OsiiCsii)", hdulist, wtb->extnam, wtb->extver,
```
MSeifert04: Needs a check if `arglist != NULL`.

MSeifert04: You can avoid this call if you use [`PyObject_CallFunction`](https://docs.python.org/3/c-api/object.html#c.PyObject_CallFunction) below.

mcara:
@MSeifert04 Thanks a lot for a thorough review. I hope I have a all your comments in the latest commit.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,145 @@ convert_rejections_to_warnings() {
...
+      PyArray_DescrFromType(NPY_DOUBLE), 0, 0, NPY_ARRAY_CARRAY, NULL);
+
+  if (arrayp == NULL) {
+    PyErr_SetString(PyExc_TypeError, "Unable to convert wtbarr callback "
+                    "result to a numpy.ndarray.");
+    Py_DECREF(result);
+    return 0;
+  }
+
+  if ((PyObject *)arrayp != result) Py_DECREF(result);
```
MSeifert04: I think you can just unconditionally `Py_DECREF(result)` after the `PyArray_FromAny` call. It seems like it always returns a new reference and doesn't steal a reference for `result`. Or was that something that didn't work?

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,145 @@ convert_rejections_to_warnings() {
...
+  if (naxis == 0) {
+    PyErr_SetString(PyExc_ValueError, "-TAB coordinate or index arrays "
+                    "cannot be 0-dimensional.");
+    Py_DECREF(arrayp);
+    return 0;
+  }
+
+  npy_naxes = PyArray_DIMS(arrayp);
+
+  if (!(naxes = calloc((size_t)naxis, sizeof(int)))) {
```
MSeifert04: We might get around the allocation if we use `NPY_MAXDIMS` for a stack-allocated array.

astropy/wcs/src/wcslib_wrap.c
mcara: If `result` is already a C contiguous array of doubles then `arrayp` is `result`. I do not think the ref counter is incremented. I would rather leave this as is.

MSeifert04: > I would rather leave this as is.

The problem is that failing to `DECREF` an owned reference will result in a memory-leak, but `DECREF` a borrowed reference will likely run into a segmentation fault. So I think we need to figure this out - if you don't have the time I can do that later today.

I'm almost sure that it increments the reference counter even if it's already the correct array (unfortunately I'm currently not on my computer so I cannot check). The routine creating the array should be [`PyArray_GetArrayParamsFromObject`](https://github.com/numpy/numpy/blob/v1.17.4/numpy/core/src/multiarray/ctors.c#L1600).

astropy/wcs/src/wcslib_tabprm_wrap.c
```diff
@@ -413,7 +414,7 @@ PyTypeObject PyTabprmType = {
   0,                            /*tp_getattr*/
   0,                            /*tp_setattr*/
   0,                            /*tp_compare*/
-  (reprfunc)PyTabprm___str__,   /*tp_repr*/
```
MSeifert04: What is the reason for removing the repr? 

If that wasn't an accidental removal we probably should also remove the function implementation of `PyTabprm___str__`

mcara: I removed `__repr__` because I hated the output of it when I was debugging my code. However, it is really a matter of taste and whether people prefer excessive vs succinct output. To understand (=see for yourself) what I am talking about, try: 

```python
>>> from astropy import wcs
>>> 5 * [wcs.WCS().wcs]
```

and now please find the `cdelt` of the 2nd `WCS` in the printed output. So, for me it was easier to see how many `tabprm` or `wtbarr` structures I had in the members `WCS.wcs.tab` and `WCS.wcs.wtb`. If I wanted to see the content of one particular structure, I would do `print(WCS.wcs.wtb[0])`.

Maybe my above example is not appropriate for a single WCS object (and that is how it is used most of the time), but for lists of `tabprm` and `wtbarr` structures, it felt overwhelming. Maybe a more compact format can be found for `__repr__` that is also more informative than a just object address.

astropy/wcs/src/wcslib_wrap.c
mcara: Good catch! Indeed, it increments reference count even when input is passed through. I'll fix that.

astropy/wcs/src/wcslib_wrap.c
mcara: We might... I doubt coordinate arrays will exceed 32 axes and anyway, proposed implementation uses `numpy` arrays so it is limited by `numpy` anyway.

mcara:
@pllim Could you please be so kind as to help figuring out what is the issue with tests on Windows: https://travis-ci.org/astropy/astropy/jobs/615643939#L1980-L1990 Thank you!

pllim:
I can tell you it probably segfaulted but I have no desire to debug it. You can try have the CI install pytest-faulthandler and see if that gives you anything useful. Otherwise, you might need to run it locally on a Windows machine with `gdb` or something equivalent. Maybe @jhunkeler got better ideas.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -3773,38 +3893,34 @@ PyWcsprm_set_velref(
...
+  PyObject* list;
+  PyObject* elem;
+  int i, nwtb;
 
-/*   nwtb = self->x.nwtb; */
+  nwtb = self->x.nwtb;
+
+  list = PyList_New(nwtb);
+  if (list == NULL) {
+    Py_DECREF(list);
```
MSeifert04: This needs to be in the `if (elem == NULL)` branch in the loop.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -170,9 +301,7 @@ note_change(PyWcsprm* self) {
 }
 
 static void
-PyWcsprm_dealloc(
```
MSeifert04: Could you also undo these style-changes?

MSeifert04:
Locally I get quite a few warnings with Windows for the wcs module, maybe one of the is related to the segfault (sorry for the german locale 😓)

```

wcsbth.l(1925): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
wcsbth.l(1943): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
wcsbth.l(2164): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
wcsbth.l(2263): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
wcsbth.l(2272): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
wcsbth.l(2504): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
wcsbth.l(2769): warning C4244: "=": Konvertierung von "__int64" in "int", möglicher Datenverlust

wcspih.l(2048): warning C4244: "=": Konvertierung von "__int64" in "int", möglicher Datenverlust
wcspih.l(2052): warning C4244: "=": Konvertierung von "__int64" in "int", möglicher Datenverlust

wcsulex.c(9032): warning C4013: "read" undefiniert; Annahme: extern mit Rückgabetyp int
wcsulex.c(9032): warning C4013: "fileno" undefiniert; Annahme: extern mit Rückgabetyp int

wcsutrn.c(5228): warning C4013: "read" undefiniert; Annahme: extern mit Rückgabetyp int
wcsutrn.c(5228): warning C4013: "fileno" undefiniert; Annahme: extern mit Rückgabetyp int

cextern\wcslib\C\cel.c(147): warning C4311: "Typumwandlung": Zeigerverkürzung von "wcserr *const " zu "unsigned long"

cextern\wcslib\C\dis.c(453): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\dis.c(460): warning C4311: "Typumwandlung": Zeigerverkürzung von "dpkey *const " zu "unsigned long"
cextern\wcslib\C\dis.c(471): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\dis.c(481): warning C4311: "Typumwandlung": Zeigerverkürzung von "int **const " zu "unsigned long"
cextern\wcslib\C\dis.c(494): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\dis.c(501): warning C4311: "Typumwandlung": Zeigerverkürzung von "double **const " zu "unsigned long"
cextern\wcslib\C\dis.c(510): warning C4311: "Typumwandlung": Zeigerverkürzung von "double **const " zu "unsigned long"
cextern\wcslib\C\dis.c(519): warning C4311: "Typumwandlung": Zeigerverkürzung von "int **const " zu "unsigned long"
cextern\wcslib\C\dis.c(522): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *" zu "unsigned long"
cextern\wcslib\C\dis.c(536): warning C4311: "Typumwandlung": Zeigerverkürzung von "double **const " zu "unsigned long"
cextern\wcslib\C\dis.c(539): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *" zu "unsigned long"
cextern\wcslib\C\dis.c(557): warning C4311: "Typumwandlung": Zeigerverkürzung von "wcserr *const " zu "unsigned long"
cextern\wcslib\C\dis.c(563): warning C4311: "Typumwandlung": Zeigerverkürzung von "int (__cdecl **const )(int,const int *,const double *,int,const double *,double *)" zu "unsigned long"
cextern\wcslib\C\dis.c(591): warning C4311: "Typumwandlung": Zeigerverkürzung von "int (__cdecl **const )(int,const int *,const double *,int,const double *,double *)" zu "unsigned long"
cextern\wcslib\C\dis.c(596): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\dis.c(601): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\dis.c(604): warning C4311: "Typumwandlung": Zeigerverkürzung von "dpkey *const " zu "unsigned long"
cextern\wcslib\C\dis.c(607): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"

cextern\wcslib\C\lin.c(464): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\lin.c(472): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\lin.c(481): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\lin.c(488): warning C4311: "Typumwandlung": Zeigerverkürzung von "disprm *const " zu "unsigned long"
cextern\wcslib\C\lin.c(491): warning C4311: "Typumwandlung": Zeigerverkürzung von "disprm *const " zu "unsigned long"
cextern\wcslib\C\lin.c(528): warning C4311: "Typumwandlung": Zeigerverkürzung von "wcserr *const " zu "unsigned long"
cextern\wcslib\C\lin.c(534): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\lin.c(539): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\lin.c(542): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\lin.c(545): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\lin.c(548): warning C4311: "Typumwandlung": Zeigerverkürzung von "disprm *const " zu "unsigned long"
cextern\wcslib\C\lin.c(551): warning C4311: "Typumwandlung": Zeigerverkürzung von "disprm *const " zu "unsigned long"

cextern\wcslib\C\prj.c(263): warning C4311: "Typumwandlung": Zeigerverkürzung von "wcserr *const " zu "unsigned long"
cextern\wcslib\C\prj.c(7248): warning C4305: "Initialisierung": Verkürzung von "double" in "const float"

cextern\wcslib\C\spc.c(201): warning C4311: "Typumwandlung": Zeigerverkürzung von "wcserr *const " zu "unsigned long"

cextern\wcslib\C\tab.c(554): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\tab.c(562): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\tab.c(570): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\tab.c(578): warning C4311: "Typumwandlung": Zeigerverkürzung von "double **const " zu "unsigned long"
cextern\wcslib\C\tab.c(581): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *" zu "unsigned long"
cextern\wcslib\C\tab.c(594): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\tab.c(616): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\tab.c(625): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\tab.c(634): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\tab.c(643): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\tab.c(665): warning C4311: "Typumwandlung": Zeigerverkürzung von "wcserr *const " zu "unsigned long"
cextern\wcslib\C\tab.c(675): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\tab.c(679): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\tab.c(683): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\tab.c(687): warning C4311: "Typumwandlung": Zeigerverkürzung von "double **const " zu "unsigned long"
cextern\wcslib\C\tab.c(692): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *" zu "unsigned long"
cextern\wcslib\C\tab.c(697): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"

cextern\wcslib\C\wcs.c(1659): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1668): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1678): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1686): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1694): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\wcs.c(1699): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\wcs.c(1717): warning C4311: "Typumwandlung": Zeigerverkürzung von "pvcard *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1724): warning C4311: "Typumwandlung": Zeigerverkürzung von "pscard *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1732): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1743): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1761): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1770): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\wcs.c(1781): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1794): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1807): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1820): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1902): warning C4311: "Typumwandlung": Zeigerverkürzung von "tabprm *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1906): warning C4311: "Typumwandlung": Zeigerverkürzung von "wtbarr *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1911): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1924): warning C4311: "Typumwandlung": Zeigerverkürzung von "wcserr *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1936): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1939): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1942): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1945): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1948): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\wcs.c(1951): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\wcs.c(1954): warning C4311: "Typumwandlung": Zeigerverkürzung von "pvcard *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1957): warning C4311: "Typumwandlung": Zeigerverkürzung von "pscard *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1960): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1963): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1967): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1970): warning C4311: "Typumwandlung": Zeigerverkürzung von "char (*const )[72]" zu "unsigned long"
cextern\wcslib\C\wcs.c(1973): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1976): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1979): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1982): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1985): warning C4311: "Typumwandlung": Zeigerverkürzung von "tabprm *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(1988): warning C4311: "Typumwandlung": Zeigerverkürzung von "wtbarr *const " zu "unsigned long"
cextern\wcslib\C\wcs.c(2006): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *" zu "unsigned long"
cextern\wcslib\C\wcs.c(2007): warning C4311: "Typumwandlung": Zeigerverkürzung von "double **" zu "unsigned long"
cextern\wcslib\C\wcs.c(2008): warning C4311: "Typumwandlung": Zeigerverkürzung von "double *" zu "unsigned long"
cextern\wcslib\C\wcs.c(2180): warning C4244: "=": Konvertierung von "__int64" in "int", möglicher Datenverlust

cextern\wcslib\C\wcserr.c(178): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich

cextern\wcslib\C\wcsfix.c(288): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
cextern\wcslib\C\wcsfix.c(641): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
cextern\wcslib\C\wcsfix.c(652): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
cextern\wcslib\C\wcsfix.c(703): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
cextern\wcslib\C\wcsfix.c(714): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
cextern\wcslib\C\wcsfix.c(786): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich
cextern\wcslib\C\wcsfix.c(798): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich

cextern\wcslib\C\wcshdr.c(1343): warning C4244: "=": Konvertierung von "__int64" in "int", möglicher Datenverlust
cextern\wcslib\C\wcshdr.c(1907): warning C4244: "=": Konvertierung von "__int64" in "int", möglicher Datenverlust

cextern\wcslib\C\wcsutil.c(55): warning C4267: "=": Konvertierung von "size_t" nach "int", Datenverlust möglich

astropy\wcs\src\wcslib_wtbarr_wrap.c(76): warning C4311: "Typumwandlung": Zeigerverkürzung von "int *const " zu "unsigned long"
astropy\wcs\src\wcslib_wtbarr_wrap.c(83): warning C4311: "Typumwandlung": Zeigerverkürzung von "double **const " zu "unsigned long"
```

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -3773,38 +3893,34 @@ PyWcsprm_set_velref(
...
+    if (elem == NULL) return NULL;
 
-/*   for (i = 0; i < nwtb; ++i) { */
-/*     subresult = (PyObject *)PyWtbarr_cnew((PyObject *)self, &(self->x.wtb[i])); */
-/*     if (subresult == NULL) { */
-/*       Py_DECREF(result); */
-/*       return NULL; */
-/*     } */
+    if (PyList_SetItem(list, i, elem) == -1) {
+      Py_DECREF(elem);
```
MSeifert04: `PyList_SetItem` steals a reference to `elem`, so it must not be decrefed again.

MSeifert04: But since we just created the list, you could use `PyList_SET_ITEM` here instead because we can garantuee that `elem` is not NULL and the `i` is within the bounds of the list.

astropy/wcs/src/wcslib_wrap.c
mcara: Ah, this is an unfortunate bug introduced in the last commit. However, I am not sure what does

astropy/wcs/src/wcslib_wrap.c
mcara: I thought I did but maybe I missed some.

astropy/wcs/src/wcslib_wrap.c
MSeifert04: Inside the loop you check if `elem == NULL`, in that case you return `NULL`. Before the return there you need to use the `PyDECREF(list)`.

mcara: oopsie... I put the decref in a wrong place. Thanks!

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,135 @@ convert_rejections_to_warnings() {
...
+      PyArray_DescrFromType(NPY_DOUBLE), 0, 0, NPY_ARRAY_CARRAY, NULL);
+
+  if (arrayp == NULL) {
+    PyErr_SetString(PyExc_TypeError, "Unable to convert wtbarr callback "
+                    "result to a numpy.ndarray.");
+    Py_DECREF(result);
+    return 0;
+  }
+
+  Py_DECREF(result);
```
MSeifert04: Just stylistic: You can do the `Py_DECREF` immediately after the call, that way you don't need to do it in the branch above and here.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,135 @@ convert_rejections_to_warnings() {
...
+      /* Allow TDIMn to be omitted for degenerate coordinate arrays. */
+      naxis = 2;
+      naxes[1] = 1;
+    } else {
+      PyErr_Format(PyExc_ValueError,
+          "An array with an unexpected number of axes was "
+          "received from the callback. Expected %d but got %d.",
+          wtb->ndim, (int) naxis);
+      Py_DECREF(arrayp);
+      free(naxes);
```
MSeifert04: Since `naxes` is stack-allocated this will be undefined behavior (see also later).

astropy/wcs/src/wcslib_wrap.c
mcara: Important, IMO.

astropy/wcs/src/wcslib_wtbarr_wrap.c
```diff
@@ -0,0 +1,233 @@
...
+  wcsprintf("     i: %d\n", wtb->i);
+  wcsprintf("     m: %d\n", wtb->m);
+  wcsprintf("  kind: %c\n", wtb->kind);
+  wcsprintf("extnam: %s\n", wtb->extnam);
+  wcsprintf("extver: %d\n", wtb->extver);
+  wcsprintf("extlev: %d\n", wtb->extlev);
+  wcsprintf(" ttype: %s\n", wtb->ttype);
+  wcsprintf("   row: %ld\n", wtb->row);
+  wcsprintf("  ndim: %d\n", wtb->ndim);
+  wcsprintf("dimlen: %#lx\n", (unsigned long int)wtb->dimlen);
```
MSeifert04: I'm not completely sure, but I think pointer addresses should be printed using `%p` and the pointer needs to be cast to `void *`.

Not sure if this could lead to real problems but at least on windows long is only 32bits (one needs to use long long for 64bit).

mcara: to the best of my knowledge (which is kind of rusty) pointers are `DWORD` on Windows (i.e., `unsigned long`) but I may be completely wrong. Also, I essentially tried to reproduce the same formatting the WCSLIB is using, see https://github.com/astropy/astropy/blob/918abcd480acdf824c5e5d856daf0573811d9b3c/cextern/wcslib/C/wcsprintf.h#L136-L141

astropy/wcs/src/wcslib_wrap.c
mcara: Latest commit will use `PyList_Append`.

MSeifert04:
I encountered the failing also locally. The weird thing is only when the complete tests are run, when I just test the wcs module then it fails only sometimes.

The faulthandler also indicates it happens in the teardown which means that something that is already disposed may be kept around. Since it's only in the complete test run and I don't really know how to make a debug build of Python on windows this is a very annoying and slow process. 😕

mcara:
@MSeifert04 Thank you so much for doing this and also for taking a look at the Windows issue.

astropy/wcs/src/wcslib_wrap.c
MSeifert04: I don't know if that will work. I think the `PyList_New` sets the size (not only the allocated size) and Append would then append after these. So we would have a list containing NULLs which will probably crash Python.

There's nothing wrong with using `PyList_SetItem` or `PyList_SET_ITEM`. The nicest thing about `PyList_SET_ITEM` here is that we don't need to account for (not possible) failures.

mcara: You are right. My tests did not catch this but I am working on a new set of tests that will catch this king of error.

astropy/wcs/src/wcslib_wtbarr_wrap.c
MSeifert04: My knowledge is also a bit lacking but I think windows uses 48bits for pointers and the type is `unsigned long long`.

But wouldn't it be best to use the `%p`placeholder and `void *`-casting - or does this have other issues? That's what [CPython also uses for `id` and the `repr` of objects without `__repr__`](https://github.com/python/cpython/blob/v3.8.0/Objects/object.c#L518-L520)

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -157,6 +159,132 @@ convert_rejections_to_warnings() {
...
+void _set_wtbarr_callback(PyObject* callback) {
+  Py_XINCREF(callback);         /* Add a reference to new callback */
+  Py_XDECREF(get_wtbarr_data);  /* Dispose of previous callback */
+  get_wtbarr_data = callback;   /* Remember new callback */
+}
+
+
+int _update_wtbarr_from_hdulist(PyObject *hdulist, struct wtbarr *wtb) {
+  PyArrayObject *arrayp=NULL;
+  PyObject *result=NULL, *arglist = NULL;
```
MSeifert04: `arglist` is now unused.

mcara:
These Windows test failures seem to be now in a different place: https://travis-ci.org/astropy/astropy/jobs/615792192#L1992-L2003

```
Thread 0x00000a30 (most recent call first):
1987  File "...\astropy\samp\client.py", line 180 in _serve_forever
1988  File "C:\tools\miniconda3\envs\test\lib\threading.py", line 870 in run
1989  File "C:\tools\miniconda3\envs\test\lib\threading.py", line 926 in _bootstrap_inner
1990  File "C:\tools\miniconda3\envs\test\lib\threading.py", line 890 in _bootstrap
1991
1992Current thread 0x000002dc (most recent call first):
1993  File "...\astropy\io\fits\card.py", line 584 in normalize_keyword
1994  File "...\astropy\io\fits\header.py", line 1695 in _cardindex
1995  File "...\astropy\io\fits\header.py", line 148 in __getitem__
1996  File "...\astropy\io\fits\header.py", line 887 in get
1997  File "...\astropy\io\fits\hdu\base.py", line 188 in name
1998  File "...\astropy\io\fits\hdu\hdulist.py", line 725 in index_of
1999  File "...\astropy\io\fits\hdu\hdulist.py", line 754 in _positive_index_of
2000  File "...\astropy\io\fits\hdu\hdulist.py", line 300 in __getitem__
2001  File "...\astropy\wcs\wcs.py", line 109 in _load_tab_bintable
2002  File "...\astropy\wcs\wcs.py", line 486 in __init__
2003  File "...\astropy\wcs\tests\test_tab.py", line 144 in test_2d_spacial_tab_vs_model
2004
```

pllim:
@mcara , you should get a Windows machine from ITSD.

MSeifert04:
I'm a bit struck with investigating the segfault. I'm not able to get a debug Python build with numpy+astropy working. I also ran out of ideas what could cause the issue. It looks like some use-after-free issue, but the fact that it's only happening on Windows and only in some circumstances make this really weird. 

Just on the off-chance that this isn't Windows-specific but just manifests on Windows does someone know how valgrind or address-sanitize (or whatever could be used to debug this) this on unix?

jdavies-st:
@pllim fyi, `pytest-faulthandler` is now built into `pytest` since `pytest 5.0`.

pllim:
Thanks, @MSeifert04 ! I spoke to @mcara offline. Given that this work is part of his sprint, he should investigate more with his team members with (a) Windows laptop (b) `valgrind` as you suggested. cc @jdavies-st and @perrygreenfield

astropy/wcs/tests/test_tab.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+
+import numpy as np
+
+from astropy import wcs
+
+from . helper import SimModelTAB
+
+
+def test_2d_spacial_tab_roundtrip():
```
jdavies-st: Can you use the `tab_wcs_2di` pytest fixture here and in the test below instead of importing `SimModelTAB`?  For consistency with your other tests?

astropy/wcs/tests/test_wtbarr.py
```diff
@@ -0,0 +1,66 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import pytest
+
+import numpy as np
+
+from astropy import wcs
+
+from . helper import SimModelTAB
```
jdavies-st: Perhaps this import is no longer needed since you are now using the fixture?

mcara: Yeah. I got distracted and forgot to remove the line.

stargaser:
FWIW I successfully tested this on a Herschel PACS spectral cube that contains -TAB for the wavelength axis. Thanks @mcara for this implementation!

mcara:
@stargaser Thank you! It is great to know that it works on real images (at least sometimes).

@MSeifert04 I wish I could also understand why Windows tests are segfaulting. I replaced 
```python
# Set coordinate extraction callback for WCS -TAB:
def _load_tab_bintable(hdulist, extnam, extver, extlev, kind, ttype, row, ndim):
    if ttype == 'wavelength':
        arr = np.array(
            [[[0.0, -0.5], [0.26, -0.5], [0.8, -0.5], [1.0, -0.5]],
             [[0.0, 0.5], [0.26, 0.5], [0.8, 0.5], [1.0, 0.5]]],
            dtype=np.float64, order='C'
        )
        return arr
    elif ttype == 'xi':
        return np.array([1.0, 51.0, 51.0, 151.0], dtype=np.float64, order='C')
    elif ttype == 'yi':
        return np.array([1.0, 201.0], dtype=np.float64, order='C')
    else:
        raise ValueError("wrong ttype")
```
This essentially shortcuts all the "callback" calls and dealing with `io.fits.HDUList` and instead simply supplies the coordinate and index arrays required for `-TAB`. (arrays are specific for the tests). As far as I can see, there should be no issues with the creation of these arrays. The tests are still segfaulting.

mcara:
@MSeifert04 I wish I could also understand why Windows tests are segfaulting. I replaced `_update_wtbarr_from_hdulist()` in `wcslib_wrap.c` with:
```c
int _update_wtbarr_from_hdulist(PyObject *hdulist, struct wtbarr *wtb) {
  PyArrayObject *arrayp=NULL;
  PyObject *result=NULL;
  int i, naxis, nelem, naxes[NPY_MAXDIMS];
  npy_intp *npy_naxes;
  npy_double *appayp_data;
  double arr1[] = { 11.0, 12.0, 61.0, 12.0, 111.0, 12.0, 1010.0, 12.0, 11.0, 211.0, 61.0, 211.0, 111.0, 211.0, 1010.0, 211.0 };
  int naxes1[] = { 2, 4, 2 };
  double arr2[] = { 1.0, 51.0, 51.0, 151.0 };
  int naxes2[] = { 4 };
  double arr3[] = { 1.0, 201.0 };
  int naxes3[] = { 2 };


  if (*wtb->ttype == 'w') {
      for (i = 0; i < 3; i++) wtb->dimlen[i] = naxes1[i];
      *(wtb->arrayp) = calloc((size_t)16, sizeof(double));
      for (i = 0; i < 16; i++) (*wtb->arrayp)[i] = arr1[i];
  }
  else if (*wtb->ttype == 'x') {
      wtb->dimlen[0] = naxes2[0];
      *(wtb->arrayp) = calloc((size_t)4, sizeof(double));
      for (i = 0; i < 4; i++) (*wtb->arrayp)[i] = arr2[i];
  }
  else {
      wtb->dimlen[0] = naxes3[0];
      *(wtb->arrayp) = calloc((size_t)2, sizeof(double));
      for (i = 0; i < 2; i++) (*wtb->arrayp)[i] = arr3[i];
  }
  return 1;
```
This essentially bypasses C-Python "callback" calls and also dealing with `io.fits.HDUList` and instead simply supplies the coordinate and index arrays required for `-TAB`. (arrays are specific for the tests). As far as I can see, there should be no issues with the creation of these arrays. The tests are still segfaulting.

MSeifert04:
This is quite difficult to debug on Windows. I don't know when I have the time to dig through that some more. First I probably need to get the debug python on windows running with pip installed C-Extension modules.

MSeifert04:
Interestingly I've ran into a segfault today that used `BinTableHDU` as well (at least it used compression which uses the binary-table). So maybe this is not related to this PR at all - that narrows it down to a bug in either CPython/cfitsio/NumPy/astropy - come to think of it ... *narrows down* probably isn't the appropriate word.

pllim:
> I've ran into a segfault today that used `BinTableHDU` as well

@MSeifert04 , are you sure it is the same segfault? Did you report it as an issue?

> So maybe this is not related to this PR at all

But somehow this PR triggers it in CI? :thinking: 

> *narrows down*

Maybe *widens up* is more correct? :sweat_smile:

mcara:
I can run the test code on Windows *outside* the pytest system. That is, if I open an interactive session in let's say `ipython`, I get no errors. Only when the code is run as part of the `pytest` I get the crash.

```python
(astro) PS C:\Users\mcara\code> ipython
Python 3.7.5 (default, Oct 31 2019, 15:18:51) [MSC v.1916 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.9.0 -- An enhanced Interactive Python. Type '?' for help.
 
In [1]: import numpy as np
   ...: from astropy.io import fits
   ...: from astropy import wcs
   ...: nx=150
   ...: ny=200
   ...: crpix=[1, 1]
   ...: crval = [1, 1]
   ...: cdelt = [1, 1]
   ...: pc = {'PC1_1': 1, 'PC2_2': 1}
 
In [2]: def create_hdulist(nx, ny, crpix, crval, cdelt, pc):
   ...:     x = np.array([[0.0, 0.26, 0.8, 1.0], [0.0, 0.26, 0.8, 1.0]])
   ...:     y = np.array([[-0.5, -0.5, -0.5, -0.5], [0.5, 0.5, 0.5, 0.5]])
   ...:     c = np.dstack([x, y])
   ...:     # index arrays
   ...:     xb = 1 + nx // 3
   ...:     px = np.array([1, xb, xb, nx + 1])
   ...:     py = np.array([1, ny + 1])
   ...:     xi = crval[0] + cdelt[0] * (px - crpix[0])
   ...:     yi = crval[1] + cdelt[1] * (py - crpix[1])
   ...:     # structured array (data) for binary table HDU:
   ...:     arr = np.array([(c, xi, yi)], dtype=[('wavelength', np.float64, c.shape),
   ...:         ('xi', np.double, (xi.size,)), ('yi', np.double, (yi.size,))
   ...:     ])
   ...:     # create binary table HDU:
   ...:     bt = fits.BinTableHDU(arr);
   ...:     bt.header['EXTNAME'] = 'WCS-TABLE'
   ...:     # create primary header:
   ...:     image_data = np.random.random((ny, nx)).astype(np.float32)
   ...:     pu = fits.PrimaryHDU(image_data)
   ...:     pu.header['ctype1'] = 'RA---TAB'
   ...:     pu.header['ctype2'] = 'DEC--TAB'
   ...:     pu.header['naxis1'] = nx
   ...:     pu.header['naxis2'] = ny
   ...:     pu.header['PS1_0'] = 'WCS-TABLE'
   ...:     pu.header['PS2_0'] = 'WCS-TABLE'
   ...:     pu.header['PS1_1'] = 'wavelength'
   ...:     pu.header['PS2_1'] = 'wavelength'
   ...:     pu.header['PV1_3'] = 1
   ...:     pu.header['PV2_3'] = 2
   ...:     pu.header['CUNIT1'] = 'deg'
   ...:     pu.header['CUNIT2'] = 'deg'
   ...:     pu.header['CDELT1'] = cdelt[0]
   ...:     pu.header['CDELT2'] = cdelt[1]
   ...:     pu.header['CRPIX1'] = crpix[0]
   ...:     pu.header['CRPIX2'] = crpix[1]
   ...:     pu.header['CRVAL1'] = crval[0]
   ...:     pu.header['CRVAL2'] = crval[1]
   ...:     pu.header['PS1_2'] = 'xi'
   ...:     pu.header['PS2_2'] = 'yi'
   ...:     for k, v in pc.items():
   ...:         pu.header[k] = v
   ...:     hdulist = fits.HDUList([pu, bt])
   ...:     return hdulist
   ...:
 
In [3]: for k in range(1000):
   ...:     h = create_hdulist(nx, ny, crpix, crval, cdelt, pc)
   ...:     w = wcs.WCS(h[0].header, h)
   ...:     w.wcs_world2pix(w.wcs_pix2world([[1, 2]], 1), 1)
   ...:
 
In [4]: w.wcs_world2pix(w.wcs_pix2world([[1, 2]], 1), 1)
Out[4]: array([[1.00000001, 2.00000002]])
 
In [5]: exit
(astro) PS C:\Users\mcara\code>     
```

pllim:
@mcara , so you get a segfault on that same machine when running `pytest` on this branch?

mcara:
I would narrow this _down_ :) to `pytest` and/or its setup, `astropy` setup on windows, and, maybe threading.

mcara:
@pllim yes

mcara:
@jdavies-st suggestion I run a couple of tests. Hopefully they will reveal the underlying issue.

astropy/wcs/tests/test_tab.py
```diff
@@ -0,0 +1,37 @@
...
+    model = SimModelTAB(nx=nx, ny=ny)
+
+    # generate FITS HDU list:
+    hdulist = model.hdulist
+
+    # create WCS object:
+    w = wcs.WCS(hdulist[0].header, hdulist)
+
+    # generate random test coordinates:
+    xy = 0.51 + [nx + 0.99, ny + 0.99] * np.random.random((100, 2))
```
pllim: @mcara says there is no need for random
```suggestion
    xy = 0.51 + [nx + 0.99, ny + 0.99] * np.zeros((100, 2))
```

mcara:
@pllim I think this is irrelevant to testing itself and to the crash. In fact, I could just leave it to be be `np.empty` filled with whatever is in the memory. The values in the image are not used for anything (well, except to get the `shape` of the array).

MSeifert04:
@mcara I can reproduce the segfault if I copy the test + test setup (without pytest) and call the `test_2d_spacial_tab_roundtrip` function. It doesn't always trigger a segfault - only sometimes. Which is why I said that it's probably a use-after-free or double-free.

MSeifert04:
@pllim Not reported yet, I wanted to investigate it in more detail before doing that.

jdavies-st:
Can folks here who have access to Windows verify that the test fails under standard `pytest` testing scenarios?  I.e. isolate this failure from astropy's test runner?

So test under an editable install
```
cd path/to/dev/astropy
pip install -e .
pytest astropy/wcs -k test_2d_spacial_tab_vs_model
```
and under an install to site-packages, making sure to get out of the git repo before running the test.
```
cd path/to/dev/astropy
pip install .
cd ~/
pytest --pyargs astropy.wcs -k test_2d_spacial_tab_vs_model
```
And what results do you get?  @mcara?

mcara:
I run @jdavies-st suggested tests both only for `test_2d_spacial_tab_vs_model` function and the whole test suit (no remote data though) as well and they all run successfully. Only when I run `astropy` tests as suggested on https://docs.astropy.org/en/stable/development/testguide.html:

MSeifert04:
@mcara Does it still work if you repeat the test multiple times in the same pytest session, e.g. something like mentioned in [this StackOverflow answer](https://stackoverflow.com/a/21857841/5393381) would probably be easiest. Or running the script I uploaded in https://github.com/astropy/astropy/pull/9641#issuecomment-561310364.

astropy/wcs/tests/helper.py
```diff
@@ -0,0 +1,112 @@
...
+        cfx = (psix - xi[i]) / (xi[i + 1] - xi[i])
+        cfy = (psiy - yi[0]) / (yi[1] - yi[0])
+
+        ra = cx[i] + cfx * (cx[i + 1] - cx[i])
+        dec = cy[0] + cfy * (cy[1] - cy[0])
+
+        return np.dstack([ra, dec])[0]
+
+    @property
+    def hdulist(self):
```
pllim: Why not make this a lazyproperty?

pllim:
Are you sure you are testing the correct function? The traceback Mihai sent me pointed to a different one:
```
test_2d_spacial_tab_roundtrip Windows fatal exception
```

Also, maybe not using a pytest fixture that computes WCS will fix the problem. Strictly speaking, it is not necessary. All the tests that use the fixture can be made into test class methods, and the WCS stuff can be set up in `setup_class` without using fixture.

MSeifert04:
Even if we work around the segmentation fault we somehow need to figure out the root cause, otherwise this will eventually fail again. Although it would be nice to get the improvement from this PR in.

mcara:
@pllim My original commit did not use fixtures and tests were still failing. And it works with fixtures for @jdavies-st tests `=>` segfault has nothing to do with fixtures.

mcara:
Also, if I extract the code from the test function and run it in an interactive mode (in an interpreter), I get no errors no matter how much I run the code. For those interested, here is the "extracted" test code:
```python
(astro) PS C:\Users\mcara\code> ipython
Python 3.7.5 (default, Oct 31 2019, 15:18:51) [MSC v.1916 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.9.0 -- An enhanced Interactive Python. Type '?' for help.
 
In [1]: import numpy as np
   ...: from astropy.io import fits
   ...: from astropy import wcs
   ...: nx=150
   ...: ny=200
   ...: crpix=[1, 1]
   ...: crval = [1, 1]
   ...: cdelt = [1, 1]
   ...: pc = {'PC1_1': 1, 'PC2_2': 1}
 
In [2]: def create_hdulist(nx, ny, crpix, crval, cdelt, pc):
   ...:     x = np.array([[0.0, 0.26, 0.8, 1.0], [0.0, 0.26, 0.8, 1.0]])
   ...:     y = np.array([[-0.5, -0.5, -0.5, -0.5], [0.5, 0.5, 0.5, 0.5]])
   ...:     c = np.dstack([x, y])
   ...:     # index arrays
   ...:     xb = 1 + nx // 3
   ...:     px = np.array([1, xb, xb, nx + 1])
   ...:     py = np.array([1, ny + 1])
   ...:     xi = crval[0] + cdelt[0] * (px - crpix[0])
   ...:     yi = crval[1] + cdelt[1] * (py - crpix[1])
   ...:     # structured array (data) for binary table HDU:
   ...:     arr = np.array([(c, xi, yi)], dtype=[('wavelength', np.float64, c.shape),
   ...:         ('xi', np.double, (xi.size,)), ('yi', np.double, (yi.size,))
   ...:     ])
   ...:     # create binary table HDU:
   ...:     bt = fits.BinTableHDU(arr);
   ...:     bt.header['EXTNAME'] = 'WCS-TABLE'
   ...:     # create primary header:
   ...:     image_data = np.random.random((ny, nx)).astype(np.float32)
   ...:     pu = fits.PrimaryHDU(image_data)
   ...:     pu.header['ctype1'] = 'RA---TAB'
   ...:     pu.header['ctype2'] = 'DEC--TAB'
   ...:     pu.header['naxis1'] = nx
   ...:     pu.header['naxis2'] = ny
   ...:     pu.header['PS1_0'] = 'WCS-TABLE'
   ...:     pu.header['PS2_0'] = 'WCS-TABLE'
   ...:     pu.header['PS1_1'] = 'wavelength'
   ...:     pu.header['PS2_1'] = 'wavelength'
   ...:     pu.header['PV1_3'] = 1
   ...:     pu.header['PV2_3'] = 2
   ...:     pu.header['CUNIT1'] = 'deg'
   ...:     pu.header['CUNIT2'] = 'deg'
   ...:     pu.header['CDELT1'] = cdelt[0]
   ...:     pu.header['CDELT2'] = cdelt[1]
   ...:     pu.header['CRPIX1'] = crpix[0]
   ...:     pu.header['CRPIX2'] = crpix[1]
   ...:     pu.header['CRVAL1'] = crval[0]
   ...:     pu.header['CRVAL2'] = crval[1]
   ...:     pu.header['PS1_2'] = 'xi'
   ...:     pu.header['PS2_2'] = 'yi'
   ...:     for k, v in pc.items():
   ...:         pu.header[k] = v
   ...:     hdulist = fits.HDUList([pu, bt])
   ...:     return hdulist
   ...:
 
In [3]: for k in range(1000):
   ...:     h = create_hdulist(nx, ny, crpix, crval, cdelt, pc)
   ...:     w = wcs.WCS(h[0].header, h)
   ...:     w.wcs_world2pix(w.wcs_pix2world([[1, 2]], 1), 1)
   ...:
 
In [4]: w.wcs_world2pix(w.wcs_pix2world([[1, 2]], 1), 1)
Out[4]: array([[1.00000001, 2.00000002]])
 
In [5]: exit
(astro) PS C:\Users\mcara\code>
```

mcara:
the same code placed into the `test_` module segfaults.

mcara:
@MSeifert04 I realized why I was "more lucky" than you in running this code successfully on Windows: I was running a single test once and then I would repeat the test many times. For example, with `pytest --count=1000` it would successfully run 1000 times but it would crash in the end. When wcs/hdulist are created multiple times within the same test, it fails much more often.

So, I almost agreed with your hypothesis about `BinTableHDU`. However, to theck this theory I have modified my test and C code in a way that: 1) eliminates/bypasses passing data (for the `WCS-TAB` creation purpose) from C<->Python; 2) feeds known coordinate and indices arrays to `wtbarr` structures without creating a `BinTableHDU`.

What I found is that the new test still crashes on Windows. So, this eliminates `BinTableHDU` as a root cause, of this crash, IMO. To repeat my test, replace `_update_wtbarr_from_hdulist()` in `wcslib_wrap.c` with:
```C
int _update_wtbarr_from_hdulist(PyObject *hdulist, struct wtbarr *wtb) {
  PyArrayObject *arrayp=NULL;
  PyObject *result=NULL;
  int i, naxis, nelem, naxes[NPY_MAXDIMS];
  npy_intp *npy_naxes;
  npy_double *appayp_data;

  double arr1[] = { 0.0, -0.5, 0.26, -0.5, 0.8, -0.5, 1.0, -0.5, 0.0, 0.5, 0.26, 0.5, 0.8, 0.5, 1.0, 0.5 };
  int naxes1[] = { 2, 4, 2 };
  double arr2[] = { 1.0, 51.0, 51.0, 151.0 };
  int naxes2[] = { 4 };
  double arr3[] = { 1.0, 201.0 };
  int naxes3[] = { 2 };

  printf("*wtb->ttype: %s\n", wtb->ttype);
  if (*wtb->ttype == 'w') {
      for (i = 0; i < 3; i++) wtb->dimlen[i] = naxes1[i];
      *(wtb->arrayp) = calloc((size_t)16, sizeof(double));
      for (i = 0; i < 16; i++) (*wtb->arrayp)[i] = arr1[i];
  }
  else if (*wtb->ttype == 'x') {
      wtb->dimlen[0] = naxes2[0];
      *(wtb->arrayp) = calloc((size_t)4, sizeof(double));
      for (i = 0; i < 4; i++) (*wtb->arrayp)[i] = arr2[i];
  }
  else {
      wtb->dimlen[0] = naxes3[0];
      *(wtb->arrayp) = calloc((size_t)2, sizeof(double));
      for (i = 0; i < 2; i++) (*wtb->arrayp)[i] = arr3[i];
  }
  return 1;
}
```
and run the following test: 
[test3.txt](https://github.com/astropy/astropy/files/3925066/test3.txt)

This procedure essentially leaves only the following change to the code in this PR that implements support for the `WCS-TAB`: https://github.com/astropy/astropy/blob/e4bb657f262f2e206e90d0a00f2a9c34403ece42/astropy/wcs/src/wcslib_wrap.c#L574-L582 

I do not see anything in this code (or modified `_update_wtbarr_from_hdulist()` shown above) that could cause memory corruption.

MSeifert04:
@mcara Could you check if there are any out-of-bound read/writes on the `p0` member of `tabprm`? I think it shouldn't matter on what OS/compiler.

It seems like MSVC detects out-of-bound read/writes on an array but only reports these (by segfault) when the array is freed. The problem (or at least one) seems to be `free(tab->p0)`. However I haven't managed to determine if there's any such access and where it is happening and I don't have much spare time left this year (work + moving to a new place + christmas will dominate the next few weeks). And you probably have a lot more expertise with the internals of `wcslib`.

mcara:
@MSeifert04 Thank you for detective work! I will look into this. I just got a Windows laptop and need to set it up.

MSeifert04:
I think I found the problem: 

https://github.com/astropy/astropy/blob/master/cextern/wcslib/C/tab.c#L1394

For debugging purposes I added a member `p0size` to the `tabprm` struct and got the following hit:

Unbenannt

mcara:
@MSeifert04 This is really great! However, while initially I thought that you found the root cause of the issue, I am now afraid this is just one of the issues. Here is a quick modification of my [previous test3](https://github.com/astropy/astropy/pull/9641#issuecomment-561960552):

1. get [test3.txt](https://github.com/astropy/astropy/files/3925066/test3.txt) and rename it to `test3.py`.
2. in `tab.c` add `printf` and `fflush` statements before https://github.com/astropy/astropy/blob/d38a925997ccfe6226eaa91984944d484258dc53/cextern/wcslib/C/tab.c#L1394
3. Run `test3.py` several times.

On Mac OSX I get that this specific branch in `tabedge()` (the one with `p0[m+1]`) get executed only on some occasions. Sometimes it does not get executed at all and sometimes it can get called 3-5 times (while computing `wcs_world2pix`). You will need to repeat this test many (10-20) times but you will see that sometimes it does not get executed for the _first_ (`k=0`) run of the test (see the last `for` loop in `test3.py`). Initially I thought that maybe the first run corrupts the memory and that for subsequent runs `p0[m+1]` gets executed "variable" (random) number of times. However, the fact that sometimes run `k=0` does not execute the line `p0[m+1]`, I think it indicates that the corruption occurs in another place (as well).

On a Windows machine I re-run the steps above, but I also commented out the last lines in `test_2d_spacial_tab_roundtrip`:
```python
def test_2d_spacial_tab_roundtrip(k):
    # ....................
    # xy_roundtripped = tab_wcs_2di.wcs_world2pix(rd, 1)
    # m = np.logical_and(*(np.isfinite(xy_roundtripped).T))
    # assert np.allclose(xy[m], xy_roundtripped[m], rtol=0, atol=1e-7)
    print("test passed\n")
```
This essentially eliminates the calls to `tabedge()` in `tab.c`. Although at the beginning the tests were passing (all 10 runs); they started failing eventually (re-run `test3.py` several times). This, IMO, shows that `p0[m+1]` is not the only problem in WCSLIB.

mcara:
It looks like there is another memory access violation somewhere. Even with bug fixes in WCSLIB 7.1  unit tests crash at random places (tests) following `test_tab` test on Windows.

mcara:
@MSeifert04 @nden I believe now unit tests are all passing. The reported failures are unrelated to this PR. I found another bug in WCSLIB due to which an element with a very large index (`(unsigned int) -1`) was being accessed. I implemented fix in `cextern` until we get a new release of WCSLIB.

I will report this bug.

MSeifert04:
@mcara Great 👍 I'm not sure what our policy is regarding extern libraries but I think we probably need to apply it as "patch" instead of fixing it in the extern code. Just to make it transparent what we applied and what is an extern library.

nden:
I don't think we have a policy about that but in the past we've applied "patches" to a library in cextern. And by "patch" I mean make the change in the same PR which updates wcslib. @MSeifert04 do you mean apply the change to wcslib in a separate PR or something more sophisticated?

MSeifert04:
> do you mean apply the change to wcslib in a separate PR or something more sophisticated?

I am actually not sure how we should patch extern libraries. But I see that my comment was misleading (probably shouldn't comment when tired 😓 ). What I meant was that the commit that fixed wcslib also changed our code-base which I think is problematic, I think the fix for an extern code-base should be restricted to the extern code-base (so that we can easily revert it with `git revert` or similar) when the new version is out. 

If the fix should be in a separate PR or how we should apply it (and what would be possible) is something I don't know.

mcara:
Now WCSLIB fix is in a separate commit.

nden:
@mcara Actually do you mind applying it in a separate PR (to be merged before this one)? That way this PR  contains only the `-TAB` implementation in astropy and the wcslib change is easy to find.

mcara:
@nden Allright. Will do.

nden:
This looks ready to be merged. One last request @mcara, could you add an entry in the "what's new in 4.1" file? There was a recent decision to add entries there with the PR and not before a release.

MSeifert04:
Yep, I think this looks okay now.

nden:
Thanks @mcara, @MSeifert04 and everyone else who contributed! 
@MSeifert04 approved in a comment so I'm merging it.

mcara:
Thank you @nden and @MSeifert04 for constructive comments and bug finding work!

mcara:
Correction: my explanation in https://github.com/astropy/astropy/pull/9641#issuecomment-571439443 is invalid and is due to progressive atrophication of C skills. `map[-1]` [simply points to the "previous" address to `map`](https://stackoverflow.com/a/50536478). That value somehow was inaccessible on Windows but not on Linux/MacOS.

PR #9644, opened 2019-11-21
Prevent NaNs from emitting a warning in Longitude

### LINKED ISSUE #9643: Longitude issues a warning when input contains NaN
### Description
Supplying `astropy.coordinates.angle.Longitude` with a NaN value raises a warning, but I don't think it should since having a NaN value is perfectly reasonable. (this is very similar to #9597 that I opened for `Distance`)

### Expected behavior
No warning.

### Actual behavior
```python
/Users/dstansby/miniconda3/lib/python3.7/site-packages/astropy/coordinates/angles.py:642: RuntimeWarning: invalid value encountered in less
  if np.any(self_angle < wrap_angle_floor) or np.any(self_angle >= wrap_angle):
/Users/dstansby/miniconda3/lib/python3.7/site-packages/astropy/coordinates/angles.py:642: RuntimeWarning: invalid value encountered in greater_equal
  if np.any(self_angle < wrap_angle_floor) or np.any(self_angle >= wrap_angle):
```

### Steps to Reproduce

```python
from astropy.coordinates.angles import Longitude
import astropy.units as u
import numpy as np

l = Longitude([0, np.nan, 1] * u.deg)
```

### System Details

Darwin-19.0.0-x86_64-i386-64bit
Python 3.7.5 (default, Oct 25 2019, 10:52:18) 
Numpy 1.17.3
Scipy 1.3.1
astropy 3.2.3

dstansby:
Fixes #9643

### DIFF
--- a/astropy/coordinates/angles.py
+++ b/astropy/coordinates/angles.py
@@ -632,10 +632,10 @@ class Longitude(Angle):
         wrap_angle_floor = wrap_angle - a360
         self_angle = self.value
         # Do the wrapping, but only if any angles need to be wrapped
-        if np.any(self_angle < wrap_angle_floor) or np.any(self_angle >= wrap_angle):
-            wrapped = np.mod(self_angle - wrap_angle, a360) + wrap_angle_floor
-            value = u.Quantity(wrapped, self.unit)
-            super().__setitem__((), value)
+        with np.errstate(invalid='ignore'):
+            wraps = (self_angle - wrap_angle_floor) // a360
+        if np.any(wraps != 0):
+            self -= (wraps * a360) << self.unit
 
     @property
     def wrap_angle(self):
--- a/astropy/coordinates/tests/test_angles.py
+++ b/astropy/coordinates/tests/test_angles.py
@@ -934,3 +934,8 @@ def test_angle_with_cds_units_enabled():
         Angle('5d')
     del _AngleParser._parser
     Angle('5d')
+
+
+def test_longitude_nan():
+    # Check that passing a NaN to Longitude doesn't raise a warning
+    Longitude([0, np.nan, 1] * u.deg)


pllim:
Please fix test failure. Thanks!

dstansby:
The failure is
```
E: Package 'graphviz' has no installation candidate
E: Unable to locate package language-pack-de
```
Which I don't think is my fault! Please could someone re-run the tests?

astropy/coordinates/angles.py
```diff
@@ -632,7 +632,11 @@ def _wrap_internal(self):
         wrap_angle_floor = wrap_angle - a360
         self_angle = self.value
         # Do the wrapping, but only if any angles need to be wrapped
-        if np.any(self_angle < wrap_angle_floor) or np.any(self_angle >= wrap_angle):
+        # Make sure NaNs don't emit a warning
+        with np.errstate(invalid='ignore'):
```
mhvk: An alternative might be to avoid the comparison (which is really rather ugly!). How about the following (slightly faster too! and a bit clearer, I think):
```
         wrap_angle_floor = wrap_angle - a360
         self_angle = self.value
         # Do the wrapping, but only if any angles need to be wrapped
-        if np.any(self_angle < wrap_angle_floor) or np.any(self_angle >= wrap_angle):
-            wrapped = np.mod(self_angle - wrap_angle, a360) + wrap_angle_floor
-            value = u.Quantity(wrapped, self.unit)
-            super().__setitem__((), value)
+        wraps = (self_angle - wrap_angle_floor) // a360
+        if np.any(wraps != 0):
+            self -= (wraps * a360) << self.unit
```

mhvk: p.s. I'm actually not all that sure that it is really necessary to convert quantities to arrays...  But out of scope here!

mhvk:
This looks good now! thanks!

astropy/coordinates/angles.py
```diff
@@ -632,10 +632,10 @@ def _wrap_internal(self):
         wrap_angle_floor = wrap_angle - a360
         self_angle = self.value
         # Do the wrapping, but only if any angles need to be wrapped
-        if np.any(self_angle < wrap_angle_floor) or np.any(self_angle >= wrap_angle):
-            wrapped = np.mod(self_angle - wrap_angle, a360) + wrap_angle_floor
-            value = u.Quantity(wrapped, self.unit)
-            super().__setitem__((), value)
+        with np.errstate(invalid='ignore'):
```
mhvk: Huh? Do you still need this? That's a pity, though definitely the right way to go if it is needed, so approval remains.

dstansby: Yes, a couple of the tests were failing without it (see https://travis-ci.org/astropy/astropy/builds/617419717?utm_source=github_status&utm_medium=notification for the failing tests before I force pushed the branch)

mhvk: OK, thanks. It seems there is a numpy version dependence. Anyway, now it is all good, so will merge. Thanks!

dstansby:
Thanks!

PR #9666, opened 2019-11-24
Update linux instructions for installing from source

mhvk:
While trying to add the `s390` distribution to travis in #9663, I noticed that quite a bit more was needed to install astropy than just `python3-dev` - which is what is listed in the documentation. So, an update.

cc @olebole, @sergiopasra - do these packages make sense? For debian, there is a bit of ground truth in #9663, but for redhat, I made the list just by searching the package manager...

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -202,7 +202,7 @@ Prerequisites for Linux
 
 On Linux, using the package manager for your distribution will usually be the
 easiest route to making sure you have the prerequisites to build ``astropy``. In
-order to build from source, you will need the Python and numpy development
+order to build from source, you will need the Python and Numpy development
 package for your Linux distribution, as well as a number of helper packages.
 
 For Debian/Ubuntu::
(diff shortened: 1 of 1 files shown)


olebole:
Quickl response (I can review only tomorrow or so): The list of packages that we install to build (and test) astropy is [here](https://packages.debian.org/source/unstable/astropy).

docs/install.rst
```diff
@@ -202,16 +202,16 @@ Prerequisites for Linux
 
 On Linux, using the package manager for your distribution will usually be the
 easiest route to making sure you have the prerequisites to build ``astropy``. In
-order to build from source, you will need the Python development package for
-your Linux distribution.
+order to build from source, you will need the Python and numpy development
```
pllim:
```suggestion
order to build from source, you will need the Python and Numpy development
```

mhvk:
OK, I added a little disclaimer following @olebole's comment and fixed the spelling of `Numpy`.

pllim:
@mhvk , if you are not waiting for @sergiopasra , feel free to merge. I approved it.

mhvk:
My sense is that this is close enough, as the main goal was simply to make clear that there are quite a few packages one needs to install. So I'll indeed go ahead and merge; if for Fedora, it is not quite right, we can fix later.

sergiopasra:
I didn't have time for checking this until now. For Fedora there are a few packages missing, like h5py or PyYAML. If you install python3-scipy, it will install also python3-numpy. What should I do now?

mhvk:
@sergiopasra - those packages are missing from the debian list as well: this is meant to be the minimum list that you need to build `astropy` and test its basic workings. The others are all optional dependencies for extra functionality.

So, the question is whether one needs anything more to build astropy from source on Fedora.

sergiopasra:
@mhvk In that case the list of packages is correct

mhvk:
Great, thanks!

PR #9667, opened 2019-11-24
Add ability to pass kwargs to writer in jsviewer writing to disk

keflavich:
The jsviewer writer doesn't pass keyword arguments onto the table writer
class, so you can't specify formatters and any other keyword arguments.

This is my attempt to get a nicely-formatted javascript table out of a
table write, but so far the attempted fix doesn't seem to work... I
might be barking up the wrong tree, so if this proves fruitless I'll
close the PR and open an issue.

### DIFF
--- a/astropy/table/jsviewer.py
+++ b/astropy/table/jsviewer.py
@@ -172,7 +172,7 @@ class JSViewer:
 
 def write_table_jsviewer(table, filename, table_id=None, max_lines=5000,
                          table_class="display compact", jskwargs=None,
-                         css=DEFAULT_CSS, htmldict=None):
+                         css=DEFAULT_CSS, htmldict=None, **kwargs):
     if table_id is None:
         table_id = 'table{id}'.format(id=id(table))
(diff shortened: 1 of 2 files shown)


taldcroft:
@keflavich - :+1: on this in general.  I recommend trying to push this forward for 4.1.

For instance I was just playing around with this and ran into a related issue that this would solve.  I wrote to an output file `junk.html`, but then the second time I got a warning that overwriting is deprecated and recommendation to supply `overwrite=True`.  But this is not an allowed arg for  `write_table_jsviewer` so that gives an exception.  Clearly pass-through kwargs are a good thing.

About formatting your JS table, for something that works right now, if you set the column `info.format` attributes that should do the trick. E.g. `t['a'].info.format = 'a{}'`  Of course that changes the table itself which may not be what you wanted, so some kind of local context manager would be needed.

keflavich:
This is still a cool feature I want but I have to leave it alone for now.

PR #9669, opened 2019-11-24
Fix bug in locking mechanism

### LINKED ISSUE #9642: Download cache gets confused in parallel (using shell bg / GNU parallel / SLURM / etc)
### Description
I'm running astropy on an offline computer.  I need the added accuracy that the de430 ephemerides provide, and so at the beginning of the utility, I call
```python
astropy.coordinates.solar_system_ephemeris.set('file://path/to/local/copy/de430.bsp')
```

Sometimes, I'm in a hurry to get things done, and so I'll fire off multiple copies of my utility from the shell (use moar cores!), and the processes will often (but not always) will decide that the file is out of date and start copying it ("downloading") from the path above into /tmp, start fighting over the lock for the cache directory, then barf, etc.  I end up with most of my runs failed, a /tmp directory full of abandoned files, and a bunch of frustration.

I have discovered that if I set XDG_CACHE_HOME to a directory that is mounted locally (/tmp is my only option) instead of through NFS, I can no longer trigger the bug.  I assume that the lock is assuming file existence is atomic, near as I can tell NFS only guarantees directory creation to be an atomic operation.

### Expected behavior
I expected the file to be "downloaded" at most once and everyone else to wait their turn.
I would prefer that there were a way to disable downloading/caching entirely (perhaps similar to how the IERS util in astropy allows pointing at a local finals2000a.all)

### Actual behavior
```bash
   File "/usr/local/lib/python2.7/dist-packages/astropy/coordinates/solar_system.py", line 175, in _get_kernel
    return SPK.open(download_file(value, cache=True))
  File "/usr/local/lib/python2.7/dist-packages/astropy/utils/data.py", line 1122, in download_file                   _acquire_download_cache_lock()
  File "/usr/local/lib/python2.7/dist-packages/astropy/utils/data.py", line 1405, in _acquire_download_cache_lock
    raise RuntimeError(msg.format(lockdir))
RuntimeError: Unable to acquire lock for cache directory (/home/mcbeth/Sites/astropy/download/py2/lock exists). You may need to delete the lock if the python interpreter wasn't shut down properly.
```

### Steps to Reproduce
```python
#!/usr/bin/env python

import astropy.coordinates as coords
from astropy.time import Time

coords.solar_system_ephemeris.set('file:///home/mcbeth/astropy_caching/de430.bsp')
print(coords.get_sun(Time.now()))
```

```bash
for x in {0..100}; do
  ./caching_test.py&
done
```

```bash
mkdir /tmp/astropy
export XDG_CACHE_HOME=/tmp
for x in {0..100}; do
  ./caching_test.py&
done
```

### System Details

Linux-5.0.0-32-generic-x86_64-with-Ubuntu-19.10-eoan
Python 3.7.5rc1 (default, Oct  8 2019, 16:47:45)
[GCC 9.2.1 20191008]
Numpy 1.16.2
Scipy 1.2.2
astropy 3.2.1

aarchiba:
Should fix #9642

### Description

This pull request is to address a bug where if the cache lock could not be acquired the failure path would incorrectly release the lock.

Fixes #9642

### DIFF
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -25,6 +25,7 @@ from tempfile import NamedTemporaryFile, TemporaryDirectory
 
 import py.path
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
 
 from astropy.utils import data
 from astropy.config import paths
@@ -34,6 +35,7 @@ from astropy.utils.data import (
     CacheDamaged,
     conf,
     _cache,
+    _cache_lock,
     compute_hash,
     download_file,
     cache_contents,
@@ -191,23 +193,30 @@ def fake_readonly_cache(tmpdir, valid_urls, monkeypatch):
 
 
 _shelve_possible_backends = ["dbm.dumb", "dbm.ndbm", "dbm.gnu"]
+for n in _shelve_possible_backends:
+    if n not in dbm._modules:
+        try:
+            dbm._modules[n] = __import__(n, fromlist=['open'])
+        except ImportError:
+            pass
 
 
 @contextlib.contextmanager
 def shelve_backend(name):
     """Ensure that shelve has access only to one backend."""
-    defaultmod = dbm._defaultmod
-    modules = dbm._modules
-    try:
+    # WARNING: there is something weird and fragile about this.
+    # Using it can cause weird order-dependent test failures in
+    # the second shelve_backend test if the two meet certain
+    # mysterious criteria.
+    with MonkeyPatch().context() as mp:
         try:
-            dbm._defaultmod = __import__(name, fromlist=['open'])
-        except ImportError:
-            pytest.skip(f"module {name} not available")
-        dbm._modules = {name: dbm._defaultmod}
+            m = dbm._modules[name]
+        except KeyError:
+            pytest.skip(f"Backend {name} not available")
+
+        mp.setattr(dbm, "_defaultmod", m)
+        mp.setattr(dbm, "_modules", {name: m})
         yield
-    finally:
-        dbm._defaultmod = defaultmod
-        dbm._modules = modules
 
 
 @pytest.mark.remote_data(source="astropy")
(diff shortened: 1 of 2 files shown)


aarchiba:
There is an unrelated problem in how I try to run different dbm backends that I have not yet tracked down. It's going to be ugly because I end up having to use monkeypatching to enforce use of a particular backend. There must be a better way.

astropy/utils/tests/test_data.py
```diff
@@ -25,6 +25,7 @@
 
 import py.path
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
```
pllim: Oooo is there a secret `_pytest` module? Should we be using a "hidden" module?

aarchiba: Probably not. This is part of the Work-In-Progress-ness - I was hoping that using the "official" monkeypatching tools might work more reliably than my own hacks, but they don't. And monkeypatching is officially provided but only in the form of a test fixture, which I don't know how to access from something that isn't a test function. So I had to pull in the secret module to get at the same tools the test fixture uses. Since it doesn't help, if I still need this function once I've fixed the problem I'll do my own manual monkeypatching.

aarchiba:
Ready for review

astropy/utils/tests/test_data.py
```diff
@@ -1827,17 +1857,44 @@ def test_lock_behaviour_if_directory_disappears(b, temp_cache):
...
+            with pytest.raises(RuntimeError):
+                with _cache_lock("astropy", need_write=True):
+                    pass
+            # Ensure that the cache doesn't accidentally get unlocked!
+            with pytest.raises(RuntimeError):
+                with _cache_lock("astropy", need_write=True):
+                    pass
+            # Download
+            with pytest.raises(RuntimeError):
+                is_url_in_cache(u), "if locked, act like cache empty"
```
pllim: Do you mean the trailing string as a comment? I am not sure what the comma is used for when there is no `assert`.

```suggestion
                is_url_in_cache(u)  # if locked, act like cache empty
```

pllim:
Depending on whether this or #9703 gets merged first, the other will need a rebase to rerun the CI before merge, just in case.

astropy/utils/tests/test_data.py
aarchiba: Oops, sorry!

aarchiba:
This patch actually includes test cases to evaluate what happens when the cache lock cannot be obtained, so you should be able to test it just by running the test suite.

There is one design decision I made here that one could argue with: if the cache lock cannot be obtained, raise an exception and stop. The other design option would be to proceed as if no cache were available. I chose the former because I thought it would be much easier to tell what was going on, and because the failure mode of hammering the server from all your parallel threads is not a good one.

Issue #9665, opened 2019-11-24
Read units row from ascii files


keflavich:
Related to https://github.com/astropy/astropy/issues/4743 and https://github.com/astropy/astropy/issues/756, and loosely related to https://github.com/astropy/astropy/issues/9639 (which provides the workaround): we should have a mechanism for reading the units from the second (or n'th) header row of an ascii table.

Presently the only (?) human-readable format supported by astropy.tables is the IPAC format, which includes additional information and a very specific set of formatting, but it should be straightforward to include a units-parsing component to the generic ascii reader.

(if we have this feature somewhere, I was unable to find it by reading the docs...)

taldcroft:
I would argue that ECSV is human-readable.  What ECSV is NOT is human-*writeable*.
```
# %ECSV 0.9
# ---
# datatype:
# - {name: a, datatype: int64}
# - {name: b, unit: m, datatype: float64}
# - {name: c, datatype: string}
# schema: astropy-2.0
a b c
1 1.0 c
2 2.0 d
```
My question related to these requests to read such files is whether the data files with units in the second line are showing up "in the wild" (from collaborators or whatnot), or if you want to purposely *write* such a file in a new, implicitly-defined format.

keflavich:
I disagree: ecsv is not very human-readable.  It's not *entirely* unreadable, but this is difficult to parse and doesn't generalize well as more fields are added (while fixed width does):
![image](https://user-images.githubusercontent.com/143715/69546207-4f9aa280-0f61-11ea-8d74-f13189689f30.png)


Re: questions:
(1) Yes, these show in the wild all the time
(2) Yes, I'd like to write to this format.  Generally I want to use this for small tables, not gigantic ones, but it's useful

pllim:
I am hesitant to support random formats on a whim. It is a slippery slope. We already have options for you to write your own parser by subclassing `Table`.

keflavich:
I am not suggesting creating or supporting new formats, really.  I just want to be able to treat a specified header row as 'units' in the general case.  Sure, my use case is mostly for human-readable formats, but the application is broad.

pllim:
I have a lot of trauma trying to parse human-written units in the past. For example, "angstroms" vs "angstrom" or "erg" vs "ergs" or "ct" vs "cts" vs "count" vs "counts". And when you get to flux or magnitude units, even worse. Where do we draw the line? Is it human reable when the unit string in your header is very very long compared to the actual data for that column?

[...]

PR #9671, opened 2019-11-25
Set units and descriptions in Table init

### LINKED ISSUE #9639: Allow specifying a list of column units to io.ascii.read()
I'm using `io.ascii.read` to read in a text table that has no column names or units, but I want to (ideally) get back a `QTable` with column names and units that I specify. I know I can use `names=[...]` to set the column names, but it would be nice if there was an equivalent for units as well. For example, if I have an ascii table that looks like this:
```
-37.417 0.011
34.846 0.010
42.567 0.010
```
in some file, it would be nice if I could do something like:
```python
ascii.read(filename, names=['a', 'b'], units=[u.one, u.kpc], cls=QTable)
```

Apologies if this functionality already exists! I couldn't find this in the `io.ascii` docs.

taldcroft:
This allows explicitly setting both the `units` and `descriptions` of specified table columns from the `Table.__init__()` and `Table.read()` methods.

Closes #9639 

It provides some mitigation for #9665 (read units row from ASCII table) by making it easier to do that:
```
    text = ['a,b',
            ',s',
            '1,2',
            '3,4']
    units = Table.read(text, format='ascii', data_start=1, data_end=2)[0]
    t = Table.read(text, format='ascii', data_start=2, units=units)
```

### DIFF
--- a/astropy/table/connect.py
+++ b/astropy/table/connect.py
@@ -33,6 +33,10 @@ class TableRead(registry.UnifiedReadWrite):
         first argument is typically the input filename.
     format : str
         File format specifier.
+    units : list, dict, optional
+        List or dict of units to apply to columns
+    descriptions : list, dict, optional
+        List or dict of descriptions to apply to columns
     **kwargs : dict, optional
         Keyword arguments passed through to data reader.
 
@@ -49,6 +53,9 @@ class TableRead(registry.UnifiedReadWrite):
 
     def __call__(self, *args, **kwargs):
         cls = self._cls
+        units = kwargs.pop('units', None)
+        descriptions = kwargs.pop('descriptions', None)
+
         out = registry.read(cls, *args, **kwargs)
 
         # For some readers (e.g., ascii.ecsv), the returned `out` class is not
@@ -62,6 +69,10 @@ class TableRead(registry.UnifiedReadWrite):
             except Exception:
                 raise TypeError('could not convert reader output to {} '
                                 'class.'.format(cls.__name__))
+
+        out._set_column_attribute('unit', units)
+        out._set_column_attribute('description', descriptions)
+
         return out
 
 
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -374,6 +374,10 @@ class Table:
         Row-oriented data for table instead of ``data`` argument.
     copy_indices : bool, optional
         Copy any indices in the input data. Default is True.
+    units : list, dict, optional
+        List or dict of units to apply to columns
+    descriptions : list, dict, optional
+        List or dict of descriptions to apply to columns
     **kwargs : dict, optional
         Additional keyword args when converting table-like object.
     """
@@ -453,6 +457,7 @@ class Table:
 
     def __init__(self, data=None, masked=False, names=None, dtype=None,
                  meta=None, copy=True, rows=None, copy_indices=True,
+                 units=None, descriptions=None,
                  **kwargs):
 
         # Set up a placeholder empty table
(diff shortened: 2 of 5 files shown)


astropy/table/table.py
```diff
@@ -618,6 +623,28 @@ def __init__(self, data=None, masked=False, names=None, dtype=None,
...
+    def _set_column_attribute(self, attr, values):
+        """Set ``attr`` for columns to ``values``, which can be either a dict (keyed by column
+        name) or a dict of name: value pairs.
+        """
+        if not values:
+            return
+
+        if isinstance(values, dict):
+            for name, value in values.items():
+                if name not in self.columns:
```
mhvk: I'd move this out of the loop and change to check that all names are column names - that way one can never change only some of the columns (since these may not be copies), i.e.,
```
if any(name not in self.columns):
    raise ...
```

taldcroft: Well, if the exception gets raised then there is no table returned nothing is partially done.  Doing it outside the loop makes it harder to know which one is bad.

taldcroft:
I added kwargs to `Table.read` as well.  In theory this is ready for final review.

astropy/table/table.py
mhvk: Do we always make copies of the columns? (not the data, just the instance) If so, indeed, no harm done.

astropy/table/table.py
```diff
@@ -374,6 +374,10 @@ class Table:
         Row-oriented data for table instead of ``data`` argument.
     copy_indices : bool, optional
         Copy any indices in the input data. Default is True.
+    units : list, dict, optional
+        List or dict of units to apply to columns
```
mhvk: Missing period at end of line (also below and in the copy in `connect.py`

docs/table/construct_table.rst
```diff
@@ -665,6 +679,43 @@ care because any modifications to the new Table data will also be seen in the
...
+units
+-----
+
+This allows setting the unit for one or more columns at the time of creating the
+table.  The input can be either a list of unit values corresponding to each of
+the columns in the table (using ``None`` for no unit), or a ``dict`` that
+provides the unit for specified column names.  For example::
+
+  >>> from astropy.table import QTable
+  >>> dat = [[1, 2] * u.m, ['hello', 'world']]
```
mhvk: In this example, probably best not to give the unit? At least, I would think that leads to a failure (cannot set a Quantity unit if it has one already).

astropy/table/table.py
taldcroft: Yes, it always gets a copy of a column.  Note that it can get a reference to a Quantity, but in this case you can't change the unit anyway so I think we're OK.

docs/table/construct_table.rst
taldcroft: Done.

astropy/table/table.py
taldcroft: Done.

astropy/table/table.py
mhvk: OK, let's stick with what you have, then!

astropy/table/connect.py
```diff
@@ -33,6 +33,10 @@ class TableRead(registry.UnifiedReadWrite):
         first argument is typically the input filename.
     format : str
         File format specifier.
+    units : list, dict, optional
+        List or dict of units to apply to columns
```
mhvk: Missing the periods at the end here too....

astropy/table/table.py
```diff
@@ -618,6 +623,39 @@ def __init__(self, data=None, masked=False, names=None, dtype=None,
...
+
+    def _set_column_attribute(self, attr, values):
+        """Set ``attr`` for columns to ``values``, which can be either a dict (keyed by column
+        name) or a dict of name: value pairs.  This is used for handling the ``units`` and
+        ``descriptions`` kwargs to ``__init__``.
+        """
+        if not values:
+            return
+
+        if isinstance(values, Row):
```
mhvk: For future, perhaps we should have `Rows.items()` so there isn't a need to special-case it.

astropy/table/table.py
```diff
@@ -618,6 +623,39 @@ def __init__(self, data=None, masked=False, names=None, dtype=None,
...
+            # For a Row object transform to an equivalent dict.
+            values = {name: values[name] for name in values.colnames}
+
+        if not isinstance(values, dict):
```
mhvk: Might even register `Row` as `collections.abc.Mapping`, and then do the `instance` on that (I'm a bit puzzled it is considered a `Sequence` now, since one cannot index with the column number anyway; similarly, `Row.__iter__` is weird.

But really out of scope here! See #9687

pllim:
@adrn , now you can do it with dev astropy!

PR #9672, opened 2019-11-25
Fixes for some of the problems uncovered on ARM bigendian/quad precision

mhvk:
@bsipocz - I found it easier to just separate out the fixes. Note that these may not be all; in particular, it doesn't get to #9657 but does fix the other problems in #9661.

### DIFF
--- a/astropy/time/tests/test_basic.py
+++ b/astropy/time/tests/test_basic.py
@@ -867,7 +867,10 @@ class TestNumericalSubFormat:
         assert t.to_value('mjd', subfmt='str') == '54321.000000000001'
         assert t.to_value('mjd', 'bytes') == b'54321.000000000001'
         expected_long = np.longdouble(54321.) + np.longdouble(1e-12)
-        assert t.to_value('mjd', subfmt='long') == expected_long
+        # Check we're the same to within the double holding jd2
+        # (which is less precise than longdouble on arm64).
+        assert np.allclose(t.to_value('mjd', subfmt='long'),
+                           expected_long, rtol=0, atol=2.**-52)
         t.out_subfmt = 'str'
         assert t.value == '54321.000000000001'
         assert t.to_value('mjd') == 54321.  # Lost precision!
@@ -875,9 +878,10 @@ class TestNumericalSubFormat:
         assert t.to_value('mjd', subfmt='bytes') == b'54321.000000000001'
         assert t.to_value('mjd', subfmt='float') == 54321.  # Lost precision!
         t.out_subfmt = 'long'
-        assert t.value == expected_long
-        assert t.to_value('mjd', subfmt=None) == expected_long
-        assert t.mjd == expected_long
+        assert np.allclose(t.value, expected_long, rtol=0., atol=2.**-52)
+        assert np.allclose(t.to_value('mjd', subfmt=None), expected_long,
+                           rtol=0., atol=2.**-52)
+        assert np.allclose(t.mjd, expected_long, rtol=0., atol=2.**-52)
         assert t.to_value('mjd', subfmt='str') == '54321.000000000001'
         assert t.to_value('mjd', subfmt='float') == 54321.  # Lost precision!
 
@@ -885,19 +889,22 @@ class TestNumericalSubFormat:
                         reason="long double is the same as float")
     def test_explicit_longdouble(self):
         i = 54321
-        f = 2.**(-np.finfo(np.longdouble).nmant) * 65536
+        # Create a different long double (which will give a different jd2
+        # even when long doubles are more precise than Time, as on arm64).
+        f = max(2.**(-np.finfo(np.longdouble).nmant) * 65536, 2.**(-52))
         mjd_long = np.longdouble(i) + np.longdouble(f)
         assert mjd_long != i, "longdouble failure!"
         t = Time(mjd_long, format='mjd')
         expected = Time(i, f, format='mjd')
-        assert t == expected
+        assert abs(t - expected) <= 20.*u.ps
         t_float = Time(i+f, format='mjd')
         assert t_float == Time(i, format='mjd')
         assert t_float != t
         assert t.value == 54321.  # Lost precision!
-        assert t.to_value('mjd', subfmt='long') == mjd_long
+        assert np.allclose(t.to_value('mjd', subfmt='long'), mjd_long,
+                           rtol=0., atol=2.**-52)
         t2 = Time(mjd_long, format='mjd', out_subfmt='long')
-        assert t2.value == mjd_long
+        assert np.allclose(t2.value, mjd_long, rtol=0., atol=2.**-52)
 
     @pytest.mark.skipif(np.finfo(np.longdouble).eps >= np.finfo(float).eps,
                         reason="long double is the same as float")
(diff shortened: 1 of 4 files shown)


mhvk:
cc @aarchiba  for the time tests.

astropy/units/tests/test_quantity_ufuncs.py
```diff
@@ -711,20 +711,16 @@ def test_onearg_test_ufuncs(self, ufunc):
         assert out.dtype == bool
         assert np.all(out == ufunc(q.value))
 
+    # Ignore RuntimeWarning raised on Windows and s390.
+    @pytest.mark.filterwarnings('ignore:.*invalid value encountered in sign')
```
pllim: I thought we agreed in #7928 to not blindly ignore this warning as such so we'll know when Numpy fixes its bug upstream? Isn't this a step backward?

astropy/time/tests/test_basic.py
```diff
@@ -867,51 +867,62 @@ def test_explicit_example(self):
         assert t.to_value('mjd', subfmt='str') == '54321.000000000001'
         assert t.to_value('mjd', 'bytes') == b'54321.000000000001'
         expected_long = np.longdouble(54321.) + np.longdouble(1e-12)
-        assert t.to_value('mjd', subfmt='long') == expected_long
+        # Check we're the same to within the double holding jd2
+        # (which is less precise than longdouble on arm64).
+        assert np.allclose(t.to_value('mjd', subfmt='long'),
+                           expected_long, rtol=0, atol=2.**-52)
```
pllim: I haven't been involved in the other discussions, so `2**-52` looks like a magic number to me. And given that it appears so many times, maybe it should be assigned to a global variable for reuse and a comment to explain why this value?

astropy/units/tests/test_quantity_ufuncs.py
mhvk: Yes, that was when it was windows only... now I just don't know anymore where it could fail, so decided we might as well ignore it, as it really isn't our problem. (But I admit to being too lazy to check how to find out all possible places where this could be wrong...)

astropy/time/tests/test_basic.py
mhvk: Good point, will do.

astropy/time/tests/test_basic.py
```diff
@@ -26,12 +26,15 @@
...
     HAS_PYTZ = False
 
-allclose_jd = functools.partial(np.allclose, rtol=2. ** -52, atol=0)
-allclose_jd2 = functools.partial(np.allclose, rtol=2. ** -52,
-                                 atol=2. ** -52)  # 20 ps atol
-allclose_sec = functools.partial(np.allclose, rtol=2. ** -52,
-                                 atol=2. ** -52 * 24 * 3600)  # 20 ps atol
-allclose_year = functools.partial(np.allclose, rtol=2. ** -52,
+
+FLOAT_EPS = np.finfo(float).eps
```
aarchiba: Why the FLOAT_EPS rather than the other? It's not many more characters and it's less shouty.

aarchiba:
(sorry for the slightly redundant review comments I had forgotten to hit "submit review")

astropy/time/tests/test_basic.py
mhvk: "the other"? you mean lower case, or `eps_float`?

astropy/time/tests/test_basic.py
```diff
@@ -867,51 +867,62 @@ def test_explicit_example(self):
...
-        f = 2.**(-np.finfo(np.longdouble).nmant) * 65536
+        # Create a different long double (which will give a different jd2
+        # even when long doubles are more precise than Time, as on arm64).
+        f = max(2.**(-np.finfo(np.longdouble).nmant) * 65536, 2.**(-52))
         mjd_long = np.longdouble(i) + np.longdouble(f)
         assert mjd_long != i, "longdouble failure!"
         t = Time(mjd_long, format='mjd')
         expected = Time(i, f, format='mjd')
-        assert t == expected
+        assert abs(t - expected) <= 20.*u.ps
```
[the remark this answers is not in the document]
mhvk: I like that idea! Though out of scope here (and not quite sure about implementation).

astropy/units/tests/test_quantity_ufuncs.py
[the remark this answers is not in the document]
mhvk: The answers are actually reasonable - and those are still tested.  All this does is to filter out the numpy warnings, which we have no control over (and do not occur on all platforms), so that these do not cause test failures.

astropy/wcs/tests/test_wcs.py
```diff
@@ -1178,6 +1178,8 @@ def test_scalar_inputs():
     assert result[0].shape == (1,)
 
 
+# Ignore RuntimeWarning raised on s390.
+@pytest.mark.filterwarnings('ignore:.*invalid value encountered in.*')
```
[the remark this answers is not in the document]
mhvk: Again, note this does not affect test failure, just that we do not fail depending on whether or not this warning is emitted.

astropy/time/tests/test_basic.py
aarchiba: I mean `np.finfo(float).eps`, it's really not longer, it's immediately apparent what it is, and it shows that it's explicitly a function of the type `float`.

astropy/time/tests/test_basic.py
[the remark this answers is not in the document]
aarchiba: What's wrong with creating a class attribute in Time whose value is `(np.finfo(float).eps*u.day).to(u.ps)`?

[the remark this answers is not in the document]
mhvk: Just that no other class has this attribute, which means it won't be easily discovered - I considered whether we could override `np.finfo` on `Time` - but that is not possible.

Anyway, probably best discussed in a separate issue.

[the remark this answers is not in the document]
mhvk: A propos overrides, what would be easy to do is to override `np.allclose` to work on `Time` instances; that could have then a reasonably sensible default for `atol` of a few times the float precision in days.

astropy/time/tests/test_basic.py
mhvk: OK, I see your point. I changed it everywhere.

astropy/time/tests/test_basic.py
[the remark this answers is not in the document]
mhvk: 64-bits are filtered out, but 80-bit would still work, since the setting of `jd2` from it is exact to 2**-52, i.e., the same rounding errors will be made. (And other pieces test that the input longs are different)

mhvk:
I think I got all comments - anything left?

astropy/units/tests/test_quantity_ufuncs.py
mhvk: Hey, it turns out I now get these warnings even on my linux box (for numpy-dev), so filtering is OK after all. How nice to have laziness rewarded!

pllim: Hah! Okay, then.

mhvk:
Thanks, I've quickly rebased #9719 - hopefully it'll now show an all-clear numpy-dev job

PR #9680, opened 2019-11-25
Fix bug in init from list of mixins

### LINKED ISSUE #9357: group_by fails with a Time column
The following code:
```python
import numpy as np

from astropy import table
from astropy.time import Time

mjds = [Time(58000, format="mjd")]

table.Table([[58000.], ["gbt"]], names=("mjd", "obs")).group_by("obs")
table.Table([mjds, ["gbt"]], names=("mjd", "obs")).group_by("obs")
```
fails on the version with a Time column but not on the version with a float column:
```
  File "../table-problem.py", line 8, in <module>
    table.Table([mjds, ["gbt"]], names=("mjd", "obs")).group_by("obs")
  File "/home/archibald/projects/astropy/astropy/astropy/table/table.py", line 2949, in group_by
    return groups.table_group_by(self, keys)
  File "/home/archibald/projects/astropy/astropy/astropy/table/groups.py", line 18, in table_group_by
    return _table_group_by(table, keys)
  File "/home/archibald/projects/astropy/astropy/astropy/table/groups.py", line 54, in _table_group_by
    table_index = get_index_by_names(table, keys)
  File "/home/archibald/projects/astropy/astropy/astropy/table/index.py", line 666, in get_index_by_names
    for index in table.indices:
  File "/home/archibald/projects/astropy/astropy/astropy/table/table.py", line 667, in indices
    for index in column.info.indices:
TypeError: 'NoneType' object is not iterable
```

taldcroft:
### Description

There was an oversight in #9165 (Allow adding a table column as a list of mixin-type objects) where the column `info.indices` was not initialized to `[]`.  This whole machinery is little fragile, but for now just doing the initialization fixes the problem noted in #9357.

Fixes #9357

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -995,6 +995,7 @@ class Table:
             try:
                 col = data[0].__class__(data)
                 col.info.name = name
+                col.info.indices = []
                 return col
             except Exception:
                 # If that didn't work for some reason, just turn it into np.array of object
--- a/astropy/table/tests/test_table.py
+++ b/astropy/table/tests/test_table.py
@@ -2510,7 +2510,10 @@ def test_data_to_col_convert_strategy():
 
 
 def test_rows_with_mixins():
-    """Test for #9165 to allow adding a list of mixin objects"""
+    """Test for #9165 to allow adding a list of mixin objects.
+    Also test for fix to #9357 where group_by() failed due to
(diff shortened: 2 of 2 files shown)


mhvk:
p.s. Agreed that the `indices` part seems a big fragile - I fear I don't quite understand it, but will try to make an effort at some point!

PR #9688, opened 2019-11-26
Fix bug in masking of invalid coordinates returned by WCSLIB

astrofrog:
Previously, the masking did not respect the exact bit mask returned by WCSLIB but masked all values if the bit mask had any bit set. This caused issues with e.g. all sky spectral cubes where even if one converts pixel coordinates outside of the sky, the spectral coordinate should always return a valid coordinate if the spectral axis is uncorrelated.

I'd like to get this into 4.0 if possible since it's a pretty frustrating bug for some use cases I work on.

And for a blast from the past, this is a bug in the original implementation of the masking from https://github.com/astropy/astropy/pull/107 - unless WCSLIB changed stat from 1 to a bit mask in the mean time.

### DIFF
--- a/astropy/wcs/src/util.c
+++ b/astropy/wcs/src/util.c
@@ -20,7 +20,6 @@ void set_invalid_to_nan(
   const int* s = stat;
   const int* s_end = stat + ncoord;
   double n;
-  int bit;
 
   #ifndef NAN
     #define INF (DBL_MAX+DBL_MAX)
@@ -34,14 +33,14 @@ void set_invalid_to_nan(
 
   for ( ; s != s_end; ++s) {
     if (*s) {
-      bit = 1;
+      int bit = 1;
       for (i = 0; i < nelem; ++i) {
-        if ((*s & bit) == bit) {
+        if (*s & bit) {
           *d++ = n;
         } else {
           d++;
         }
-        bit *= 2;
+        bit <<= 1;
       }
     } else {
       d += nelem;
--- a/astropy/wcs/src/util.c
+++ b/astropy/wcs/src/util.c
@@ -20,18 +20,28 @@ void set_invalid_to_nan(
   const int* s = stat;
   const int* s_end = stat + ncoord;
   double n;
+  int bit;
 
   #ifndef NAN
     #define INF (DBL_MAX+DBL_MAX)
     #define NAN (INF-INF)
   #endif
 
+  // Note that stat is a bit mask, so we need to mask only some of
+  // the coordinates depending on the bit mask values.
+
   n = NAN;
 
   for ( ; s != s_end; ++s) {
     if (*s) {
+      bit = 1;
       for (i = 0; i < nelem; ++i) {
-        *d++ = n;
+        if ((*s & bit) == bit) {
+          *d++ = n;
+        } else {
+          d++;
+        }
+        bit *= 2;
       }
     } else {
       d += nelem;
--- a/astropy/wcs/src/util.c
+++ b/astropy/wcs/src/util.c
@@ -36,10 +36,12 @@ void set_invalid_to_nan(
       int bit = 1;
       for (i = 0; i < nelem; ++i) {
         if (*s & bit) {
-          *d++ = n;
-        } else {
-          d++;
+          *d = n;
         }
+        d++;
+        // We don't need to worry about overflow here because the WCS
+        // class cannot be used for naxis > 15 so nelem will always
+        // be <=15.
         bit <<= 1;
       }
     } else {
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1154,6 +1154,13 @@ astropy.wcs
 - ``FITSWCSAPIMixin`` now returns tuples not lists from ``pixel_to_world`` and
   ``world_to_pixel``. [#9678]
 
+- Fix NaN-masking of world coordinates when some but not all of the coordinates
+  were flagged as invalid by WCSLIB. This occurred for example with WCS with >2
+  dimensions where two of the dimensions were celestial coordinates and pixel
+  coordinates ouside of the 'sky' were converted to world coordinates -
+  previously all world coordinates were masked even if uncorrelated with the
+  celestial axes, but this is no longer the case. [#9688]
+
 Other Changes and Additions
 ---------------------------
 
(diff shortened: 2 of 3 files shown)


astropy/wcs/src/util.c
```diff
@@ -20,18 +20,28 @@ void set_invalid_to_nan(
...
     if (*s) {
+      bit = 1;
       for (i = 0; i < nelem; ++i) {
-        *d++ = n;
+        if ((*s & bit) == bit) {
+          *d++ = n;
+        } else {
+          d++;
+        }
+        bit *= 2;
```
MSeifert04: I would say a left-shift would be more comprehensible here.

astropy/wcs/src/util.c
```diff
@@ -20,18 +20,28 @@ void set_invalid_to_nan(
   const int* s = stat;
   const int* s_end = stat + ncoord;
   double n;
+  int bit;
```
MSeifert04: You can do the declaration in the loop (as long as it's at the beginning of the loop block).

astropy/wcs/src/util.c
MSeifert04: Also do we need to check for overflow here? I'm a bit hazy on the details but signed integer overflow could lead to undefined behavior, right?

astropy/wcs/src/util.c
```diff
@@ -20,18 +20,28 @@ void set_invalid_to_nan(
...
+  // the coordinates depending on the bit mask values.
+
   n = NAN;
 
   for ( ; s != s_end; ++s) {
     if (*s) {
+      bit = 1;
       for (i = 0; i < nelem; ++i) {
-        *d++ = n;
+        if ((*s & bit) == bit) {
```
MSeifert04: I'm more used to the `if (*s & bit)` but preferences may vary. Is there any specific reason to check for equality to bit?

astropy/wcs/src/util.c
```diff
@@ -20,18 +20,28 @@ void set_invalid_to_nan(
...
+          *d++ = n;
```
MSeifert04: Even if it may sound stupid: What does this line do? 😕

astropy/wcs/src/util.c
astrofrog: Oh yes of course, I was being silly

astropy/wcs/src/util.c
pllim: Just so you know, I was wondering the same thing!

astrofrog: I *think* it basically increments the pointer to the data, then dereferences it and assigns it to NaN all in one go? It was written by the great wizard @mdboom back in the days :)

astropy/wcs/src/util.c
astrofrog: I've emailed Mark Calabretta to ask whether there is a maximum supported number of dimensions in WCSLIB, to see if we can assume we are safe against overflow.

astropy/wcs/src/util.c
astrofrog: Done

astropy/wcs/src/util.c
astrofrog: I changed to left shift

astropy/wcs/src/util.c
astrofrog: I think it's just doing:

```
d++;
*d = n
```

mhvk: The other way around, I think:
```
*d = n;
d++;
```
(otherwise it would be `*++d`)

mhvk: I.e., it sets the result to `nan` and then goes to the next one, while in the `else` clause one just goes to the next one without making a change. Personally, I think it would be way clearer to write `*d = n` here and then not have an else clause and either do `d++` below or in the `for` statement.  But out of scope here!

astrofrog: Ah yeah that makes more sense :)

MSeifert04: Thank you for the clarification!

> Personally, I think it would be way clearer to write `*d = n` here and then not have an else clause and either do `d++` below or in the for statement. But out of scope here!

I don't think it would be out of scope we already touch these lines and the `else` wasn't present before. It's subjective but what you propose would make it much clearer to me.

astropy/wcs/src/util.c
MSeifert04: At least for NumPy the maximum number of dimensions is [32](https://github.com/numpy/numpy/blob/v1.17.4/numpy/core/include/numpy/ndarraytypes.h#L40) which could overflow a signed 32bit-integer (even if extremely unlikely):

```python
>>> import numpy as np
>>> np.array([[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[1]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]).ndim
32
```

astrofrog: This is not a risk, naxis is limited to <=15:

```
In [2]: wcs = WCS(naxis=50)                                                     
<ipython-input-2-9c7a363c3ea0> in <module>
----> 1 wcs = WCS(naxis=50)

~/Code/astropy/astropy/wcs/wcs.py in __init__(self, header, fobj, key, minerr, relax, naxis, keysel, colsel, fix, translate_units, _do_set)
    366                 naxis = 2
    367             wcsprm = _wcs.Wcsprm(header=None, key=key,
--> 368                                  relax=relax, naxis=naxis)
    369             self.naxis = wcsprm.naxis
    370             # Set some reasonable defaults.

ValueError: naxis must be in range 1-15
```

I added a comment about this.

astropy/wcs/src/util.c
astrofrog: Ok I've done this now and it is indeed cleaner, thanks!

astropy/wcs/src/util.c
MSeifert04: Thank you

MSeifert04:
LGTM from the technical perspective. I'll let @nden approve the PR if it also makes sense logically 😄

astropy/wcs/src/util.c
```diff
@@ -26,12 +26,21 @@ void set_invalid_to_nan(
...
+  // the coordinates depending on the bit mask values.
+
   n = NAN;
 
   for ( ; s != s_end; ++s) {
     if (*s) {
+      int bit = 1;
       for (i = 0; i < nelem; ++i) {
-        *d++ = n;
+        if (*s & bit) *d = n;
```
MSeifert04: I'm not in favor of dropping braces (even if allowed) and PEP7 also says "braces are required everywhere, even where C permits them to be omitted". But this is pure nitpicking and probably very subjective as well. Feel free to leave this as is if you've a different opinion.

astrofrog: Fixed!

astrofrog:
@nden - could you review? Thanks!

PR #9707, opened 2019-11-30
Tabular1d inverse flags

nden:
This pull request is to address an issue with `Tabular1D.inverse` which is set to use the default flags.
This change sets the inverse to reuse the flags of the original model. For example, if `bounds_error=False` in the forward model, it is also set to `False` in the inverse model. 
There's really no way to give the user the option to change he flags of the inverse. If that's desirable then they should reassign the inverse. 

@Cadair Does this solve the issue you were having? If not could you describe it?


nden:
@bsipocz Milestone changed to 4.0

PR #9712, opened 2019-12-03
Allow table row to interact like a dict, e.g., making it splattable

### LINKED ISSUE #9687: Make table's `Row` a mapping?
Right now, `Row` is registered as a `collections.abc.Sequence`, but arguably it makes more sense as a `Mapping`, since the way `__getitem__` is implemented is much more like a `dict`. Indeed, `Row` could gain `keys()`, `items()`, and `.values()` methods which would make it very similar to a `dict` (and allow "splatting" with `**row`). The only inconsistency might be in `Row.__iter__`, which currently iterates over values rather than keys.

mhvk:
Fixes #9687.

With this:
```

qt = QTable([[1., 2.], [3., 4.]], names=['a', 'b'])
dict(qt[0])
# {'a': 1.0, 'b': 3.0}
```
Using a row for keyword arguments, as in `f(**row)`, works too.

This is nicely symmetric with what already works for a `Table`.

Questions/notes:
- `Table.items()` does not exist - I could add that as well, or remove it from `Row` - it does not influence the main functionality (which only requires `.keys()`).
- I could also register `Row` as a mapping, via `collections.abc.Mapping.register(Row)` (right now, it is registered only as a sequence). Not quite sure what the (dis)advantages are.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -57,9 +57,9 @@ astropy.table
   the ``unit`` and ``description`` for the table columns at the time of
   creating or reading the table. [#9671]
 
-- Make table ``Row`` work as mappings, by adding ``.keys()``, ``.values()`` and
-  ``.items()`` methods. With this ``**row`` becomes possible, as does, more
-  simply, turning a ``Row`` into a dictionary with ``dict(row)``. [#9712]
+- Make table ``Row`` work as mappings, by adding ``.keys()`` and ``.values()``
+  methods. With this ``**row`` becomes possible, as does, more simply, turning
+  a ``Row`` into a dictionary with ``dict(row)``. [#9712]
 
 astropy.tests
 ^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


mhvk:
Hmm, `.items()` causes tests to fail, so I just removed it for now - though I think we could reinstate it by a more careful look at `_insert_row`, especially its `_is_mapping` method. But perhaps best done separately.

taldcroft:
About the questions:
- Agreed it would be good to have `items()` but not required.
- I'm also not quite sure of the benefits / problems with adding this `abc` Mapping.

mhvk:
I opened #9732 about having `.items()` - my sense is that it is worth doing and it is a very nice first issue.

saimn:
There is a little inconsistency here: `.values` returns a generator while `.keys` return a list. Maybe better to change one or the other ?

mhvk:
Hmm, you're right, it is not quite consistent - I just did the easiest I could think of... I guess ideally they are as similar as possible to the equivalents on `dict`.

saimn:
So `.keys()` should return an iterator ? That would also make it easier to switch `Table.columns` to use `dict` instead of `orderedDict` in the future (not sure if there is a plan for it ?).

mhvk:
Makes sense for `.keys()` to be the dict-like iterator indeed.

PR #9730, opened 2019-12-06
Use environment variables to select system libraries

astrofrog:

Previously, the ``--use-system-...`` command-line options required patching the setup.py commands, and these were also not easy to set when doing pip installs. The environment variable approach no longer requires patching the commands, and works regardless of whether pip or setup.py is used. This approach is described [in the APE](https://github.com/astropy/astropy-APEs/pull/52).

I checked with @olebole who confirmed this approach would not pose issues for Debian packaging.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -248,9 +248,14 @@ astropy.wcs
 Other Changes and Additions
 ---------------------------
 
-- Nothing changed yet.
-
-
+- The way in which users can specify whether to build astropy against
+  existing installations of C libraries rather than the bundled one
+  has changed, and should now be done via environment variables rather
+  than setup.py flags (e.g. --use-system-erfa). The available variables
+  are ``ASTROPY_USE_SYSTEM_CFITSIO``, ``ASTROPY_USE_SYSTEM_ERFA``,
+  ``ASTROPY_USE_SYSTEM_EXPAT``, ``ASTROPY_USE_SYSTEM_WCSLIB``, and
+  ``ASTROPY_USE_SYSTEM_ALL``. These should be set to ``1`` to build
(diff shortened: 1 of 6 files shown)


astrofrog:
@mhvk - I split the lines in my latest commit

astrofrog:
@eteq @nden - just want to make sure you are happy with the changes that touch ERFA and WCSLIB?

astrofrog:
Since @mhvk has already approved this, @olebole is happy with this approach for Debian packaging, and given that people approved APE 17 which describes this approach, in the interest of finishing up https://github.com/astropy/astropy/pull/9726 I am going to go ahead and merge this. If absolutely needed, we can consider adding back command-line flags to ``setup.py`` by just checking ``sys.argv`` and updating ``os.environ``, but I'd rather avoid it if possible.

MSeifert04:
Is this something that needs a "What's new"?

astrofrog:
@MSeifert04 - it may only be relevant for distribution maintainers, so it might make more sense to just ping them all directly?

MSeifert04:
Not sure, that's why I asked. Having some sort of "if you had x before, now you need to do y" in a whatsnew could help those who had the need or urge to build against system libraries - but if that's something that only distribution maintainers used then pinging them directly makes more sense.

astrofrog:
I'll label in just in case, and we can decide later :)

mhvk:
I think a what's-new is needed, but would do it as part of the larger astropy-helpers removal (i.e., a what's new that includes a checklist of related changes)

pllim:
Wait a minute... Didn't @astrofrog insisted that what's new entry should be added as part of the PR instead of waiting till release time?

PR #9733, opened 2019-12-06
DOC: Clarify IAU year inconsistencies

pllim:
### Description

This pull request is to address potential year typos in IAU constant descriptions. They were both added in #6083 .

I really don't know if these are typo or intentional. Someone familiar with how IAU update their stuff need to review. However, it does seem weird that `IAU2012` somehow would know about "IAU 2015" (3 years into the future).

### DIFF
--- a/astropy/constants/iau2012.py
+++ b/astropy/constants/iau2012.py
@@ -35,7 +35,7 @@ kpc = IAU2012('kpc', "Kiloparsec",
               1000. * au.uncertainty / np.tan(np.radians(1. / 3600.)),
               "Derived from au", system='si')
 
-# Luminosity
+# Luminosity not defined till 2015 (https://arxiv.org/abs/1510.06262)
 L_bol0 = IAU2012('L_bol0', "Luminosity for absolute bolometric magnitude 0",
                  3.0128e28, "W", 0.0, "IAU 2015 Resolution B 2", system='si')
 
--- a/astropy/constants/iau2015.py
+++ b/astropy/constants/iau2015.py
@@ -20,7 +20,7 @@ class IAU2015(Constant):
 
 # DISTANCE
 
-# Astronomical Unit
+# Astronomical Unit (did not change from 2012)
 au = IAU2015('au', "Astronomical Unit", 1.49597870700e11, 'm', 0.0,
              "IAU 2012 Resolution B2", system='si')
 


astropy/constants/iau2015.py
```diff
@@ -22,7 +22,7 @@ class IAU2015(Constant):
 
 # Astronomical Unit
 au = IAU2015('au', "Astronomical Unit", 1.49597870700e11, 'm', 0.0,
-             "IAU 2012 Resolution B2", system='si')
+             "IAU 2015 Resolution B2", system='si')
```
mhvk: This one is actually from 2012 - it has not changed.

astropy/constants/iau2012.py
```diff
@@ -37,7 +37,7 @@ class IAU2012(Constant):
 
 # Luminosity
 L_bol0 = IAU2012('L_bol0', "Luminosity for absolute bolometric magnitude 0",
-                 3.0128e28, "W", 0.0, "IAU 2015 Resolution B 2", system='si')
+                 3.0128e28, "W", 0.0, "IAU 2012 Resolution B 2", system='si')
```
mhvk: I think this one is worse - the value is indeed from 2015, so in 2012 it must have been another one (with an error bar still)....

pllim: Hmm. I don't see this constant mentioned in https://www.iau.org/static/resolutions/IAU2012_English.pdf

pllim: Could it be that this thing didn't exist until 2015 and it is inserted into `iau2012` for API compatibility? Maybe @stargaser remembers. https://arxiv.org/abs/1510.06262

mhvk: Hmm, doing a bit of digging, I see that in astropy 1.0 and 1.3, we had
```
L_sun = Constant('L_sun', "Solar luminosity", 3.846e26, 'W', 0.0005e26,
                 "Allen's Astrophysical Quantities 4th Ed.", system='si')
```
So, I think that is what `iau2012` should have as well.

pllim: But... it is a breaking change. What to do?

pllim: @mhvk , but the line in question is `L_bol0`, not `L_sun`.

mhvk: Duh, clearly not reading very carefully. So, `L_bol0` is there in astropy 1.3, with the same docstring. That was 2016, so by then `L_bol0` was indeed known; I think it just ended up in `iau2012`, since that is mostly what was used in 1.3.

But in any case, it was fixed in 2015!

pllim: @mhvk , so do you think the comments added are okay then? I agree that no code change is needed and not a bug (phew!) but hopefully these comments would prevent this confusion from resurfacing in the future.

PR #9739, opened 2019-12-08
Change ascii.read with converters and names both specified

### LINKED ISSUE #9701: ASCII reader `converters` uses guessed column names
### Description
When using `ascii.read` with both the `names` and `converters` arguments, the `converters` dictionary ignores the user-supplied names and requires that you know the guessed names. I think it would be clearer if the keys of `converters` were meant to be the values in `names`.

### Steps to Reproduce

```python
from astropy.io import ascii

csv_text = '''a,b,c
1,2,3
4,5,6'''
names = ['A', 'B', 'C']

converters_bad = {
    'A': [ascii.convert_numpy(float)], 
    'B': [ascii.convert_numpy(int)],
    'C': [ascii.convert_numpy(str)]
}
t_bad = ascii.read(csv_text, format='csv', names=names, converters=converters_bad)
print(t_bad.dtype)  # all columns are incorrectly parsed as integers

converters_good = {
    'a': [ascii.convert_numpy(float)], 
    'b': [ascii.convert_numpy(int)],
    'c': [ascii.convert_numpy(str)]
}
t_good = ascii.read(csv_text, format='csv', names=names, converters=converters_good)
print(t_good.dtype)  # columns are parsed as intended
```

### System Details

macOS-10.15.1-x86_64-i386-64bit
Python 3.8.0 (default, Nov  6 2019, 15:49:01) 
Numpy 1.17.4
astropy 4.1.dev27074

taldcroft:
### Description

Change the behavior when reading a table where both the ``names`` argument is provided (to specify the output column names) and the ``converters``  argument is provided (to specify column conversion functions). Previously the ``converters`` dict names referred to the *input* table column names, but now
  they refer to the *output* table column names.

Fixes #9701

### DIFF
--- a/astropy/io/ascii/core.py
+++ b/astropy/io/ascii/core.py
@@ -618,6 +618,48 @@ class BaseHeader:
         return tuple(col.name if isinstance(col, Column) else col.info.name
                      for col in self.cols)
 
+    def remove_columns(self, names):
+        """
+        Remove several columns from the table.
+
+        Parameters
+        ----------
+        names : list
+            A list containing the names of the columns to remove
+        """
+        colnames = self.colnames
+        for name in names:
+            if name not in colnames:
+                raise KeyError(f"Column {name} does not exist")
+
+        self.cols = [col for col in self.cols if col.name not in names]
+
+    def rename_column(self, name, new_name):
+        """
+        Rename a column.
+
+        Parameters
+        ----------
+        name : str
+            The current name of the column.
+        new_name : str
+            The new name for the column
+        """
+        try:
+            idx = self.colnames.index(name)
+        except ValueError:
+            raise KeyError(f"Column {name} does not exist")
+
+        col = self.cols[idx]
+
+        # For writing self.cols can contain cols that are not Column.  Raise
+        # exception in that case.
+        if isinstance(col, Column):
+            col.name = new_name
+        else:
+            raise TypeError(f'got column type {type(col)} instead of required '
+                            f'{Column}')
+
     def get_type_map_key(self, col):
         return col.raw_type
 
@@ -1046,11 +1088,14 @@ def _is_number(x):
 
 def _apply_include_exclude_names(table, names, include_names, exclude_names):
     """
-    Apply names, include_names and exclude_names to a table.
+    Apply names, include_names and exclude_names to a table or BaseHeader.
+
+    For the latter this relies on BaseHeader implementing ``colnames``,
+    ``rename_column``, and ``remove_columns``.
 
     Parameters
     ----------
-    table : `~astropy.table.Table`
+    table : `~astropy.table.Table`, `~astropy.io.ascii.BaseHeader`
         Input table
     names : list
         List of names to override those in table (set to None to use existing names)
@@ -1201,10 +1246,11 @@ class BaseReader(metaclass=MetaBaseReader):
         self.data.masks(cols)
         if hasattr(self.header, 'table_meta'):
             self.meta['table'].update(self.header.table_meta)
-        table = self.outputter(cols, self.meta)
-        self.cols = self.header.cols
 
-        _apply_include_exclude_names(table, self.names, self.include_names, self.exclude_names)
+        _apply_include_exclude_names(self.header, self.names, self.include_names, self.exclude_names)
+
+        table = self.outputter(self.header.cols, self.meta)
+        self.cols = self.header.cols
 
         return table
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -109,6 +109,12 @@ astropy.extern
 astropy.io.ascii
 ^^^^^^^^^^^^^^^^
 
+- Changed the behavior when reading a table where both the ``names`` argument
+  is provided (to specify the output column names) and the ``converters``
+  argument is provided (to specify column conversion functions). Previously the
+  ``converters`` dict names referred to the *input* table column names, but now
+  they refer to the *output* table column names. [#9739]
+
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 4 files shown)


astropy/io/ascii/core.py
```diff
@@ -618,6 +618,48 @@ def colnames(self):
...
+        """
+        Remove several columns from the table.
+
+        Parameters
+        ----------
+        names : list
+            A list containing the names of the columns to remove
+        """
+        colnames = self.colnames
+        for name in names:
```
mhvk: Maybe just
```
if any(name not in colnames for name in names):
   raise(...)
```

dhomeier: In principle, with the new behaviour
`converters={'ntp_seconds': [convert_numpy(np.int64)]}` 
becomes the correct usage.
But in a way this is complementary to my comment in #9701 on pathological `converters` with conflicting values - meaning one could consider as a convenience and backwards-compatibility option to consider any converters passed by default, automatic _or_ user names, as in
```
converters.set_default('A', converters['a'])
converters.set_default('A', converters['col1'])
```
so that additional tests
```
    converters = {
        'a': [ascii.convert_numpy(float)],
        'b': [ascii.convert_numpy(int)],
        'c': [ascii.convert_numpy(str)]
    }
    t = ascii.read(csv_text, format='csv', names=names, converters=converters)
    assert t['A'].dtype.kind == 'f'
    assert t['B'].dtype.kind == 'i'
    assert t['C'].dtype.kind == 'U'

    converters = {
        'col1': [ascii.convert_numpy(float)],
        'col2': [ascii.convert_numpy(int)],
        'col3': [ascii.convert_numpy(str)]
    }
    t = ascii.read(csv_text, format='csv', names=names, converters=converters)
    assert t['A'].dtype.kind == 'f'
    assert t['B'].dtype.kind == 'i'
    assert t['C'].dtype.kind == 'U'
```
would pass.
Unfortunately I don't see any obvious place to implement this where all the different colnames would be readily available.

mhvk: Yes, good to have the name (walrus to the rescue eventually?).

And agreed with choosing one over the other, even if it is somewhat painful.

taldcroft:
Added test and fixed failing test and (I think) addressed review comments.

astropy/io/ascii/core.py
dhomeier: Agreed, too. The user after all selected the custom names, so it's reasonable to expect them to use those names for the converters, too.

pllim:
This broke my code. My table (provided by someone else):
```
 WAVECAT.DAT -- A list of wavelength sets that cover the range of each
# observation mode. Spectrographic modes have the wavelength spacing set
# to equal the detector spacing. Wavelength sets can either be specified
# as a file or a set of coefficients. In the latter case, the coefficients
# are surrounded by parentheses and separated by commas. There are up to
# four coefficients. In order, they are the short wavelength, long
# wavelength, dispersion at the short wavelength, and dispesion at the long
# wavelength. Either one or both of the dispersions are optional. Units
# of the wavelengths are in angstroms and of the dispersions are angstroms
# per pixel.
#
#_OBSMODE_______________FILENAME_______________________
cos,fuv                 (900.0,3000.0,1.0)
cos,nuv                 (1000.0,12000.0,1.0)
fgs            (2000.0,8000.0,1.0)
foc            (1000.0,8000.0,1.0)
hsp            (1000.0,12000.0,1.0)
```

This used to work up to `astropy` 4.0 but no longer works with dev:
```python
import numpy as np
from astropy.io import ascii
t = ascii.read(
        filename, header_start=None, names=(str('OBSMODE'), str('FILENAME')),
        converters={'OBSMODE': np.str, 'FILENAME': np.str})
```

How do I make it work again?

mhvk:
Damn, that's really weird: this one works now by passing in 'col1' and 'col2' for the keys in converters, which is exactly what we were trying to get rid off! So, definitely a bug (do make it a new issue...). @taldcroft - any ideas???

pllim:
Hmm... astropy dev is able to read when I change the call to:
```python
t = ascii.read(
        filename, names=('OBSMODE', 'FILENAME'),
        guess=False, format='no_header',
        converters={'OBSMODE': [ascii.convert_numpy(str)],
                    'FILENAME': [ascii.convert_numpy(str)]})
```

Was I using `ascii.read()` incorrectly? Do you still want me to open an issue? 🤷‍♀

taldcroft:
Definitely the original `np.str` was not correct, and I think it was "accidentally working" in the old version because the explicit conversion was not actually necessary and because of using the wrong column names (for pre-4.0) it was silently just ignoring the incorrect convert specifications.
```
In [11]: ascii.read(txt, header_start=None, names=(str('OBSMODE'), str('FILENAME')),
    ...:         converters={'jack': np.str, 'jill': np.str}, format='no_header', guess=False)
    ...: 
    ...:         
Out[11]: 
<Table length=5>
OBSMODE       FILENAME      
  str7         str20        
------- --------------------
cos,fuv   (900.0,3000.0,1.0)
cos,nuv (1000.0,12000.0,1.0)
    fgs  (2000.0,8000.0,1.0)
    foc  (1000.0,8000.0,1.0)
    hsp (1000.0,12000.0,1.0)
```

The incantation in the comment above is correct usage if you think the `str` conversion is in general necessary.  For the specific example it is not, a simple `ascii.read(txt, format='no_header', names=..)` gives the right answer.

So... I *think* everything is OK.

mhvk:
Ah, I completely missed that! So, the "reversion" was just that I made the converters irrelevant again.... (the use of `converters` is not too intuitive...)

taldcroft:
> (the use of converters is not too intuitive...)

Yeah, I would do something simpler today than my design from almost 10 years ago when just learning Python.

mhvk:
Very recognizable sentiment! But it is really a good language if over time one leans to make things simpler!

taldcroft:
I've just learned to start simple and not over engineer things.

Issue #9610, opened 2019-11-16
FITS header "CONTINUE" keyword is now part of the FITS standard


demitri:
### Description

In the documentation [here](https://docs.astropy.org/en/stable/io/fits/usage/headers.html#continue-cards) it notes that `CONTINUE` keywords are not part of the FITS specification, but are supported by Astropy. This is no longer true; the proposal is now officially part of the 4.0 standard. For details, see Section 4.2.1.2 of [the specification](https://fits.gsfc.nasa.gov/fits_standard.html).

As noted in the docs, the convention is supported in Astropy, but it would be nice to update them to reflect that it is part of the format.

I also recently encountered a `fits.VerifyError` warning, saying that the`CONTINUE` keyword is invalid and it was inserting a missing "=" character (which is not present with the `CONTINUE` keyword), but I can't find it in the source so maybe that has been removed.

msrdinesh:
Hey, I am a student and would to contribute to the project. I have created a pull request for the issue, please look at it. [9751](https://github.com/astropy/astropy/pull/9751)

pllim:
Update: Documentation is clarified in #9475 

However, the report on `VerifyError` might need a standalone code snippet so others can attempt to reproduce the problem.

saimn:
With #9751 merged, addressing the documentation issue, I would be tempted to close this. We can address the `VerifyError` in another ticket if you have a way to reproduce @demitri ?

PR #9751, opened 2019-12-10
Small correction in doc

msrdinesh:
### Description

This pull request is to address  doc correction about CONTINUE keyworkd

EDIT: Address some of #9610

### DIFF
--- a/docs/io/fits/usage/headers.rst
+++ b/docs/io/fits/usage/headers.rst
@@ -336,8 +336,8 @@ for keywords, a proposal was made in:
     https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/ofwg_recomm/r13.html
 
 by using the CONTINUE keyword after the regular 80 column containing the
-keyword. ``astropy`` does support this convention, even though it is not a FITS
-standard.
+keyword. ``astropy`` does support this convention, which is a part of the FITS
+standard since version 4.0.
 
 Examples
 --------


MSeifert04:
Thank you @msrdinesh  for your contribution! 👍 

> Fixes 9610

The issue mentioned also a VerifyError. If there's indeed such a message we probably should leave the issue open.

docs/io/fits/usage/headers.rst
```diff
@@ -336,7 +336,7 @@ for keywords, a proposal was made in:
     https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/ofwg_recomm/r13.html
 
 by using the CONTINUE keyword after the regular 80 column containing the
-keyword. ``astropy`` does support this convention, even though it is not a FITS
+keyword. ``astropy`` does support this convention, which is a part of latest FITS
```
MSeifert04: "latest" is a moving target. Even though the FITS standard changes very rarely I would feel better if we write the version explicitly "which is a part of the FITS standard since version 4" (it is 4, right?).

Also the trailing whitespace at the end of the line can be removed.

msrdinesh: Yes, it's written 4.0 on the website. I have made the change you suggested. Thank's @MSeifert04 for your remarks.

docs/io/fits/usage/headers.rst
```diff
@@ -336,8 +336,7 @@ for keywords, a proposal was made in:
     https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/ofwg_recomm/r13.html
 
 by using the CONTINUE keyword after the regular 80 column containing the
-keyword. ``astropy`` does support this convention, even though it is not a FITS
-standard.
+keyword. ``astropy`` does support this convention, which is a part of the FITS standard since version 4.0.
```
MSeifert04: This looks much better now. However we generally require lines to be <= 80 characters long. Could you include a newline between "FITS" and "standard"?

Hint: For documentation-only changes you can add `[docs only]` to the commit message so the test-infrastructure knows that it doesn't have to run the full-test-suite.

If possible amend it to your latest commit - but if you don't have much experience with git and don't know how to amend then feel free to ignore this.

pllim:
Hello, thank you for your contribution! Given the one-line change, I am going to take the liberty to squash this into one commit and wrap this up. There is no action necessary on your side. FYI.

msrdinesh:
Thankyou @saimn and @pllim

PR #9754, opened 2019-12-10
Flag examples in config sub-package

lglattly:
### Description

This pull request is to flag examples in the (previously copy edited) config sub-package for possible use in the example gallery.


pllim:
cc @astropy/coordinators

PR #9761, opened 2019-12-11
Change import Exception for missing bz2/lzma to ModuleNotFoundError

### LINKED ISSUE #9713: ENH: raise consistent exceptions on missing bz2/lzma modules
### Description
#9652 has introduced a more concise `ModuleNotFoundError` for `io.fits` if bzip2 is not available. For consistency, `get_readable_fileobj` should raise the same exception, analogously for missing `lzma` support (this had already been changed, but reverted in 7795c8d in order not  to change the API in a bugfix PR). It should be reintroduced on an appropriate deprecation scheme.

dhomeier:
Fixes #9713 - test adapted accordingly, though the `not HAS_BZ2`/`not HAS_XZ` branches are probably rarely ever run.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -160,6 +160,10 @@ astropy.units
 astropy.utils
 ^^^^^^^^^^^^^
 
+- Changed the exception raised by ``get_readable_fileobj`` on missing
+  compression modules (for ``bz2`` or ``lzma``/``xz`` support) to
+  ``ModuleNotFoundError``, consistent with ``io.fits`` file handlers. [#9761]
+
 astropy.visualization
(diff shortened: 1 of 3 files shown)


pllim:
I don't think the CI failure is related. I cleared the cache and reran the CI, just in case.

However, this PR does need a change log. Change log only commit can have a `[docs only]` directive in the commit message to skip irrelevant jobs from rerunning. Thanks!

dhomeier:
Thanks, yes, I did not see anything related to the compression modules in the Windows failure (was wondering  initially if that might be one test system that actually has no `bz2` or `lzma`), and it is obviously not consistent. Had just been waiting for the milestone.

PR #9764, opened 2019-12-12
Add a counter to keep track of SkyCoord initializations

### LINKED ISSUE #9744: Have a counter in SkyCoord initializations and raise a warning in the event of a lot of scalar initializations
Idea from @astrofrog from the 2019 coordination meeting: have a "counter" in `SkyCoord` that keeps track of how many times a scalar SkyCoord it's initialized and raises a warning along the lines of "Are you doing scalar initialization of SkyCoord in for loops? Maybe you should use vectors - see LINK".

Suggestion from @perrygreenfield  - is it really a "rate" (initializations / time) vs "total count" that we care about?

cc @adrn @taldcroft @perrygreenfield

adrn:
I opted for an absolute number instead of a rate because I think this should be a very simple check.

This fixes #9744.

### DIFF
--- a/astropy/coordinates/__init__.py
+++ b/astropy/coordinates/__init__.py
@@ -21,3 +21,22 @@ from .sky_coordinate import *
 from .funcs import *
 from .calculation import *
 from .solar_system import *
+
+
+from astropy import config as _config
+
+
+class Conf(_config.ConfigNamespace):
+    """
+    Configuration parameters for `astropy.coordinates`.
+    """
+
+    skycoord_init_counter_warn_threshold = _config.ConfigItem(
+        100,
+        'This controls the (globally-counted) number of SkyCoord '
+        'initializations that will trigger a warning to the user. To disable '
+        'this counter and warning, set this config item to 0.'
+    )
+
+
+conf = Conf()
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -1,6 +1,7 @@
 
 import re
 import copy
+import warnings
 
 import numpy as np
 
@@ -10,6 +11,7 @@ from astropy import units as u
 from astropy.constants import c as speed_of_light
 from astropy.wcs.utils import skycoord_to_pixel, pixel_to_skycoord
 from astropy.utils.data_info import MixinInfo
+from astropy.utils.exceptions import AstropyWarning
 from astropy.utils import ShapedLikeNDArray
 from astropy.time import Time
 
@@ -206,11 +208,35 @@ class SkyCoord(ShapedLikeNDArray):
     # info property.
     info = SkyCoordInfo()
 
+    # A counter to keep track of the (global) number of times SkyCoord has been
+    # initialized
+    _init_counter = 0
+
     def __init__(self, *args, copy=True, **kwargs):
 
         # these are frame attributes set on this SkyCoord but *not* a part of
         # the frame object this SkyCoord contains
         self._extra_frameattr_names = set()
+
+        # Increment the SkyCoord init counter:
+        SkyCoord._init_counter += 1
+
+        # If there is a threshold in the config, check the number of times
+        # SkyCoord has been initialized and possibly issue a warning
+        from . import conf
+        _counter_lim = conf.skycoord_init_counter_warn_threshold
+        if _counter_lim > 0 and SkyCoord._init_counter > _counter_lim:
+            url = "http://docs.astropy.org/en/latest/coordinates/skycoord.html#array-operations"
+            warnings.warn("We noticed that you have initialized SkyCoord more "
+                          f"than {_counter_lim} times. It is generally much "
+                          "more efficient to instead create a single SkyCoord "
+                          "instance with array-valued coordinate components, "
+                          f"as demonstrated in this documentation page: {url} "
+                          "However, if you do need to initialize many SkyCoord "
+                          "instances and would like to remove this check, set "
+                          "the astropy.coordinates configuration item "
+                          "'skycoord_init_counter_warn_threshold=0'",
+                          AstropyWarning)
 
         # If all that is passed in is a frame instance that already has data,
         # we should bypass all of the parsing and logic below. This is here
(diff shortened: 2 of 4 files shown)


adrn:
BTW: I arbitrarily set the default threshold to 100, but happy to change to something else (maybe 1000?).

pllim:
Wait a minute, I thought this was a "no-go"?

adrn:
@pllim If that's the case, let's discuss here so I remember why :) 

But given the number of test failures this creates, I can see at least one argument for rejecting 😬

astropy/coordinates/__init__.py
```diff
@@ -21,3 +21,22 @@
...
+from astropy import config as _config
+
+
+class Conf(_config.ConfigNamespace):
+    """
+    Configuration parameters for `astropy.coordinates`.
+    """
+
+    skycoord_init_counter_warn_threshold = _config.ConfigItem(
+        100,
```
taldcroft: I would have thought this would be much higher, more like 10000 or 100000.  I see a time of 260 usec / init, so that means 2.6 to 26 seconds in SkyCoord before raising a flag.

astropy/coordinates/sky_coordinate.py
```diff
@@ -206,12 +208,36 @@ class or the corresponding string alias.  The frame classes that are built in
...
+        # Increment the SkyCoord init counter:
+        SkyCoord._init_counter += 1
+
+        # If there is a threshold in the config, check the number of times
+        # SkyCoord has been initialized and possibly issue a warning
+        from . import conf
+        _counter_lim = conf.skycoord_init_counter_warn_threshold
+        if _counter_lim > 0 and SkyCoord._init_counter > _counter_lim:
+            url = "http://docs.astropy.org/en/latest/coordinates/skycoord.html#array-operations"
+            warnings.warn("We noticed that you have initialized SkyCoord more "
```
taldcroft: This is not filtering on scalar inits, so the user may well have made 100 (or 10000) array SkyCoords, so this message will be confusing.

astropy/coordinates/sky_coordinate.py
```diff
@@ -206,12 +208,36 @@ class or the corresponding string alias.  The frame classes that are built in
...
+                          f"than {_counter_lim} times. It is generally much "
+                          "more efficient to instead create a single SkyCoord "
+                          "instance with array-valued coordinate components, "
+                          f"as demonstrated in this documentation page: {url} "
+                          "However, if you do need to initialize many SkyCoord "
+                          "instances and would like to remove this check, set "
+                          "the astropy.coordinates configuration item "
+                          "'skycoord_init_counter_warn_threshold=0'",
```
taldcroft: Better to give the full story, `coordinates.conf.skycoord_init_counter_warn_threshold = 0`.  Most users probably won't know offhand how to set a configuration item.  I don't remember now if the config items need to be typed, since setting this to `None` for no threshold would be more natural.

eteq: Alternatively, explain the whole thing like:
```suggestion
                          "'from astropy.coordinates import conf",
                          "conf.skycoord_init_counter_warn_threshold=0'"
```

astropy/coordinates/sky_coordinate.py
```diff
@@ -206,12 +208,36 @@ class or the corresponding string alias.  The frame classes that are built in
...
         self._extra_frameattr_names = set()
 
+        # Increment the SkyCoord init counter:
+        SkyCoord._init_counter += 1
+
+        # If there is a threshold in the config, check the number of times
+        # SkyCoord has been initialized and possibly issue a warning
+        from . import conf
+        _counter_lim = conf.skycoord_init_counter_warn_threshold
+        if _counter_lim > 0 and SkyCoord._init_counter > _counter_lim:
```
taldcroft: This will be hitting this branch every time after the first time it goes over the limit.  You could write this instead as below, which also reduces the number of compares to 1.
```
if SkyCoord._init_counter == _counter_lim:
    ...
```

taldcroft:
Doing the rate should just be a few more lines of actual code.  Start with:
```
_init_counter_time0 = time.time()
```
Then
```
if SkyCoord._init_counter == _counter_lim:
    rate = _counter_lim / (time.time() - SkyCoord._init_counter_time0)
    if rate > 500:  # Fixed limit of 500 per second
        # Issue warning, but no other action, so SkyCoord._init_counter keeps
        # incrementing and we never get inside again.
    else:
        # Rate too low, try again.
        SkyCoord._init_counter = 0
        SkyCoord._init_counter_time0 = time.time()
```

astropy/coordinates/sky_coordinate.py
```diff
@@ -206,12 +208,36 @@ class or the corresponding string alias.  The frame classes that are built in
...
         # the frame object this SkyCoord contains
         self._extra_frameattr_names = set()
 
+        # Increment the SkyCoord init counter:
+        SkyCoord._init_counter += 1
+
+        # If there is a threshold in the config, check the number of times
+        # SkyCoord has been initialized and possibly issue a warning
+        from . import conf
+        _counter_lim = conf.skycoord_init_counter_warn_threshold
```
pllim: Add two more chars so people don't blame me when this variable misbehaves? 😝 
```suggestion
        _counter_limit = conf.skycoord_init_counter_warn_threshold
```

astropy/coordinates/__init__.py
astrofrog: FWIW I would personally lean towards ~10000 or so

astropy/coordinates/sky_coordinate.py
```diff
@@ -206,12 +208,36 @@ class or the corresponding string alias.  The frame classes that are built in
...
+    _init_counter = 0
+
     def __init__(self, *args, copy=True, **kwargs):
 
         # these are frame attributes set on this SkyCoord but *not* a part of
         # the frame object this SkyCoord contains
         self._extra_frameattr_names = set()
 
+        # Increment the SkyCoord init counter:
+        SkyCoord._init_counter += 1
```
astrofrog: Only increment in the else clause of the if statement below?

astropy/coordinates/sky_coordinate.py
astrofrog: What is the performance impact of getting the configuration value?

PR #9771, opened 2019-12-13

taldcroft:
### Description

This PR makes the table subpackage pass:
```
```

Justification for exceptions:

```
income = (gross_wages
          + taxable_interest)
```
- E721: *Do not compare types, use 'isinstance()'.*  In testing it is frequently necessary to do direct type comparisons.  (There are around 40 instances of this in table).


astropy/table/np_utils.py
```diff
@@ -107,7 +103,7 @@ def get_descrs(arrays, col_name_map):
         # Make sure all input shapes are the same
         uniq_shapes = set(col.shape[1:] for col in in_cols)
         if len(uniq_shapes) != 1:
-            raise TableMergeError(f'Key columns {name!r} have different shape')
+            raise TableMergeError(f'Key columns have different shape')
```
taldcroft: This is a real bug that showed up where an undefined variable was used in an exception.  Bad coverage...

taldcroft:
@bsipocz @aarchiba - travis CI is failing on a cache problem.  Do I need #9703?
```
E               astropy.utils.data.CacheDamaged: Apparently abandoned files: {'/tmp/pytest-of-travis/pytest-0/test_check_download_cache0/astropy/download/py3/6fa27f095b6862050fdc465f0c29c850', '/tmp/pytest-of-travis/pytest-0/test_check_download_cache0/astropy/download/py3/eb145695fe8cc89a8ea7e5c74108d768'}
```

pllim:
Ignore it for now. Temporarily skipping it in #9776. Still unresolved.

taldcroft:
@mhvk - ping to have a look at this.

taldcroft:
@mhvk - another ping to have a look at this please.  I'm looking at #9962 but get annoyed because my editor now is strict with flake8 so I see of a ton of false-positive errors in the table code.

PR #9777, opened 2019-12-13
TST: Clean up tmpdir usage in utils.data tests

pllim:
### Description

This pull request is to make `tmpdir` usage more consistent in `test_data.py`. I wondered if this will naturally fix the non-deterministic failure reported in #9765 but I am not sure.

cc @aarchiba


pllim:
Nope. Didn't fix the problem. Not worth doing.

PR #9787, opened 2019-12-16
Add torr unit

### LINKED ISSUE #9783: Add Torr unit to units packages
### Description

The LSST camera vacuum engineers wish to use the torr unit.

This is *almost* the same as mmHg which is in the CDS units package but it's not quite that (7 parts in a million). See https://en.wikipedia.org/wiki/Torr

I believe that torr is defined as:

```python
torr = u.def_unit("Torr", cds.atm/760.0)
```

I'm happy to do a pull request for this but I'm not completely sure where I should define the unit. It seems that `cds` might be the best place but I'm not sure whether `cds` has some restrictive definition that it must only come from some master list held at CDS.  Please advise.

timj:
### Description

This pull request is to address the missing torr unit.

I support mTorr == millitorr and allow all prefixes since I have seen mention of k, m, and micro prefixes.

No unit tests since I'm not entirely sure where to put them. We don't seem to have tests for specific unit conversions. For example, I'd like to show that torr approximates mmHg and that mTorr and millitorr both parse to the same unit.

Fixes #9783

### DIFF
--- a/astropy/units/astrophys.py
+++ b/astropy/units/astrophys.py
@@ -169,6 +169,14 @@ def_unit(['littleh'], namespace=_ns, prefixes=False,
          doc="Reduced/\"dimensionless\" Hubble constant",
          format={'latex': r'h_{100}'})
 
+# The torr is almost the same as mmHg but not quite.
+# See https://en.wikipedia.org/wiki/Torr
+# Define the unit here despite it not being an astrophysical unit.
+# It may be moved if more similar units are created later.
+def_unit(['Torr', 'torr'], _si.atm.value/760. * si.Pa, namespace=_ns,
+         prefixes=[(['m'], ['milli'], 1.e-3)],
+         doc="Unit of pressure based on an absolute scale, now defined as "
+             "exactly 1/760 of a standard atmosphere")
 
 ###########################################################################
 # CLEANUP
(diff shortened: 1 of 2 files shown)


pllim:
@mhvk , if you decide that a "what's new" entry is necessary, the new policy is that the entry now is added together with the PR, not before a release; FYI. Also, does this need a new test?

astropy/units/astrophys.py
```diff
@@ -169,6 +169,14 @@
          doc="Reduced/\"dimensionless\" Hubble constant",
          format={'latex': r'h_{100}'})
 
+# The torr is almost the same as mmHg but not quite.
+# See https://en.wikipedia.org/wiki/Torr
+# Define the unit here despite it not being an astrophysical unit.
+# It may be moved if more similar units are created later.
+def_unit(['Torr', 'torr'], _si.atm.value/760. * si.Pa, namespace=_ns,
+         prefixes=True,
```
mhvk: I would do only `milli`, since there is no need to pollute the namespace with units that are never used.

timj: I'm asking the vacuum engineers whether they are okay with Torr and mTorr.

timj:
@mhvk I've updated the change to limit it to mTorr.

PR #9788, opened 2019-12-17
Flag examples in convolution sub-package

lglattly:
### Description

This pull request is to flag examples in the (previously copy edited) convolution sub-package for possible use in the example gallery.

cc @astropy/coordinators

### DIFF
--- a/docs/convolution/index.rst
+++ b/docs/convolution/index.rst
@@ -153,7 +153,14 @@ algorithm, while :func:`~astropy.convolution.convolve_fft` uses a Fast Fourier
 Transform (FFT). Thus, the former is better for small kernels, while the latter
 is much more efficient for larger kernels.
 
-For example, to convolve a 1D dataset with a user-specified kernel, you can do::
+Example
+-------
+
+..
+  EXAMPLE START
+  Convolution for User-Specified Kernels
+
+To convolve a 1D dataset with a user-specified kernel, you can do::
 
     >>> from astropy.convolution import convolve
     >>> convolve([1, 4, 5, 6, 5, 7, 8], [0.2, 0.6, 0.2])  # doctest: +FLOAT_CMP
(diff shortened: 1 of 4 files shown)


keflavich:
Looks fine at a glance; can I review this in final form somewhere?  These are example snippets that are going into the framework jonathansick showed, right?

lglattly:
That's right! I'm working on putting in the markup on all of the examples so they can be reviewed for possible use. @jonathansick is there a way to preview these examples as they would look in the gallery or not at this stage?

docs/convolution/using.rst
```diff
@@ -65,27 +69,37 @@ interpolated using the kernel::
     >>> convolve([1, 4, 5, 6, np.nan, 7, 8], [0.2, 0.6, 0.2], boundary='extend')  # doctest: +FLOAT_CMP
     array([1.6 , 3.6 , 5.  , 5.75, 6.5 , 7.25, 7.8 ])
 
+..
+  EXAMPLE END
+
+..
+  EXAMPLE START
+  Constructing Arrays as Lists
```
astrofrog: I think this should be Constructing Kernels from Lists

jonathansick:
Sorry, there isn't a great way to preview these yet. Ultimately the comments will get converted into `example` directives, but we should wait on getting the Sphinx extension into the regular Astropy docs stack before doing that.

docs/convolution/using.rst
lglattly: Thanks for the suggestion!

docs/convolution/index.rst
```diff
@@ -153,7 +153,14 @@ algorithm, while :func:`~astropy.convolution.convolve_fft` uses a Fast Fourier
 Transform (FFT). Thus, the former is better for small kernels, while the latter
 is much more efficient for larger kernels.
 
-For example, to convolve a 1D dataset with a user-specified kernel, you can do::
+Example
+-------
+
+..
+  EXAMPLE START
+  Convolution for User-Specified Kernels
```
astrofrog: @jonathansick - just out of curiosity, is there any way we can make the markup more concise, e.g.

```
.. EXAMPLE START: Title

...

.. EXAMPLE END
```

?

jonathansick: That'd be fine. The comment syntax is just a human-readable/searchable placeholder for the actual syntax, which is

```rst
.. example:: title

   Example content.
```

lglattly: Good to know! So I should change the comment markup to:

.. EXAMPLE START: Title

Content

.. EXAMPLE END

with just one colon after "example start" correct?

jonathansick: Sounds good :) Starting by next week at the latest I'll start pushing again to get the extension shipped so you can use the real syntax at some point.

PR #9791, opened 2019-12-17
Clarify wording about location of sun in Galactocentric example

adrn:
It was pointed out to me that the wording in this example was a little confusing, so I tried to clarify!

I also simplified the v_sun definitions as now this can be specified as a quantity instead of a `CartesianDifferential`.

### DIFF
--- a/examples/coordinates/plot_galactocentric-frame.py
+++ b/examples/coordinates/plot_galactocentric-frame.py
@@ -80,10 +80,10 @@ print(gc1.v_x, gc1.v_y, gc1.v_z)
 # respectively. The velocity of the sun can be specified as an
 # `~astropy.units.Quantity` object with velocity units and is interepreted as a
 # Cartesian velocity, as in the example below. Note that, as with the positions,
-# the Galactocentric frame is a right-handed system (i.e. the Sun is at negative
+# the Galactocentric frame is a right-handed system (i.e., the Sun is at negative
 # x values) so ``v_x`` is opposite of the Galactocentric radial velocity:
 
-v_sun = [11.1, 244, 7.25] * u.km/u.s  # [vx, vy, vz]
+v_sun = [11.1, 244, 7.25] * (u.km / u.s)  # [vx, vy, vz]
 gc_frame = coord.Galactocentric(
     galcen_distance=8*u.kpc,
     galcen_v_sun=v_sun,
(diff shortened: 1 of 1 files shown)


examples/coordinates/plot_galactocentric-frame.py
```diff
@@ -77,16 +77,17 @@
...
-# ``v_x`` is opposite of the Galactocentric radial velocity:
-
-v_sun = coord.CartesianDifferential([11.1, 244, 7.25]*u.km/u.s)
-gc_frame = coord.Galactocentric(galcen_distance=8*u.kpc,
-                                galcen_v_sun=v_sun,
-                                z_sun=0*u.pc)
+# respectively. The velocity of the sun can be specified as an
+# `~astropy.units.Quantity` object with velocity units and is interepreted as a
+# Cartesian velocity, as in the example below. Note that, as with the positions,
+# the Galactocentric frame is a right-handed system (i.e. the Sun is at negative
```
pllim:
```suggestion
# the Galactocentric frame is a right-handed system (i.e., the Sun is at negative
```

examples/coordinates/plot_galactocentric-frame.py
```diff
@@ -77,16 +77,17 @@
...
+# x values) so ``v_x`` is opposite of the Galactocentric radial velocity:
+
+v_sun = [11.1, 244, 7.25] * u.km/u.s  # [vx, vy, vz]
```
pllim: It is good practice to create the composite unit first, right? This would also apply to other similar usage in this file.
```suggestion
v_sun = [11.1, 244, 7.25] * (u.km / u.s)  # [vx, vy, vz]
```

examples/coordinates/plot_galactocentric-frame.py
```diff
@@ -108,9 +109,10 @@
 # We still have to assume a line-of-sight velocity for the Galactic center,
 # which we will again take to be 11 km/s:
 vx = 11.1 * u.km/u.s
+v_sun2 = u.Quantity([vx, vy, vz])
```
pllim: The need to cast Quantity here is not obvious.
```suggestion
v_sun2 = u.Quantity([vx, vy, vz])  # List of Quantity -> a single Quantity
```

adrn:
> Does this need copy-editing?

@lglattly hasn't done the examples yet, so I think that can wait until she gets to them.

adrn:
32biy failure looks unrelated

pllim:
Pretty sure I fixed that `io.ascii` failure a while ago. Did you not create this PR off latest upstream `master`?

PR #9793, opened 2019-12-17
Improve performance of rotation_matrix for common angle units

mcara:
### Description

This pull request is to improve performance of `astropy.coordinates.matrix_utilities.rotation_matrix()` by about 10x when using default (degree) units for angles, 6x for radians.

### DIFF
--- a/astropy/coordinates/matrix_utilities.py
+++ b/astropy/coordinates/matrix_utilities.py
@@ -61,12 +61,13 @@ def rotation_matrix(angle, axis='z', unit=None):
     rmat : `numpy.matrix`
         A unitary rotation matrix.
     """
-    if unit is None:
-        angle = np.deg2rad(angle)
+    if isinstance(angle, u.Quantity):
+        angle = angle.to_value(u.radian)
     else:
-        unit = u.Unit(unit)
-        if unit is not u.rad:
-            angle = Angle(angle, unit=unit)
+        if unit is None:
+            angle = np.deg2rad(angle)
+        else:
+            angle = u.Unit(unit).to(u.rad, angle)
 
     s = np.sin(angle)
     c = np.cos(angle)
@@ -90,7 +91,7 @@ def rotation_matrix(angle, axis='z', unit=None):
     else:
         a1 = (i + 1) % 3
         a2 = (i + 2) % 3
-        R = np.zeros(np.shape(angle) + (3, 3))
+        R = np.zeros(getattr(angle, 'shape', ()) + (3, 3))
         R[..., i, i] = 1.
         R[..., a1, a1] = c
         R[..., a1, a2] = s
--- a/astropy/coordinates/matrix_utilities.py
+++ b/astropy/coordinates/matrix_utilities.py
@@ -62,9 +62,11 @@ def rotation_matrix(angle, axis='z', unit=None):
         A unitary rotation matrix.
     """
     if unit is None:
-        unit = u.degree
-
-    angle = Angle(angle, unit=unit)
+        angle = np.deg2rad(angle)
+    else:
+        unit = u.Unit(unit)
+        if unit is not u.rad:
+            angle = Angle(angle, unit=unit)
 
     s = np.sin(angle)
     c = np.cos(angle)
@@ -88,7 +90,7 @@ def rotation_matrix(angle, axis='z', unit=None):
     else:
         a1 = (i + 1) % 3
         a2 = (i + 2) % 3
-        R = np.zeros(angle.shape + (3, 3))
+        R = np.zeros(np.shape(angle) + (3, 3))
         R[..., i, i] = 1.
         R[..., a1, a1] = c
         R[..., a1, a2] = s


pllim:
Looks like the code in question was added in #5104

pllim:
I'll let the reviewer decide whether a change log is necessary or not.

astropy/coordinates/matrix_utilities.py
```diff
@@ -62,9 +62,11 @@ def rotation_matrix(angle, axis='z', unit=None):
         A unitary rotation matrix.
     """
     if unit is None:
-        unit = u.degree
-
-    angle = Angle(angle, unit=unit)
+        angle = np.deg2rad(angle)
```
mhvk: Hmm, I appreciate the speed-up, but wonder about legibility. In particular, I now really have to think what happens when `angle` is already a Quantity. This branch will do the right thing, but it will be rather slow compared to just getting the Quantity in radians. And the second branch will fail in the admittedly unlikely case that someone, mistakenly, passes in a quantity not in radians and sets `unit=u.rad` (admittedly unlikely!).

That said, the current default action when one passes in a `Quantity` is quite awful: first convert to degrees then to radian...

How about this instead?
```
if isinstance(angle, u.Quantity):
    angle = angle.to_value(u.radian)
else:
    unit = u.Unit(unit) if unit is not None else u.degree
    angle = unit.to(u.rad, angle)
    # the above is 1 us if unit equals u.rad; does not
    # seem worth optimizing away.
```

astropy/coordinates/matrix_utilities.py
```diff
@@ -88,7 +90,7 @@ def rotation_matrix(angle, axis='z', unit=None):
     else:
         a1 = (i + 1) % 3
         a2 = (i + 2) % 3
-        R = np.zeros(angle.shape + (3, 3))
+        R = np.zeros(np.shape(angle) + (3, 3))
```
mhvk: If we're micro-optimizing anyway, use `getattr(angle, 'shape', ()` (58 ns instead of 1.73 us for a scalar)

mcara:
Ah, I am sorry. I missed that `unit` if only used when `angle` is not a quantity. Of course this PR will not work.

mcara:
I just wanted something fast for the most common cases. The quantity slows this code down *a lot*.

mhvk:
@mcara - I think the suggestion I made in-line would still give you almost all the speed-up, so if you want to make the change, we can get that in.  Re-opening in case you want to pursue it (since you already have a PR...)

astropy/coordinates/matrix_utilities.py
mcara: I love this optimization

saimn: Or non-optimization ?
```
In [21]: %timeit np.shape(a)       
326 ns ± 0.521 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [22]: %timeit a.shape           
23.2 ns ± 0.0369 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [23]: %timeit getattr(a, 'shape', ())                              
50.5 ns ± 0.69 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

mcara:
@mhvk I know the `if` block is more complicated than what you have originally proposed, but in my tests it is 30% faster for `degrees` matching the speed of `rad`.

astropy/coordinates/matrix_utilities.py
mcara: @saimn The above tests do not include the rest of the code. For example `a.shape` will not work if `a = 15` (i.e., degrees). For the most typical use when the angle is a scalar, @mhvk optimization is great. For the same reason I switched to `np.shape()` so that if input angles are in radians and provided such as `(angle=[1, 2], unit=np.rad)` the angles can be used "as is".

mcara:
please wait.

astropy/coordinates/matrix_utilities.py
mhvk: @saimn - indeed, the reason this became necessary is that now scalars can go all the way down to here. I think the `getattr` way is the fastest one to deal with both cases.

mcara:
I think in the new code it is guaranteed that `angle` is an object with `shape` attribute. Then @saimn test can help.

mhvk:
OK, removed the "merge-when-CI-passes".

astropy/coordinates/matrix_utilities.py
```diff
@@ -61,10 +61,13 @@ def rotation_matrix(angle, axis='z', unit=None):
...
-        unit = u.degree
-
-    angle = Angle(angle, unit=unit)
+    if isinstance(angle, u.Quantity):
+        angle = angle.to_value(u.radian)
+    else:
+        if unit is None:
+            angle = np.deg2rad(angle)
+        else:
+            angle = u.Unit(unit).to(u.rad, angle)
```
mhvk: If passed in a float, this one will return a float

mcara: You are right.

mcara:
@mhvk Thank you! Now speed-up is > x10 for both degrees and radians

astropy/coordinates/matrix_utilities.py
saimn: Ah ok, I missed the scalar context ;).
But still I'm not sure 200ns are worth the complexity added by this `getattr`.

saimn: Ok, if `a` is a Python scalar I get the same 1us as @mhvk .

mhvk:
Great, an order of magnitude is well worth the if statement!  Merging...

PR #9800, opened 2019-12-18
Fix doctest race condition

### LINKED ISSUE #9796: Test failure in docs/io/ascii/index.rst
From #9711: One failure in the 32bits build : https://circleci.com/gh/astropy/astropy/56597
```
______________________ [doctest] docs/io/ascii/index.rst _______________________
[gw3] linux -- Python 3.7.5 /opt/python/cp37-cp37m/bin/python
148   Writing Data Tables as Formatted ASCII Tables
149 
150 The following writes a table as a simple space-delimited file::
151 
152   >>> import numpy as np
153   >>> from astropy.table import Table, Column, MaskedColumn
154   >>> x = np.array([1, 2, 3])
155   >>> y = x ** 2
156   >>> data = Table([x, y], names=['x', 'y'])
157   >>> ascii.write(data, 'values.dat')
UNEXPECTED EXCEPTION: AstropyDeprecationWarning("values.dat already exists. Automatically overwriting ASCII files is deprecated. Use the argument 'overwrite=True' in the future.")

  File "/opt/python/cp37-cp37m/lib/python3.7/doctest.py", line 1329, in __run
    compileflags, 1), test.globs)

  File "<doctest index.rst[13]>", line 1, in <module>

  File "/tmp/astropy-test-ur8e8wgu/lib/python3.7/site-packages/astropy/io/ascii/ui.py", line 750, in write
    output), AstropyDeprecationWarning)

astropy.utils.exceptions.AstropyDeprecationWarning: values.dat already exists. Automatically overwriting ASCII files is deprecated. Use the argument 'overwrite=True' in the future.
```

I think it happens because this is the parallelized build, and `docs/io/ascii/write.rst` also writes a file with the same name (but with `overwrite=True`). And since tests are run in parallel I guess that `write.rst` was run before `index.rst`.

astrofrog:
I've seen a failure happen in the 32-bit parallel build which is due to two doctests both needing to write to a file called ``values.dat``. I think until recently the tests were always run such that the one with ``overwrite=True`` came second but now things have flipped so an error is happening.

Ideally we should run doctests in an isolated way that prevents this from happening, but for now this is a quick fix. I'm also renaming the CircleCI build for 32-bit to mention it is also a parallel build since this is often a source of confusion.

Fixes #9796

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
       - checkout
       - run: python setup.py egg_info
 
-  32bit:
+  32bit-and-parallel:
     docker:
       - image: quay.io/pypa/manylinux1_i686
     steps:
@@ -160,7 +160,7 @@ workflows:
     jobs:
       - egg-info-37
       - html-docs
-      - 32bit
+      - 32bit-and-parallel
       - image-tests-mpl212
       - image-tests-mpl222
       - image-tests-mpl302
--- a/docs/io/ascii/index.rst
+++ b/docs/io/ascii/index.rst
@@ -154,7 +154,7 @@ The following writes a table as a simple space-delimited file::
   >>> x = np.array([1, 2, 3])
   >>> y = x ** 2
   >>> data = Table([x, y], names=['x', 'y'])
-  >>> ascii.write(data, 'values.dat')
+  >>> ascii.write(data, 'values.dat', overwrite=True)
 
 The ``values.dat`` file will then contain::
 


MSeifert04:
Related to #9796 ?

astrofrog:
@MSeifert04 - indeed, good catch! I've added a mention of that in the description so that the issue auto-closes.

mhvk:
@astrofrog - for this example, the output file is not checked, so it is fine if it changes underneath. Is that true for the other example as well?

astrofrog:
@mhvk - yes, the overwrite is safe for the other example too, the content of the files is not checked.

mhvk:
I'll merge too since those failures are very distracting.

PR #9812, opened 2019-12-21
Fix issue changing format with incompatible out_subfmt

### LINKED ISSUE #9810: Cannot access mjd property on Time with out_subfmt set to date
### Description

Accessing the ```mid``` property on a ```Time``` with custom ```out_subfmt``` set raises an exception.

### Steps to Reproduce

```python
In [3]: from astropy.time import Time                                                                                

In [4]: t = Time('2019-12-20', out_subfmt='date_hm')                                                                 

In [5]: t.mjd                                                                                                        
<ipython-input-5-bd7732b99f6e> in <module>
----> 1 t.mjd

~/anaconda3/envs/astropy_est/lib/python3.7/site-packages/astropy/time/core.py in __getattr__(self, attr)
   1694 
   1695         elif attr in self.FORMATS:
-> 1696             return self.to_value(attr, subfmt=None)
   1697 
   1698         elif attr in TIME_SCALES:  # allowed ones done above (self.SCALES)

~/anaconda3/envs/astropy_est/lib/python3.7/site-packages/astropy/time/core.py in to_value(self, format, subfmt)
    999                     raise ValueError('subformat must match one of {}'
   1000                                      .format([option[0] for option in
-> 1001                                               tm._time.subfmts])) from None
   1002                 else:
   1003                     raise

ValueError: subformat must match one of ['float', 'long', 'decimal', 'str', 'bytes']
```

### System Details

Darwin-19.2.0-x86_64-i386-64bit
Numpy 1.17.4
Scipy 1.3.1
astropy 4.1.dev27242

### LINKED ISSUE #9860: Time.to_value() can raise unexpected exception for legacy/custom Format classes
### Description

Working on #9812, I noticed that  `t.to_value('jyear_str', subfmt='adsf')` gives an unexpected result.  I'm planning to fix this in #9812 because that PR is touching this code anyway, but wanted to document this bug independently.

This is a case that is supposed to be handled more gracefully, as evidenced by the comment in `Time.to_value()`:
```
            # Custom TimeFormat subclasses created in astropy versions <= 4.0
            # may not be able to handle being passes on [sic] a out_subfmt.
            # But those do supposedly deal with `self.out_subfmt` internally,
            # so if subfmt is the same, we do not pass it on.
```

### Steps to Reproduce
```
In [2]: t = Time('B2000')

In [4]: t.to_value('jyear_str')
Out[4]: 'J1999.999'

In [5]: t.to_value('jyear_str', subfmt='adsf')
<ipython-input-5-dece9251ad92> in <module>()
----> 1 t.to_value('jyear_str', subfmt='adsf')

~/git/astropy/astropy/time/core.py in to_value(self, format, subfmt)
    994             try:
    995                 value = tm._shaped_like_input(tm._time.to_value(
--> 996                     parent=tm, **kwargs))
    997             except ValueError as exc:
    998                 if 'subformats match' in str(exc):

TypeError: to_value() got an unexpected keyword argument 'out_subfmt'
```

### Expected behavior

This exception message is confusing so I expected `TypeError` to be caught and changed into a more informative message, e.g.
```
ValueError: to_value() method for format 'jyear_str' does not support passing a 'subfmt' argument
```

### System Details

Darwin-19.2.0-x86_64-i386-64bit
Python 3.6.7 |Anaconda, Inc.| (default, Oct 23 2018, 14:01:38) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
Numpy 1.17.4
astropy 4.1.dev27320
Scipy 1.3.1

taldcroft:
### Description

This pull request is to fix an exception when changing Time format in the case where ``out_subfmt`` is a value that is incompatible with available `subfmts` in the new format class.  See #9810 for examples and discussion.  The fix here is within the ``_apply`` method, which is used for copy, replicate and many other ways that format can get changed.

It also fixes a bug I just discovered where `tm.to_value(fmt, subfmt='long')` fails for `fmt` in 'byear', 'jyear', or 'decimalyear'.

This also changes two minor problems in the changelog entry for #9679.

Fixes #9810
Fixes #9860

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -1477,12 +1477,14 @@ class Time(ShapedLikeNDArray):
         tm._format = new_format
         tm.SCALES = self.SCALES
 
-        # If current output subformat does not match any subfmts in new format
-        # then replace with default '*'.
-        try:
-            tm._time._select_subfmts(tm.out_subfmt)
-        except ValueError:
-            tm.out_subfmt = '*'
+        # If format has changed and current output subformat does not match any
+        # subfmts in new format then replace with default '*'.  If format is
+        # unchanged then the original out_subfmt is by definition OK.
+        if new_format != self.format:
+            try:
+                tm._time._select_subfmts(tm.out_subfmt)
+            except ValueError:
+                tm.out_subfmt = '*'
 
         return tm
 
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -661,18 +661,19 @@ class Time(ShapedLikeNDArray):
                              .format(list(self.FORMATS)))
         format_cls = self.FORMATS[format]
 
-        # If current output subformat is not in the new format then replace
-        # with default '*'
-        if hasattr(format_cls, 'subfmts'):
-            subfmt_names = [subfmt[0] for subfmt in format_cls.subfmts]
-            if self.out_subfmt not in subfmt_names:
-                self.out_subfmt = '*'
-
         self._time = format_cls(self._time.jd1, self._time.jd2,
                                 self._time._scale, self.precision,
                                 in_subfmt=self.in_subfmt,
                                 out_subfmt=self.out_subfmt,
                                 from_jd=True)
+
+        # If current output subformat does not match any subfmts in new format
+        # then replace with default '*'
+        try:
+            self._time._select_subfmts(self.out_subfmt)
+        except ValueError:
+            self.out_subfmt = '*'
+
         self._format = format
 
     def __repr__(self):
@@ -1468,12 +1469,21 @@ class Time(ShapedLikeNDArray):
                              .format(list(tm.FORMATS)))
 
         NewFormat = tm.FORMATS[new_format]
+
         tm._time = NewFormat(tm._time.jd1, tm._time.jd2,
                              tm._time._scale, tm.precision,
                              tm.in_subfmt, tm.out_subfmt,
                              from_jd=True)
         tm._format = new_format
         tm.SCALES = self.SCALES
+
+        # If current output subformat does not match any subfmts in new format
+        # then replace with default '*'.
+        try:
+            tm._time._select_subfmts(tm.out_subfmt)
+        except ValueError:
+            tm.out_subfmt = '*'
+
         return tm
 
     def __copy__(self):
--- a/astropy/time/formats.py
+++ b/astropy/time/formats.py
@@ -129,6 +129,7 @@ class TimeFormat(metaclass=TimeFormatMeta):
     """
 
     _default_scale = 'utc'  # As of astropy 0.4
+    subfmts = ()
 
     def __init__(self, val1, val2, scale, precision,
                  in_subfmt, out_subfmt, from_jd=False):
--- a/astropy/time/formats.py
+++ b/astropy/time/formats.py
@@ -293,15 +293,43 @@ class TimeFormat(metaclass=TimeFormatMeta):
         """
         raise NotImplementedError
 
-    def to_value(self, parent=None):
+    def to_value(self, parent=None, out_subfmt=None):
         """
-        Return time representation from internal jd1 and jd2.  This is
-        the base method that ignores ``parent`` and requires that
-        subclasses implement the ``value`` property.  Subclasses that
-        require ``parent`` or have other optional args for ``to_value``
-        should compute and return the value directly.
+        Return time representation from internal jd1 and jd2 in specified
+        ``out_subfmt``.
+
+        This is the base method that ignores ``parent`` and uses the ``value``
+        property to compute the output. This is done by temporarily setting
+        ``self.out_subfmt`` and calling ``self.value``. This is required for
+        legacy Format subclasses prior to astropy 4.0  New code should instead
+        implement the value functionality in ``to_value()`` and then make the
+        ``value`` property be a simple call to ``self.to_value()``.
+
+        Parameters
+        ----------
+        parent : obj
+            Parent `~astropy.time.Time` object associated with this
+            `~astropy.time.TimeFormat` object
+        out_subfmt : str or `None`
+            Output subformt (use existing self.out_subfmt if `None`)
+
+        Returns
+        -------
+        value : numpy.array, numpy.ma.array
+            Array or masked array of formatted time representation values
         """
-        return self.mask_if_needed(self.value)
+        # Get value via ``value`` property, overriding out_subfmt temporarily if needed.
+        if out_subfmt is not None:
+            out_subfmt_orig = self.out_subfmt
+            try:
+                self.out_subfmt = out_subfmt
+                value = self.value
+            finally:
+                self.out_subfmt = out_subfmt_orig
+        else:
+            value = self.value
+
+        return self.mask_if_needed(value)
 
     @property
     def value(self):
@@ -488,7 +516,7 @@ class TimeDecimalYear(TimeNumeric):
         dt_end = (jd1_end - jd1_start) + (jd2_end - jd2_start)
         decimalyear = iy_start + dt / dt_end
 
-        return super().to_value(jd1=decimalyear, jd2=0., **kwargs)
+        return super().to_value(jd1=decimalyear, jd2=np.float64(0.0), **kwargs)
 
     value = property(to_value)
 
@@ -773,7 +801,7 @@ class TimeDatetime(TimeUnique):
                               *iterator.operands[1:])
         self.jd1, self.jd2 = day_frac(jd1, jd2)
 
-    def to_value(self, timezone=None, parent=None):
+    def to_value(self, timezone=None, parent=None, out_subfmt=None):
         """
         Convert to (potentially timezone-aware) `~datetime.datetime` object.
 
@@ -790,6 +818,11 @@ class TimeDatetime(TimeUnique):
         `~datetime.datetime`
             If ``timezone`` is not ``None``, output will be timezone-aware.
         """
+        if out_subfmt is not None:
+            # Out_subfmt not allowed for this format, so raise the standard
+            # exception by trying to validate the value.
+            self._select_subfmts(out_subfmt)
+
         if timezone is not None:
             if self._scale != 'utc':
                 raise ScaleValueError("scale is {}, must be 'utc' when timezone "
@@ -1390,7 +1423,7 @@ class TimeEpochDate(TimeNumeric):
     def to_value(self, **kwargs):
         jd_to_epoch = getattr(erfa, self.jd_to_epoch)
         value = jd_to_epoch(self.jd1, self.jd2)
-        return super().to_value(jd1=value, jd2=0., **kwargs)
+        return super().to_value(jd1=value, jd2=np.float64(0.0), **kwargs)
 
     value = property(to_value)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -399,8 +399,12 @@ astropy.time
 
 - Fix inaccuracy when converting between TimeDelta and datetime.timedelta. [#9679]
 
-- Fix exception when changing ``format`` in the case when ``out_subfmt`` is
+- Fixed exception when changing ``format`` in the case when ``out_subfmt`` is
   defined and is incompatible with the new format. [#9812]
+
+- Fixed exceptions in ``Time.to_value()``: when supplying any ``subfmt`` argument
+  for string-based formats like 'iso', and for ``subfmt='long'`` for the formats
+  'byear', 'jyear', and 'decimalyear'. [#9812]
 
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -397,6 +397,11 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- Fix inaccuracy when converting between TimeDelta and datetime.timedelta. [#9679]
+
+- Fix exception when changing ``format`` in the case when ``out_subfmt`` is
+  defined and is incompatible with the new format. [#9812]
+
 astropy.timeseries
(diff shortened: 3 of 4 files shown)


mhvk:
Looks OK, except that, as I noted, the `subfmt` are allowed to be globs:  https://github.com/astropy/astropy/blob/064ba7fa65fe47f05c9486d80c809215ee818246/astropy/time/core.py#L351-L352

I think this will break on this (admittedly fairly contrived) example 
```
Time('2012-12-12', format='isot', out_subfmt='*hm').iso
```
This is partially why I suggested punting it to `TimeFormat`. Though perhaps easier is just to pass it through `_select_subfmts` and set to '*' on `ValueError`.

taldcroft:
Good catch @mhvk.  It turns out that this `out_subfmt` checking code was already in the `format` setter property (which I had vaguely recalled but somehow didn't get there).  That code was basically the same as my new version and suffered from the same problem.  So now both places are fixed to handle globs.

This does make a somewhat bigger/out-of-scope change, namely adding `subfmts = ()` as a base attribute so that `subfmts` always exists on every TimeFormat subclass.  That seems reasonable to me, but could always make this patch work without that, albeit uglier code.

StuartLittlefair:
Nice work @taldcroft. I'm not 100% sure this fixes the issue though - bear with me because I didn't write this production code, but imagine the following...

```python
>>> t = Time(58837.72643888889, format='mjd')                                          
>>> t.out_subfmt = 'date_hm'                                                           
>>> t.iso                                                                              
'2019-12-20 17:26'
>>> t.mjd                                                                              
<ipython-input-44-bd7732b99f6e> in <module>
----> 1 t.mjd

~/anaconda3/envs/astropy_est/lib/python3.7/site-packages/astropy/time/core.py in __getattr__(self, attr)
   1704 
   1705         elif attr in self.FORMATS:
-> 1706             return self.to_value(attr, subfmt=None)
   1707 
   1708         elif attr in TIME_SCALES:  # allowed ones done above (self.SCALES)

~/anaconda3/envs/astropy_est/lib/python3.7/site-packages/astropy/time/core.py in to_value(self, format, subfmt)
   1000                     raise ValueError('subformat must match one of {}'
   1001                                      .format([option[0] for option in
-> 1002                                               tm._time.subfmts])) from None
   1003                 else:
   1004                     raise
```

Now, I think that's a bug in the *user's* code but I'd argue that you shouldn't be able to set an invalid ```out_subfmt```. I think the ```t.out_subfmt = 'date_hm'``` line should raise an Exception.

taldcroft:
About doing the `out_subfmt` checking / changing to `*` in `TimeFormat`, I think that makes trouble for validation of `out_subfmt` (which I just added in `Time.__init__` ). 

The `_select_subfmts` method is an instance method, so it is more convenient to make the `TimeFormat` object first and then check if `out_subfmt` is acceptable.  So that requires making `TimeFormat` with a potentially invalid `out_subfmt` but leaving it alone.  If `TimeFormat` just changed that unacceptable value to `*` then one could create a `Time` object with any unacceptable `out_subfmt` and it would likewise get silently changed to `*`.

taldcroft:
@StuartLittlefair - I edited my last comment, please refresh.  Some current code is relying on the ability to have an invalid `out_subfmt`, temporarily.  I'll try putting in validation when setting `out_subfmt`, but if it breaks anything else like YAML then will need to punt that out as a bigger scope problem.

taldcroft:
Nope, adding validation of `Time.out_subfmt` to ensure that you get an exception immediately when setting a bad value breaks everything (as in you can't even import Time).  That is a bigger problem where the current permissive behavior has been relied upon in other code.

StuartLittlefair:
Fair enough. In that case, my feeling is that *if* setting an invalid ```out_subfmt``` doesn't raise an error, then the code I pasted [above](https://github.com/astropy/astropy/pull/9812#issuecomment-568174466) ought to work...

That snippet tests the edge case where the ```format``` argument in ```to_value``` is equal to ```self.format``` so the call to replicate is bypassed. Therefore I think we need an additional check inside ```to_value``` along the lines of...

```python
            if format == self.format:
                tm = self
                if subfmt is None and subfmt not in self._time.subfmts:
                    subfmt = '*'
```

Ugly, but needed?

mhvk:
@taldcroft - would changing `_select_subfmt` to either a `classmethod` or a [`sharedmethod`](https://github.com/astropy/astropy/blob/064ba7fa65fe47f05c9486d80c809215ee818246/astropy/utils/decorators.py#L770) help? I guess the problem with mucking around too much with `TimeFormat` is that people have created subclasses and those should continue to work.

The other option is to go wtih @StuartLittlefair's suggestion and just check in `to_value` only. It might be conceptually similar to stating that there is a default fall-back for each format. When I looked at the issue initially, I also wondered about having a special token for `subfmt` in `to_value()` that does that, and which is used for internal calls only.

taldcroft:
> @taldcroft - would changing _select_subfmt to either a classmethod or a sharedmethod help?

I'm not too excited about making a change that might break custom subclasses, so let's say that option is off the table.  (And I have not tried to see if that would actually help).

> The other option is to go wtih @StuartLittlefair's suggestion and just check in to_value only. 

I spent a little time looking into this and thinking on it.  The real problem right now is that there is a blocker (the YAML serialization) to doing the right solution which is up-front validation of `out_subfmt`.  At this moment we require being able to set an invalid value, and I don't want to dig into fixing that just now.

But if I understand, the option that is being suggested is to convert any invalid value to '*' and press on.  I think that option is worse than the current problem of latent (on-demand) validation that occurs only when you try to use `out_subfmt` by getting a `.value`.  Basically this leads to:
```
>>> t = Time('2000-01-02')
>>> t.out_subfmt = 'asdfasdf'
>>> t.value  # out_subfmt is not valid, gets converted to '*'
'2000-01-02 00:00:00.000'
```
Current behavior is better:
```
ValueError: subformat must match one of ['date_hms', 'date_hm', 'date']
```

So this current PR fixes the real problem in #9810 while leaving the on-demand validation in place. 

Note that currently when calling `tm.value` that calls `tm.to_value(self.format, None)`, so the `subfmt` argument is always `None` in this situation.  That was a point of confusion for me initially, and I think that @StuartLittlefair suggestion is imagining that `subfmt` is not always `None`.

StuartLittlefair:
@taldcroft I'm not sure I agree that the current behaviour is better. Adding a check in ```to_value``` which converts invalid sub formats to '*' does create the behaviour you indicate above, but that is not *wrong*, it's just that the output sub format is not what is expected. 

OTOH, the current behaviour means that something I consider to be valid code - eg the [snippet](https://github.com/astropy/astropy/pull/9812#issuecomment-568174466) I posted actually causes an exception. IMO this is a worse sin. Bear in mind that I posted this issue because someone sent me some reasonable code that uses astropy in the wild. It used to work, but now crashes with up-to-date versions of astropy...

taldcroft:
> OTOH, the current behaviour means that something I consider to be valid code - eg the snippet I posted actually causes an exception.

Here is where we do disagree.  Here is what *should* be the behavior, namely immediate validation of `out_subfmt` to be a legal subfmt for the `mjd` format:
```
>>> t = Time(58837.72643888889, format='mjd')                                          
>>> t.out_subfmt = 'date_hm'                                                           
ValueError: 'date_hm'  is not an allowed value of out_subfmt for 'mjd'
```
So the user would **immediately** get an exception saying that the subfmt is invalid.  Totally reasonable validation of object attributes and consistent with normal practices.

What happens instead is that they get the same exception, but only later "on-demand" when actually doing `t.mjd` or `t.value`.  So it is certainly confusing, but essentially equivalent to the desired behavior above that an invalid value of `out_subfmt` (for the relevant format) causes an exception.

taldcroft:
Conversely, silently converting an invalid value for the current format to `*` is a bad idea.  Imagine a "fat-finger" typo:
```
t = Time('2001-01-01 12:14:16', out_subfmt='date_hn')
print(t)
2001-01-01 12:14:16.000
```
User is confused why the output is not `2001-01-01 12:14`, stares and code and doesn't notice the tiny typo.  Converting to `*` (if necessary) when format changes is a different story, basically because there is no other reasonable option.

StuartLittlefair:
Ok - fair enough. Perhaps it would be possible to raise a warning when invalid sub formats are set without breaking YAML serialisation? 

This would help people with code that is now broken to work out what the offending line is...

taldcroft:
A warning like that seems reasonable.  At this point though I will probably start by giving a quick look at what it would take to make YAML work if there is an exception for setting a bad value.  I would need to touch that code anyway to suppress the warning.

taldcroft:
I let myself crawl down a rabbit hole of trying to validate `in_subfmt` and `out_subfmt` at the level of the `TimeFormat` subclasses.  It seems like a good thing, but a fair bit of code in astropy.time seems to depend on allowing invalid values to hang out there.  Anyway, tests are not yet passing and I feel like a bit of whack-a-mole, but maybe it will all sort out in the end.

In any case, I have basically achieved the initial goal of validating `out_subfmt` at the `Time` class level, so I can probably back off to that if necessary.   It just requires a lot more code repetition because the `out_subfmt` validation needs to be done in many separate places.

BTW the YAML fix turned out to be easy!

mhvk:
@taldcroft - this does look quite good - especially if it is indeed possible to move the validation inside `TimeFormat` (where you'll automatically get more of a symmetry with `in_subfmt`).

astropy/time/core.py
```diff
@@ -984,24 +991,35 @@ def to_value(self, format, subfmt='*'):
...
+                # Try validating subfmt, e.g. for formats like 'jyear_str' that
+                # do not implement out_subfmt in to_value() (because there are
+                # no allowed subformats).  If subfmt is not valid this gives the
+                # same exception as would have occurred if the call to
+                # `to_value()` had succeeded.
+                tm._time._select_subfmts(subfmt)
+
+                # Subfmt was valid, so fall back to the original exception to see
+                # if it was lack of support for out_subfmt as a call arg.
+                if "unexpected keyword argument 'out_subfmt'" in str(exc):
```
StuartLittlefair: Perhaps this is a separate issue to the one addressed in this PR, but shouldn't we do more than raise an exception here? i.e shouldn't the following actually work, rather than raise an exception?

```python
t = Time(58837.72643888889, format='mjd')
t.to_value('iso', subfmt='date_hm') 
```

taldcroft: Good catch, that is definitely a case that should work!  That is failing in 4.0 as well.  Looks like the existing test coverage for `to_value` needs some work since this version is passing tests now.  On it...

taldcroft:
And of course I realize now that this PR (with all 4 commits) is now no longer OK for 4.0.1 (and LTS in general) because it changes API with the up-front validation of in/out_subfmt.  It would be good to fix the `t.to_value('iso', subfmt='date_hm')` bug as well in LTS.

Probably we can just go with the first commit here which does (I think) fully address #9810, base a new PR of that commit for the up-front validation and code re-organization that results.

mhvk:
Can I be the devil's advocate and say this does fix at least a dormant bug?!

taldcroft:
Well... this definitely has non-small potential to break code, as evidenced by all the code in our core itself that need fixing.  Putting this PR right now as a bug fix seems like not a good idea.

mhvk:
True...  let's break it into two PRs then...

taldcroft:
OK this is now a bug-fix only PR for 4.0.1, with the subfmt validation now in #9868.

taldcroft:
Agh, unrelated FFT test fails on this.   Any clues?  This is based off master from Jan 13 (f7c6c2af3).

astropy/time/core.py
```diff
@@ -1468,12 +1469,21 @@ def _apply(self, method, *args, format=None, **kwargs):
...
         NewFormat = tm.FORMATS[new_format]
+
         tm._time = NewFormat(tm._time.jd1, tm._time.jd2,
                              tm._time._scale, tm.precision,
                              tm.in_subfmt, tm.out_subfmt,
                              from_jd=True)
         tm._format = new_format
         tm.SCALES = self.SCALES
+
+        # If current output subformat does not match any subfmts in new format
```
mhvk: Can we put this in an `if new_format is not self.format:` clause to avoid slowing down regular operations like indexing and reshaping?

astropy/time/formats.py
```diff
@@ -487,7 +516,7 @@ def to_value(self, **kwargs):
         dt_end = (jd1_end - jd1_start) + (jd2_end - jd2_start)
         decimalyear = iy_start + dt / dt_end
 
-        return super().to_value(jd1=decimalyear, jd2=0., **kwargs)
+        return super().to_value(jd1=decimalyear, jd2=np.float64(0.0), **kwargs)
```
mhvk: Why this change? (Out of curiosity, doesn't matter too much)

taldcroft: There are a couple places, e.g. `utils.twoval_to_longdouble` that use `val2.astype(..)`.  A standard Python float does not have `.astype()` and this crashes.  The original testing did not cover this case.

astropy/time/core.py
taldcroft: Done.  I used `!=` not `is not` since the user might supply a `format` string which is the same value as the current format.

astropy/time/formats.py
mhvk: If we want to ensure things are always arrays (probably good, since they can be `.view`d, etc.), maybe just do `np.array(0.)`?

taldcroft: Does it matter?  Both cases in question are just `val2=..` args in `to_value()` which combine `val1` and `val2` into the single "value".  The `val2` value is lost in that process.

mhvk: OK, good point. Let's keep it as you have it.

taldcroft:
@mhvk @StuartLittlefair  - this is passing all tests and from what I can see there are no outstanding issues from the review.  Is it ready?

PR #9813, opened 2019-12-21
Remove old erfa cruft needed for numpy < 1.16

mhvk:
While working on #9804, I noticed that the numpy >=1.16 cleanup had missed some stuff in `_erfa`. This removes parts of work-arounds that now never are called any more.


pllim:
`mpldev` failure is unrelated. FYI.

mhvk:
Tests do pass, would be nice to get this in; cc @astrofrog, @eteq.

PR #9838, opened 2020-01-05
Fix a memory corruption issue in WCS due to uninitialized wcsprm.flag before wcscopy

mcara:
This PR fixes a memory corruption issue in WCS. This memory corruption might be the likely reason for unit tests failures in https://github.com/astropy/astropy/pull/9641.

Fundamentally, currently the code does not set the flag of `Wcsprm` _before_ calling `wcscopy`. The most important change in this PR is the addition of the line:

```c
self->x.flag = -1;
```

before `wcscopy()`. `memset` was added out of caution.

CC: @nden, @MSeifert04 

I will rebase https://github.com/astropy/astropy/pull/9641 after this one is merged.

### DIFF
--- a/astropy/wcs/src/wcslib_wrap.c
+++ b/astropy/wcs/src/wcslib_wrap.c
@@ -181,6 +181,7 @@ static PyWcsprm*
 PyWcsprm_cnew(void) {
   PyWcsprm* self;
   self = (PyWcsprm*)(&PyWcsprmType)->tp_alloc(&PyWcsprmType, 0);
+  memset(&self->x, 0, sizeof(struct wcsprm));
   return self;
 }
 
@@ -192,6 +193,7 @@ PyWcsprm_new(
 
   PyWcsprm* self;
   self = (PyWcsprm*)type->tp_alloc(type, 0);
+  memset(&self->x, 0, sizeof(struct wcsprm));
   return (PyObject*)self;
 }
 
@@ -257,7 +259,6 @@ PyWcsprm_init(
       return -1;
     }
 
-    note_change(self);
     self->x.flag = -1;
     status = wcsini(1, naxis, &self->x);
 
@@ -434,6 +435,7 @@ PyWcsprm_init(
       return -1;
     }
 
+    self->x.flag = -1;
     if (wcscopy(1, wcs + i, &self->x) != 0) {
       wcsvfree(&nwcs, &wcs);
       PyErr_SetString(
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,6 +91,9 @@ astropy.visualization
 astropy.wcs
 ^^^^^^^^^^^
 
+- Fix a memory corruption issue in WCS due to uninitialized value of
+  ``wcsprm.flag`` before ``wcscopy``. [#9838]
+
 
 API Changes
 -----------


astropy/wcs/src/wcslib_wrap.c
```diff
@@ -181,6 +181,7 @@ static PyWcsprm*
 PyWcsprm_cnew(void) {
   PyWcsprm* self;
   self = (PyWcsprm*)(&PyWcsprmType)->tp_alloc(&PyWcsprmType, 0);
+  memset(&self->x, 0, sizeof(struct wcsprm));
```
MSeifert04: `tp_alloc` initializes the memory with zeros:

>  It should return a pointer to a block of memory of adequate length for the instance, suitably aligned, and initialized to zeros[...]

So `self->x` should be `NULL`, shouldn't it? And I thought `memset` would be undefined behavior if *dest* is NULL.

However this is just educated guess, I haven't looked into this in detail.

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -257,7 +259,6 @@ PyWcsprm_init(
       return -1;
     }
 
-    note_change(self);
```
MSeifert04: Good catch, makes no sense to set flag to 0 if it's set to -1 in the following line 👍

astropy/wcs/src/wcslib_wrap.c
```diff
@@ -434,6 +435,7 @@ PyWcsprm_init(
       return -1;
     }
 
+    self->x.flag = -1;
```
MSeifert04: I think that's wrong, but I may be mistaken. `self->x` should already have allocated memory, so having a -1 would lead to a memory leak.

If this is necessary for `wcscopy` (I haven't seen something like that in the [documentation](https://www.atnf.csiro.au/people/mcalabre/WCS/wcslib/wcs_8h.html#ac55946dadc53ac592cb686275902ae7b)) then we probably have to adjust a few more calls we have in our extension and it would be good to ask for a clarification on this in the documentation.

Or am I missing something?

MSeifert04:
However there seems to be one `x.flag = -1` missing in the `PyWcsprm_copy` function.

astropy/wcs/src/wcslib_wrap.c
MSeifert04: It could be problematic because `__init__` may be called multiple times on the same wcs object and then the flag shouldn't be set to `-1` for subsequent calls. I'm not sure if that's possible through our `WCS` wrapper.

MSeifert04:
I have edited most of my comments because my statements were partially incorrect. 

More general note: After reading about `wcssub` and `wcscopy` I believe there could be more bugs lurking in our wrapper because we're not always setting the flag correctly there. However this is really tricky stuff and I (or someone else) may have to check the coverage results to verify that we really exercise the problematic branches.

astropy/wcs/src/wcslib_wrap.c
mcara: Ah, yes. It was in my "C equivalent" (that excludes Python from the equation) that the block was not initialized.

mcara:
@MSeifert04 Hm, I may have to retract this PR if what you said about `tp_alloc` is true. I will not close it for now but I am investigating more what is going on.

By the way, setting x.flag = -1 _for the destination WCS object_ will not produce memory memory leaks (I do not think so).

MSeifert04:
From the documentation on `wcssub` which is called by `wcscopy`:

> [in,out] | wcsdst | Struct describing the subimage. wcsprm::flag should be set to -1 if wcsdst was not previously initialized (memory leaks may result if it was previously initialized).

MSeifert04:
I think setting the `flag` to -1 where you did seems appropriate. However I'm really not very familiar with the subtleties of wcslib, so take that statement with a grain of salt.

pllim:
@mcara , did you try to apply this patch to your other PR and see if the Windows segfault really goes away?

mcara:
Yeah, this PR does nothing to fix crashes in tests in #9641. Even though `self->x.flag = -1;` is likely the correct thing to do here, it practically has no effect because, if the `wcsprm` structure is initialized to 0 (as `tp_alloc` does) then this would be effectively equivalent to setting flag to 1. My mistake was that my C code did not simulate correctly the behavior of `tp_alloc`.

mcara:
I will just remove the unnecessary `note_change(self);` line in  #9641.

PR #9845, opened 2020-01-08
Fix memory corruption bugs in WCSLIB/wcscopy for WCS-TAB

mcara:
This PR implements bug fixes in WCSLIB that lead to memory corruption in wcscopy when a WCS has -TAB. This should fix memory corruption issues in https://github.com/astropy/astropy/pull/9641 PR. No unit tests added in this PR but the affected code gets invoked *only when* -TAB is present in the WCS so it should not affect existing functionality.

### DIFF
--- a/cextern/wcslib/C/wcs.c
+++ b/cextern/wcslib/C/wcs.c
@@ -1260,7 +1260,7 @@ int wcssub(
     for (m = 0; m < wcssrc->tab[itab].M; m++) {
       i = wcssrc->tab[itab].map[m];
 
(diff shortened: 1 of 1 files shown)


mcara:
CC: @nden @MSeifert04

nden:
The bug and fix were reported upstream. I will merge this patch meanwhile so we can proceed with the 
`-TAB` PR.

pllim:
Now you can finally sleep at night, @mcara ! 😅 👏

PR #9849, opened 2020-01-09
Update docstring & keywords for isclose and allclose in units

namurphy:
I corrected the one-liner at the top of the docstring for `astropy.units.isclose`, since it was the one-liner that was meant for `np.allclose`.  While I was at it, I expanded the docstrings to list the parameters, describe the requirements on the units of `rtol` and `atol`, and add a section on what exceptions are raised.

I made the `equal_nan` keyword argument explicit since it has been in numpy for several years now (I think).  Since `isclose` and `allclose` both took `**kwargs` already, I don't think it changes anything.

I set the default value of `atol` to zero instead of `None` in order to be more explicit about what the defaults are.  I updated the error messages to f-strings because they are awesome.

### DIFF
--- a/astropy/units/quantity.py
+++ b/astropy/units/quantity.py
@@ -1736,30 +1736,44 @@ class SpecificTypeQuantity(Quantity):
         super()._set_unit(unit)
 
 
-def isclose(a, b, rtol=1.e-5, atol=None, **kwargs):
+def allclose(a, b, rtol=1.e-5, atol=0, equal_nan=False, **kwargs):
     """
+    Return `True` if two arrays are element-wise equal within a
+    tolerance, and `False` otherwise.
+
+    Parameters
+    ----------
+    a, b : array_like or :class:`~astropy.units.Quantity`
+        Input values or arrays to compare
+    rtol : array_like or dimensionless :class:`~astropy.units.Quantity`
+        The relative tolerance for the comparison, which defaults to
+        ``1e-5``.  If ``rtol`` is a :class:`~astropy.units.Quantity`,
+        then it must be dimensionless.
+    atol : number or :class:`~astropy.units.Quantity`
+        The absolute tolerance for the comparison.  The units (or lack
+        thereof) of ``a``, ``b``, and ``atol`` must be consistent with
+        each other.  If `None`, ``atol`` defaults to zero.
+    equal_nan : `bool`
+        Whether to compare NaN’s as equal. If `True`, NaNs in ``a`` will
+        be considered equal to NaN’s in ``b``.
+
     Notes
     -----
-    Returns True if two arrays are element-wise equal within a tolerance.
-
-    This is a :class:`~astropy.units.Quantity`-aware version of
-    :func:`numpy.isclose`.
-    """
-    return np.isclose(*_unquantify_allclose_arguments(a, b, rtol, atol),
-                      **kwargs)
-
-
-def allclose(a, b, rtol=1.e-5, atol=None, **kwargs):
-    """
-    Notes
-    -----
-    Returns True if two arrays are element-wise equal within a tolerance.
-
     This is a :class:`~astropy.units.Quantity`-aware version of
     :func:`numpy.allclose`.
+
+    Raises
+    ------
+    UnitsError
+        If the dimensions of ``a``, ``b``, or ``atol`` are incompatible,
+        or if ``rtol`` is not dimensionless.
+
+    See also
+    --------
+    isclose
     """
-    return np.allclose(*_unquantify_allclose_arguments(a, b, rtol, atol),
-                       **kwargs)
+    unquantified_args = _unquantify_allclose_arguments(a, b, rtol, atol)
+    return np.allclose(*unquantified_args, equal_nan=equal_nan, **kwargs)
 
 
 def _unquantify_allclose_arguments(actual, desired, rtol, atol):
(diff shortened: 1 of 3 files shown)


pllim:
I think Travis CI failure is related but you can ignore `mpldev`.

astropy/units/quantity.py
```diff
@@ -1769,21 +1825,25 @@ def _unquantify_allclose_arguments(actual, desired, rtol, atol):
     try:
         desired = desired.to(actual.unit)
     except UnitsError:
-        raise UnitsError("Units for 'desired' ({}) and 'actual' ({}) "
-                         "are not convertible"
-                         .format(desired.unit, actual.unit))
+        raise UnitsError(
+            f"Units for `desired` ({desired.unit}) and `actual` "
```
mhvk: I'd stick with the regular single quotes around 'desired' and 'actual'

astropy/units/quantity.py
```diff
@@ -1736,30 +1736,86 @@ def _set_unit(self, unit):
...
-    return np.isclose(*_unquantify_allclose_arguments(a, b, rtol, atol),
-                      **kwargs)
+    unquantified_args = _unquantify_allclose_arguments(a, b, rtol, atol)
+    return np.isclose(*unquantified_args, equal_nan=equal_nan, **kwargs)
 
 
-def allclose(a, b, rtol=1.e-5, atol=None, **kwargs):
+def allclose(a, b, rtol=1.e-5, atol=None, equal_nan=False, **kwargs):
     """
+    Return `True` if two arrays are element-wise equal within a
```
mhvk: I realize the first line comes directly from numpy, but how about:
```
Whether two arrays are element-wise equal within a tolerance.
```

astropy/units/quantity.py
```diff
@@ -1736,30 +1736,86 @@ def _set_unit(self, unit):
         super()._set_unit(unit)
 
 
-def isclose(a, b, rtol=1.e-5, atol=None, **kwargs):
+def isclose(a, b, rtol=1.e-5, atol=None, equal_nan=False, **kwargs):
     """
+    Returns a boolean array where two arrays are element-wise equal
```
mhvk: Similarly, maybe
```
Whether elements of two arrays are equal within a tolerance
```
I think the numpy docstring is not very good...

But absolutely fine to stick with what you have!

astropy/units/quantity.py
namurphy: Sounds good; I'll make the change.

astropy/units/quantity.py
mhvk: Actually, this is also why the tests fail - the exception message is explicitly checked.

astropy/units/quantity.py
namurphy: I'm wondering if this change could potentially lead to some ambiguity (like making it seem like `isclose` is doing what `allclose` does), so I'd prefer something closer to the original from NumPy.

astropy/units/quantity.py
namurphy: Looks like that was the cause of the test failure!  I suppose I'm getting too used to ReST docstrings...

astropy/units/quantity.py
mhvk: No worries, just do what you think is best.

astropy/units/quantity.py
```diff
@@ -1769,26 +1824,30 @@ def _unquantify_allclose_arguments(actual, desired, rtol, atol):
...
-                         "are not convertible"
-                         .format(desired.unit, actual.unit))
+        raise UnitsError(
+            f"Units for 'desired' ({desired.unit}) and 'actual' "
+            f"({actual.unit}) are not convertible"
+        )
 
     if atol is None:
-        # by default, we assume an absolute tolerance of 0
+        # By default, we assume an absolute tolerance of 0.  The default
```
mhvk: THis is not strictly true: values of 0 are treated as being allowed to have any unit.

mhvk:
Changes look good - just need to chase down the testing error...

astropy/units/quantity.py
namurphy: The code after that converts `atol` into a `Quantity`, and then tries to convert `atol` to the same units as `actual`.  That means that if I do `u.isclose(1 * u.m, 1.1 * u.m, atol=0)` then I get a `u.UnitsError`.  I think this way is a bit more robust, since it's more explicit that the absolute tolerance does need consistent units.

namurphy:
Huh...looks like there are tests of `allclose` and `isclose` in both:

 - `astropy/units/tests/test_quantity_helpers.py` (tests of functionality)
 - `astropy/tests/tests/test_quantity_helpers.py` (tests of exceptions)

While I'm at it, I think I'll move the exceptions tests from the latter file into the former file so that they're all in one place.  There's no chance I'll cause more tests to fail, right? :joy_cat:

astropy/units/quantity.py
mhvk: Ah, yes, that right, because we make a `Quantity` first, the 0 doesn't work. In any case, I agree that `None` is a clearer default here.

mhvk:
I thought those tests in `tests/tests` are specifically for `assert_quantity_allclose`, which is defined in `tests` - I think it is easiest to just keep things there.

namurphy:
> I thought those tests in `tests/tests` are specifically for `assert_quantity_allclose`, which is defined in `tests` - I think it is easiest to just keep things there.

Oh!  I got confused since the test failures were arising there.  The main problem was me changing the quotes from ` to ' and vice versa, and its in `astropy/tests/tests` that the error messages were being tested.  In that case, since the tests are passing, I'm happy for it to be merged if you think it's ready.  Thank you for the helpful code reviews!

PR #9856, opened 2020-01-12
Add get_constellation edge tests

### LINKED ISSUE #9855: get_constellation() should use exact HMS, DMS boundaries, not 4-decimal approximations
### Description

IAU's official constellation boundaries are defined using HMS / DMS coordinates in 
E. Delporte | Délimitation scientifique des constellations, cartes, Cambridge University Press, London 1930.

But the data that drives the astropy `get_constellation()` routine is derived by Roman et al. (1987) from Delporte's work, using a lossy conversion to decimal hourangles and degrees with just 4 decimal places.

Roman's data is archived at http://cdsarc.u-strasbg.fr/viz-bin/Cat?VI/42
astropy's copy is in `constellation_data_roman87.dat` file.

For example, the 4th line is:

https://github.com/astropy/astropy/blob/51c619e5ffe7aaf748e7c34d08c6336762decc74/astropy/coordinates/data/constellation_data_roman87.dat#L4

It represents an edge between Ursa Minor (`UMi` there) and Cepheus.

It matches data on page 14 of Delporte's book, as reproduced at http://www.atlascoelestis.com/del%20inice%20cost.htm, where
 the corresponding data shows up as the 2nd line after the Ursa Minor sequence starts:

> Paralléle de 86° 10' de 21 h. o m. à 23 h. o m.

We thus see that the original parallel at dec 86° 10' was rounded off by Roman to 86.1667.

(Note that edges are listed twice in Delporte's book, once for each constellation, so it also shows up on page 39 for UMi.)

A better digital representation of Delporte's IAU constellation standard is provided in multiple formats at [Pierre Barbier's Constellation Boundaries](http://pbarbier.com/constellations/boundaries.html)

For example, line 195 of his edges_18.txt file shows the same edge represented exactly in the native sexagesimal format:

```
318:319 P+ 21:00:00 +86:10:00 23:00:00 +86:10:00 CEP UMI
```

Based on the roundoff, astropy thus would find the wrong constellation for coordinates like 22h 86.16668°

While trying to demonstrate this, I ran across what seems to be another issue that I understand less well. `get_constellation()` calls `PrecessedGeocentric(equinox='B1875')` which throws the coordinates off a bit as I document below, interpreting them as geocentric with a distance inferred somehow I guess.

https://github.com/astropy/astropy/blob/b1addf5ae4ae8c18b048940c460a3be2735f5392/astropy/coordinates/funcs.py#L235

While this might make sense for a SkyCoord pointing to a comet or other solar system object, it makes no sense to me when trying to figure out which constellation a distant variable star is in. Perhaps there is a more appropriate way to call the functions, but the methods I'm using below are similar to those in `test_constellations()` in `test_funcs.py`, so I'm still confused.  Note that those tests don't test the edge case (pun intended) that I'm discussing there. I have some additional tests which will, which I'll add to that file in a PR.

To be precise:

### Actual behavior

```
from astropy.coordinates import SkyCoord, FK5, get_constellation, PrecessedGeocentric

print('Coordinate\tTransformed\tConstellation\tHand annotation')

for dec in [86.16, 86.1666, 86.16668, 86.1668]:
    c = FK5(22*u.hour, dec*u.deg, equinox='B1875')
    constel = get_constellation(c)
    pg = c.transform_to(PrecessedGeocentric(equinox='B1875'))
    print(f'{dec:<10.8}\t{pg.dec:.8}\t{constel}')
```

Annotated output:
```
Coordinate      Transformed     Constellation   Hand annotation
86.16           86.164512 deg   Cepheus     # Right
86.1666         86.171112 deg   Ursa Minor  # Wrong
86.16668        86.171192 deg   Ursa Minor  # Right for wrong reason
86.1668         86.171312 deg   Ursa Minor  # Right
```

### Expected behavior

I expected the 2nd row (86.1666) to properly be located in Cepheus.
I expected the transformed column (showing the values that are actually looked up in the table, currently generated when PrecessedGeocentric has been applied to the coordinate) above to match the Coordinate column.
After the issue with PrecessedGeocentric is resolved, I expect the 3rd row to continue to show up in Ursa Minor.

### Steps to Reproduce
See code above.

### Notes

I also note that at http://pbarbier.com/terms.html Barbier says "All the original content on his site is licensed under a Creative Commons Attribution-ShareAlike 4.0 International License, so we should be able to re-use it.

### System Details

Linux-4.15.0-72-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
Numpy 1.18.1
astropy 4.0

nealmcb:
WIP!

### Description

The new tests expose two problems in original get_constellation code:
Next step: fix them....

- [x] Add test case near rounded-off constellation edge
- [ ] Fix decimal rounding errors in boundaries from Roman's data
- [ ] Fix improper (?) use of PrecessedGeocentric(equinox='B1875') to precess to equinox of 1875

See also: #9855

EDIT: Fix #9855

### DIFF
--- a/astropy/coordinates/tests/test_funcs.py
+++ b/astropy/coordinates/tests/test_funcs.py
@@ -32,6 +32,7 @@ def test_sun():
     assert np.all(np.abs(gcrs2.dec - [23.5, 0, -23.5]*u.deg) < 1*u.deg)
 
 
+@pytest.mark.xfail
 def test_constellations(recwarn):
     from astropy.coordinates import ICRS, FK5, SkyCoord
     from astropy.coordinates.funcs import get_constellation
--- a/astropy/coordinates/tests/test_funcs.py
+++ b/astropy/coordinates/tests/test_funcs.py
@@ -62,6 +62,23 @@ def test_constellations(recwarn):
     assert boores == 'Boötes'
     assert isinstance(boores, str) or getattr(boores, 'shape', None) == tuple()
 
+    # Test edge cases close to borders, using B1875.0 coordinates
+    # Look for HMS / DMS roundoff-to-decimal issues from Roman (1987) data
+    # as documented in https://github.com/astropy/astropy/issues/9855
+    # The actual boundary on the west side of Orion at Dec +3.0 is
+    # 06h14m30 == 6.2416666666666...
+    ras = [6.24100, 6.24160, 6.24166, 6.24168]
+    # aka ['6h14m27.6s' '6h14m29.76s' '6h14m29.976s' '6h14m30.048s']
+    decs = [3.0, 3.0, 3.0, 3.0]
+    shortnames = ['Ori', 'Ori', 'Ori', 'Mon']
+
+    testcoos = FK5(ras*u.hour, decs*u.deg, equinox='B1875')
+    npt.assert_equal(get_constellation(testcoos, short_name=True), shortnames,
+      "get_constellation() uses Roman approximations, not IAU boundaries from Delporte")
+
+    # When that's fixed, add other tests with coords that are in different constellations
+    # depending on equinox
+
 
 def test_concatenate():
     from astropy.coordinates import FK5, SkyCoord, ICRS
--- a/astropy/coordinates/tests/test_funcs.py
+++ b/astropy/coordinates/tests/test_funcs.py
@@ -67,22 +67,34 @@ def test_constellations(recwarn):
 def test_constellation_edge_cases():
     from astropy.coordinates import FK5
     from astropy.coordinates.funcs import get_constellation
-    # Test edge cases close to borders, using B1875.0 coordinates
-    # Look for HMS / DMS roundoff-to-decimal issues from Roman (1987) data
-    # as documented in https://github.com/astropy/astropy/issues/9855
-    # The actual boundary on the west side of Orion at Dec +3.0 is
-    # 06h14m30 == 6.2416666666666...
-    ras = [6.24100, 6.24160, 6.24166, 6.24168]
 
-    # aka ['6h14m27.6s' '6h14m29.76s' '6h14m29.976s' '6h14m30.048s']
+    # Test edge cases close to borders, using B1875.0 coordinates
+    # Look for HMS / DMS roundoff-to-decimal issues from Roman (1987) data,
+    # and misuse of PrecessedGeocentric, as documented in
+    # https://github.com/astropy/astropy/issues/9855
+
+    # Define eight test points.
+    # The first four cross the boundary at 06h14m30 == 6.2416666666666... hours
+    # with Monoceros on the west side of Orion at Dec +3.0.
+    ras = [6.24100, 6.24160, 6.24166, 6.24171]
+    # aka ['6h14m27.6s' '6h14m29.76s' '6h14m29.976s' '6h14m30.156s']
+
     decs = [3.0, 3.0, 3.0, 3.0]
+
+    # Correct constellations for given RA/Dec coordinates
     shortnames = ['Ori', 'Ori', 'Ori', 'Mon']
+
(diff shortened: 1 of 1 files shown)


pllim:
Thank you, @nealmcb ! Let's hold off on the milestone (and hence, change log entry) until the fix is in place, assuming you do plan to add the actual fix to this pull request.

nealmcb:
Thanks, @pllim. My next goal is to understand the use of PrecessedGeocentric, and if that is just a mistake, and if the fix I listed at https://github.com/astropy/astropy/issues/9855#issuecomment-573382503 would work, I could easily commit that also.

I noticed that there were other test cases that failed, so I wondered if you might
actually accept just this failing test case as is.

Otherwise I'm in a crunch right now so can't do anything more for a week or perhaps more.

pllim:
If you just want to add tests without a code patch, we can mark them as `xfail` for now until a follow-up PR can be made to fix them.

Or if you'd rather wait for the patch to go with the tests, this PR can remain open for a few months before our bot complains about it.

Your call. 😄

astrofrog:
@nealmcb - just to check, do you have any time to work on this to wrap it up? This seems like a very useful fix!

nealmcb:
I would love to wrap this up. Though I actually got more busy after finding this, and the crunch will last well in to May. But then I should be able to get back to it.

But in the meantime, I'd appreciate it if someone who really knows the transforms would just chime in on whether the fix I [noted above](https://github.com/astropy/astropy/issues/9855#issuecomment-573382503) is the proper way to use the library.

nealmcb:
@pllim, I've marked it as xfail for now and rebased.

I still need guidance on how it is intended to work as noted above.
Thanks!

pllim:
Thanks, @nealmcb ! I have requested coordinates maintainers to review this. 🤞

nstarman:
@nealmcb mark as draft?

adrn:
Sorry for the long wait on this! This is way outside of my wheelhouse, but would be good to close the loop on this before v5.0 (feature freeze in ~1 month). I'll add this to our list of things to discuss at our next astropy.coordinates check-ins.

nealmcb:
Thank you @adrn. I'm still interested in this, and was planning to keep it alive via a simple push in the next day or two. I guess removing the Close? label means I don't have to?

I am still looking for insights whether my proposal makes sense, or would break something.

pllim:
Removing "Close?" will just re-trigger the labeling the next bot run, as you have seen above. Rebasing will give you more time. I really do not recommend using the `keep-open` label unless this is blocking the next release or something.

nealmcb:
Rebased. As is, this documents a problematic edge case.

eteq:
I'm milestoning this to 5.0 in the optimistic case that we fix it in time, @nealmcb - note that since this is a bug in principle it isn't subject to the feature freeze. I also made a PR into @nealmcb's repo breaking this out to a *separate* test so that this isn't xfailing existing working tests, which you may want to merge if you agree!

That said, I'm sorry I never had a chance to read over this before, as I think your solution in https://github.com/astropy/astropy/issues/9855#issuecomment-573382503 should work fine! See my comment https://github.com/astropy/astropy/issues/9855#issuecomment-955014498 . 

Based on that, I think you have what you'd need to do this. If you think you can do the FK5 switch *and* the correct boundaries that would be great.  But I think if you do just the FK5 switch for 5.0 and we get the boundaries later, that could also be good!

eteq (approved):
oops, missed that there was a follow-up here, but this is looking good now, so I'm going ahead and merging.  CI failures are definitely a red herring here.

eteq:
(and this is no-changelog-needed because it's adding a test that xfails.  the bug fix will need a changelog entry though)

pllim:
@meeseeksdev backport to v5.0.x

Issue #9660, opened 2019-11-23
Option to show table column descriptions in show_in_notebook()


hcferguson:
### Description

It would be nice to have the ability to reveal the column descriptions in the table headers if they are there. I'd probably put this as the default, but if not, it would be good to at least include it as an option.

### Additional context

As far as I can tell the only way to see the column descriptions is with table.info() or by introspecting or saving to a format that saves the metadata. At least one of the print options ought to have the ability to put the descriptions in the header. `show_in_notebook` and even `show_in_browser` seem like the simplest solution, since it could be implemented as a keyword option.

taldcroft:
My main concern on this is that in my experience the "description" tends to be on the long side and so putting it into the table header will generally make the plaintext render of the table useless.  However, for HTML-based output, one could imagine having a hover-over on the column name that provides the description if available.

FYI, the following location is where one might go to add an option to show description in the header:

https://github.com/astropy/astropy/blob/918abcd480acdf824c5e5d856daf0573811d9b3c/astropy/table/pprint.py#L364

hcferguson:
I like the idea of a hover option.

PR #9870, opened 2020-01-20
added a new feature of showing description on hoverover column name o…

Harshil-C:
HOVER OVER TABLE COLUMN SHOWING DESC OF COLUMN

### Description

@taldcroft  and @hcferguson  suggested an enhancement idea in #9660 issue,after this commit i was able to hoverover the html based table column header from jupyter notebook  to show the column description .
*i only added this to the show_in_notebook() method and for a whole table(not for a single column)
  i did that because i just wanna make sure it is perfect from you maintainers and would add feature in to columns in seperate pr
*i only enabled the hoverover for table which contains atleast a column with description(again to make sure ,can change on demand),if the table does contain atleast one desc it shows tooltip on every column and show desc if it exists else it writes none in desc .
the code is mainly altered on tables.pprint.py(at_pformatable()).


taldcroft:
@Harshil-C - this one got away from me, sorry!  This was a good start and since we are at feature freeze I just took a crack at finishing it.

taldcroft:
@astrofrog - rebased!

taldcroft:
This is causing other test failures so pushed off to 4.2.

Issue #9898, opened 2020-01-28
In-place data modification when changing from SphericalRepresentation to PhysicsSphericalRepresentation 


niess:
The following looks like a bug. At least the behaviour is rather surprising.

When representing a `SphericalRepresentation` as a `PhysicsSphericalRepresentation` some data are changed in-place, i.e. the two representations actually share some data.  For example:

```python
>>> r = SphericalRepresentation(0 * u.deg, 0 * u.deg, 1)
>>> r.lat.flags.writeable = False
>>> r.lon.flags.writeable = False
>>> r.distance.flags.writeable = False
>>> r.represent_as(PhysicsSphericalRepresentation)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/.mount_pythonSLjLgF/usr/python/lib/python3.7/site-packages/astropy/coordinates/representation.py", line 1633, in represent_as
    r=self.distance, copy=False)
  File "/tmp/.mount_pythonSLjLgF/usr/python/lib/python3.7/site-packages/astropy/coordinates/representation.py", line 1734, in __init__
    self._phi.wrap_at(360 * u.deg, inplace=True)
  File "/tmp/.mount_pythonSLjLgF/usr/python/lib/python3.7/site-packages/astropy/coordinates/angles.py", line 385, in wrap_at
    self[()] = wrapped
  File "/tmp/.mount_pythonSLjLgF/usr/python/lib/python3.7/site-packages/astropy/units/quantity.py", line 1060, in __setitem__
    self.view(np.ndarray).__setitem__(i, self._to_own_unit(value))
ValueError: assignment destination is read-only
```

Another example:

```python
>>> r0 = SphericalRepresentation(0 * u.deg, 0 * u.deg, 1)
>>> r1 = r0.represent_as(PhysicsSphericalRepresentation)                             
>>> r1
(0., 90., 1.) (deg, deg, )                                                                                                                   
>>> r0.lon[()] = 90 * u.deg                                                          
>>> r1
(90., 90., 1.) (deg, deg, )
```

mhvk:
Actually, this looks like two somewhat separate issues/problems: 

1. It is tried to wrap the input longitude if it is not already in the right class, and this fails for `copy=False` and a non-writeable array:
```
from astropy import units as u
from astropy.coordinates import UnitSphericalRepresentation
lon = 0*u.degree
lon.flags.writeable = False
[...]

PR #9900, opened 2020-01-29
Avoid wrapping input longitudes to EarthLocation

mhvk:
This follows on the report of @niess (https://github.com/astropy/astropy/issues/9898#issuecomment-579769528) that initializing `EarthLocation` with read-only arrays gives errors because the in-place wrapping of `Longitude` cannot be done. But there really is no need to wrap, since the resulting longitude is never exposed to the user, and only its sin and cos are taken. So, here we initialize an `Angle` instead.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -304,6 +304,9 @@ astropy.coordinates
 - Fixed a bug with ``TransformGraph.remove_transform()`` when the "from" and
   "to" frame classes are not explicitly specified. [#9815]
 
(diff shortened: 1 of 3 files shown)


mhvk:
Tests now pass (except mpldev which is unrelated); @adrn, @eteq, any suggestions? This is quite small & trivial.

niess:
@mhvk - Thanks for the patch :) This will be in the next release, e.g. on PyPI?

mhvk:
Yes, it will be in 4.0.1.

niess:
Great. Thanks!

Issue #9795, opened 2019-12-18
Incomplete repr/str for polynomial models


jehturner:
**Not a bug -- see below.**

This happens with at least Chebyshev1D/2D in the latest `master` branch `54f9e54d`). I haven't tried other models yet. The resulting model reports incorrect parameters *but* gets evaluated as if the correct parameters are being used (am I doing anything unusually stupid?):

```python
>>> import numpy as np
>>> from astropy.modeling import models, fitting

>>> x = np.arange(0, 10)

>>> fitter = fitting.LinearLSQFitter()

>>> c1 = models.Chebyshev1D(1, c0=0.1, c1=0.5)

>>> # This should reproduce c1 within rounding errors, but appears not to:
>>> c2 = fitter(models.Chebyshev1D(1), x, c1(x))
>>> print('C2', repr(c2))
C2 <Chebyshev1D(1, c0=2.35, c1=2.25)>

>>> # This has the same parameter values reported by c2 (so should do the same):
>>> c3 = models.Chebyshev1D(1, c0=2.35, c1=2.25)
>>> print('C3', repr(c3))
C3 <Chebyshev1D(1, c0=2.35, c1=2.25)>

>>> print('C1(1)', c1(1))
C1(1) 0.6
>>> print('C2(1)', c2(1))  # behaves like c1, not according to its parameters
C2(1) 0.5999999999999972
>>> print('C3(1)', c3(1))  # behaves like c2 would be expected to
C3(1) 4.6
```

Linux-3.10.0-957.27.2.el7.x86_64-x86_64-with-centos-7.6.1810-Core
Python 3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 19:07:31) 
[GCC 7.3.0]
Numpy 1.17.3
Scipy 1.3.1

jehturner:
@nden, @perrygreenfield

jehturner:
I see the same behaviour when I `git checkout v4.0rc1`, @bsipocz. But to my surprise I'm also seeing it when I check out 2.0 & 3.0. Am I imagining things?

jehturner:
OK, this is not a bug, there's just a difference in `domain` for `c1`/`c3` (which default to `[0,1]`) and `c2` (which gets a domain spanning the input co-ordinates). Sorry for the false alarm, but I think before closing this we should check whether this is documented clearly enough, since there are probably other users who are just as slow as I am. I'm not seeing much in the docs, but I'm not as familiar with the reorganized version.

[...]

PR #9906, opened 2020-01-30
fixed serialization of orthonormal polynomials

### LINKED ISSUE #9809: ASDF serialization of polynomial models missing domain (causing incorrect evaluation)
When saving & reloading a polynomial model, the `domain` information gets lost (also `window`?), causing the reloaded model to produce different results from the original one.

```python
>>> import numpy as np
>>> from astropy.modeling import models, fitting
>>> import asdf

>>> x = np.arange(0, 9)

>>> model = models.Chebyshev1D(1, c0=2.35, c1=2.25, domain=(0,9))
>>> model.inverse = models.Chebyshev1D(1, c0=4.5, c1=4.5, domain=[0.1, 4.6])

>>> af = asdf.AsdfFile({"model" : model})
>>> af.write_to("testdom.asdf")

>>> newmodel = asdf.open("testdom.asdf")['model']

>>> print('Fwd before {}, after {}'.format(model.domain, newmodel.domain))
Fwd before (0, 9), after None

>>> print('Inv before {}, after {}'.format(model.inverse.domain,
>>>                                        newmodel.inverse.domain))
Inv before [0.1, 4.6], after None

>>> print('y before:', model(x))
y before: [0.1 0.6 1.1 1.6 2.1 2.6 3.1 3.6 4.1]

>>> print('y after:', newmodel(x))
y after: [ 2.35  4.6   6.85  9.1  11.35 13.6  15.85 18.1  20.35]
```

See also #9795.

Linux-3.10.0-957.27.2.el7.x86_64-x86_64-with-centos-7.6.1810-Core
[GCC 7.3.0]
Numpy 1.17.3
Scipy 1.3.1

perrygreenfield:
### Description

This pull request is to address the fact that previously orthonormal polynomials were not serialized with necessary information about domain or window if those attributes were not the defaults

Fixes #9809

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,10 @@ astropy.io.ascii
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
+- Added serializaton to ASDF of orthonormal polynomial models the 
+  domain and window attributes, without which the serialized model
+  will not properly evaluate its arguments [#xxxx]
+
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,4 @@
+
 4.1 (unreleased)
 ================
 
@@ -27,10 +28,6 @@ astropy.io.ascii
 
 astropy.io.misc
 ^^^^^^^^^^^^^^^
-
-- Added serializaton to ASDF of orthonormal polynomial models the 
-  domain and window attributes, without which the serialized model
-  will not properly evaluate its arguments [#9906]
 
 astropy.io.fits
 ^^^^^^^^^^^^^^^
@@ -322,6 +319,10 @@ astropy.io
 
 - Fixed a bug that prevented the unified I/O infrastructure from working with
   datasets that are represented by directories rather than files. [#9866]
+
+- Added serializaton to ASDF of orthonormal polynomial models the 
+  domain and window attributes, without which the serialized model
+  will not properly evaluate its arguments [#9906]
 
 astropy.io.misc
 ^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,7 @@ astropy.io.misc
 
 - Added serializaton to ASDF of orthonormal polynomial models the 
   domain and window attributes, without which the serialized model
-  will not properly evaluate its arguments [#xxxx]
+  will not properly evaluate its arguments [#9906]
 
 astropy.io.fits
 ^^^^^^^^^^^^^^^
--- a/astropy/io/misc/asdf/tags/transform/compound.py
+++ b/astropy/io/misc/asdf/tags/transform/compound.py
@@ -42,6 +42,7 @@ _tag_to_method_mapping = {
 class CompoundType(TransformType):
     name = ['transform/' + x for x in _tag_to_method_mapping.keys()]
     types = [CompoundModel]
+    version = '1.2.0'
     handle_dynamic_subclasses = True
 
     @classmethod
--- a/astropy/io/misc/asdf/tags/transform/polynomial.py
+++ b/astropy/io/misc/asdf/tags/transform/polynomial.py
@@ -176,18 +176,26 @@ class OrthoPolynomialType(TransformType):
         n_dim = coefficients.ndim
         poly_type = node['polynomial_type']
         if n_dim == 1:
-            model = cls.types[cls.typemap[poly_type]](coefficients.size - 1)
+            domain = node.get('domain')
+            window = node.get('window')
+            model = cls.types[cls.typemap[poly_type]](coefficients.size - 1,
+                              domain=domain, window=window)
             model.parameters = coefficients
         elif n_dim == 2:
+            x_domain, y_domain = node.get('domain', (None, None))
+            x_window, y_window = node.get('window', (None, None))
+
             coeffs = {}
             shape = coefficients.shape
             x_degree = shape[0] - 1
-            y_degree = shape[1] -1
+            y_degree = shape[1] - 1
             for i in range(x_degree + 1):
                 for j in range(y_degree + 1):
                     name = f'c{i}_{j}'
                     coeffs[name] = coefficients[i, j]
-            model = cls.types[cls.typemap[poly_type]+1](x_degree, y_degree, **coeffs)
+            model = cls.types[cls.typemap[poly_type] + 1](x_degree, y_degree,
+                        x_domain=x_domain, y_domain=y_domain,
+                        x_window=x_window, y_window=y_window, **coeffs)
         else:
             raise NotImplementedError(
                 "Asdf currently only supports 1D or 2D polynomial transforms.")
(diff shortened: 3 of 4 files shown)


pllim:
@perrygreenfield , please move change log entry to 4.0.1 since this is a bug fix.

p.s. Not sure why tests failed. If you haven't already, please try rebase on top of latest `master`. Thanks!

jehturner:
Do I also need to update anything else to go with this (where does the schema come from again)? Thanks.

perrygreenfield:
@jehturner That's a fair question about schemas. It turns out (I believe) that there aren't any schemas for bounding_boxes or things like that. So any errors in specifying the value in ASDF that way won't be caught by using the schema machinery, but rather by the object machinery in Python (preferably in creating the object, but it possible in some cases if they are assigned to non-property attributes the error will only arise in using the object). It is something that I think deserves a small schema eventually (i.e., defining a rectangular region)

jehturner:
Thanks a lot for this @perrygreenfield. The following code (and hence my real `gwcs` serialization test) fails for me with a long `jsonschema` validation error (only when the `domain` is defined):
```python
from astropy.modeling import models
import asdf

fwcal = models.Chebyshev1D(2, c0=500.075, c1=0.05, c2=0.001, domain=(0, 9))
wavtrans = fwcal & models.Identity(1)

af = asdf.AsdfFile({"model" : wavtrans})
```
It looks like this is probably the pertinent line:
```jsonschema.exceptions.ValidationError: mismatched tags, wanted 'tag:stsci.edu:asdf/transform/domain-1.0.0', got 'tag:yaml.org,2002:int'```

**Comment by perrygreenfield** [comment]
Hmmm, that may be a versioning error since domain was supposedly deprecated. @nden, do you understand how this may have happened, offhand?

**Comment by nden** [comment]
@perrygreenfield Could you try adding `version = 1.2.0` to the `CompoundType` tag? I have it working locally but I also have other changes which may not be needed. I'm thinking this may be the minimal change needed. I can't test it because I need to reinstall everything now.

**Comment by jehturner** [comment]
:+1: That seemed to fix the failure I was seeing from `AsdfFile`.

I'm still seeing an error from the domain comparison in my test because the `domain` is a tuple when the model is instantiated in my example above, but it becomes a list after the model is reloaded (at L225). Wouldn't it be better if it were always converted to one type or the other? Thanks.

**Comment by perrygreenfield** [comment]
@jehturner Good point, but before fixing this I'd like a little more detail on how the model is instantiated in your example. In my quick look, it appears that the current polynomial models don't enforce anything about domain or windows so that however one sets those values, there is a possibility that ASDF will return an inconsistent value. That can be changed, and I would agree that tuple is the natural form they should take and keep.

**Comment by jehturner** [comment]
I was doing this:
```python
>>> fwcal = models.Chebyshev1D(2, c0=500.075, c1=0.05, c2=0.001, domain=(0, 9))
>>> fwcal.domain
(0, 9)
```
whereas with a list it does this:
```python
>>> fwcal = models.Chebyshev1D(2, c0=500.075, c1=0.05, c2=0.001, domain=[0, 9])
>>> fwcal.domain
[0, 9]
```
Is that what you were asking?

It turns out that one can even do this:
```python
>>> fwcal = models.Chebyshev1D(2, c0=500.075, c1=0.05, c2=0.001, domain='fred')
>>> fwcal.domain
'fred'
```
So indeed I think the model should enforce something like a tuple (presumably converting whatever sequence of numbers you give it into one), in which case your tag code should do the right thing automatically -- so perhaps this is really a separate issue from this PR, come to think of it.

perrygreenfield:
Right! I was just looking at what it should do to enforce permitted values (None complicates it a bit). Might as well fold that in.

eteq:
I got here by way of #9795 and I'm a bit confused: Wasn't @jehturner's suggestion in  https://github.com/astropy/astropy/issues/9809#issuecomment-568078984 to have this come from `OrthoPolynomial1D` instead of `hasattr`?  I think I like that suggestion more because it gives a way to address the point from #9795 that this is not just ASDF but *other* use cases.

jehturner:
Contrary to my thinking in #9809, @eteq, it turns out that Perry's solution here uses neither `hasattr` nor `OrthoPolynomial1D`, because it only involves changes to the "tag" code. For better or worse, the tag classes are structured differently from the model hierarchy itself and enumerate the supported/applicable models (in this case `Legendre1D, Legendre2D, Chebyshev1D, Chebyshev2D, Hermite1D, Hermite2D`) -- so the code can simply assume that 1D models have `domain` & 2D models have `x_domain` & `y_domain`. If someone wanted to add another `PolynomialModel`, they'd have to write or adapt a tag to do the right thing, so there's no forward compatibility problem in that sense if they add one with no `domain`, though this scheme does imply some guaranteed extra work to add new models.

jehturner:
This raises the point that the `PolynomialType` tag also needs changing, doesn't it @perrygreenfield, since those models have `domain` & `window` too? I know that wasn't in my issue title but still :wink:.

perrygreenfield:
@jehturner Yes they would. The more I looked into this, the more surprises jumped out at me. I was thinking of tackling all this at once, but given the issues regarding domain/window in general, perhaps the quick solution here should be done and more thought and input on object interface issues done without rushing.

nden:
To clarify a bit on the implementation. It follows the implementation of polynomials in numpy in having `domain` and `window`. Their purpose is mostly to support fitting. And yes, general `Polynomial` models define the defaults as they are in order to provide the identity transform
but also provide the option to use a different transform. And an attempt to summarize all opinions in the several issues:

- Modify the `__init__` signature of all polynomials to define `domain` and `window` as None and if so set the defaults in the `__init__` to their current default values. (4.0.1)
- Serialize `domain` and `window` (4.0.1)
- Improve the documentation to better explain what they are. (4.0.1)
- Fix `repr` and `str` (4.1)

Am I missing something?

perrygreenfield:
To possibly add to the @nden list: provide an alternate interface for defining the transformation of the input value. Given all the involved issues, this is likely non-trivial and would go into 4.1 or 5.0 if done at all. The main motivation is to make it clearer what is going on since I think many find the existing terminology misleading (even if numpy is using it).

jehturner:
> Am I missing something?

I think also converting the `domain` & `window` arguments from whatever sequences the user provides to 2-tuples in the corresponding attributes (which will also fix their type changing from a tuple to a list after serialization, such that the pre/post-serialized models don't quite match).

jehturner:
I wonder whether you might prefer to make the `domain` & `window` attributes lists rather than tuples though, since that's what the documentation currently says the arguments should be? Either way.

So where are we with this particular PR? I think it's OK, except that `domain` & `window` still need adding to `PolynomialType` as well. If you want, I could probably do that now I have an example (I suppose by submitting another PR to Perry's branch, or is that getting too silly?).

perrygreenfield:
@jehturner The fix is awaiting changes being made to fix lint issues and I will then base the fix on that PR. I'm not sure of the consequences of lists vs tuples other than if explicitly in the default of an argument list, the tuple is much preferred since it isn't mutable, but I'll have to see in the context of other changes.

perrygreenfield:
This will be done as a new PR. I'll provide the number when it is ready to submit.

perrygreenfield:
replaced py PR  #9941

PR #9915, opened 2020-01-31
Deprecate imshow_only_kwargs keyword in imshow_norm

### LINKED ISSUE #9861: Deprecate imshow_only_kwargs in imshow_norm
As discussed in #9851, in Matplotlib 3.2, passing ``vmin/vmax`` to ``imshow_only_kwargs`` in ``imshow_norm`` is deprecated since Matplotlib no longer allows ``norm`` and ``vmin/vmax`` to be set. I don't think there are other useful kwargs that could be passed to ``imshow_only_kwargs`` so we should probably deprecate it?

larrybradley:
Closes #9861. 

Note that I didn't use `deprecated_renamed_argument` because of the bug I discovered in #9914.

### DIFF
--- a/astropy/visualization/tests/test_norm.py
+++ b/astropy/visualization/tests/test_norm.py
@@ -7,6 +7,7 @@ import numpy as np
 from numpy import ma
 from numpy.testing import assert_allclose, assert_equal
 
+from astropy.utils.exceptions import AstropyDeprecationWarning
 from astropy.visualization.mpl_normalize import ImageNormalize, simple_norm, imshow_norm
 from astropy.visualization.interval import ManualInterval, PercentileInterval
 from astropy.visualization.stretch import SqrtStretch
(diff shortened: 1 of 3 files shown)


astropy/visualization/tests/test_norm.py
```diff
@@ -238,15 +238,6 @@ def test_imshow_norm():
 
     imshow_norm(image, ax=ax, vmin=0, vmax=1)
 
-    # Note that the following is deprecated in Matplotlib 3.2
```
pllim: I think we should keep the tests until the keyword is actually removed. Although we should add a `pytest.warns` to ensure the deprecation warning is raised correctly too.

pllim:
Note to maintainer: After this is merged, please open an issue to remind ourselves to remove the keyword after a deprecation period.

astropy/visualization/tests/test_norm.py
larrybradley: Ah yes, thanks!  Done.

PR #9920, opened 2020-02-03
Copy editing: stats

lglattly:
### Description

This pull request is to copy edit the stats sub-package.

### DIFF
--- a/docs/stats/robust.rst
+++ b/docs/stats/robust.rst
@@ -4,9 +4,9 @@
 Robust Statistical Estimators
 *****************************
 
-Robust statistics provide reliable estimates of basic statistics for complex
-distributions.  The statistics package includes several robust statistical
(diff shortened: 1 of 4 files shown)


docs/stats/robust.rst
```diff
@@ -71,13 +85,23 @@ To perform the same sigma clipping with the
      >>> filtered_data = sigclip(y)
 
 Note that once the ``sigclip`` instance is defined above, it can be
-applied to other data, using the same, already-defined, sigma-clipping
+applied to other data using the same already defined, sigma-clipping
```
larrybradley:
```suggestion
applied to other data using the same already defined sigma-clipping
```

larrybradley: I don't think this comma is necessary.

PR #9937, opened 2020-02-11
Fix indentation for the "Combining Models > Other Topics" section

saimn:
This section is indented and taken with the code example, in the compound models docs. Just before the "Indexing and slicing" subsection (http://docs.astropy.org/en/latest/modeling/compound-models.html#indexing-and-slicing).


saimn:
Looks better now:
https://59608-2081289-gh.circle-artifacts.com/0/docs/_build/html/modeling/compound-models.html#other-topics

PR #9941, opened 2020-02-12
Fix serialization of polynomials; improve documentation.

### LINKED ISSUE #9809: ASDF serialization of polynomial models missing domain (causing incorrect evaluation)

perrygreenfield:
### Description

This pull request is to address the fact that previously orthonormal polynomials were not serialized with necessary information about domain or window if those attributes were not the defaults

Documentation was improved to clarify the meaning of the domain and window keywords and attributes of polynomials since there is evident confusion about how they are applied.

The  repr output now includes the values for domain and window if not default values and str does as well if not None.

Previously the polynomial constructors used default lists for some of the values of domain and window, which is generally discouraged. These have been replaced with None and a check for None in the constructor is used to set the value to a list. A future PR will change the usage to tuple because it is planned to back port this PR, and making that change now would break the API for a bugfix.

This PR replaces #9906

Fixes #9809 #9795

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -503,8 +503,6 @@ astropy.io.misc
 
 - Fixed serialization of Time objects with location under time-1.0.0
   ASDF schema. [#9983]
-- Fixed serialization of polynomial models to include non default values of
-  domain and window values. [#9941]
 
 astropy.io.fits
 ^^^^^^^^^^^^^^^
@@ -553,9 +551,6 @@ astropy.modeling
 
 - Fixed a bug in setting default values of parameters of orthonormal
   polynomials when constructing a model set. [#9987]
-- Improvements to documentation clearing up how domain and window attributes are
-  used for polynomials; added domain and window attributes to repr and str. Fixed
-  bug with _format_repr in core.py. [#9941]
 
 astropy.table
 ^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -50,6 +50,9 @@ astropy.modeling
 - Added ``UnitsMapping`` model and ``Model.coerce_units`` to support units on otherwise
   unitless models. [#9936]
 
+- Added ``domain`` and ``window`` attributes to ``repr`` and ``str``. Fixed bug with
+  ``_format_repr`` in core.py. [#9941]
+
 astropy.nddata
 ^^^^^^^^^^^^^^
 
@@ -500,6 +503,8 @@ astropy.io.misc
 
 - Fixed serialization of Time objects with location under time-1.0.0
   ASDF schema. [#9983]
+- Fixed serialization of polynomial models to include non default values of
+  domain and window values. [#9941]
 
 astropy.io.fits
 ^^^^^^^^^^^^^^^
@@ -548,6 +553,9 @@ astropy.modeling
 
 - Fixed a bug in setting default values of parameters of orthonormal
   polynomials when constructing a model set. [#9987]
+- Improvements to documentation clearing up how domain and window attributes are
+  used for polynomials; added domain and window attributes to repr and str. Fixed
+  bug with _format_repr in core.py. [#9941]
 
 astropy.table
 ^^^^^^^^^^^^^
--- a/astropy/io/misc/asdf/tags/transform/basic.py
+++ b/astropy/io/misc/asdf/tags/transform/basic.py
@@ -13,7 +13,7 @@ __all__ = ['TransformType', 'IdentityType', 'ConstantType']
 
 
 class TransformType(AstropyAsdfType):
-    version = '1.1.0'
+    version = '1.2.0'
     requires = ['astropy']
 
     @classmethod
--- a/astropy/modeling/polynomial.py
+++ b/astropy/modeling/polynomial.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """
 This module contains models representing polynomials and polynomial series.
+
 """
 # pylint: disable=invalid-name
 import numpy as np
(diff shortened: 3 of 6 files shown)


CHANGES.rst
```diff
@@ -341,6 +341,9 @@ astropy.io.ascii
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
+- Fixed serialization of polynomial models to include non default values of
+  domain and window values. #9941
```
nden: Need brackets around the PR number

nden:
@jehturner I think this fixes the issues you reported. Please take a look if you have time.

astropy/modeling/polynomial.py
```diff
@@ -1,6 +1,50 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """
 This module contains models representing polynomials and polynomial series.
+
+Notes regarding usage of domain and window
```
bsipocz: This text is not exposed anywhere for users. It's a bit verbose for docstrings, but maybe it can be included in the narrative docs, somewhere?

nden: I'll see if I can find a better place.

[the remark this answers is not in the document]
jehturner: Sorry, I posted this before the comment from @bsipocz but it doesn't show up until the review is complete.

jehturner:
One more thing... I'm now seeing the following behaviour when playing with it a bit, but it looks like this is some unrelated bug that has been in `master` for the past year or so?
```python
>>> from astropy.modeling.models import Chebyshev2D

>>> c = Chebyshev2D(2, 2, n_models=2)
>>> print(c)

  File "test5.py", line 4, in <module>
    print(c)
  File "/home/jturner/repos/astropy/astropy/modeling/polynomial.py", line 1271, in __str__
    ('Y_Window', self.y_window)])
  File "/home/jturner/repos/astropy/astropy/modeling/core.py", line 2194, in _format_str
    param_table = Table(columns, names=self.param_names)
  File "/home/jturner/repos/astropy/astropy/table/table.py", line 616, in __init__
    init_func(data, names, dtype, n_cols, copy)
  File "/home/jturner/repos/astropy/astropy/table/table.py", line 951, in _init_from_list
    self._init_from_cols(cols)
  File "/home/jturner/repos/astropy/astropy/table/table.py", line 1150, in _init_from_cols
    lengths = set(len(col) for col in cols)
  File "/home/jturner/repos/astropy/astropy/table/table.py", line 1150, in <genexpr>
    lengths = set(len(col) for col in cols)
TypeError: len() of unsized object
```

astropy/io/misc/asdf/tags/transform/basic.py
```diff
@@ -13,7 +13,7 @@
 
 
 class TransformType(AstropyAsdfType):
-    version = '1.1.0'
+    version = '1.2.0'
```
nden: Currently this change raises an error because transform v 1.2.0 is not in the currently released `version_mapping` file.

astropy/io/misc/asdf/tags/transform/compound.py
```diff
@@ -42,6 +42,7 @@
 class CompoundType(TransformType):
     name = ['transform/' + x for x in _tag_to_method_mapping.keys()]
     types = [CompoundModel]
+    version = '1.2.0'
```
nden: Unfortunately this change means `fix_inputs` is looking for a schema with version number 1.2.0 (only version 1.1.0 is available) and asdf raises a warning before using version 1.1.0. astropy turns the warning into an error. I think the way around this is to catch the warning in the test for now but have an asdf public release with the cleaned up versioning problems soon.

bsipocz: I'm not so sure then about the 4.0.1 milestone. Bumping up required dependency versions in bugfix releases is not a thing we should do. Maybe it's possible to split the PR into two then? One to backport, too and one to 4.1 only?

nden:
@jehturner [This issue](https://github.com/astropy/astropy/pull/9941#issuecomment-586440559) was fixed and most comments addressed. I think this is ready to merge once tests finish. And just to confirm this is going into 4.1 only.

jehturner:
OK, thanks @nden for all the work on this. We should be able to use `gwcs` with `asdf` now. Do the pending minor comments (I think the condition for serializing `window` and also `__str__` not knowing the default for `window`) want noting anywhere for later? I think there's also polynomial models converting `domain` & `window` args to a tuple (or list) still pending? Just so they don't get lost.

nden:
@jehturner Would you be able to file a ticket for the rest of the polynomial issues?

pllim:
Failure is seen on `master` too, so it's not related. Not sure where it crept in from but I suspect #9929 .

bsipocz:
The unrelated failure is gone, and this is all approved, so I go ahead and merge.

bsipocz:
Thanks @perrygreenfield!

PR #9948, opened 2020-02-13
Fix for #9932 by checking for collections.abc.Sequence in units.quantity_decorator

### LINKED ISSUE #9932: quantity_input fails  on Enum arguments
### Description

I am trying to use quantity_support on the `__init__` of a class which takes an Enum as an argument. If it's annotated something in the type resolution explodes. 

### Expected behavior

Enums should be supported like any other class, or with a special case to ignore them in the worst case. Mypy seems to handle them fine.

### Steps to Reproduce

Running the following code produces the error below. If the Enum annotation is removed, it seems to work fine.

```python
import astropy.units as u
from enum import Enum, auto

class BasicEnum(Enum):
    AThing = auto()
    AnotherThing = auto()

class UnitsTest:
    @u.quantity_input
    def __init__(self, e: BasicEnum, x: u.m):
        self.e = e
        self.x = x


u = UnitsTest(BasicEnum.AThing, 1 << u.m)
```

Error message:
```
~/Lightweaver/lightweaver/units_test.py in <module>
     13 
     14 
---> 15 u = UnitsTest(BasicEnum.AThing, 1 << u.m)

/local0/scratch/cmo/miniconda3/envs/LightweaverDev/lib/python3.8/site-packages/astropy/units/decorators.py in wrapper(*func_args, **func_kwargs)
    211                 # Check for None in the supplied list of allowed units and, if
    212                 #   present and the passed value is also None, ignore.
--> 213                 elif None in targets:
    214                     if arg is None:
    215                         continue

/local0/scratch/cmo/miniconda3/envs/LightweaverDev/lib/python3.8/enum.py in __contains__(cls, member)
    308     def __contains__(cls, member):
    309         if not isinstance(member, Enum):
--> 310             raise TypeError(
    311                 "unsupported operand type(s) for 'in': '%s' and '%s'" % (
    312                     type(member).__qualname__, cls.__class__.__qualname__))

TypeError: unsupported operand type(s) for 'in': 'NoneType' and 'EnumMeta'

```
So, a workaround for now is to not annotate the argument, but it would be nice to not need to do this long term.

### System Details

Linux-3.10.0-1062.12.1.el7.x86_64-x86_64-with-glibc2.10
Python 3.8.1 | packaged by conda-forge | (default, Jan 29 2020, 14:55:04) 
[GCC 7.3.0]
Numpy 1.18.1
astropy 4.0
Scipy 1.4.1

Goobley:
### Description

It is checked if an object is a Sequence rather than simply iterable. I would expect a sequence of types to be provided in a list or tuple, whereas objects like Enum are also iterable, but can't hold multiple types for checking here. Added test for Enum case.

Tested in clean venvs on 3.8 and 3.7

Fixes #9932

### DIFF
--- a/astropy/units/tests/test_quantity_annotations.py
+++ b/astropy/units/tests/test_quantity_annotations.py
@@ -222,3 +222,17 @@ def test_return_annotation_none():
 
     solarx = myfunc_args(1*u.arcsec)
     assert solarx is None
+
+
+def test_enum_annotation():
+    # Regression test for gh-9932
+    from enum import Enum, auto
+
+    class BasicEnum(Enum):
+        AnOption = auto()
+
+    @u.quantity_input
+    def myfunc_args(a: BasicEnum, b: u.arcsec) -> None:
+        pass
+
+    myfunc_args(BasicEnum.AnOption, 1*u.arcsec)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -399,6 +399,11 @@ astropy.uncertainty
 astropy.units
 ^^^^^^^^^^^^^
 
+- Fix for ``quantity_input`` annotation raising an exception on iterable
+  types that don't define a general ``__contains__`` for checking if ``None``
+  is contained (e.g. Enum as of python3.8), by instead checking for instance of
+  Sequence. [#9948]
+
 astropy.utils
 ^^^^^^^^^^^^^
 
--- a/astropy/units/decorators.py
+++ b/astropy/units/decorators.py
@@ -3,9 +3,9 @@
 
 __all__ = ['quantity_input']
 
+from collections.abc import Sequence
 import inspect
 from astropy.utils.decorators import wraps
-from astropy.utils.misc import isiterable
 
 from .core import Unit, UnitBase, UnitsError, add_enabled_equivalencies
 from .physical import _unit_physical_mapping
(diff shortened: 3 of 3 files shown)


pllim:
Thanks! I'll let the subpackage maintainers decide if a change log is needed.

mhvk:
I think this counts as a bug fix and it would be good to have a changelog entry, just in case people have been trying to work around it. So, @Goobley, could you add an entry in `CHANGES.rst` under 4.01/bug fixes/units?

astropy/units/tests/test_quantity_annotations.py
```diff
@@ -222,3 +222,16 @@ def myfunc_args(solarx: u.arcsec) -> None:
 
     solarx = myfunc_args(1*u.arcsec)
     assert solarx is None
+
+
+def test_enum_annotation():
+    from enum import Enum, auto
```
mhvk: While adding a changelog entry, maybe also add a comment here along the lines of "Regression test for gh-9932".

mhvk:
p.s. If you have the time/ability, might as well rebase and squash all commits in one at the same time as you're adding the changelog entry.

Goobley:
Unless somebody wants to make those modifications, I'll get them done this weekend.
Thanks all!

PR #9950, opened 2020-02-14
Fixed overwriting bug in HDF5 tables

danielrd6:
### Description

Overwriting a table within a HDF5 file is achieved by setting `append=True` and `overwrite=True`. 
However, if the table was written with the option `serialize_meta=True`, attempting to overwrite the 
table causes an error because the node which contains the meta data is not removed.

The error is reproduced by the following example code:

    from astropy import table

    t = table.Table([[1, 2, 3]])
    t.write('test_file.hdf5', path='data', overwrite=True, serialize_meta=True)
    t.write('test_file.hdf5', path='data', overwrite=True, append=True, serialize_meta=True)

The expected behaviour is that both the table in path `data` and its meta data, which is recorded at 
the path `data.__table_column_meta__`, should be overwritten. This was achieved by adding a 
conditional, after deleting `data` from the output_group, which checks if the `serialize_meta` option 
is set to `True` and if a path with the meta data is already present. If it evaluates to `True`, then the 
column meta data is also removed from `output_group`.

### DIFF
--- a/astropy/io/misc/tests/test_hdf5.py
+++ b/astropy/io/misc/tests/test_hdf5.py
@@ -871,3 +871,26 @@ def test_round_trip_masked_table_default(tmpdir):
         t[name].mask = False
         t2[name].mask = False
         assert np.all(t2[name] == t[name])
+
+
+@pytest.mark.skipif('not HAS_YAML or not HAS_H5PY')
+def test_overwrite_serialized_meta():
+    # This used to cause an error because the meta data table
+    # was not removed from the existing file.
+
+    with h5py.File('test_data.h5', 'w', driver='core', backing_store=False) as out:
+        t1 = Table()
+        t1.add_column(Column(data=[4, 8, 15], unit='cm'))
+        t1.write(out, path='data', serialize_meta=True)
+
+        t2 = Table.read(out, path='data')
+        assert all(t1 == t2)
+        assert t1.info(out=None) == t2.info(out=None)
+
+        t3 = Table()
+        t3.add_column(Column(data=[16, 23, 42], unit='g'))
+        t3.write(out, path='data', serialize_meta=True, append=True, overwrite=True)
+
+        t2 = Table.read(out, path='data')
+        assert all(t3 == t2)
+        assert t3.info(out=None) == t2.info(out=None)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -350,6 +350,10 @@ astropy.io.misc
 
 - Fixed serialization of polynomial models to include non default values of
   domain and window values. [#9956]
+- Fixed a bug which affected overwriting tables within ``hdf5`` files.
+  Overwriting an existing path with associated column meta data now also
+  overwrites the meta data associated with the table. [#9950]
+
 
 astropy.io.fits
 ^^^^^^^^^^^^^^^
--- a/astropy/io/misc/hdf5.py
+++ b/astropy/io/misc/hdf5.py
@@ -317,6 +317,8 @@ def write_table_hdf5(table, output, path=None, compression=False,
         if append and overwrite:
             # Delete only the dataset itself
             del output_group[name]
+            if serialize_meta and name + '.__table_column_meta__' in output_group:
+                del output_group[name + '.__table_column_meta__']
         else:
             raise OSError(f"Table {path} already exists")
 


mhvk:
Was bitten by this recently, so very nice to see a PR to fix it before I even had a chance to look at the code! 

The fix looks good, though @astrofrog or @matteobachetti should have a look as well. For now, could you add a test case, to ensure we don't inadvertently regress? Also, we'd want an entry in `CHANGES.rst` (under the bug-fix/io-misc for 4.0.1). Thanks!

matteobachetti:
Hi @danielrd6, thanks!
Indeed, the fix looks fine. 
Thinking out loud @mhvk and @astrofrog: I'm wondering if there might be some corner case where we lose information during the append. Is there any case where the meta in the data we are appending is not exactly equal to that already in the file? Should we throw a specific exception in this case?

mhvk:
@matteobachetti - I think in this case `append` means "including in an existing hdf5 file" and `overwrite` allows overwriting of a given path, i.e., the existing table as stored in the hdf5 file is completely replaced, and thus its metadata should be replaced as well.

Obviously, it would be good to be very clear about that in the documentation!

danielrd6:
Hi @mhvk,

I am writing the test and adding the change to CHANGES.rst now, but I am note sure on how to proceed after that. Should I ammend the previous commit or start a new pull request with all the chages?

danielrd6:
Hi @mhvk ,

I just added the change to CHANGES.rst and included a test within test_table.py.

@matteobachetti I agree with @mhvk on the point you brought up. The meta data only makes sense for a given table. If that table is being overwritten, its metadata should follow suit.

astropy/table/tests/test_table.py
```diff
@@ -4,6 +4,7 @@
 import gc
 import sys
 import copy
+import h5py
```
mhvk: I think you'll need to move this test to `io/misc/tests/test_hdf5.py` - this in particular since `h5py` and `pyyaml` are optional dependencies for astropy, i.e., we cannot count on them being present (the tests in `test_hdf5` will show how to account for that - basically you'll need `@pytest.mark.skipif('not HAS_H5PY or not HAS_YAML')`

astropy/table/tests/test_table.py
```diff
@@ -2693,3 +2694,14 @@ def test_items():
...
+
+def test_overwrite_serialized_meta():
+    # This used to cause an error because the meta data table
+    # was not removed from the existing file.
+
+    with h5py.File('test_data.h5', 'w', driver='core', backing_store=False) as out:
+        t = table.Table([[1, 2, 3]])
+
+        t.write(out, path='data', serialize_meta=True)
+        t.write(out, path='data', serialize_meta=True, append=True, overwrite=True)
```
mhvk: Ideally, make this a test with two tables and ones that have meta data, and then check that the table and metadata are in fact replaced. in `test_hdf5`, `test_read_write_existing_overwrite` and `test_preserve_serialized` might be good examples to use for making a new test.

danielrd6:
@mhvk 

Thanks for warning me about the optional dependencies. Also, I am not familiar with the testing protocols, so thanks for pointing out the skip decorator.

I implemented your suggestions concerning both the squashing of commits and the moving of the test to `test_hdf5.py`. The test now also checks that the table and its associated meta are both written and read correctly, and that the overwriting actually took place.

Unlike the test functions you offered as examples, I thought it best to keep everything in the memory instead of writing a temporary file.

astropy/io/misc/tests/test_hdf5.py
```diff
@@ -871,3 +871,26 @@ def test_round_trip_masked_table_default(tmpdir):
...
         t2[name].mask = False
         assert np.all(t2[name] == t[name])
+
+
+@pytest.mark.skipif('not HAS_YAML or not HAS_H5PY')
+def test_overwrite_serialized_meta():
+    # This used to cause an error because the meta data table
+    # was not removed from the existing file.
+
+    with h5py.File('test_data.h5', 'w', driver='core', backing_store=False) as out:
```
mhvk: Ah, I now see how you do things in memory with `backing_store=False`. That seems reasonable.

pllim:
There are unrelated commits and merge commits in this PR now that needs to be rebased and possibly squashed out. FYI.

```
git fetch upstream master
git rebase upstream/master
git log
```

If there are still undesired commits:

```
git rebase -i HEAD~n
```

where `n` is the number of commits you need to squash. Then, carefully follow screen instructions. When in doubt, make a backup branch before doing this. Thanks!

astropy/io/misc/tests/test_hdf5.py
danielrd6: I did not know about that until a few days ago when I saw the way astropy uses StringIO for some tests. Since keeping everything in memory seemed a good idea, I went out searching for something similar to HDF5 files.

mhvk:
Merged!  Thanks!

danielrd6:
My pleasure! :)

PR #9959, opened 2020-02-19
Add tags for Gaussian1D and Gaussian2D models. 

shannnnnyyy:
This creates a new file functional_models.py in io/misc/asdf/tags/transform, in which all the functional models in modeling will eventually have tags. It also contains tags for Gaussian1D and Gaussian2D models, so they are serializable. I submitted a separate PR to asdf-standard for the schemas. 

A few notes:  

1. You have the option to initialize the Gaussian2D model with the ‘cov_matrix’ parameter, which is used to calculate and set (x_stddev, y_stddev, theta), or you can omit cov_matrix and set those values individually (or not at all to retain the defaults). You can’t simultaneously create a model with any of those parameters AND cov_matrix. For that reason, even though cov_matrix is a parameter that can be used to initialize, only amplitude, x_mean, y_mean, x_stddev, y_stddev, and theta determine the behavior of the model and so cov_matrix is not included in the schema or tag. I did consider and try to do this, but interestingly even if you initialize the model with cov_matrix instead of (x_stddev, y_stddev, theta), model.cov_matrix is not an available attribute. Another this I found (that may warrant an issue?) is that if you initialize with cov_matrix, you can’t use units/quantites since some math with non-quantities is done to calculate (x_stddev, y_stddev, theta). 

2.  None of the serializable models currently retain any of the fitting constraints (fixed/tied/bounds). This needs to be fixed inside of TransformType I believe, but for now I am following the current convention that other writeable models use and not implementing this. 

3. I could roll Gaussian1DType and Gaussian2DType into a GaussianType if that would be preferable, but for now they are separate.

### DIFF
--- a/astropy/io/misc/asdf/tags/transform/functional_models.py
+++ b/astropy/io/misc/asdf/tags/transform/functional_models.py
@@ -1,0 +1,82 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+from numpy.testing import assert_array_equal
+
+from asdf import yamlutil
+
+from astropy.modeling import functional_models
+from .basic import TransformType
+from . import _parameter_to_value
+
+
+__all__ = ['Gaussian1DType', 'Gaussian2DType']
+
+
+class Gaussian1DType(TransformType):
+    name = "transform/gaussian1d"
+    version = '1.0.0'
+    types = ['astropy.modeling.functional_models.Gaussian1D']
+
+    @classmethod
+    def from_tree_transform(cls, node, ctx):
+        return functional_models.Gaussian1D(amplitude=node['amplitude'],
+                                            mean=node['mean'],
+                                            stddev=node['stddev'])
+
+    @classmethod
+    def to_tree_transform(cls, model, ctx):
+        node = {'amplitude': _parameter_to_value(model.amplitude),
+                'mean': _parameter_to_value(model.mean),
+                'stddev': _parameter_to_value(model.stddev)}
+        return yamlutil.custom_tree_to_tagged_tree(node, ctx)
+
+    @classmethod
+    def assert_equal(cls, a, b):
+        # TODO: If models become comparable themselves, remove this.
+        TransformType.assert_equal(a, b)
+        assert (isinstance(a, functional_models.Gaussian1D) and
+                isinstance(b, functional_models.Gaussian1D))
+        assert_array_equal(a.amplitude, b.amplitude)
+        assert_array_equal(a.mean, b.mean)
+        assert_array_equal(a.stddev, b.stddev)
+
+
+class Gaussian2DType(TransformType):
+    name = "transform/gaussian2d"
+    version = '1.0.0'
+    types = ['astropy.modeling.functional_models.Gaussian2D']
+
+    @classmethod
+    def from_tree_transform(cls, node, ctx):
+        return functional_models.Gaussian2D(amplitude=node['amplitude'],
+                                            x_mean=node['x_mean'],
+                                            y_mean=node['y_mean'],
+                                            x_stddev=node['x_stddev'],
+                                            y_stddev=node['y_stddev'],
+                                            theta=node['theta'])
+
+    @classmethod
+    def to_tree_transform(cls, model, ctx):
+        node = {'amplitude': _parameter_to_value(model.amplitude),
+                'x_mean': _parameter_to_value(model.x_mean),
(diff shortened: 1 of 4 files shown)


astropy/io/misc/asdf/tags/transform/functional_models.py
```diff
@@ -0,0 +1,82 @@
...
+                'mean': _parameter_to_value(model.mean),
+                'stddev': _parameter_to_value(model.stddev)}
+        return yamlutil.custom_tree_to_tagged_tree(node, ctx)
+
+    @classmethod
+    def assert_equal(cls, a, b):
+        # TODO: If models become comparable themselves, remove this.
+        TransformType.assert_equal(a, b)
+        assert (isinstance(a, modeling.functional_models.Gaussian1D) and
+                isinstance(b, modeling.rotations.Gaussian1D))
```
nden: typo: `functional_models`

astropy/io/misc/asdf/tags/transform/functional_models.py
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+        return yamlutil.custom_tree_to_tagged_tree(node, ctx)
+
+    @classmethod
+    def assert_equal(cls, a, b):
+        # TODO: If models become comparable themselves, remove this.
+        TransformType.assert_equal(a, b)
+        assert (isinstance(a, modeling.functional_models.Gaussian2D) and
+                isinstance(b, modeling.rotations.Gaussian2D))
```
nden: `functional_models` not `rotations`

nden:
@shannnnnyyy Could you file a ticket for constraints serialization? Thanks!

astropy/io/misc/asdf/tags/transform/functional_models.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class Gaussian1DType(TransformType):
+    name = "transform/gaussian1d"
+    version = '1.0.0'
+    types = ['astropy.modeling.functional_models.Gaussian1D']
+
+    @classmethod
+    def from_tree_transform(cls, node, ctx):
+        return modeling.functional_models.Gaussian1D(node['amplitude'],
```
eslavich: I would use keyword parameters here, since the order of amplitude, mean, and stddev is defined by the order of the parameter definitions in the class, and not an explicit `__init__` method -- it seems like it'd be easy for someone to switch them around without realizing that they changed the interface.

astropy/io/misc/asdf/tags/transform/functional_models.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class Gaussian2DType(TransformType):
+    name = "transform/gaussian2d"
+    version = '1.0.0'
+    types = ['astropy.modeling.functional_models.Gaussian2D']
+
+    @classmethod
+    def from_tree_transform(cls, node, ctx):
+        return modeling.functional_models.Gaussian2D(node['amplitude'],
```
eslavich: Same comment here about keyword parameters

eslavich:
Can you add some serialization tests?  There's an asdf helper that makes this easy:

```python
from asdf.tests import helpers

@pytest.mark.parametrize("model", [
    Gaussian1D(),
    Gaussian1D(amplitude=10, mean=5, stddev=.3),
    Gaussian2D(),
    # and etc
])
def test_serialization(tmpdir, model):
    helpers.assert_roundtrip_tree({"model": model}, tmpdir)
```

The `assert_roundtrip_tree` function writes the tree to an ASDF file, reads it back again, and compares the result to the original tree.  It uses the `assert_equal` method defined on the type class to compare the model instances.

shannnnnyyy:
> Can you add some serialization tests? There's an asdf helper that makes this easy:
> 
> ```python
> from asdf.tests import helpers
> 
> @pytest.mark.parametrize("model", [
>     Gaussian1D(),
>     Gaussian1D(amplitude=10, mean=5, stddev=.3),
>     Gaussian2D(),
>     # and etc
> ])
> def test_serialization(tmpdir, model):
>     helpers.assert_roundtrip_tree({"model": model}, tmpdir)
> ```
> 
> The `assert_roundtrip_tree` function writes the tree to an ASDF file, reads it back again, and compares the result to the original tree. It uses the `assert_equal` method defined on the type class to compare the model instances.

I added Gaussian1D and Gaussian2D to the `test_models` list in test/test_transform.py where this same test you described is done on all models in that list (with the `test_single_model` function), so I think that should cover it and will pass once the schema is there.

PR #9961, opened 2020-02-19
Revive ASDF serialization with older polynomial schemas

eslavich:
### Description

This pull request revives support for ASDF serializing with older versions of the polynomial schema.  Serializing objects under polynomial-1.0.0 and polynomial-1.1.0 has started to fail since #9956 was merged, due to the new `domain` property failing to validate under the older transform-1.0.0 base schema.


nden:
#9956 fixed the "domain"  serialization problem but effectively removed support for older versions of the standard which this PR adds. This is the first time we have to provide support for older versions of the standard. 
@eslavich Could you add this PR number to the change log for #9956.

PR #9969, opened 2020-02-20
Fix Time init problems for location

### LINKED ISSUE #9962: Instantiating a Time from a Time looses the location
As found by @luojing1211 in https://github.com/nanograv/PINT/issues/609, initializing a table column from a list of time instances loses the location. Further investigation showed that more generally, initializing a `Time` from another `Time` instance causes loss of location, and that likely the fact that it now happens in `Table` reflects a change in how tables initialize from list:
```
from astropy.time import Time
t = Time('J2010', location=(45, 45))
print(t.location)
# (3194419.14506057, 3194419.14506057, 4487348.40886592) m
print(Table([[t]])['col0'][0].location)
# None
print(Time(t).location)
# None
```
Looking back, in versions prior to 4.0 (checked 3.2.x and 3.1), the `Table([[t]])` kept the location, but the `Time(t)` already did not.

taldcroft:
### Description

This pull request fixes some problems in losing the `location` when creating a `Time` object from one or more existing `Time` objects, as noted in #9962.

This is presently a work in progress with no tests, but it does address the specific examples:
```
In [7]: from astropy.time import Time 
   ...: from astropy.table import Table 
   ...: t = Time('J2010', location=(45, 45)) 
   ...: print(t.location) 
   ...: # (3194419.14506057, 3194419.14506057, 4487348.40886592) m 
   ...: print(Table([[t]])['col0'][0].location) 
   ...: # None 
   ...: print(Time(t).location) 
   ...: # None                                                                                                                                                  
(3194419.14506057, 3194419.14506057, 4487348.40886592) m
[(3194419.14506057, 3194419.14506057, 4487348.40886592)]
(3194419.14506057, 3194419.14506057, 4487348.40886592) m
```

This implementation may not be the most general path forward, and does rely on a somewhat unseemly trick of putting a (temporary) hidden `_location` attribute on the `TimeFormat` object to pass the locations back up to the `Time` object.  Just putting it up as a first quick idea for discussion.

Fixes #9962

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -486,6 +486,9 @@ class Time(ShapedLikeNDArray):
                                         precision, in_subfmt, out_subfmt)
         self._format = self._time.name
 
+        # Hack from #9969 to allow passing the location value that has been
+        # collected by the TimeAstropyTime format class up to the Time level.
+        # TODO: find a nicer way.
         if hasattr(self._time, '_location'):
             self.location = self._time._location
             del self._time._location
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -403,6 +403,18 @@ class Time(ShapedLikeNDArray):
                  precision=None, in_subfmt=None, out_subfmt=None,
                  location=None, copy=False):
 
+        if location is not None:
+            from astropy.coordinates import EarthLocation
+            if isinstance(location, EarthLocation):
+                self.location = location
+            else:
+                self.location = EarthLocation(*location)
+            if self.location.size == 1:
+                self.location = self.location.squeeze()
+        else:
+            if not hasattr(self, 'location'):
+                self.location = None
+
         if isinstance(val, self.__class__):
             # Update _time formatting parameters if explicitly specified
             if precision is not None:
@@ -415,17 +427,6 @@ class Time(ShapedLikeNDArray):
             if scale is not None:
                 self._set_scale(scale)
         else:
-            if location is not None:
-                from astropy.coordinates import EarthLocation
-                if isinstance(location, EarthLocation):
-                    self.location = location
-                else:
-                    self.location = EarthLocation(*location)
-                if self.location.size == 1:
-                    self.location = self.location.squeeze()
-            else:
-                self.location = None
-
             self._init_from_vals(val, val2, format, scale, copy,
                                  precision, in_subfmt, out_subfmt)
             self.SCALES = TIME_TYPES[self.scale]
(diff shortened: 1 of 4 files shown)


astropy/time/core.py
```diff
@@ -403,17 +403,6 @@ def __init__(self, val, val2=None, format=None, scale=None,
                  precision=None, in_subfmt=None, out_subfmt=None,
                  location=None, copy=False):
 
-        if location is not None:
```
taldcroft: This was being run even for initializing from an existing `Time` object, in which case `self` already has a `location`.  But actually looking at this now maybe I need to just put something in the `else` block at the end to check if `location` already exists and not override.  That allows overriding the existing location, e.g. `Time(tm_object, location=...)`.

taldcroft: Done.

taldcroft:
Passing tests locally FWIW.

mhvk:
@taldcroft - not the prettiest indeed, but very nicely contained, so I think perfect for a bug fix. Hence, my sense would be to go with this for 4.0.1, and see if for 4.1 we can solve this in `__new__` directly, removing `TimeAstropyTime` altogether.

taldcroft:
OK, I'll go forward with some tests and leave bread crumb comments in the code to help with tidying this up for a better solution in 4.1.

mhvk:
@bsipocz - this would be one PR I'd love to have included in 4.0.1! (This is no longer WIP, really, but does need tests and changelog)

taldcroft:
@mhvk - I think this is ready for review and is passing tests locally.

astropy/time/formats.py
```diff
@@ -738,18 +738,38 @@ def __new__(cls, val1, val2, scale, precision,
...
+            if any(tm.location is not None for tm in val1):
+                if any(tm.location is None for tm in val1):
+                    raise ValueError('cannot concatenate times unless all locations '
+                                     'are set or no locations are set')
+                locations = []
+                for tm in val1:
+                    location = np.broadcast_to(tm.location, tm._time.jd1.shape,
+                                               subok=True)
+                    locations.append(np.atleast_1d(location))
+                location = np.concatenate(locations)
```
mhvk: This now works for numpy 1.17 but not yet for numpy 1.16 without explicit opt-in to numpy function overrides. The easiest way to solve it should be to just use `location = u.Quantity(locations)` instead (or, inside an `if NUMPY_LT_1_17` branch).

mhvk: Hmm, my solution does not actually work...  I'll push up a more detailed fix in a bit.

taldcroft:
@mhvk - thanks for the quick review and the help!  Looks good, and the comment should be sufficient to find this and to use the one-liner when we drop 1.16 support.

mhvk:
Yes, I initially had a whole `if NUMPY_LT_1_17` section, but decided that was overkill. Nice to have this fixed!!

PR #9971, opened 2020-02-20
very minor changes to support Functional units in CDS tables

### LINKED ISSUE #9946: Functional units in CDS tables
CDS units use square brackets to indicate log10-based units (e.g., `[cm/s2]`), which would be nice to support in the units table.

See https://github.com/astropy/astropy/blob/6a231eaaa9d4aad529f554cf95cc5ad6b372798b/astropy/units/format/cds.py#L31-L33

DShivansh:
### Description

changes to support functional units in CDS tables.

Fixes #9946

### DIFF
--- a/astropy/units/format/cds.py
+++ b/astropy/units/format/cds.py
@@ -45,6 +45,8 @@ class CDS(Base):
         'DIVISION',
         'OPEN_PAREN',
         'CLOSE_PAREN',
+        'OPEN_BRACKET',
+        'CLOSE_BRACKET',
         'X',
         'SIGN',
         'UINT',
@@ -88,9 +90,12 @@ class CDS(Base):
         t_DIVISION = r'/'
         t_OPEN_PAREN = r'\('
         t_CLOSE_PAREN = r'\)'
+        t_OPEN_BRACKET = r'\['
+        t_CLOSE_BRACKET = r'\]'
 
         # NOTE THE ORDERING OF THESE RULES IS IMPORTANT!!
         # Regular expression rules for simple tokens
+
         def t_UFLOAT(t):
             r'((\d+\.?\d+)|(\.\d+))([eE][+-]?\d+)?'
             if not re.search(r'[eE\.]', t.value):
@@ -157,11 +162,15 @@ class CDS(Base):
             '''
             main : factor combined_units
                  | combined_units
+                 | OPEN_BRACKET combined_units CLOSE_BRACKET
                  | factor
             '''
             from astropy.units.core import Unit
+            from astropy.units import dex
             if len(p) == 3:
                 p[0] = Unit(p[1] * p[2])
+            elif len(p) == 4:
+                p[0] = dex(p[2])
             else:
                 p[0] = Unit(p[1])
 
--- a/astropy/units/function/logarithmic.py
+++ b/astropy/units/function/logarithmic.py
@@ -150,6 +150,12 @@ class DexUnit(LogUnit):
     def _quantity_class(self):
         return Dex
 
+    def to_string(self, format='generic'):
+        if format == 'cds':
+            return "[{}]".format(self.physical_unit.to_string(format=format))
+        else:
+            return super(DexUnit, self).to_string()
+
 
 class DecibelUnit(LogUnit):
     """Logarithmic physical units expressed in dB
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,7 @@ astropy.extern
 
 astropy.io.ascii
 ^^^^^^^^^^^^^^^^
+- Functional Units can now be processed in CDS-tables. [#9971]
 
 astropy.io.misc
 ^^^^^^^^^^^^^^^
@@ -2588,6 +2589,7 @@ astropy.io.ascii
 
 - Units in CDS-formatted tables are now parsed correctly by the units
   module. [#7348]
+
 
 astropy.io.misc
 ^^^^^^^^^^^^^^^
--- a/astropy/io/ascii/tests/data/cdsFunctional.dat
+++ b/astropy/io/ascii/tests/data/cdsFunctional.dat
@@ -1,0 +1,68 @@
+J/A+A/557/A19   Mass and age of extreme low-mass white dwarfs   (Althaus+, 2013)
+================================================================================
+New evolutionary sequences for extremely low-mass white dwarfs: Homogeneous mass
+and age determinations and asteroseismic prospects.
+    Althaus L.G., Miller Bertolami M.M., Corsico A.H.
+   <Astron. Astrophys. 557, A19 (2013)>
+   =2013A&A...557A..19A
+================================================================================
+ADC_Keywords: Stars, white dwarf ; Models, evolutionary ; Stars, masses ;
+              Stars, ages ; Effective temperatures
+Keywords: white dwarfs - binaries: general - stars: evolution -
+          stars: oscillations - stars: interiors
+
+Abstract:
+    The number of detected extremely low mass (ELM) white dwarf stars has
+    increased drastically in recent year thanks to the results of many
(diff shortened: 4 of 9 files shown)


astropy/io/ascii/cds.py
```diff
@@ -119,6 +119,10 @@ def get_cols(self, lines):
                 unit = match.group('units')
                 if unit == '---':
                     col.unit = None  # "---" is the marker for no unit in CDS table
+                elif unit[0] == '[' and unit[-1] == ']':
+                    temp = re.compile(r'\[(?P<un>.*?)\]')
```
pllim: Maybe the compile can be done once at the module level or is that premature optimization?

astropy/io/ascii/cds.py
```diff
@@ -119,6 +119,10 @@ def get_cols(self, lines):
                 unit = match.group('units')
                 if unit == '---':
                     col.unit = None  # "---" is the marker for no unit in CDS table
+                elif unit[0] == '[' and unit[-1] == ']':
```
pllim: Is this comparison going to catch other non-CDS stuff and cause unintended side effects?

mhvk:
Hmm, while this works, I think it is more logically done inside the parser/lexer in `units/format/cds.py`, so that all the unit machinery benefits. Also, this should round-trip, i.e., I should be able to do `u.dex(u.cm/u.s**2).to_string(format='cds')` and get out `[cm/s2]`.

astropy/io/ascii/cds.py
DShivansh: No because this function is only run when the program guesses that the provided format is CDS.

astropy/io/ascii/cds.py
DShivansh: Yes, I think I should do the compile at the module level because each time it sees the functional unit it will recompile.

DShivansh:
@mhvk parser and lexer have no role when making the table when provided with the data.
I am now working on the ``to_string`` functionality.

hamogu:
@DShivansh : @mhvk is right. The Units are parsed here: https://github.com/astropy/astropy/blob/master/astropy/io/ascii/cds.py#L123 That line parses a string to the `Unit` and then it is the parser/lexer inside the units module that does the work. So, the code that you wrote here should just be moved into the units module, as @mhvk suggests (into the part that deals with "cds" inside the units stuff).

DShivansh:
@hamogu yes, sorry while debugging I stepped over that function 😅, I'll move my code inside
format/cds.py.

DShivansh:
@hamogu I still think that although all the work is done inside the units/format/cds.py but while converting the column info to units to units there is no use of parser <https://github.com/astropy/astropy/blob/72728efafe1aa06f99a551f4b9a2dc787ad2b22f/astropy/units/format/cds.py#L81>
and lexer <https://github.com/astropy/astropy/blob/72728efafe1aa06f99a551f4b9a2dc787ad2b22f/astropy/units/format/cds.py#L142>

Please correct me if I am wrong.

mhvk:
@DShivansh - I will have to correct you, sorry! The line `Unit(unit, format='cds', parse_strict='warn')` ends up going to `format.cds.CDS.parse` which uses the CDS parser/lexer.

So, I think you'll need to define `[` and `]` tokens and associated grammar. Your previous PR on the coordinate directions should give you some idea on how to proceed. The simple test case will be if something like `Unit('[cm/s2]', format='cds')` gives back `u.dex(cm/s2)`. 

p.s. So far, only the `format.generic.Generic` class has support for logarithmic units, but looking there may still help a little in the implementation.

DShivansh:
@mhvk is it done correctly now? Should I now implement test cases?

mhvk:
@DShivansh - this looks OK, but I must admit that with the parser, I just never know... So, yes, do add test cases, both in `units/tests/test_format` in `test_cds_grammar` and (likely a new method) in `TestRoundtripCDS`, and in `io/ascii/tests/test_read.py` + `test_cds_header_from_readme.py`. I fear that adding the test cases may be most of the work for this PR, but it is super-important.

Plus an entry under new features for 4.1. And of course be sure to remove the `TODO` entry in `cds.py` (see #9946)!

DShivansh:
@mhvk can you please check whether my to_string implementation is correct or not?

astropy/units/function/core.py
```diff
@@ -360,10 +360,12 @@ def to_string(self, format='generic'):
             The name of a format or a formatter object.  If not
             provided, defaults to the generic format.
         """
-        if format not in ('generic', 'unscaled', 'latex'):
+        if format not in ('generic', 'unscaled', 'latex', 'cds'):
             raise ValueError("Function units cannot be written in {} format. "
                              "Only 'generic', 'unscaled' and 'latex' are "
                              "supported.".format(format))
+        if format == 'cds':
```
mhvk: I think this is the right idea, but it should not be here, but rather in `DexUnit` (which is the only one that is allowed to be overridden; CDS cannot reproduce `mag` and `dB` function units). I.e., create a new method `to_string` there which has your special case for CDS, but otherwise just does `return super().to_string(....)`

mhvk:
@DShivansh - quick comment in-line. Please do add the test cases even if the code is not working yet, it makes thinking about changes much easier (and will tell you whether you got it right!).

p.s.  I'll be traveling the coming week, so responses may be slow.

astropy/units/function/core.py
DShivansh: Inside UnitDex what should I do if the format is not cds?

mhvk: `return super().to_string(....)`

DShivansh:
@mhvk can you please provide the readme with glob_header and multi_header that have the functional unit.
Can you please tell what all units should I check in ``TestRoundtripCDS `` 
Can you also guide me as to what kind of testing is required in ``test_read.py``?

mhvk:
@DShivansh - sorry, very quickly only: there are examples scattered throughout this PR and the original issue. The latter also has a link to a Vizier table which has them, might as well use the corresponding `ReadMe`.

astropy/units/format/cds.py
```diff
@@ -157,11 +169,15 @@ def p_main(p):
...
                  | combined_units
+                 | OPEN_BRACKET combined_units CLOSE_BRACKET
                  | factor
             '''
             from astropy.units.core import Unit
+            from astropy.units import dex
             if len(p) == 3:
                 p[0] = Unit(p[1] * p[2])
+            elif len(p) == 4:
+                p[0] = dex(Unit(p[2]))
```
mhvk: It doesn't matter too much but I think this can just be `p[0] = dex(p[2])` (`combined_unit` is guaranteed to be a unit instance already).

astropy/units/function/logarithmic.py
```diff
@@ -150,6 +150,21 @@ def _default_function_unit(self):
     def _quantity_class(self):
         return Dex
 
+    def to_string(self, format = 'generic'):
+        if format == 'cds':
+            return "[{}]".format(self.physical_unit.to_string(format=format))
+        else:
+            self_str = self.function_unit.to_string(format)
```
mhvk: No need to repeat code from above; as I noted in the original comment, you can just do
```
return super().to_string(format=format)
```

astropy/units/function/logarithmic.py
```diff
@@ -150,6 +150,21 @@ def _default_function_unit(self):
     def _quantity_class(self):
         return Dex
 
+    def to_string(self, format = 'generic'):
```
mhvk: No spaces around `=` for argument defaults.

DShivansh:
@mhvk thank you for correcting my PR. Can you please tell what is CR? sorry I forgot about the round-trip test case.

mhvk:
@DShivansh - CR=carriage return (was one missing at the end of the data file; really didn't matter much at all).  As tests pass, I'll merge this. Thanks again!

PR #7226, opened 2018-02-21
Adding TDISP support in FITS binary table serialisation of Table columns, and Table columns back to FITS Binary Table

SaOgaz:
Issue #7165 
@saimn, as far as I can tell, the TDISPn keywords are getting skipped over completely when the FITS binary table is read into an Astropy Table (as opposed to being put into the table meta data). They're in the KEYWORD_NAMES list imported from the io/fits/column file, meaning the is_column_keyword argument evaluates to true for the TDISPn keywords, and subsequently just get a pass in the loop:
https://github.com/astropy/astropy/blob/master/astropy/io/fits/connect.py#L261

It looks like getting TDISP into the table.format attribute won't take too many changes. But I'm not sure how best to take the TDISP format specification and translate that into the appropriate Python String formatting syntax.

No idea at the moment how much work the other direction, getting the column format attribute back out to the TDISPn keywords will be.

PR #9978, opened 2020-02-24
Add verification of disp (TDISP) keyword to fits.column and tests for TFORM

dhomeier:
### Description
`fits.Column._verify_keywords` checks its `disp` kwarg for type `str` but not much else, allowing just about any string to be quietly written as the `TDISPn` key to the FITS file header. Conversely, reading in a table per #7226 parses that key and raises an exception if not conforming to the FITS standard (case in question: I have produced lots of files lazily copying `TDISP2=E` from the corresponding `TFORM2` value, which are now rejected by `io.fits`).

This pull request is adding the same verification of the `disp` argument done on reading by calling `_parse_tdisp_format(disp)`.

Additional notes:

1. On adding the tests I found that the `format` (`TFORMn`) argument is implicitly checked, but not handled by the `_verify_keywords` framework and thus could not be tested in the same way as the other keywords. The second commit adds a failed `format` verification to the `invalid_kwargs` list.

2. `_parse_tdisp_format` actually only tests the beginning of the string for a conforming format specifier, therefore kwargs like `format='Exponential', disp='E14.6D+02'` are still accepted for writing - not conforming to the FITS standard specifications, but just as readily accepted on reading. Reading them might be OK, since they are unambiguous, but perhaps the writer should implement a stricter validation.

3. The `_verify_keywords` error messages generally state "The invalid keyword will be ignored for the purpose of formatting this column", sounding like the `Column` will still be created without optional kwargs or with default values – in that case treating `format` differently would be justified, since it is a required argument. But in fact any `invalid_kwargs` raise an exception, so currently there seems no need to discriminate between mandatory and optional arguments.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -373,6 +373,9 @@ astropy.io.fits
 
 - Fix writing noncontiguous data to a compressed HDU. [#9958]
 
+- Added verification of ``disp`` (``TDISP``) keyword to ``fits.Column`` and
+  extended tests for ``TFORM`` and ``TDISP`` validation. [#9978]
+
 astropy.io.registry
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


saimn:
Looks good, thanks @dhomeier .

PR #9981, opened 2020-02-24
BUG: Fix deprecated_renamed_argument bug when no new name given

### LINKED ISSUE #9914: deprecated_renamed_argument fails for a removed keyword (not solved: when it is called as positional argument)

pllim:
### Description

This pull request is to address a bug in `deprecated_renamed_argument` when a keyword is deprecated without replacement, the user value is ignored and default value is kept regardless.

The feature of deprecating a keyword without replacement was introduced in #8324 (v2.0.12).

Address part of #9914

p.s. With this patch, maybe a follow-up PR for #9915 can be made to use this decorator again. I did not do this here because that PR is milestoned for v4.1.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -414,7 +414,7 @@ astropy.time
 - Fixed exceptions in ``Time.to_value()``: when supplying any ``subfmt`` argument
   for string-based formats like 'iso', and for ``subfmt='long'`` for the formats
   'byear', 'jyear', and 'decimalyear'. [#9812]
-  
+
 - Fixed bug where the location attribute was lost when creating a new ``Time``
   object from an existing ``Time`` or list of ``Time`` objects. [#9969]
 
@@ -434,6 +434,9 @@ astropy.units
 
 astropy.utils
 ^^^^^^^^^^^^^
+
+- Fixed ``deprecated_renamed_argument`` not passing in user value to
+  deprecated keyword when the keyword has no new name. [#9981]
 
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


pllim:
@larrybradley , I noticed that about the positional argument but looking at the tests, maybe that use case is not meant to be supported? A positional argument is mandatory, which means, that warning will always appear until the maintainer removes it from API; how is such a warning useful to users?

pllim:
Ah, wait, you mean it is a keyword in the signature but user calls it as positional. As that is not a good practice to begin with, I am not inclined to fix it as part of this PR. Maybe a follow up PR in the future?

larrybradley:
@pllim Yes, I meant keyword in signature, but called as positional.  This PR fixes the biggest issue, so I'll go ahead and merge.

pllim:
Sounds good. Thanks! I'll keep that issue open until the other use case has a decision.

Issue #10000, opened 2020-03-01
Parameters fill_include_names and fill_exclude_names in astropy.io.ascii.read don't work as expected


Gabriel-p:
### Description
The parameters `fill_include_names` and `fill_exclude_names` in [astropy.io.ascii.read](https://docs.astropy.org/en/stable/api/astropy.io.ascii.read.html#astropy.io.ascii.read) don't work as expected.

### Expected behavior
Setting a column name in `fill_exclude_names` should exclude that column from the `fill_values` mask.

### Actual behavior
It works in a confusing way.

### Steps to Reproduce
```python
from astropy.io import ascii

test_data = """
ID  x  y
1   99  3.0
99  5.2  6.1
7   1.8  99
"""
fill_msk = [('', '0'), ('99', '0')]
```

1. Works as expected: all columns are masked
```
data = ascii.read(test_data, fill_values=fill_msk)
 ID  x   y 
--- --- ---
  1  -- 3.0
 -- 5.2 6.1
  7 1.8  --
```
1.  Works as expected: only the `x` column is excluded from the mask
```
data = ascii.read(test_data, fill_values=fill_msk, fill_exclude_names='x')
 ID  x    y 
--- ---- ---
  1 99.0 3.0
 --  5.2 6.1
  7  1.8  --
```
1.  Works as expected: only the `y` column is excluded from the mask
```
data = ascii.read(test_data, fill_values=fill_msk, fill_exclude_names='y')
 ID  x   y  
--- --- ----
  1  --  3.0
 -- 5.2  6.1
  7 1.8 99.0
```
1.  Works as expected: only the `ID` column is excluded from the mask
```
data = ascii.read(test_data, fill_values=fill_msk, fill_include_names=('x', 'y'), fill_exclude_names='ID')
 ID  x   y 
--- --- ---
  1  -- 3.0
 99 5.2 6.1
  7 1.8  --
```
1. **Does not** work as expected: the `y` column is excluded from the mask when a different name was given
```
data = ascii.read(test_data, fill_values=fill_msk, fill_exclude_names='nny')
 ID  x   y  
--- --- ----
  1  --  3.0
 -- 5.2  6.1
  7 1.8 99.0
```
1. **Does not** work as expected: the `ID` column is **not** excluded from the mask
```
data = ascii.read(test_data, fill_values=fill_msk, fill_exclude_names='ID')
 ID  x   y 
--- --- ---
  1  -- 3.0
 -- 5.2 6.1
[...]

PR #10005, opened 2020-03-02
Validate type of key arguments in io.ascii read() and write()

taldcroft:
### Description

Inspired by #10000, where the Python interpretation of a string input as a iterable of characters caused confusion, this PR adds explicit up-front type validation for many of the arguments to `io.ascii.read()` and `io.ascii.write()`.

This passes tests locally, and in getting there I uncovered variations in argument types that had been undocumented.  That was fixed and some elaboration on argument usages added.

I do have a worry that this explicit approach to argument types might break some custom classes or break for usages that are legal but not tested.  Comments welcome.
[EDIT] - the current implementation is more robust and more in the line of duck typing, basically asking if the object can be converted to the required type.

### DIFF
--- a/astropy/io/ascii/docs.py
+++ b/astropy/io/ascii/docs.py
@@ -96,12 +96,12 @@ READ_KWARG_TYPES = {
     # 'Reader'
     # 'Inputter'
     # 'Outputter'
-    'delimiter': (str, np.str_),
-    'comment': (str, np.str_),
-    'quotechar': (str, np.str_),
-    'header_start': (int, np.integer),
-    'data_start': (int, np.integer, str, np.str_),  # CDS allows 'guess'
-    'data_end': (int, np.integer),
+    'delimiter': str,
+    'comment': str,
+    'quotechar': str,
+    'header_start': int,
+    'data_start': (int, str),  # CDS allows 'guess'
+    'data_end': int,
     'converters': dict,
     # 'data_Splitter'
     # 'header_Splitter'
@@ -111,8 +111,8 @@ READ_KWARG_TYPES = {
     'fill_values': 'list-like',
     'fill_include_names': 'list-like',
     'fill_exclude_names': 'list-like',
-    'fast_reader': (bool, np.bool_, str, np.str_, dict),
-    'encoding': (str, np.str_),
+    'fast_reader': (bool, str, dict),
+    'encoding': str,
 }
 
 
@@ -170,16 +170,16 @@ WRITE_DOCSTRING = """
 WRITE_KWARG_TYPES = {
     # 'table'
     # 'output'
-    'format': (str, np.str_),
-    'delimiter': (str, np.str_),
-    'comment': (str, np.str_, bool, np.bool_),
-    'quotechar': (str, np.str_),
-    'header_start': (int, np.integer),
+    'format': str,
+    'delimiter': str,
+    'comment': (str, bool),
+    'quotechar': str,
+    'header_start': int,
     'formats': dict,
-    'strip_whitespace': (bool, np.bool_),
+    'strip_whitespace': (bool),
     'names': 'list-like',
     'include_names': 'list-like',
     'exclude_names': 'list-like',
-    'fast_writer': (bool, np.bool_, str, np.str_),
-    'overwrite': (bool, np.bool_),
+    'fast_writer': (bool, str),
+    'overwrite': (bool),
 }
--- a/astropy/io/ascii/docs.py
+++ b/astropy/io/ascii/docs.py
@@ -86,9 +86,11 @@ READ_DOCSTRING = """
 
     """
 
-# Specify allowed types for core read() keyword arguments.
-#   The commented-out kwargs are too flexible for a useful check
-#   'list-list' is basically an iterable that is not a string.
+# Specify allowed types for core write() keyword arguments.  Each entry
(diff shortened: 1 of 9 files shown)


astropy/io/ascii/docs.py
```diff
@@ -48,13 +50,13 @@
...
         List of names to exclude from output (applied after ``include_names``)
-    fill_values : dict
+    fill_values : tuple, list of tuple
         specification of fill values for bad or missing table values
     fill_include_names : list
         List of names to include in fill_values.
     fill_exclude_names : list
         List of names to exclude from fill_values (applied after ``fill_include_names``)
-    fast_reader : bool or dict
+    fast_reader : bool, str or dict
```
pllim: Oxford comma. :smile_cat: 
```suggestion
    fast_reader : bool, str, or dict
```

pllim:
cc @gabriel-p

docs/io/ascii/read.rst
```diff
@@ -120,11 +120,14 @@ Parameters for ``read()``
...
+**fill_include_names** : list of column names affected by ``fill_values``.
+  From the list of column names found from the header or the ``names``
+  parameter, select for filling only columns within this list. If not supplied,
+  then apply ``fill_values`` to all columns.
 
-**fill_exclude_names** : list of column names, which are not affected by ``fill_values``.
-  If not supplied, then ``fill_values`` can affect all columns.
+**fill_exclude_names** : list of column names not affected by ``fill_values``.
+  Exclude these names from the list of columns for filling. This is applied *after*
+  the ``fill_include_names`` filtering. If not specified then no columns are excluded.
```
hamogu: Is there any reason to even allow `fill_exclude_names` at all if `fill_include_names` is used? `fill_include_names` must be an exhaustive list of all column names to be filtered, so if you want to exclude one of those, just don't list it in `fill_include_names`. Your clarification of the doc string is correct, but it's still complicated. I think form a user perspective it would be much simpler, to use only include or exclude, but not both.

(For backwards compatibility, we might not enforce that really only one of them is used, but if we document it that way, users know that they are on your own if they do so.)

Another approach is to describe more when these options make sense and less how it's acutally done in the code:
```
**fill_include_names** : list of column names affected by ``fill_values``.
  This is a list of column names (found from the header or the ``names``
  parameter) for all columns where values will be filled. `None` (the default) will
  apply ``fill_values`` to all columns.

**fill_exclude_names** : list of column names not affected by ``fill_values``.
  This is a list of column names (found from the header or the ``names``
  parameter) for all columns where values will be **not** be filled. Listing a 
  few columns in `fill_exclude_names` has   the same effect as listing all the 
  other columns in `fill_include_names`. If a column name is listed in both
  `fill_include_names` and `fill_exclude_names`, its values are **not** filled.
  `None` (default) does not exclude any columns.
```
(Note that the formatting is probably screwed up because I type this in the github web form.)

astropy/io/ascii/ui.py
```diff
@@ -201,13 +203,54 @@ def _get_fast_reader_dict(kwargs):
...
+                try:
+                    new_val = cls(val)
+                    assert new_val == val
+                except Exception:
+                    ok = False
+                else:
+                    ok = True
+        return ok
+
+    kwarg_types = READ_KWARG_TYPES if read_write == 'read' else WRITE_KWARG_TYPES
```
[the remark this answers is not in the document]
taldcroft: This is not a user-facing function so I'm not worried about the input being wrong.

astropy/io/ascii/ui.py
```diff
@@ -201,13 +203,54 @@ def _get_fast_reader_dict(kwargs):
...
+        # This reflects that some readers allow additional arguments that may not
+        # be well-specified, e.g. ```__init__(self, **kwargs)`` is an option.
+        if arg not in kwarg_types or val is None:
+            continue
+
+        types = kwarg_types[arg]
+        err_msg = (f"{read_write}() argument '{arg}' must be a "
+                   f"{types} object, got {type(val)} instead")
+
+        if not isinstance(types, tuple):
```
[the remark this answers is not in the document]
taldcroft: I can do that, but the current definition matches `isinstance(object, classinfo)`, where `types` can be either a single type or a tuple of types.  It's just a bit tedious and visually noisy to have `(int,)` everywhere.   I'll add more comments (overall the code is under-commented, so I'll fix that).

docs/io/ascii/read.rst
taldcroft: This is sort of a common issue, e.g. in rsync you have `--include` and `--exclude` that are mostly useful on their own but are not strictly mutually exclusive.  Generally speaking I would imagine most users would use one or the other, but using both should not be an exception.  An example (perhaps contrived) where you might want to use both is if you have a long list of column names and you want to use fill values for every column starting with `s`, except for the `s1` and `s2 ` column.
```
fill_include_names = [colname for colname in colnames if colname.startswith('s')]
fill_exclude_names = ['s1', 's2']
```
How about your version but with `fill_exclude_names` like:
```
**fill_exclude_names** : list of column names not affected by ``fill_values``.
  This is a list of column names (found from the header or the ``names``
  parameter) for all columns where values will be **not** be filled.
  This parameter takes precedence over ``fill_include_names``.  A value
  of `None` (default) does not exclude any columns.

hamogu: Sounds good.

PR #10012, opened 2020-03-05
Fix checksum verification to process all HDUs

### LINKED ISSUE #8115: Checksum not verified until loading HDUs
Since the addition of lazy loading, checksum are verified only for the primary HDU when opening the file, and the other HDUs are verified only when (and if) there are loaded. So if an invalid HDU is not explicitly loaded, it will not be verified.
I think this is wrong, and specifying `checksum=True` in `fits.open` should load all HDUs to verify them. The documentation also states that opening is enough to verify the file: http://docs.astropy.org/en/latest/io/fits/usage/verification.html#verification-using-the-fits-checksum-keyword-convention

(example after adding prints in `_ValidHDU._verify_checksum_datasum`).
```
In [1]: %astropy
Numpy 1.15.4
Astropy 3.2.dev23320

In [2]: hdu = fits.ImageHDU(data=np.arange(10))

In [3]: hdu.writeto('test.fits', checksum=True, overwrite=True)

In [4]: hdul = fits.open('test.fits', checksum=True)
verify checksum <astropy.io.fits.hdu.image.PrimaryHDU object at 0x7fedcefbf240>
verify datasum <astropy.io.fits.hdu.image.PrimaryHDU object at 0x7fedcefbf240>

In [5]: len(hdul)
verify checksum <astropy.io.fits.hdu.image.ImageHDU object at 0x7fedcefc6320>
verify datasum <astropy.io.fits.hdu.image.ImageHDU object at 0x7fedcefc6320>
Out[5]: 2
```

saimn:
Fix #8115.

### DIFF
--- a/astropy/io/fits/hdu/hdulist.py
+++ b/astropy/io/fits/hdu/hdulist.py
@@ -1079,7 +1079,7 @@ class HDUList(list, _Verify):
 
             raise OSError('Empty or corrupt FITS file')
 
-        if not lazy_load_hdus:
+        if not lazy_load_hdus or kwargs.get('checksum') is True:
             # Go ahead and load all HDUs
             while hdulist._read_next_hdu():
                 pass
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -383,6 +383,9 @@ astropy.io.fits
 - Added verification of ``disp`` (``TDISP``) keyword to ``fits.Column`` and
   extended tests for ``TFORM`` and ``TDISP`` validation. [#9978]
 
+- Fix checksum verification to process all HDUs instead of only the first one
+  because of the lazy loading feature. [#10012]
+
 astropy.io.registry
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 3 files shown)


astropy/io/fits/hdu/hdulist.py
```diff
@@ -1079,7 +1079,7 @@ def _readfrom(cls, fileobj=None, data=None, mode=None,
 
             raise OSError('Empty or corrupt FITS file')
 
-        if not lazy_load_hdus:
+        if not lazy_load_hdus or kwargs.get('checksum') is True:
```
MSeifert04: Why the `is True`? Wouldn't `kwargs.get('checksum)` be sufficient?

saimn: Because it can set to 'remove', in which case the checksums are not verified.

MSeifert04: Ah, right. Still weird 🤔

PR #10016, opened 2020-03-07
Fix technically correct, but non-sensical example in docstring

hamogu:
In astronomical coordinates, we put in RA, DEC and typically RA is given
as hourangle and DEC in degrees. The example did te opposite.
While in principle, the math allows us to specify the RA
in degrees and the DEC in hourangle, that's never ever done.

Thus the example should also give units in (hourangle, deg)
and not the other way around.
Note that the example as give would fail, if DEC was > 24, so
just to make sure, I also changed a number.

### DIFF
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -124,7 +124,7 @@ class SkyCoord(ShapedLikeNDArray):
       >>> c = SkyCoord([1, 2, 3], [-30, 45, 8], frame="icrs", unit="deg")  # 3 coords
 
       >>> coords = ["1:12:43.2 +1:12:43", "1 12 43.2 +1 12 43"]
-      >>> c = SkyCoord(coords, frame=FK4, unit=(u.deg, u.hourangle), obstime="J1992.21")
+      >>> c = SkyCoord(coords, frame=FK4, unit=(u.hourangle, u.deg), obstime="J1992.21")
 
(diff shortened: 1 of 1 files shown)


pllim:
Good catch!

Should we straight up disallow `hourangle` being defined in the Dec field?

> I also changed a number

I didn't see any number changing in the diff. Did I misunderstand something?

hamogu:
@pllim I wanted to change a number, but I forgot to save the file before I commited other other change. When I noticed that, @taldcroft had already merged and I decided not to bother to open a second PR for that.

I'm in favor of not allowing hourangle for Dec because it makes no sense, but I'm not sure how to do that. My understanding is that the units machinery abstracts away the input format and converts things automatically so that the SkyCoord does not even know what type of angle (degree, hourangle, rad) the user put in?

taldcroft:
I'm -1 on disallowing hourangle.  We should not assume that we can think of every use case in the universe of possibilities.  Special casing this one unit just makes the code more complicated and needs to include extra documentation and tests.

pllim:
Still, maybe a follow up PR anyway to change the number? I feel like it would be a waste to lose that improvement to doctest.

hamogu:
Thinking about it again, I don't think there is any possibility in the universe of possibilities that hourangle should be used for Dec and if there is, it's a one-in-a-million case. In that case, people presumably know why they are doing and could convert the angle to a different units by hand before.

However, I also agree with @taldcroft that it's not worth the effort to make an extra code change. 

I'll do a second PR to add the number that I forgot to include in this one. It just makes it visually clear that the number that stars with -30 cannot be RA.

PR #10069, opened 2020-03-24
Fix column access by attribute with FITS_rec

### LINKED ISSUE #10065: Fits data field access via dict-like and dotted notation seems inconsistent
### Description

Using dotted notation (`hdu.data.name` vs `hdu.data['name']`) seems inconsistent and varies with the type of extension type (acsii, binary table).

### Expected behavior

Accessing data fields by either method should be equivalent or raise and error when using unsupported access method (dotted in this case).

### Actual behavior

|Extension Type | Status | Description |
|----------------|-------|------------|
|AsciiTable | ❌ | returns different (uncalibrated) values depending on method |
|BinTable (fixed length) | ✅ | seems to work
|BinTable (variable length) | ❌| returns different arrays? |

#### Ascii
```python
from astropy.io import fits 
ascii_file = fits.util.get_testdata_filepath('ascii.fits') 
ascii_table_hdul = fits.open(ascii_file)
```

Dict-like access works correctly
```python
ascii_table_hdul[1].data['a']
array([ 10.123,   5.2  ,  15.61 ,   0.   , 345.   ])
```

dotted access returns unconverted data
```
ascii_table_hdul[1].data.a
array([b'.10123E+02', b'.52000E+01', b'.15610E+02', b'*         ',
       b'.34500E+03'], dtype='|S10')
```

#### BinTable (fixed length)
```
bintable_file = fits.util.get_testdata_filepath('btable.fits')
bintable_hdul = fits.open(bintable_file)
```

Both dict-like and dotted access return the same data
```python
bintable_hdul[1].data.order
array([1, 2, 3], dtype=int16)

bintable_hdul[1].data['order']
array([1, 2, 3], dtype=int16)
```

#### BinTable (variable  length)
```python
col1 = fits.Column(name='x', format='PI()', array=np.array([[45, 56], [11, 12, 13]], dtype=np.object_))
varbin_hdu = fits.BinTableHDU.from_columns([col1])
```
Dict-like access works
```python
varbin_hdu.data['x']
_VLF([array([45, 56], dtype=int16), array([11, 12, 13], dtype=int16)],
     dtype=object)
```

dotted access doesn't, not sure what it is returning here
```python
varbin_hdu.data.x
array([[2, 0],
       [3, 4]], dtype=int32)
```

### System Details

Darwin-19.3.0-x86_64-i386-64bit
Python 3.7.0 (default, Nov  7 2018, 00:55:12)
[Clang 10.0.0 (clang-1000.11.45.5)]
Numpy 1.18.2
astropy 4.0
Scipy 1.4.1

saimn:
Fix #10065: columns with scaling or columns from ASCII tables where not
properly converted when accessed by attribute name.
cc @mhvk, would be great if you can have a look, since you have more experience with ndarray subclasses.

### DIFF
--- a/astropy/io/fits/fitsrec.py
+++ b/astropy/io/fits/fitsrec.py
@@ -479,6 +479,20 @@ class FITS_rec(np.recarray):
         # one added for recarray in Numpy 1.10) for backwards compat
         return np.ndarray.__repr__(self)
 
+    def __getattribute__(self, attr):
+        # See if ndarray has this attr, and return it if so. (note that this
+        # means a field with the same name as an ndarray attr cannot be
+        # accessed by attribute). NOTE: this is Numpy's default behavior.
+        try:
+            return object.__getattribute__(self, attr)
+        except AttributeError:  # attr must be a fieldname
+            pass
+
+        if self._coldefs is None:
+            return super().__getattribute__(attr)
+
+        return self.field(attr)
+
     def __getitem__(self, key):
         if self._coldefs is None:
             return super().__getitem__(key)
--- a/astropy/io/fits/fitsrec.py
+++ b/astropy/io/fits/fitsrec.py
@@ -480,21 +480,24 @@ class FITS_rec(np.recarray):
         return np.ndarray.__repr__(self)
 
     def __getattribute__(self, attr):
-        # See if ndarray has this attr, and return it if so. (note that this
-        # means a field with the same name as an ndarray attr cannot be
-        # accessed by attribute). NOTE: this is Numpy's default behavior.
+        # First, see if ndarray has this attr, and return it if so. Note that
+        # this means a field with the same name as an ndarray attr cannot be
+        # accessed by attribute, this is Numpy's default behavior.
+        # We avoid using np.recarray.__getattribute__ here because after doing
+        # this check it would access the columns without doing the conversions
+        # that we need (with .field, see below).
         try:
             return object.__getattribute__(self, attr)
-        except AttributeError:  # attr must be a fieldname
+        except AttributeError:
             pass
 
-        if self._coldefs is None:
-            return super().__getattribute__(attr)
-
-        if attr in self.columns.names:
+        # attr might still be a fieldname.  If we have column definitions,
+        # we should access this via .field, as the data may have to be scaled.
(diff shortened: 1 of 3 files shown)


saimn:
Hmm, this breaks some tests, will investigate later.

saimn:
Should be good now :crossed_fingers:

astropy/io/fits/fitsrec.py
```diff
@@ -479,6 +479,23 @@ def __repr__(self):
         # one added for recarray in Numpy 1.10) for backwards compat
         return np.ndarray.__repr__(self)
 
+    def __getattribute__(self, attr):
```
mhvk: Hmm, at first I thought that is really weird, to try `object.__getattribute__` and if that fails, do something else -- this is exactly what `__getattr__` is for! But then, of course, I realized that since `np.recarray` overrides `__getattribute__`, you really have no choice but to override it as well. That's definitely silly in `np.recarray`...

Anyway, a minimal request would be to add a comment to that extent! But since I was playing with it anyway, I also thought it could be a bit shorter:
```
    def __getattribute__(self, attr):
        # Unfortunately, np.recarray overrides __getattribute__, which
        # means we have to override it too, to change its behaviour.
        # Since this is called before anything else, as the first step,
        # we *have* to check whether the attribute simply exists.
        # (Normally, one would use __getattr__ to avoid that.)
        try:
            return object.__getattribute__(self, attr)
        except AttributeError:
            # attr might still be a fieldname.  If we have column
            # definitions, we should access this via .field, as the
            # data may have to be scaled.
            if self._coldefs is not None and attr in self.columns.names:
                return self.field(attr)

        # If not, just let the usual np.recarray override deal with it.
        return super().__getattribute__(attr)
```

saimn: Yes sorry, I should have explained this. I also need to call it first to keep the same behavior as Numpy, finding ndarray attributes first (so a `'var'` column will not be accessible with the dot notation, `np.var` is returned instead). But then we need to check for columns before calling `recarray.__getattribute__`.
Thanks for the shorter version, that's much better indeed!

mhvk:
All tests passed, so merging. Thanks, @saimn!

PR #10071, opened 2020-03-24
Fix Documentation Bug in io.ascii

### LINKED ISSUE #9999: Inconsistencies in the docs regarding the `guess` parameter
1. 
The https://docs.astropy.org/en/stable/io/ascii/index.html#guess-table-format url in https://docs.astropy.org/en/stable/io/ascii/index.html#reading-tables does not exist. The correct link is https://docs.astropy.org/en/stable/io/ascii/read.html#guess-table-format

2. 
In https://docs.astropy.org/en/stable/io/ascii/read.html#guess-table-format it is stated that:

> If the guess parameter in read() is set to True (which is the default) then...

but the `guess` parameters is set to `None` [by default](https://docs.astropy.org/en/stable/api/astropy.io.ascii.read.html#astropy.io.ascii.read), not `True`.

robelgeda:
This PR provides doc fixes as mentioned by #9999 

Fixes #9999

### DIFF
--- a/docs/io/ascii/index.rst
+++ b/docs/io/ascii/index.rst
@@ -70,7 +70,7 @@ The first argument to the |read| function can be the name of a file, a string
 representation of a table, or a list of table lines. The return value
 (``data`` in this case) is a :ref:`Table <astropy-table>` object.
 
-By default, |read| will try to `guess the table format <#guess-table-format>`_
+By default, |read| will try to :ref:`guess the table format <guess_formats>`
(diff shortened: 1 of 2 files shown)


pllim:
Thanks!

cc @Gabriel-p

hamogu:
This changes only docs, not code, but since it's only going to be a few more minutes for the Travis job to finish, I might as well wait for that before merging.

PR #10085, opened 2020-03-30
Replace uses of tostring with tobytes

saimn:
`ndarray.tostring` will be deprecated in Numpy 1.19 : numpy/numpy#15867

### DIFF
--- a/astropy/units/tests/test_quantity_array_methods.py
+++ b/astropy/units/tests/test_quantity_array_methods.py
@@ -513,6 +513,8 @@ class TestArrayConversion:
         with pytest.raises(NotImplementedError):
             q1.tostring()
         with pytest.raises(NotImplementedError):
+            q1.tobytes()
+        with pytest.raises(NotImplementedError):
             q1.tofile(0)
         with pytest.raises(NotImplementedError):
             q1.dump('a.a')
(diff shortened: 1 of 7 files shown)


astropy/units/tests/test_quantity_array_methods.py
```diff
@@ -511,6 +511,7 @@ def test_not_implemented(self):
         with pytest.raises(NotImplementedError):
             q1.tolist()
         with pytest.raises(NotImplementedError):
+            # FIXME: tobytes instead ?
             q1.tostring()
```
saimn: @mhvk - What to do about this one? Should `tobytes` be added here and in `astropy/units/quantity.py` ?

pllim:
As long as tests pass, the `votable` change is fine by me. But I don't see the Travis CI job listed here. Where did it go?

astropy/units/tests/test_quantity_array_methods.py
mhvk: Yes, I think it was an oversight not to have those. If you could add those, that would be great!

saimn: @mhvk - Ok, done.

PR #10095, opened 2020-03-31
Add bit flag name map

mcara:
### Description
This PR adds support for using mnemonic bit field names in the `nddata.bitmask` module when specifying which bit fields can be used (or ignored) when converting, i.e., a DQ array to a boolean mask.

CC: @perrygreenfield @crawfordsm 

**TODO**

- [x] Get approval for the proof of concept to proceed.
- [x] Add tests.

### DIFF
--- a/astropy/nddata/bitmask.py
+++ b/astropy/nddata/bitmask.py
@@ -633,14 +633,14 @@ good_mask_value=False, dtype=numpy.bool_)
         ...                                  dtype=int)
         array([[0, 0, 0, 1, 0, 0, 1, 0],
                [1, 1, 0, 0, 0, 0, 1, 0]])
-        >>> bitmask.bitfield_to_boolean_mask(dqarr, ignore_flags='CR,CLOUDY',
-        ...                                  flip_bits=True, good_mask_value=0,
-        ...                                  dtype=int, flag_name_map=flag_map)
+        >>> bitmask.bitfield_to_boolean_mask(dqarr, ignore_flags='~(CR,CLOUDY)',
+        ...                                  good_mask_value=0, dtype=int,
+        ...                                  flag_name_map=flag_map)
         array([[0, 0, 0, 1, 0, 0, 1, 0],
                [1, 1, 0, 0, 0, 0, 1, 0]])
-        >>> bitmask.bitfield_to_boolean_mask(dqarr, ignore_flags='CR+CLOUDY',
-        ...                                  flip_bits=True, good_mask_value=0,
-        ...                                  dtype=int, flag_name_map=flag_map)
+        >>> bitmask.bitfield_to_boolean_mask(dqarr, ignore_flags='~(CR+CLOUDY)',
+        ...                                  good_mask_value=0, dtype=int,
+        ...                                  flag_name_map=flag_map)
         array([[0, 0, 0, 1, 0, 0, 1, 0],
                [1, 1, 0, 0, 0, 0, 1, 0]])
 
--- a/astropy/nddata/bitmask.py
+++ b/astropy/nddata/bitmask.py
@@ -218,10 +218,10 @@ def extend_bit_flag_map(cls_name, base_cls=BitFlagNameMap, **kwargs):
         >>> from astropy.nddata.bitmask import extend_bit_flag_map
         >>> ST_DQ = extend_bit_flag_map('ST_DQ', __version__='1.0.0', CR=1, CLOUDY=4, RAINY=8)
         >>> ST_CAM1_DQ = extend_bit_flag_map('ST_CAM1_DQ', ST_DQ, HOT=16, DEAD=32)
-        >>> JWST_MIRI_DQ['HOT']  # <-- Access flags as dictionary keys
-        256
-        >>> JWST_MIRI_DQ.HOT  # <-- Access flags as class attributes
-        256
+        >>> ST_CAM1_DQ['HOT']  # <-- Access flags as dictionary keys
+        16
+        >>> ST_CAM1_DQ.HOT  # <-- Access flags as class attributes
+        16
 
     """
     new_cls = BitFlagNameMeta.__new__(
@@ -246,7 +246,7 @@ def interpret_bit_flags(bit_flags, flip_bits=None, flag_name_map=None):
     Converts input bit flags to a single integer value (bit mask) or `None`.
 
     When input is a list of flags (either a Python list of integer flags or a
-    sting of comma-, ``'|'``-, or ``'+'``-separated list of flags),
+    string of comma-, ``'|'``-, or ``'+'``-separated list of flags),
     the returned bit mask is obtained by summing input flags.
 
     .. note::
@@ -313,7 +313,7 @@ def interpret_bit_flags(bit_flags, flip_bits=None, flag_name_map=None):
         >>> "{0:016b}".format(0xFFFF & interpret_bit_flags('~(4+8+16)'))
         '1111111111100011'
         >>> "{0:016b}".format(0xFFFF & interpret_bit_flags('~(CLOUDY+RAINY+HOT)',
-        ... flag_name_map=ST_DQ)))
+        ... flag_name_map=ST_DQ))
         '1111111111100011'
         >>> "{0:016b}".format(0xFFFF & interpret_bit_flags([4, 8, 16]))
         '0000000000011100'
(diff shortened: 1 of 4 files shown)


astropy/nddata/bitmask.py
```diff
@@ -92,11 +276,15 @@ def interpret_bit_flags(bit_flags, flip_bits=None):
...
+        >>> ST_DQ = extend_bit_flag_map('ST_DQ', CR=1, CLOUDY=4, RAINY=8, HOT=16, DEAD=32)
         >>> "{0:016b}".format(0xFFFF & interpret_bit_flags(28))
         '0000000000011100'
         >>> "{0:016b}".format(0xFFFF & interpret_bit_flags('4,8,16'))
         '0000000000011100'
+        >>> "{0:016b}".format(0xFFFF & interpret_bit_flags('CLOUDY,RAINY,HOT', flag_name_map=ST_DQ))
+        '0000000000011100'
         >>> "{0:016b}".format(0xFFFF & interpret_bit_flags('~4,8,16'))
         '1111111111100011'
         >>> "{0:016b}".format(0xFFFF & interpret_bit_flags('~(4+8+16)'))
```
pllim: Can you do this `'~(4+8+16)'` with named bits too?

pllim:
Thanks, @mcara ! You might need to rebase against latest upstream master to fix the CI.

Question for maintainers: Does this need a what's new entry?

mhvk:
Implementation question: would it make sense to use/move to the new [`enum.Flag`](https://docs.python.org/3/library/enum.html#flag)? Would certainly be good to make it similar to that in the API (if it isn't already - I'll admit just glancing at the implementation)

astropy/nddata/bitmask.py
mcara: @pllim Yes, one can. It is identical: `name<->value`

pllim: Maybe add an example?

pllim:
p.s. I also wonder if it accepts emoji name. 🤔

mcara:
@mhvk I do not think `Flag`'s behavior is what I had in mind. The most important difference is that I can't subclass a `Flag`. The most typical use of this class that I think of is by somebody in the "instrument" teams of observatories what would define important flags for their instruments. For example, they could define in one of their software:
```python
from super_observatory.instrument import MAGIC_CAM_DQ
```
and then just use it as is. In general it is rare to want to add new flags but not unconceivable. For example, in the `drizzlepac` (one of our packages) we add `1024` to the DQ array to mark pixels that we have detected as being cosmic rays.

The software developer of the `super_observatory`, in collaboration with the instrument teams, may do the following:
```
# in instrument.py
from astropy.nddata.bitmask import BaseBitFlagNameMap
class SUPER_OBS_DQ(BaseBitFlagNameMap):
    POWER_ON = 1
    CLOUDY = 2

class MAGIC_CAM_DQ(SUPER_OBS_DQ):
    RABBIT_IN_HAT = 4

class REGULAR_CAM_DQ(SUPER_OBS_DQ):
    SHUTTER_OPEN = 4
```
and so users can easily import pre-defined mappings. Subclassing or adding new flags is allowed but results in a new class so that base(s) is not modified. Also, I wanted to indexing by keys/flags.

Probably I should rename `BaseBitFlagNameMap` to simply `BitFlagNameMap`. Also, looking at `enum`, I could make error messages (and types) similar to `enum` and maybe add support for using dictionaries when adding new flags using `+` (not just lists of tuples).

I do not think that support for bitwise operators is needed here: the purpose of this object is to "translate" instead of serving as a mask.

mhvk:
@mcara - ah, yes, the subclassing would indeed not work for `Enum`. I certainly have found before that new & nifty features sometimes just aren't quite what one hoped...

mcara:
Question/Note to reviewers. I know this PR does not have unit tests. I will add the tests after I get a nod to proceed if this PR is OK at conceptual level.

perrygreenfield:
OK with me to proceed, but it would be good to get acknowledgement from those with a clear interest in this. Do we have a list of those people?

mcara:
@crawfordsm ?

pllim:
@mcara , I added checkboxes in your original post above. You can tick them off when they are done. Hope this helps!

Also, now that those flags have names, how do you "version control" them? I have seen instrument team repurpose an old flag for something else half-way through the mission.

perrygreenfield:
I'd also add an LSST person ( @timj ?)

mcara:
> how do you "version control" them? I have seen instrument team repurpose an old flag for something else half-way through the mission.


Oh oh oh oh

PaulPrice:
Based on a skim reading, and zero familiarity with current astropy mask code:
* I didn't see anywhere the requirement that the values be powers of two is enforced. Why not have the user specify the power of two, rather than the actual value?
* Does this (or should this) support derived values, e.g., `IGNORE = HOT+SATURATED`?
* Many of us think of bitmasks in binary, so consider supporting `|` in addition to `+`. Maybe `&` too.
* You should consider requiring (or at least allowing) a documentation string (in code, i.e., not a `#` comment) for each value. A good name helps, but having an extended explanation can be helpful. In case it's useful, [here](https://github.com/Subaru-PFS/datamodel/blob/master/python/pfs/datamodel/pfsConfig.py#L14-L44) is an similarish implementation using `enum.IntEnum`.
* Having to call `interpret_bit_flags` each time is annoying. Could you make the bitmask names first-class objects (`enum.IntFlags` might be useful for this), and allow them to operate with bitmasks directly, e.g.:
```
    from .myMasks import BAD, SATURATED
    if source.flags & (BAD|SATURATED):
        # do stuff
```

crawfordsm:
Thanks for the contribution @mcara !   Adding a comment that this is related to closing #8243. 

Overall, I think this does conceptional work for what is being described there.   On my high level review, the capability of using the bit names for processing is there in the example you are giving, and the usage of it as a class enables some of the things that you point out along with other features, so I think that is a good solution.

crawfordsm:
To answer some of @pllim questions:

1. Yes this will need a what's new entry

2. As to version control for the names, I think that is best left to the developers that use this feature and how they want to handle that.   In this implementation, there is no set names.   Adding a class attribute of `version` might be helpful to encourage that practice to make it more reproducible, but I don't think would be required.

pllim:
Thanks, @crawfordsm ! So, does this completely resolve #8243 or just parts of it?

mcara:
I actually completely forgot about https://github.com/astropy/astropy/issues/8243 Let's keep it open for now please. There are some things that I should at least look into but I do not have time right now.

mcara:
I have added unit tests. I have also added support for `|` as bit flag separator (in addition to `+`, `,`). Updated documentation accordingly.

mcara:
To address some points in @PaulPrice comments:

> * I didn't see anywhere the requirement that the values be powers of two is enforced. Why not have the user specify the power of two, rather than the actual value?

Mostly because of how users used to describe flags HST(in fact, I found IRAF code doing the same) and now that is how [JWST](http://www.stsci.edu/files/live/sites/www/files/home/jwst/documentation/technical-documents/_documents/JWST-STScI-002082.pdf) (see section 4.3.1) defines its flags. It appears that even LSST uses *flag values* to describe DQ flags - see "Table 4-2: Meanings of Source Data Quality Flags" in [LSST DATA CHALLENGE HANDBOOK](https://project.lsst.org/sciencewiki/images/DC_Handbook_v1.1.pdf).

> * Does this (or should this) support derived values, e.g., `IGNORE = HOT+SATURATED`?

This code does not support this kind of usage mostly because a "derived value" is not a (single) *bit* flag.

> * Many of us think of bitmasks in binary, so consider supporting `|` in addition to `+`. Maybe `&` too.

Added support for `|` in the last commit. I do not see how `&` could be useful with bit flags.

> * You should consider requiring (or at least allowing) a documentation string (in code, i.e., not a `#` comment) for each value. A good name helps, but having an extended explanation can be helpful. In case it's useful, [here](https://github.com/Subaru-PFS/datamodel/blob/master/python/pfs/datamodel/pfsConfig.py#L14-L44) is an similarish implementation using `enum.IntEnum`.

I think this is a useful feature and may/should be added in the future. I'll open an issue.

> * Having to call `interpret_bit_flags` each time is annoying. Could you make the bitmask names first-class objects (`enum.IntFlags` might be useful for this), and allow them to operate with bitmasks directly, e.g.:
> 
> ```
>     from .myMasks import BAD, SATURATED
>     if source.flags & (BAD|SATURATED):
>         # do stuff
> ```

I am not sure I understand this. In principle there is nothing that prevents one to do `source.flags & (BAD|SATURATED)` directly on a `numpy.ndarray` (if I understood the idea). Using `interpret_bit_flags` is not even required to convert a DQ array to a boolean array - one should be using `bitfield_to_boolean_mask()` directly.

astropy/nddata/bitmask.py
```diff
@@ -49,13 +50,204 @@ def _is_int(n):
...
+
+
+def interpret_bit_flags(bit_flags, flip_bits=None, flag_name_map=None):
     """
     Converts input bit flags to a single integer value (bit mask) or `None`.
 
     When input is a list of flags (either a Python list of integer flags or a
-    sting of comma- or '+'-separated list of flags), the returned bit mask
-    is obtained by summing input flags.
+    sting of comma-, ``'|'``-, or ``'+'``-separated list of flags),
```
perrygreenfield, line 249: Typo sting -> string

PaulPrice:
If your scheme is that all the values must be a power of two, then allowing people to write the values rather than the power of two is inviting mistakes. LSST does not use flag values in the code: the implementation of powers of two is hidden from the user (see [here](https://github.com/lsst/afw/blob/master/include/lsst/afw/image/Mask.h#L463) and [here](https://github.com/lsst/afw/blob/master/src/image/detail/MaskDict.cc#L206-L216); we're not completely happy with this implementation but it does the job); the table you cite is a description of what the code produces, for people who don't use our code. I don't think using direct values is wrong, but I think abstracting away the notion that these values are powers of two is helpful, and makes the code easier for people to use who aren't familiar with bitmasks.

astropy/nddata/bitmask.py
```diff
@@ -347,10 +596,14 @@ def bitfield_to_boolean_mask(bitfield, ignore_flags=0, flip_bits=None,
 
     Examples
     --------
+
         >>> from astropy.nddata import bitmask
         >>> import numpy as np
         >>> dqbits = np.asarray([[0, 0, 1, 2, 0, 8, 12, 0],
```
perrygreenfield, line 602: Is dqbits a good name for this quantity? I would favor dqvals or something like that

perrygreenfield:
@PaulPrice It seems there are perhaps two issues here: first, whether bits should be defined by their value or bit position. This is one of those things that has adherents on both sides, both with reasonable points of view. I don't think one or the other satisfies everyone. I haven't looked at this closely, but my sense is that nothing in this design precludes a different interface for specifying the actual bits. E.g., bitpos or bit_position being the alternate to bit value. So why can't both be supported? And adding that later shouldn't prevent this from being merged, unless there is a irreconcilable conflict between the two approaches (e.g., if all bitpos values are converted to the integer value internally, is that a problem?). Secondly, shielding the user from the actual values, whether bit value or bit position, doesn't this allow it as well (or at least not preclude future methods that basically say, find me then next free bit to associate with this name?)

The issue here is not whether all these capabilities should be part of this commit, but whether they are precluded from this commit. We don't want to make this unavailable for an additional 6 months to wait for these additions.

PaulPrice:
@perrygreenfield Agreed.

pllim:
Err... So the 32 bit failure was unrelated?

mcara:
> Err... So the 32 bit failure was unrelated?

If they are: how???

pllim:
>  how???

My peeps will call your peeps. 😉

PR #10103, opened 2020-04-02
Fix bugs / performance when sorting a table

### LINKED ISSUE #6545: Sorting a Table with index fails
```python
In [1]: %astropy
Numpy 1.13.1
Astropy 3.0.dev19860

In [2]: t = Table([[1,3,2], [6,4,5]], names=('a', 'b'))

In [3]: t.add_index('a')

In [4]: t.sort('a')

In [5]: t.sort('b')
<ipython-input-5-e7e0172e4efe> in <module>()
----> 1 t.sort('b')

~/lib/astropy/astropy/table/table.py in sort(self, keys)
   2462
   2463         for col in self.columns.values():
-> 2464             col[:] = col.take(indexes, axis=0)
   2465
   2466         if sort_index is not None:

~/lib/astropy/astropy/table/column.py in __setitem__(self, index, value)
    935
    936         # update indices
--> 937         self.info.adjust_indices(index, value, len(self))
    938
    939         # Set items using a view of the underlying data, as it gives an

~/lib/astropy/astropy/utils/data_info.py in adjust_indices(self, index, value, col_len)
    502         for key, val in zip(keys, value):
    503             for col_index in self.indices:
--> 504                 col_index.replace(key, self.name, val)
    505
    506     def slice_indices(self, col_slice, item, col_len):

~/lib/astropy/astropy/table/index.py in replace(self, row, col, val)
    537     def replace(self, row, col, val):
    538         if not self._frozen:
--> 539             self.index.replace(self.orig_coords(row), col, val)
    540
    541     def copy(self):

~/lib/astropy/astropy/table/index.py in replace(self, row, col_name, val)
    336         key = [c[row] for c in self.columns]
    337         key[self.col_position(col_name)] = val
--> 338         self.data.add(tuple(key), row)
    339
    340     def replace_rows(self, col_slice):

~/lib/astropy/astropy/table/sorted_array.py in add(self, key, row)
     73                              'unique index'.format(key))
     74         self.data.insert_row(pos, key)
---> 75         self.row_index = self.row_index.insert(pos, row)
     76
     77     def _get_key_slice(self, i, begin, end):

AttributeError: 'numpy.ndarray' object has no attribute 'insert'
```
After playing a bit to reproduce, it seems related to sorting a column after sorting the column with an index.
cc @taldcroft

### LINKED ISSUE #9473: TimeSeries sorting fails when sorting by time
When sorting a TimeSeries by the time column after having sorted by another column, the sorting fails:

```
In [15]: ts = TimeSeries(time=['2016-01-01', '2018-01-01', '2017-01-01'])                   

In [16]: ts['flux'] = [3, 2, 1]                                                             

In [17]: ts                                                                                 
Out[17]: 
<TimeSeries length=3>
          time           flux
         object         int64
----------------------- -----
2016-01-01 00:00:00.000     3
2018-01-01 00:00:00.000     2
2017-01-01 00:00:00.000     1

In [18]: ts.sort('flux')                                                                    

In [19]: ts                                                                                 
Out[19]: 
<TimeSeries length=3>
          time           flux
         object         int64
----------------------- -----
2017-01-01 00:00:00.000     1
2018-01-01 00:00:00.000     2
2016-01-01 00:00:00.000     3

In [20]: ts.sort('time')                                                                    

In [21]: ts                                                                                 
Out[21]: 
<TimeSeries length=3>
          time           flux
         object         int64
----------------------- -----
2017-01-01 00:00:00.000     1
2016-01-01 00:00:00.000     3
2018-01-01 00:00:00.000     2
```

This may be related to #6545 but I haven't had a chance to investigate.

taldcroft:
### Description

This pull request is to address various bugs / problems when sorting.  In theory these could be three separate PRs, but they kind of go together and should all be backported to LTS.

#### Sorting indexed table on the indexed column after first sorting on another column

As shown in #6545 and #9473.

In looking at the code I never was able to understand the root cause of the problem because in all honesty I have very little understanding of all the indexing machinery.  But just grasping at straws I tried replacing the "by-hand" freezing of the index with the documented context manager, et voila this did the trick.

I checked performance for sorting a large (1e7) table of random numbers, both with and without an index, and this fix did not impact the timing relative to 4.0.

#### Argsort with reverse=True on a table with an index

This is just an obvious mistake ignoring the `reverse` arg in this case.

#### Sorting a table on one column

The most common table sort was slow by a factor of 5-10 because `np.argsort` on a structured array is 5-10 slower than on an ndarray, even if the structured array has only one column.

Fixes #9473 
Fixes #6545

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -2875,7 +2875,8 @@ class Table:
         if keys is not None:
             index = get_index(self, names=keys)
             if index is not None:
-                return index.sorted_data()
+                idx = np.asarray(index.sorted_data())
+                return idx[::-1] if reverse else idx
 
         kwargs = {}
         if keys:
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -2880,21 +2880,25 @@ class Table:
 
         kwargs = {}
         if keys:
-            kwargs['order'] = keys
+            # For multiple keys return a structured array which gets sorted,
+            # while for a single key return a single ndarray.  Sorting a
+            # one-column structured array is much slower than ndarray, e.g. a
+            # factor of ~6 for a 10 million long random array.
+            if len(keys) > 1:
+                kwargs['order'] = keys
+                data = self.as_array(names=keys)
+            else:
+                data = self[keys[0]].view(np.ndarray)
+        else:
+            # No keys provided so sort on all columns.
+            data = self.as_array()
+
         if kind:
             kwargs['kind'] = kind
 
-        if keys:
-            data = self.as_array(names=keys)
-        else:
-            data = self.as_array()
-
         idx = data.argsort(**kwargs)
 
-        if reverse:
-            return idx[::-1]
-
-        return idx
+        return idx[::-1] if reverse else idx
 
     def sort(self, keys=None, reverse=False):
         '''
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -2956,34 +2956,22 @@ class Table:
         if reverse:
             indexes = indexes[::-1]
 
-        sort_index = get_index(self, names=keys)
+        with self.index_mode('freeze'):
+            for name, col in self.columns.items():
+                # Make a new sorted column.  This requires that take() also copies
+                # relevant info attributes for mixin columns.
+                new_col = col.take(indexes, axis=0)
 
-        if sort_index is not None:
-            # avoid inefficient relabelling of sorted index
-            prev_frozen = sort_index._frozen
-            sort_index._frozen = True
-
-        for name, col in self.columns.items():
-            # Make a new sorted column.  This requires that take() also copies
-            # relevant info attributes for mixin columns.
-            new_col = col.take(indexes, axis=0)
-
-            # First statement in try: will succeed if the column supports an in-place
-            # update, and matches the legacy behavior of astropy Table.  However,
-            # some mixin classes may not support this, so in that case just drop
-            # in the entire new column. See #9553 and #9536 for discussion.
-            try:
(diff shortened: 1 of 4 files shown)


taldcroft:
I fixed two more problems I noticed, one a similar bug and one a performance "bug".

astropy/table/table.py
```diff
@@ -2875,25 +2875,30 @@ def argsort(self, keys=None, kind=None, reverse=False):
         if keys is not None:
             index = get_index(self, names=keys)
             if index is not None:
-                return index.sorted_data()
+                idx = np.asarray(index.sorted_data())
```
taldcroft, line 2878: There was also API inconsistency previously, since `index.sorted_data()` can return a `Column` or `list`, in contrast to the non-indexed case which is an `ndarray`.

taldcroft:
Oops, I forgot to update the changelog, will do that once CI gets through.

astropy/table/table.py
```diff
@@ -2956,34 +2961,22 @@ def sort(self, keys=None, reverse=False):
...
-        if sort_index is not None:
-            # avoid inefficient relabelling of sorted index
-            prev_frozen = sort_index._frozen
-            sort_index._frozen = True
-
-        for name, col in self.columns.items():
-            # Make a new sorted column.  This requires that take() also copies
-            # relevant info attributes for mixin columns.
-            new_col = col.take(indexes, axis=0)
+        with self.index_mode('freeze'):
```
mhvk, line 2964: Lovely shortening!

mhvk:
But do remember completing the changelog entry!

PR #10113, opened 2020-04-04
Correct treatment of size-1 array times and coordinates.

mhvk:
Found this while working with PINT, which does a lot with object `ndarray` of times (since their locations and scales can differ), which at times get turned into size-1 arrays. This became a problem because current master special-cases scalars by simply looking whether an attribute had `size == 1`, but this means that indexing an size-1 array item might leave an attribute with higher dimensions (but still size 1, so many things just work, but not our new `TimeAstropyTime`):
```
from astropy.time import Time
t = Time(['J2010', 'J2011'], location=([10., 20.], [10., 10.]))[:1]                           
t                                                                                             
# <Time object: scale='tt' format='jyear_str' value=['J2010.000']>
t.location                                                                                    
# <EarthLocation [(6186437.06603022, 1090835.76919604, 1100248.54773536)] m>
t[0]                                                                                          
# <Time object: scale='tt' format='jyear_str' value=J2010.000>
t[0].location  # should not give shape=(1,) result                                                                       
# <EarthLocation [(6186437.06603022, 1090835.76919604, 1100248.54773536)] m>
Time([t[0]])
# ... exception - TimeAstropyTime cannot handle it.
```
This PR removes the mistake (did I do it for speed? I cannot remember), ensuring that shapes remain consistent.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -340,6 +340,10 @@ astropy.convolution
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
 
+- Ensure that for size-1 array ``SkyCoord`` and coordinate frames
+  the attributes also properly become scalars when indexed with 0.
+  [#10113]
+
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 
@@ -395,6 +399,10 @@ astropy.tests
 
 astropy.time
 ^^^^^^^^^^^^
+
+- Ensure that for size-1 array ``Time``, the location also properly becomes
+  a scalars when indexed with 0. [#10113]
+
 
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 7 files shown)


CHANGES.rst
```diff
@@ -396,6 +400,10 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- Ensure that for size-1 array ``Time``, the location also properly becomes
+  a scalars when indexed with 0. [#10113]
```
taldcroft, line 404: Copy typo - `a scalars when`

mhvk:
Yes, I cannot really believe I made that mistake myself. Though in fairness it almost never matters, as broadcasting takes care of it. But even by "look", the current version is clearly better...

Anyway, thanks for the quick review; merging now.

PR #10123, opened 2020-04-07
Allow more unit types for pixel_scale

### LINKED ISSUE #10110: Add a new unit equivalency to handle pixel scale specified with length unit.
I came across this situation that I need to build a transformation to describe the projection of a detector array specified in `u.um` to some fiducial plane which is used in lab testing the array.

We construct maps from such tests and would like to build a GWCS object as the WCS of these maps.

In this case, the WCS is an AffineTransform2D that takes inputs unit 'u.um' to 'u.pix', which requires pixel_scales specified as u.m / u.pix as the input_units_equivalencies.

I was wondering if this makes sense and I am glad to contribute a initial pull request. Mostly it gonna be a duplication of u.pixel_scale (say, u.pixel_scale_phys), but just replace u.deg there with u.m. Or it makes more sense to extend the current pixel_scale equivalency to cover this situation.

Jerry-Ma:
### Description

This pull request is to address the need of pixel equivalencies on images where pixels are not related to distances on the sky.

This proposed implementation does not assume an equivalency of the `pixscale` unit to `u.rad/u.pix` (or `u.pix/u.rad`), instead, it constructs the correct scaling relation by decomposing the unit and inspect the dimensionality of `u.pix` in the unit of `pixscale`.
  

Fixes #10110

### DIFF
--- a/astropy/units/equivalencies.py
+++ b/astropy/units/equivalencies.py
@@ -744,24 +744,29 @@ def assert_is_spectral_unit(value):
 
 def pixel_scale(pixscale):
     """
-    Convert between pixel distances (in units of ``pix``) and angular units,
+    Convert between pixel distances (in units of ``pix``) and other units,
     given a particular ``pixscale``.
 
     Parameters
     ----------
     pixscale : `~astropy.units.Quantity`
-        The pixel scale either in units of angle/pixel or pixel/angle.
+        The pixel scale either in units of <unit>/pixel or pixel/<unit>.
     """
-    if pixscale.unit.is_equivalent(si.arcsec/astrophys.pix):
-        pixscale_val = pixscale.to_value(si.radian/astrophys.pix)
-    elif pixscale.unit.is_equivalent(astrophys.pix/si.arcsec):
-        pixscale_val = (1/pixscale).to_value(si.radian/astrophys.pix)
-    else:
-        raise UnitsError("The pixel scale must be in angle/pixel or "
-                         "pixel/angle")
 
-    return Equivalency([(astrophys.pix, si.radian,
-                         lambda px: px*pixscale_val, lambda rad: rad/pixscale_val)],
+    decomposed = pixscale.unit.decompose()
+    dimensions = dict(zip(decomposed.bases, decomposed.powers))
+    pix_power = dimensions.get(astrophys.pix, 0)
+
+    if pix_power == -1:
+        physical_unit = Unit(pixscale * astrophys.pix)
+    elif pix_power == 1:
+        physical_unit = Unit(astrophys.pix / pixscale)
+    else:
+        raise UnitsError(
+                "The pixel scale unit must have"
+                " pixel dimensionality of 1 or -1.")
+
+    return Equivalency([(astrophys.pix, physical_unit)],
                        "pixel_scale", {'pixscale': pixscale})
 
 
--- a/astropy/units/tests/test_equivalencies.py
+++ b/astropy/units/tests/test_equivalencies.py
@@ -791,6 +791,35 @@ def test_pixel_scale():
     assert_quantity_allclose(asec.to(u.pix, u.pixel_scale(pixscale2)), pix)
 
 
+def test_pixel_scale_invalid_scale_unit():
+
+    pixscale = 0.4 * u.arcsec
+    pixscale2 = 0.4 * u.arcsec / u.pix ** 2
+
+    with pytest.raises(u.UnitsError, match="pixel dimension"):
+        u.pixel_scale(pixscale)
+    with pytest.raises(u.UnitsError, match="pixel dimension"):
+        u.pixel_scale(pixscale2)
+
+
+def test_pixel_scale_acceptable_scale_unit():
+
+    pix = 75 * u.pix
+    v = 3000 * (u.cm / u.s)
+
+    pixscale = 0.4 * (u.m / u.s / u.pix)
+    pixscale2 = 2.5 * (u.pix / (u.m / u.s))
+
+    assert_quantity_allclose(pix.to(u.m / u.s, u.pixel_scale(pixscale)), v)
+    assert_quantity_allclose(pix.to(u.km / u.s, u.pixel_scale(pixscale)), v)
+
+    assert_quantity_allclose(pix.to(u.m / u.s, u.pixel_scale(pixscale2)), v)
+    assert_quantity_allclose(pix.to(u.km / u.s, u.pixel_scale(pixscale2)), v)
+
+    assert_quantity_allclose(v.to(u.pix, u.pixel_scale(pixscale)), pix)
+    assert_quantity_allclose(v.to(u.pix, u.pixel_scale(pixscale2)), pix)
+
+
 def test_plate_scale():
     mm = 1.5*u.mm
     asec = 30*u.arcsec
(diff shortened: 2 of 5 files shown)


astropy/units/tests/test_equivalencies.py
```diff
@@ -791,6 +791,32 @@ def test_pixel_scale():
...
+        u.pixel_scale(pixscale)
+
+
+def test_pixel_scale_acceptable_scale_unit():
+
+    pix = 75 * u.pix
+    v = 3000 * u.cm / u.s
+
+    pixscale = 0.4 * u.m / u.s / u.pix
+    pixscale2 = 2.5 * u.pix / (u.m / u.s)
```
pllim, line 808:
```suggestion
    pixscale = 0.4 * (u.m / u.s / u.pix)
    pixscale2 = 2.5 * (u.pix / (u.m / u.s))
```

nden:
@Jerry-Ma Could you add a test that it serializes to asdf?
Adding an example to the list here will do it. Thanks.

https://github.com/astropy/astropy/blob/6191daaab9df363c39c1e61acac4955e5ff5e64c/astropy/io/misc/asdf/tags/unit/tests/test_equivalency.py#L14

Jerry-Ma:
@nden Just added the Asdf round trip test.

However, there is a can of worm that I opted not to open by choosing the test case, because I think the issue is bigger than the scope of this pull request.

In short, the Asdf round trip test would fail on a test case like this: `pixel_scale(100. * u.pix / u.imperial.inch)`.

The reason is that Asdf serialize the argument in an alternated form for some units. In this case, the scaled form of `u.imperial.inch` is used. This results in the serialized file as follows:

```
#ASDF 1.0.0
#ASDF_STANDARD 1.4.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
  name: asdf, version: 2.5.2}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension.BuiltinExtension
    software: {name: asdf, version: 2.5.2}
  - !core/extension_metadata-1.0.0
    extension_class: astropy.io.misc.asdf.extension.AstropyAsdfExtension
    software: {name: astropy, version: 4.1.dev917+g8267d16fa.d20200407}
  - !core/extension_metadata-1.0.0
    extension_class: astropy.io.misc.asdf.extension.AstropyExtension
    software: {name: astropy, version: 4.1.dev917+g8267d16fa.d20200407}
eq: !<tag:astropy.org:astropy/units/equivalency-1.0.0>
- kwargs_names: [pixscale]
  kwargs_values:
  - !unit/quantity-1.1.0 {unit: !unit/unit-1.0.0 0.39370079cm-1 pixel, value: 100.0}
  name: pixel_scale
inch: !unit/unit-1.0.0 2.54cm
per_inch: !unit/unit-1.0.0 0.39370079cm-1
...
```

Then, it is easy to see that when this file is read and the new pixel_scale is constructed, the kwargs will be different due to the round tripping of float pointing numbers. In this particular case, we have:

```
>>> dpi_rt.kwargs[0]  # the round tripped value
{'pixscale': <Quantity 100. 0.393701 pix / cm>}
>>> dpi.kwargs[0]  # the input value
{'pixscale': <Quantity 100. pix / inch>}
```

This will fail the assertion of equality because the ``__eq__`` of Equivalency class not setup to handle float comparison.

astropy/units/equivalencies.py
```diff
@@ -744,24 +744,32 @@ def assert_is_spectral_unit(value):
...
+    elif pix_power == 1:
+        value_per_pix = 1. / _scale
+    else:
+        raise UnitsError(
+                "The pixel scale unit must have"
+                " pixel dimensionality of 1 or -1.")
+    u_phys = value_per_pix.unit * astrophys.pix
+
+    return Equivalency([(astrophys.pix, u_phys,
+                         lambda px: px * value_per_pix,
```
mhvk, line 771: Multiplication and division are the defaults for equivalencies, so in principle there should be no need to define these `lambda` expressions. However, that relies on getting the replacement unit - here `u_phys` - right. The way to do that is to ensure `u_phys` also contains the value. To get that, one can replace as follows (where I also adjusted according to the comment above):
```
    pixscale_unit = Unit(pixscale)
    decomposed = pixscale_unit.decompose()
    dimensions = dict(zip(decomposed.bases, decomposed.powers))
    pix_power = dimensions.get(astrophys.pix, 0)
    if pix_power == -1:
        physical_unit = pixscale_unit * astrophys.pix
    elif pix_power == 1:
        physical_unit = astrophys.pix / pixscale_unit
    else:
        raise UnitsError(
                "The pixel scale unit must have"
                " pixel dimensionality of 1 or -1.")

    return Equivalency([(astrophys.pix, physical_unit)],
                       "pixel_scale", {'pixscale': pixscale})
```

astropy/units/tests/test_equivalencies.py
```diff
@@ -791,6 +791,32 @@ def test_pixel_scale():
     assert_quantity_allclose(asec.to(u.pix, u.pixel_scale(pixscale2)), pix)
 
 
+def test_pixel_scale_invalid_scale_unit():
+
+    pixscale = 0.4 * u.arcsec
+
+    with pytest.raises(Exception):
```
mhvk, line 798: Maybe be a bit more precise in catching the exception: `pytest.raises(u.UnitsError, match="pixel dimension")`

I'd also add one with `u.pix**2` or so.

docs/units/equivalencies.rst
```diff
@@ -354,6 +354,17 @@ and you want to know how big your pixels need to be to cover half an arcsecond::
     >>> (0.5*u.arcsec).to(u.micron, tel_platescale)  # doctest: +FLOAT_CMP
     <Quantity 18.9077335632719 micron>
 
+The pixel scale equivalency can also work in more general context, where the
+scale is specified as any quantity that is reducible to ``<composite
+unit>/u.pix`` or ``u.pix/<composite unit>`` (that is, the dementionality of
```
mhvk, line 359: typo: dimensionality

astropy/units/equivalencies.py
Jerry-Ma: Thanks for the suggestion. I made the changes accordingly, and the code now become quite neat. I did some reordering of the variable defs to make it more readable.

astropy/units/tests/test_equivalencies.py
Jerry-Ma: Good suggestion. Added a new case with unexpected u.pix dimensionality.

docs/units/equivalencies.rst
```diff
@@ -354,6 +354,17 @@ and you want to know how big your pixels need to be to cover half an arcsecond::
     >>> (0.5*u.arcsec).to(u.micron, tel_platescale)  # doctest: +FLOAT_CMP
     <Quantity 18.9077335632719 micron>
 
+The pixel scale equivalency can also work in more general context, where the
+scale is specified as any quantity that is reducible to ``<composite
+unit>/u.pix`` or ``u.pix/<composite unit>`` (that is, the dimentionality of
```
mhvk, line 359: Still a small typo: `dimensionality`!

Jerry-Ma:
@mhvk Did the squash and rebase, forgot to pin you.

PR #10126, opened 2020-04-08
Handle auto_max_age = None consistently

weaverba137:
### Description

This PR fixes a "bare" use of `None` in a numerical computation and ensures
that `None` is converted to a valid float throughout the module.  See also #9479.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -117,8 +117,6 @@ astropy.units
 astropy.utils
 ^^^^^^^^^^^^^
 
-- Fix case where ``None`` could be used in a numerical computation. [#10126]
-
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
 
@@ -426,6 +424,8 @@ astropy.utils
 ^^^^^^^^^^^^^
 
 - Fixed memory allocation on 64-bit systems within ``xml.iterparse`` [#10076]
+
+- Fix case where ``None`` could be used in a numerical computation. [#10126]
 
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -117,6 +117,8 @@ astropy.units
 astropy.utils
 ^^^^^^^^^^^^^
 
+- Fix case where ``None`` could be used in a numerical computation. [#10126]
```
pllim, line 120: The entry might be moved depending on the milestone set. FYI.

weaverba137: No problem!

astropy/utils/iers/tests/test_leap_second.py
```diff
@@ -109,6 +109,11 @@ def test_auto_open_simple(self):
         ls = iers.LeapSeconds.auto_open([iers.IERS_LEAP_SECOND_FILE])
         assert ls.meta['data_url'] == iers.IERS_LEAP_SECOND_FILE
 
+    def test_auto_open_max_age_none(self):
+        with iers.conf.auto_max_age.set_temp(None):
```
mhvk, line 113: I'm not sure this test actually tests what you want. I think you'd want to create a file that is for sure out of date, and pass both that file and the real file in, i.e., `auto_open([outdated_file, iers.IERS_LEAP_SECOND_FILE])`, and then check that it loads `outdated_file`. It may be easiest to copy things from `test_fake_expired_file` (or even just to tag it on there).

weaverba137: Good point.  I was only checking to make sure an exception was not thrown.  I'll take a look.

weaverba137: See latest commit.

weaverba137:
I moved the change log entry to 4.0.2.  If that looks good, I can rebase/squash.

PR #10129, opened 2020-04-09
Fixed error in the obliquity for *TrueEcliptic coordinate frames

ayshih:
The *TrueEcliptic coordinate frames (added in #8394) have a calculation error in their transformations, specifically in the function `_true_ecliptic_rotation_matrix()`.  The function correctly makes use of the transformation matrix from ICRF to "true" equatorial (obtained using `erfa.pnm06a()`), but then rotates it by the "mean" obliquity of the ecliptic (obtained using `erfa.obl06()`) instead of the "true" obliquity of the ecliptic.  In other words, it doesn't account for the effect of nutation on the obliquity of the ecliptic.  This PR modifies the function to use the true obliquity of the ecliptic.

Here's an illustrative before/after example.  Before this PR, a point on the ecliptic plane in `GeocentricMeanEcliptic` (latitude of 0 deg) ends up ~6 arcsec off of the ecliptic plane in `GeocentricTrueEcliptic`.  However, while nutation should change the apparent ecliptic longitude (because the location of the apparent vernal equinox changes), it shouldn't affect ecliptic latitude; objects on the ecliptic plane should stay on the ecliptic plane, regardless of what the Earth is doing.
```python
>>> import astropy.units as u
>>> s = SkyCoord(90*u.deg, 0*u.deg, 1*u.AU, frame='geocentricmeanecliptic')
>>> s.geocentrictrueecliptic
<SkyCoord (GeocentricTrueEcliptic: equinox=J2000.000, obstime=J2000.000): (lon, lat, distance) in (deg, deg, AU)
    (89.99613, -0.00160261, 1.)>
>>> s.geocentrictrueecliptic.lat.to('arcsec')
<Latitude -5.76939806 arcsec>
```
After this PR, the ecliptic latitude does not change (and the ecliptic longitude still does change):
```python
>>> import astropy.units as u
>>> s = SkyCoord(90*u.deg, 0*u.deg, 1*u.AU, frame='geocentricmeanecliptic')
>>> s.geocentrictrueecliptic
<SkyCoord (GeocentricTrueEcliptic: equinox=J2000.000, obstime=J2000.000): (lon, lat, distance) in (deg, deg, AU)
    (89.99613, 6.36110936e-15, 1.)>
>>> s.geocentrictrueecliptic.lat.to('arcsec')
<Latitude 2.28999937e-11 arcsec>
```
I've added a test for this preservation of ecliptic latitude.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -432,6 +432,10 @@ astropy.coordinates
 - Ensure that the ``lon`` values in ``SkyOffsetFrame`` are wrapped correctly at
   180 degree regardless of how the underlying data is represented. [#10163]
 
+- Fixed an error in the obliquity of the ecliptic when transforming to/from the
+  ``*TrueEcliptic`` coordinate frames. The error would primarily result in an
+  inaccuracy in the ecliptic latitude on the order of arcseconds. [#10129]
+
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 
--- a/astropy/coordinates/builtin_frames/ecliptic_transforms.py
+++ b/astropy/coordinates/builtin_frames/ecliptic_transforms.py
@@ -43,7 +43,8 @@ def _true_ecliptic_rotation_matrix(equinox):
     # (see https://github.com/astropy/astropy/pull/6508).
     jd1, jd2 = get_jd12(equinox, 'tt')
     rnpb = erfa.pnm06a(jd1, jd2)
-    obl = erfa.obl06(jd1, jd2)*u.radian
+    _, nut_obl = erfa.nut06a(jd1, jd2)*u.radian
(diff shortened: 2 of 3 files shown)


ayshih:
I'll point out that I discovered this when I was refactoring SunPy code to make use of `GeocentricTrueEcliptic` and it broke our tests against published values in _The Astronomical Almanac_.

pllim:
Thank you! This needs a change log.

ayshih:
Changelog entry added

astrofrog:
@ayshih - can you rebase to resolve the conflict? Also it'd be good to get this in soon. Maybe @StuartLittlefair would be able to review this?

ayshih:
@astrofrog Sure, now rebased

adrn:
Thanks @ayshih!

astrojuanlu:
I'm late, but this looks perfect - good catch @ayshih!

Issue #10070, opened 2020-03-24
LombScargle occasionally returns all-NaN powers


megbedell:
### Description
I want to use LombScargle to calculate the spectral window function of my observations, but sometimes I get an array of NaNs returned for powers. I'm not sure what governs the conditions for this error, but so far I've only encountered it when the input `y` grid is populated with a single value.

### Expected behavior
Input an array `t` and a same-sized array `y` where all y-values are 1, get back some powers corresponding to the window function.

### Actual behavior
Sometimes this works as expected. Other times, the output powers are an array of NaNs and the following warning is raised:

> /Users/mbedell/miniconda3/envs/exoplanet/lib/python3.8/site-packages/astropy/timeseries/periodograms/lombscargle/implementations/fast_impl.py:125: RuntimeWarning: invalid value encountered in true_divide
>  power /= YY

### Steps to Reproduce
Here is a snippet of code that reproduces the error:

```python
from astropy.timeseries import LombScargle
np.random.seed(42)
N = 180
xs = np.arange(N) + np.random.normal(0., 0.1, N)
ys = np.ones_like(xs)
frequency, power = LombScargle(xs, ys).autopower(method='fast')
np.isfinite(power).any()
```
(should raise a warning and return `False`)

Changes to the above code that do not fix the error:
- changing autopower method (I have tried 'slow' and 'chi2' with same results)
- changing the random seed/rerunning the randomization
- specifying a frequency grid as an input
- making the `xs` grid perfectly uniformly spaced (no random perturbations)

Changes to the above code that *do* fix the error and return sensible powers:
- changing `N` to nearly any other integer (I haven't tried an exhaustive list but for example 179 or 181 work just fine, as do other random ones I've tried. However, `N = 180` is definitely not the only number that can reproduce this bug - `N = 256` behaves similarly.)
- scaling `ys` by some factor - this fixes the error if and only if the factor is not a multiple of 180, as far as I can tell. That is, `ys = np.ones_like(xs)/100` fixes the error, but `ys = np.ones_like(xs)/180` does not. Again, I haven't tried factors exhaustively, so take that theory with a grain of salt.
- making the values of `ys` different from each other, e.g.:
`ys = np.ones_like(xs) * np.random.normal(1., 0.01, N)`

### System Details

macOS-10.15.3-x86_64-i386-64bit
Python 3.8.2 | packaged by conda-forge | (default, Mar 23 2020, 17:55:48) 
[Clang 9.0.1 ]
[...]

PR #10134, opened 2020-04-10
Intercept all-zero input signals to LombScargle.power() to catch NaNs

dhomeier:
### Description
This pull request presents some suggestions to improve `LombScargle` output when the signal (after centring to the mean, if selected) is zero everywhere, and avoid returning a power spectrum of `NaN`s after normalisation.

Addresses #10070, where a constant (but nonzero) input to `autopower` inconsistently returned all-`NaN` `power` arrays.

1. The default settings of `fit_mean=True`, `center_data=True`, are centring this input to exactly or almost exactly 0.0 before proceeding to computing `p`.

2. The default `normalization='standard'` then returns, depending on the exact result of the centring, either `NaN` powers or a spectrum of numerical noise (which might however not be immediately recognisable as such).

The proposed fix checks for such inputs and directly returns the exact resulting zero spectrum without calling the `method` implementation (the `fast` methods might still produce a noise spectrum on zero input). Thus no further normalising is applied, but a warning raised if a `normalization` method was selected that would produce NaNs.

### Questions

1. Should the check rather be moved separately into the different implementations and after the actual centring was done (ensuring that the exact input that would have been used in the method is tested)? That would let some inputs pass that are mathematically, but not numerically zero.

2. Are the warnings on unsuitable `normalization` methods appropriate, should they provide more information (e.g. recommend `'psd'` as an alternative), return `NaN` after the warning or raise an exception instead?

### DIFF
--- a/astropy/timeseries/periodograms/lombscargle/implementations/main.py
+++ b/astropy/timeseries/periodograms/lombscargle/implementations/main.py
@@ -7,7 +7,11 @@ statement for the various implementations available in this submodule
 
 __all__ = ['lombscargle', 'available_methods']
 
+import warnings
+
 import numpy as np
+
+from astropy.utils.exceptions import AstropyWarning
 
 from .slow_impl import lombscargle_slow
 from .fast_impl import lombscargle_fast
@@ -154,7 +158,7 @@ def lombscargle(t, y, dy=None,
         or 'fast'.
     normalization : str, optional
         Normalization to use for the periodogram.
-        Options are 'standard' or 'psd'.
+        Options are 'standard', 'model', 'log' or 'psd'.
     fit_mean : bool, optional
         if True, include a constant offset as part of the model at each
         frequency. This can lead to more accurate results, especially in the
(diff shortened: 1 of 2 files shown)


pllim:
Thanks! Does this completely fix #10070?

dhomeier:
> Thanks! Does this completely fix #10070?

Depends on what a complete fix is.

1. It avoids inconsistent output depending on the exact scaling of the input.

2. It does not return `NaN` (in theory; there is no check before the actual normalisation when the denominator `YY` with floating-point inaccuracy _could_ perhaps become 0 even if `y` is not 0 everywhere. Also there might be cases where the `YY-p` term in the `'model'` normalisation still becomes 0...)

3. Does it give the user enough information how to set up their calculation to avoid the centring to 0?

The original use case to calculate the windowing function with a constant input seems legitimate enough, and obviously it is not evident what is happening internally with the default settings. I guess a zero spectrum should make it clear enough that it is not the desired result, and from there the correct settings can in principle be found in the docs. But perhaps it should be pointed out clearly somewhere that one needs both `fit_mean=False, center_data=False` to avoid the shifting.

PR #10140, opened 2020-04-13
TST: Jail skyfield tests in a temporary directory

### LINKED ISSUE #10133: TST: Still some leftover doctest outputs left
A reminder to investigate https://github.com/astropy/astropy/pull/9439#issuecomment-611770982 . These files not cleaned up after a remote data tests on Windows.

```
        Leap_Second.dat
        de421.bsp
        deltat.data
        deltat.preds
```

pllim:
### Description

This pull request is to address creation of untracked output files after running `pytest astropy/coordinates --remote-data` (I ran it on Windows but I don't see why it is OS specific).

Fixes #10133 

p.s. The issue reported at https://github.com/astropy/astropy/issues/10133#issuecomment-613099443 is out of scope for this PR.

p.p.s. I welcome a less hacky solution too.

### DIFF
--- a/astropy/coordinates/tests/test_solar_system.py
+++ b/astropy/coordinates/tests/test_solar_system.py
@@ -1,4 +1,6 @@
 import os
+import shutil
+import tempfile
 
 import pytest
 import numpy as np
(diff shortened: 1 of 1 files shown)


mhvk:
@pllim - since the files are due to `skyfield`, I think it is quite a neat rather than hacky way to just jail them. My only suggestion would be to define this using `setup_function` and `teardown_function` for `test_positions_skyfield` (or perhaps make `test_postitions_skyfield` a class and use `setup_class` and `teardown_class`), so that it remains clear that `skyfield` is the problem. (Alternatively, even easier, just define them as an `else` in the `HAS_SKYFIELD` determination.)

mhvk:
p.s. I confirm that your pytest invocation downloads all four files, that it is specific to `test_solar_system.py` and that if I remove `skyfield`, none of the four are produce (not even `Leap_Second.dat`).

mhvk:
Hmm, looking at the skyfield docs at https://rhodesmill.org/skyfield/api-iokit.html#skyfield.iokit.Loader it does seem one should be able to do
```
...
from skyfield.api import Loader
...
def test_positions_skyfield(tmpdir):
    load = Loader(tmpdir)
```
with no further changes...

...

OK, that actually works - see #10142

jdavies-st:
Yeah, #10142 looks to be the more elegant solution that uses existing pytest `tmpdir` infrastructure.  👍

pllim:
Thanks! I did this with kids screaming in the house, so I didn't try very hard. I'll check out #10142.

mhvk:
I liked the jailing...

PR #10142, opened 2020-04-14
Let skyfield use a temporary directory for files it downloads.

### LINKED ISSUE #10133: TST: Still some leftover doctest outputs left

mhvk:
Alternative to #10140. fixes #10133.


eteq:
Oh, but if implementing the test I suggested above is going to be a huge pain given that it's sort of deep in the testing machinery, I think it'd be ok to just get a comment here from @pllim or @jdavies-st that this fixes the underlying problem...

pllim:
>  ensure the files are not left in

@eteq , I think that is out of scope for this PR. If it were to be done, it would be for the whole repo. Maybe like a `git status` check after a remote data job with all dependencies installed on CI or something like that.

eteq:
Point taken @pllim.  Since you confirmed it's working I'll merge this as it stands.  If it re-surfaces we should think more about how to test it properly, but hopefully that won't happen.

mhvk:
Thanks, @pllim, for IDing the culprit and figuring out a first solution!

PR #10145, opened 2020-04-14
fix issues with domain and window in polynomial models

### LINKED ISSUE #9992: Minor issues related to polynomial domain & window
Following PRs #9906, #9941, #9956 to fix issues #9942, #9795 & #9809 (not a one-to-one correspondence), a few minor outstanding issues have been left for later. I think the following are still pending , if I have kept track of them correctly:

- Polynomial models don't do any checking of their `domain` & `window` (or `x_domain` / `y_domain` / `x_window` / `y_window`) arguments, so those can be set to whatever value you like, causing errors when you later try to use the model. In particular, if the user provides a tuple, it works as expected but then gets serialized to `asdf` as a list instead, so if a model is saved and reloaded, its `domain` and `window` are not identical before and afterwards (eg. in a test comparison `[-2, 2] != (-2, 2)`). The tentative consensus was that the `domain` & `window` arguments should be checked and converted to 2-tuples in the corresponding attributes (though the current documentation defines their values as lists).

- In <a href="https://github.com/astropy/astropy/blob/7807e33bc1ba1eb65d0557b571efbb1608295688/astropy/modeling/core.py#L2296">`Model._format_repr`</a>, an optional model argument/attribute gets included in the printed output only if its value is not the default (as provided by the calling `__repr__` implementation). In <a href="https://github.com/astropy/astropy/blob/7807e33bc1ba1eb65d0557b571efbb1608295688/astropy/modeling/core.py#L2324">`Model._format_str`</a>, however, the same argument/attribute gets printed whenever the value is `not None`, which is not correct the default eg. for `window`. Thus (once #9941 is merged), `window` will **always** get printed by `str`, unlike `domain` and also unlike `window` in `repr`. Presumably this is not the intended behaviour and `_format_str` should have a `defaults` parameter like the one in `_format_repr`, which should then get passed by `PolynomialModel.__str__` and `OrthoPolynomialBase.__str__` () (or the equivalent methods of their child classes).

- Similar to the previous point, in <a href="https://github.com/astropy/astropy/blob/7807e33bc1ba1eb65d0557b571efbb1608295688/astropy/io/misc/asdf/tags/transform/polynomial.py#L168">the asdf tag code for `PolynomialTypeBase`</a>, the `window` will **always** get serialized given the condition `not None` -- unlike the `domain` -- because the value is never None, defaulting to `[-1, 1]`. This should be harmless but is obviously not the intended behaviour. Likewise for `x_window` / `y_window`. For `Polynomial1D`/`Polynomial2D`, it looks like the `domain` will always get serialized as well.

These might be good problems for a new contributor, eg. at PIA?

nden:
Fixes #9992

Summary of changes:

- `domain` and `window` are now tuples of size 2 and are validated
- `__str__` and `__repr__` print the domain and window only if their values are different from the defaults.
- ~Only non-default values of `domain` and `window` are written to file~
  After discussion default values are written to file to ensure reproducibility if defaults change.
- general clean-up

### DIFF
--- a/astropy/io/misc/asdf/tags/transform/polynomial.py
+++ b/astropy/io/misc/asdf/tags/transform/polynomial.py
@@ -118,10 +118,8 @@ class PolynomialTypeBase(TransformType):
                                                  domain=domain, window=window)
             model.parameters = coefficients
         elif n_dim == 2:
-            x_domain = node.get('x_domain', None)
-            y_domain = node.get('y_domain', None)
-            x_window = node.get('x_window', None)
-            y_window = node.get('y_window', None)
+            x_domain, y_domain = tuple(node.get('domain', (None, None)))
+            x_window, y_window = tuple(node.get('window', (None, None)))
             shape = coefficients.shape
             degree = shape[0] - 1
             if shape[0] != shape[1]:
@@ -165,15 +163,15 @@ class PolynomialTypeBase(TransformType):
             # property.  We can't serialize the new domain values with those
             # versions because they don't validate.
             if ndim == 1:
-                if model.domain != model._default_domain_window['domain']:
+                if model.domain is not None:
                     node['domain'] = model.domain
-                if model.window != model._default_domain_window['window']:
+                if model.window is not None:
                     node['window'] = model.window
             else:
-                for attr in ['x_domain', 'y_domain', 'x_window', 'y_window']:
-                    attr_obj = getattr(model, attr)
-                    if attr_obj != model._default_domain_window[attr]:
-                        node[attr] = attr_obj
+                if model.x_domain or model.y_domain is not None:
+                    node['domain'] = (model.x_domain, model.y_domain)
+                if model.x_window or model.y_window is not None:
+                    node['window'] = (model.x_window, model.y_window)
 
         return yamlutil.custom_tree_to_tagged_tree(node, ctx)
 
@@ -240,10 +238,8 @@ class OrthoPolynomialType(TransformType):
                                                       domain=domain, window=window)
             model.parameters = coefficients
         elif n_dim == 2:
-            x_domain = node.get('x_domain', None)
-            y_domain = node.get('y_domain', None)
-            x_window = node.get('x_window', None)
-            y_window = node.get('y_window', None)
+            x_domain, y_domain = tuple(node.get('domain', (None, None)))
+            x_window, y_window = tuple(node.get('window', (None, None)))
             coeffs = {}
             shape = coefficients.shape
             x_degree = shape[0] - 1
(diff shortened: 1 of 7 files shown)


nden:
@jehturner

astropy/io/misc/asdf/tags/transform/polynomial.py
```diff
@@ -274,15 +276,25 @@ def to_tree_transform(cls, model, ctx):
...
+            domain = []
+            window = []
+            if model.x_domain != model._default_domain_window['x_domain']:
+                domain.append(model.x_domain)
+            if model.y_domain != model._default_domain_window['y_domain']:
+                domain.append(model.y_domain)
+            if model.x_window != model._default_domain_window['x_window']:
+                window.append(model.x_window)
+            if model.y_window != model._default_domain_window['y_window']:
+                window.append(model.y_window)
```
perrygreenfield, line 293: Does this do the right thing if only one of the two x or y values passes their test? It looks like if either of the x or y values equals its default, it doesn't get appended, leaving a list with only one value.

nden: Only the one that is non-default is written to file.
DO you think we should write out the default value as well?

perrygreenfield: I may be missing the rest of the context, but it looks like there is no information in domain or window to indicate which coordinate the domain is associated with. If there is only one domain for a 2d case, is it for x or y?

nden: Nope, I missed that. I'll write them out separately.

nden: This is fixed now but let's wait for tests. I noticed, unrelated, that default values of constraints were serialized. Fixed it in this PR.

astropy/modeling/fitting.py
```diff
@@ -1347,7 +1347,10 @@ def _fitter_to_model_params(model, fps):
     if has_tied:
         for idx, name in enumerate(model.param_names):
             if model.tied[name]:
-                value = model.tied[name](model)
+                oldval = getattr(model, name).value
+                print('oldval', oldval)
+                value = model.tied[name](model) #oldval) #model)
+                print('value', value)
```
perrygreenfield, line 1353: Is it intended to leave the print statements in?

astropy/io/misc/asdf/tags/transform/polynomial.py
```diff
@@ -274,15 +278,16 @@ def to_tree_transform(cls, model, ctx):
...
         else:
-            if model.x_domain or model.y_domain is not None:
-                node['domain'] = (model.x_domain, model.y_domain)
-            if model.x_window or model.y_window is not None:
-                node['window'] = (model.x_window, model.y_window)
+            for attr in ['x_domain', 'y_domain', 'x_window', 'y_window']:
+                attr_obj = getattr(model, attr)
+                if attr_obj != model._default_domain_window[attr]:
+                    node[attr] = attr_obj
+
```
perrygreenfield, line 290: Does the change in the serialization attributes need to be reflected in the schema? I forget what was put in previously as far as the schema goes (perhaps ignored).

perrygreenfield: Also, did we release the previous serialization? If so, do we have a backwards compatibility issue?

jehturner:
I see this has already been merged, but it LGTM (and now my serialization test comparison passes whether I specify the domain as a tuple or a list). It's still the case that the `window` always gets serialized (eg. I think [this line](https://github.com/nden/astropy/blob/8767dc05a84b5c7823aa72a78bb8f5549b701309/astropy/io/misc/asdf/tags/transform/polynomial.py#L279) is always `True`), but that seems a small detail compared with the other two things and should be harmless.

Thanks a lot for fixing these details, which I think will help avoid mistakes!

nden:
@jehturner Perry pointed out correctly that for an archival format we should serialize the defaults. If defaults in the future changed this will ensure we reproduce the object with the new defaults.

nden:
And thanks for testing and reviewing :-)

jehturner:
I did briefly wonder about that, but then shouldn't `domain` always get serialized too? I suppose its default of `None` wouldn't add anything to the serialization, meaning whatever the default value is in a given version, as opposed to `window==[-1,1]`, which is explicit, so fair enough.

nden:
Yes, `None` is a special case, it doesn't roundtrip in asdf and is dealt with in the tag. For `Polynomial1D` where the default domain is `(-1, 1)` it is serialized.

PR #10152, opened 2020-04-15
ENH: astropy.io.votable.exceptions.conf.max_warnings

### LINKED ISSUE #7529: Make io.votable.exceptions.MAX_WARNINGS configurable
Currently, `io.votable.exceptions.MAX_WARNINGS` is hardcoded to 10. There might be cases where users wish to increase/decrease that value. Perhaps we should make it a configurable item in `io/votable/__init__.py`; but if that introduces ugly circular imports, it can instead be a configurable item within `exceptions.py` itself.

Whoever wants to tackle this also has to write tests for it (and a change log).

pllim:
### Description

This pull request is to add a new configuration item to replace hardcoded `MAX_WARNINGS` in `astropy.io.votable.exceptions`.

Fixes #7529

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,9 @@ astropy.io.registry
 astropy.io.votable
 ^^^^^^^^^^^^^^^^^^
 
+- New ``exceptions.conf.max_warnings`` configuration item to control the number of times a
+  type of warning appears before being suppressed. [#10152]
+
 astropy.modeling
 ^^^^^^^^^^^^^^^^
 - Added Plummer1D model to ``functional_models``. [#9896]
--- a/astropy/io/votable/exceptions.py
+++ b/astropy/io/votable/exceptions.py
@@ -12,8 +12,8 @@ Warnings
     tools that produced the VOTable file.
 
     To control the warnings emitted, use the standard Python
-    :mod:`warnings` module.  Most of these are of the type
-    `VOTableSpecWarning`.
+    :mod:`warnings` module and the ``astropy.io.votable.exceptions.conf.max_warnings``
+    configuration item.  Most of these are of the type `VOTableSpecWarning`.
 
 {warnings}
 
@@ -31,7 +31,6 @@ Exceptions
 {exceptions}
 """
 
-
 # STDLIB
 import io
 import re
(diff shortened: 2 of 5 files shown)


CHANGES.rst
```diff
@@ -47,6 +47,9 @@ astropy.io.registry
 astropy.io.votable
 ^^^^^^^^^^^^^^^^^^
 
+- New ``conf.max_warnings`` configuration item to control the number of times a
```
pllim, line 50: Maybe this is clearer?
```suggestion
- New ``exceptions.conf.max_warnings`` configuration item to control the number of times a
```

astropy/io/votable/exceptions.py
```diff
@@ -31,25 +31,36 @@
...
 from textwrap import dedent
 from warnings import warn
 
+from astropy import config as _config
 from astropy.utils.exceptions import AstropyWarning
 
-
 __all__ = [
-    'warn_or_raise', 'vo_raise', 'vo_reraise', 'vo_warn',
+    'Conf', 'conf', 'warn_or_raise', 'vo_raise', 'vo_reraise', 'vo_warn',
```
tomdonaldson, line 45: Is the capital C `Conf` needed here?  I didn't notice it being referenced, but I'm probably missing some of the black magic needed for setup.

pllim: I just copy-pasted that part from the `__init__.py` file. If you feel strongly about it, I can remove it. I think all this affects is the `from astropy.io.votable.exceptions import *` command and also when Sphinx scrapes the files for API doc.

pllim: Ah, actually, I do need to change the doc a little to make it render.

pllim: I pushed the doc update. Given that the doc doesn't use `automodapi` for `exceptions`, this addition only affects star-imports. Are you okay with leaving this in or you want me to take it back out? Please advise, @tomdonaldson . Thanks!

tomdonaldson: I'm fine leaving it in since I'm not sure enough that it would never be useful.  It's certainly not harmful in any important way.

pllim:
Thanks for the review, @tomdonaldson !

PR #10158, opened 2020-04-16
fix units of compound models

### LINKED ISSUE #9921: CompoundModel.return_units assumes that outputs will have have same names as inputs
### Description
When a CompoundModel is constructed from a component model that has different input and output names, its return_units method fails with a KeyError.

### Steps to Reproduce
```python
from astropy.modeling import Model

class ExampleModel(Model):
    n_inputs = 1
    n_outputs = 1
    def __init__(self):
        self._input_units = {"x": u.m}
        self._return_units = {"y": u.s}
        super().__init__()
 
    def evaluate(self, input):
        return u.Quantity(1, u.s)

m = ExampleModel()

(m & m).return_units
```
Result:
```
<ipython-input-1-6d0cb817f1ae> in <module>
     15 m = ExampleModel()
     16
---> 17 (m & m).return_units

~/repos/astropy/astropy/modeling/core.py in return_units(self)
   3129         inputs_map = self.inputs_map()
   3130         return {key: inputs_map[key][0].return_units[orig_key]
-> 3131                 for key, (mod, orig_key) in inputs_map.items()
   3132                 if inputs_map[key][0].return_units is not None}
   3133

~/repos/astropy/astropy/modeling/core.py in <dictcomp>(.0)
   3130         return {key: inputs_map[key][0].return_units[orig_key]
   3131                 for key, (mod, orig_key) in inputs_map.items()
-> 3132                 if inputs_map[key][0].return_units is not None}
   3133
   3134     def outputs_map(self):

KeyError: 'x'
```

### System Details
```
Darwin-17.7.0-x86_64-i386-64bit
Python 3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 22:05:27)
[Clang 9.0.1 ]
Numpy 1.18.1
astropy 4.1.dev503+g1cb6c81e8.d20200129
Scipy 1.4.1
```

nden:
Fixes #9921

Map `input_units` and `return_units` of compound models correctly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -396,6 +396,9 @@ astropy.io.votable
 astropy.modeling
 ^^^^^^^^^^^^^^^^
 
+- Fixed a problem with mapping ``input_units`` and ``return_units``
+  of a ``CompoundModel`` to the units of the constituent models. [#10158]
+
 astropy.nddata
 ^^^^^^^^^^^^^^
 
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -3089,6 +3089,14 @@ class CompoundModel(Model):
                         inputs_map[inp] = r_inputs_map[self.right.inputs[i - len(self.left.inputs)]]
                     else:
                         inputs_map[inp] = self.right, self.right.inputs[i - len(self.left.inputs)]
+        elif self.op == 'fix_inputs':
+            fixed_ind = list(self.right.keys())
+            ind = [list(self.left.inputs).index(i) if isinstance(i, str) else i for i in fixed_ind]
+            inp_ind = list(range(self.left.n_inputs))
+            for i in ind:
+                inp_ind.remove(i)
+            for i in inp_ind:
+                inputs_map[self.left.inputs[i]] = self.left, self.left.inputs[i]
         else:
             if isinstance(self.left, CompoundModel):
                 l_inputs_map = self.left.inputs_map()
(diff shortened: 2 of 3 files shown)


pllim:
cc @eslavich

nden:
I think this should be ported to the LTS. Is this 4.0.2? There's no 4.0.2 in CHANGES.rst. Am I suppose to put the change log under 4.1?

nden:
Oh, never mind, I didn't see the 4.0.2 milestone.

astropy/modeling/core.py
```diff
@@ -3089,6 +3089,14 @@ def inputs_map(self):
                         inputs_map[inp] = r_inputs_map[self.right.inputs[i - len(self.left.inputs)]]
                     else:
                         inputs_map[inp] = self.right, self.right.inputs[i - len(self.left.inputs)]
+        elif self.op == 'fix_inputs':
+            fixed_ind = list(self.right.keys())
+            ind = [list(self.left.inputs).index(i) if isinstance(i, str) else i for i in fixed_ind]
+            inp_ind = list(range(self.left.n_inputs))
+            for i in ind:
```
perrygreenfield, line 3096: Don't these indices need to be removed in reverse order?

CHANGES.rst
```diff
@@ -387,6 +387,8 @@ astropy.io.votable
 astropy.modeling
 ^^^^^^^^^^^^^^^^
 
+- Fixed a problem with mapping ``input_units`` and ``return_units``
+  of a ``CompoundModel`` to the units of the constituent models. [#10158]
```
pllim, line 391: This would fix the doc failure.
```suggestion
  of a ``CompoundModel`` to the units of the constituent models. [#10158]

```

astropy/modeling/core.py
nden: I'll do it to be on the safe side. But doesn't python now have some magic to deal with this issue?

pllim:
The failure is real but unrelated. Let's hold off on merging until after #10160 is fixed, so we can be sure the full suite would pass before merge. Thanks for your patience.

nden:
I rebased to pick up #10160. Any idea why tests are failing? The failures seem unrelated.

pllim:
GitHub having issues right now. Might want to try again later.

PR #10165, opened 2020-04-18
Added paragraph in astropy.io.fits index.rst describing non-standard  FITS

### LINKED ISSUE #10153: Error reading Solar Dynamics Observatory, HMI continuum, FITS file with "nan" fields
### Description

I am unable to read a FITS file with astropy.io.fits produced by the [Solar Dynamics Observatory](https://sdo.gsfc.nasa.gov/data/aiahmi/) for the HMI continuum data product. I obtained the FITS files from  the [SDO JSOC](http://jsoc.stanford.edu/ajax/lookdata.html?ds=hmi.Ic_720s) but this requires users to register email and learn how to request data. I have the file temporarily available [here ](https://arg.usask.ca/wheels/SDO/hmi.Ic_720s.20170530_000000_TAI.3.continuum.fits) if you want to replicate the problem. 

I actually think this is not really a bug in astropy.io.fits but is due to the SDO HMI team using NaN values which is beyond the FITS standard (I think). Unfortunately this is a good idea and they obviously have FITS software that lets them create these files.  From a pragmatic point of view it might make sense for astropy.io.fits to also support the option, which seems to be relatively straight forward,  so it can keep up with other FITS software out there. 

I have done some diagnostics outlined below and I have hacked my version of astropy to allow me to keep on working with the SDO HMI data.

### Expected behavior

I hoped astropy.io.fits  would read the SDO HMI continuum FITS file  without issue. It consists of one 2-D image, 4096x4096 numbers. 

The problem seems to be that the SDO HMI people have encoded "nan" values into some of the  CARD values used in their FITS file.  This is not supported by astropy.io.fits and may not be part of the FITS standard. Unfortunately it is used by the HMI instrument.

Other people have seen this problem and reported issues on [Stack Overflow](https://stackoverflow.com/questions/59450462/getting-nan-values-while-trying-to-convert-fits-file-to-ndarray-using-astropy). 

The sample code is straight-forward but must use SDO HMI data files.  The example is shown below. The fits.open line works without issue but crashes on the line data = hdul[1].data.

```
    from astropy.io import fits
    
    fullname='hmi.Ic_720s.20170530_000000_TAI.3.continuum.fits'
    hdul =  fits.open(fullname)
    data =  hdul[1].data
```
Curiously the code does not always crash when using the PyCharm/Python debugger environment. I suspect the debugger is calling many @Property based values when it displays variable values and this is circumventing the normal internal caching and checking going on inside the FITS code. Definitely complicates debugging.

The code always crashes when it runs without a debugger. The following errors are output when executing the last line,

```
    WARNING: VerifyWarning: Verification reported errors: [astropy.io.fits.verify]
    WARNING: VerifyWarning: Card 'CRDER1' is not FITS standard (invalid value string: 'nan').  Fixed 'CRDER1' card to meet the FITS standard. [astropy.io.fits.verify]
    WARNING: VerifyWarning: Note: astropy.io.fits uses zero-based indexing.
    [astropy.io.fits.verify]
    [astropy.io.fits.verify]
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 744, in __get__ 
    val = self.fget(obj)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\compressed.py", line 1363, in data
    data = compression.decompress_hdu(self)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 744, in __get__
    val = self.fget(obj)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\compressed.py", line 1415, in compressed_data
    compressed_data = super().data
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 744, in __get__
    val = self.fget(obj)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\table.py", line 399, in data
    data = self._get_tbdata()
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\table.py", line 156, in _get_tbdata
    columns = self.columns
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 744, in __get__
    val = self.fget(obj)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\table.py", line 395, in columns
    return self._columns_type(self)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\column.py", line 1383, in __init__
    self._init_from_table(input)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\column.py", line 1448, in _init_from_table
    for keyword, value in hdr.items():
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\header.py", line 983, in items
    yield (card.keyword, card.value)
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\card.py", line 285, in value
    value = self._value = self._parse_value()
      File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\card.py", line 767, in _parse_value
    ".verify('fix').".format(self.keyword))
    astropy.io.fits.verify.VerifyError: Unparsable card (CRDER2), fix it first with .verify('fix').
    
    
    Process finished with exit code 1
```

### System Details

Numpy 1.18.1
astropy 4.0.1.post1
Scipy 1.3.1

### Diagnostics
I have debugged the code to the point that I have a temporary fix on my own machine. The problem is in file  `astropy.io.fits.card.py` in function `Card._parse_value(self)`.

Around line 764 there is a piece of code which is where the problem occurs.

```
        m = self._value_NFSC_RE.match(self._split()[1])

        if m is None:
            raise VerifyError("Unparsable card ({}), fix it first with "
                              ".verify('fix').".format(self.keyword))
```

The line `m = self._value_NFSC_RE.match(self._split()[1])` is trying to ensure that a CARD value is valid and 'nan' is not an acceptable value. The match returns as None and the exception is thrown.

The quick fix I have implemented changes the regular expression so it passes the match. 

1) The first action is to change the regular expression around line 111 of `card.py` to include a new match group called _nan_. In my implementation I only match 'nan'. More general implementations might also match 'NaN' and 'NAN'. 

```
    _value_NFSC_RE = re.compile(
        r'(?P<valu_field> *'
            r'(?P<valu>'
                r'\'(?P<strg>([ -~]+?|\'\'|) *?)\'(?=$|/| )|'
                r'(?P<bool>[FT])|'
                r'(?P<nan>nan)|'
                r'(?P<numr>' + _numr_NFSC + r')|'
                r'(?P<cplx>\( *'
                    r'(?P<real>' + _numr_NFSC + r') *, *'
                    r'(?P<imag>' + _numr_NFSC + r') *\))'
            r')? *)'
        r'(?P<comm_field>'
            r'(?P<sepr>/ *)'
            r'(?P<comm>(.|\n)*)'
        r')?$')
```

2) The second action is to edit function `Card._parse_value(self)` around line 764 to manage this new match group (nan) and use the `numpy.nan` function to assign a value,

```
        m = self._value_NFSC_RE.match(self._split()[1])

        if m is None:
            raise VerifyError("Unparsable card ({}), fix it first with "
                              ".verify('fix').".format(self.keyword))

        if m.group('bool') is not None:
            value = m.group('bool') == 'T'
        if m.group('nan') is not None:
            value = np.nan
        elif m.group('strg') is not None:
            value = re.sub("''", "'", m.group('strg'))
        ...
```
With these fixes, the code on my machine is working. In practice a robust fix would check some of the other places where CARD values are parsed.

 Please let me know if I can be of any assistance. Astropy is a great package and I would be more than happy to help with this problem.

ndl303:
### Description

This pull request is to address documentation for handling non-standard FITS files. It adds a short
paragraph quickly reviewing how non-standard FITS files are handled a simple example for users who need to read a non-standard FITS file. It includes a link to the FITS Verification page.

Changes.rst has not been edited (yet).

Fixes #10153

### DIFF
--- a/docs/io/fits/index.rst
+++ b/docs/io/fits/index.rst
@@ -181,16 +181,21 @@ a pre-existing file that is being written to.
 
 Working with non-standard files
 """""""""""""""""""""""""""""""
-`astropy.io.fits` adheres to the FITS standard and verifies that all fields within a FITS file conform
-to the standard. It may raise ``VerifyError`` if a field is deemed to be non-compliant. This action can be modified
-using method ``HDUList.verify``. It is typically invoked after opening the file but before reading any headers or data::
+When `astropy.io.fits` reads a FITS file which does not conform to the FITS standard it will try
+to make an educated interpretation of non-compliant fields. This may not always
+succeed and may trigger warnings when accessing headers or exceptions when writing
+to file. Verification of fields written to an output file can be controlled with
+the ``output_verify`` parameter of :func:`open`. Files opened for reading can be
+verified and fixed with method ``HDUList.verify``. This method is typically invoked
+after opening the file but before accessing any headers or data::
 
-    >>> hdul = fits.open(fits_image_filename)
-    >>> hdul.verify('silentfix+warn')
-    >>> data = hdul.data
+    >>> with fits.open(fits_image_filename) as hdul:
+    >>>    hdul.verify('fix')
+    >>>    data = hdul[1].data
 
-In the above example, the call to ``hdul.verify(silentfix+warn)`` requests that `astropy.io.fits` stay silent for any non-compliances
-it can fix but issue warnings for unfixable errors. Other options are described under FITS :ref:`fits_io_verification`
+In the above example, the call to ``hdul.verify("fix")`` requests that `astropy.io.fits`
+fix non-compliant fields and print informative messages. Other options in addition to ``"fix"``
+are described under FITS :ref:`fits_io_verification`
 
 .. seealso:: FITS :ref:`fits_io_verification`.
 
--- a/docs/io/fits/index.rst
+++ b/docs/io/fits/index.rst
@@ -190,8 +190,8 @@ verified and fixed with method ``HDUList.verify``. This method is typically invo
 after opening the file but before accessing any headers or data::
 
     >>> with fits.open(fits_image_filename) as hdul:
-    >>>    hdul.verify('fix')
-    >>>    data = hdul[1].data
+    ...    hdul.verify('fix')
+    ...    data = hdul[1].data
 
 In the above example, the call to ``hdul.verify("fix")`` requests that `astropy.io.fits`
 fix non-compliant fields and print informative messages. Other options in addition to ``"fix"``
--- a/docs/io/fits/index.rst
+++ b/docs/io/fits/index.rst
@@ -178,6 +178,19 @@ When writing a file (e.g., with the :func:`writeto` function), compression will
 be determined based on the filename extension given, or the compression used in
 a pre-existing file that is being written to.
 
+
+Working with non-standard files
+"""""""""""""""""""""""""""""""
+`astropy.io.fits` adheres to the FITS standard and verifies that all fields within a FITS file conform
+to the standard. It may raise ``VerifyError`` if a field is deemed to be non-compliant. This action can be modified
+using method :meth:`HDUList.verify`. It is typically invoked after opening the file but before reading any headers or data::
+
+    >>> hdul = fits.open(fits_image_filename)
+    >>> hdul.verify('silentfix+warn')
+    >>> data = hdul.data
+
+.. seealso:: FITS :ref:`fits_io_verification`.
+
 Working with FITS Headers
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/docs/io/fits/index.rst
+++ b/docs/io/fits/index.rst
@@ -183,11 +183,14 @@ Working with non-standard files
 """""""""""""""""""""""""""""""
 `astropy.io.fits` adheres to the FITS standard and verifies that all fields within a FITS file conform
 to the standard. It may raise ``VerifyError`` if a field is deemed to be non-compliant. This action can be modified
-using method :meth:`HDUList.verify`. It is typically invoked after opening the file but before reading any headers or data::
+using method ``HDUList.verify``. It is typically invoked after opening the file but before reading any headers or data::
 
     >>> hdul = fits.open(fits_image_filename)
     >>> hdul.verify('silentfix+warn')
     >>> data = hdul.data
+
+In the above example, the call to ``hdul.verify(silentfix+warn)`` requests that `astropy.io.fits` stay silent for any non-compliances
+it can fix but issue warnings for unfixable errors. Other options are described under FITS :ref:`fits_io_verification`
 
 .. seealso:: FITS :ref:`fits_io_verification`.
 
--- a/docs/io/fits/usage/verification.rst
+++ b/docs/io/fits/usage/verification.rst
@@ -1,5 +1,7 @@
 .. currentmodule:: astropy.io.fits
 
+..  _fits_io_verification:
+
 Verification
 ************
 


dhomeier:
This looks very useful! Maybe a line could be added explaining that the given example would have `io.fits` interpret the non-standard conforming data the best it can, while emitting only a single warning (for each violation, if I recall the `silentfix+warn` setting right)?

saimn:
Thanks for the PR. 
I think this needs to be rephrased a bit to highlight a few things: io.fits does not verify inputs, except in a few cases (typically when printing a header in the console), so its the user responsibility to verify the file by calling `.verify` if one needs to know if the file is valid or not. But io.fits will run the verification when writing a file. The idea is to be tolerant with input and strict with output. This is the role of the `output_verify` keyword (which defaults to `exception`) for `fits.open` (added recently), `close`, `flush`, `writeto`.

ndl303:
Thanks for the comments **saimn**.
 
I am slightly confused by your statements as they are contrary to my experience. In my experience, when reading Solar Dynamics Observatory HMI Continuum FITS data, ``io.fits`` threw ``VerifyError`` exceptions when reading _nan_ values_ from CARD values from inside method ``Card._parse_value``.  The only way I and others (e.g. sunpy) have found to successfully read the HMI FITS files is to insert **verify** in the code immediately after the  ``open`` statement; the code then works flawlessly.

My observation is that a call to **verify()** is essential when reading (some?) HMI non-compliant files. I think this is contrary to your statements which suggests (to me) it is only a diagnostic when reading. Please let me know if I have misunderstood your comments. As it stands right now I would not know how to change what I have written in the directions you have indicated. Perhaps we can iterate a couple of times as I agree that we don't want to put (or keep) bad information out there. I would be happy to run a simple test or two if we need clarification of any issues. I am not a FITS expert by any stretch of the imagination so finger trouble at my end is not out of the question.

On another note: I was speaking to the SDO HMI people and they have already started the process of making their FITS files compliant, which is probably a good thing.

saimn:
Indeed, it is a bit of a special case, in your case it fails because when accessing `.data` the code parses the header, and it actually parses all keywords *and values*, while it actually does not need to do that. And in other parts of the code, like when using an `ImageHDU` instead of a compressed one, the same error would not have failed. 
In general the idea is to avoid a crash like yours when just reading a file, unless there is a good reason, like a mandatory keyword that is missing or cannot be interpreted. Here the keyword causing the crash is not useful to read the data, so it should not have been read, and you would have noticed the error only when printing the header (with warnings instead of an exception), or when saving the file (with an exception this time). 
Hope it clarifies things a bit, which is not simple, sorry!

ndl303:
Hi @saimn.  Sorry for the long response. I agree that in general you want to avoid a crash (who does not?). But how do I actually pull that off?  I tried fixing the problem cards before reading the data but I still get exceptions and errors when reading the data field.  To illustrate this, I tried three examples:

1.  open the file and load the cards to find out the problems cards. This identifies 4 bad cards.
2.  open the file and fix the 4 problem cards before loading the data. This claims to fix the cards but still fails when I load the data. 
3. open the file , call `hdul.verify` and then load the data. This works flawlessly.

```
hdul=fits.open(filename)
cards = hdul[1].header.cards
WARNING: VerifyWarning: Verification reported errors: [astropy.io.fits.verify]
WARNING: VerifyWarning: Card 'CRDER1' is not FITS standard (invalid value string: 'nan').  Fixed 'CRDER1' card to meet the FITS standard. [astropy.io.fits.verify]
WARNING: VerifyWarning: Note: astropy.io.fits uses zero-based indexing.
 [astropy.io.fits.verify]
WARNING: VerifyWarning: Card 'CRDER2' is not FITS standard (invalid value string: 'nan').  Fixed 'CRDER2' card to meet the FITS standard. [astropy.io.fits.verify]
WARNING: VerifyWarning: Card 'CSYSER1' is not FITS standard (invalid value string: 'nan').  Fixed 'CSYSER1' card to meet the FITS standard. [astropy.io.fits.verify]
WARNING: VerifyWarning: Card 'CSYSER2' is not FITS standard (invalid value string: 'nan').  Fixed 'CSYSER2' card to meet the FITS standard. [astropy.io.fits.verify]
hdul.close()
```
```
hdul=fits.open(filename)
hdul[1].header.cards['CRDER2'].verify('silentfix+warn')
hdul[1].header.cards['CRDER1'].verify('silentfix+warn')
hdul[1].header.cards['CSYSER1'].verify('silentfix+warn')
hdul[1].header.cards['CSYSER2'].verify('silentfix+warn')
data = hdul[1].data
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\IPython\core\interactiveshell.py", line 3331, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-31-4b341e9665e2>", line 1, in <module>
    data = hdul[1].data
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 758, in __get__
    val = self.fget(obj)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\compressed.py", line 1363, in data
    data = compression.decompress_hdu(self)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 758, in __get__
    val = self.fget(obj)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\compressed.py", line 1415, in compressed_data
    compressed_data = super().data
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 758, in __get__
    val = self.fget(obj)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\table.py", line 399, in data
    data = self._get_tbdata()
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\table.py", line 156, in _get_tbdata
    columns = self.columns
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\utils\decorators.py", line 758, in __get__
    val = self.fget(obj)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\hdu\table.py", line 395, in columns
    return self._columns_type(self)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\column.py", line 1405, in __init__
    self._init_from_table(input)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\column.py", line 1470, in _init_from_table
    for keyword, value in hdr.items():
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\header.py", line 983, in items
    yield (card.keyword, card.value)
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\card.py", line 285, in value
    value = self._value = self._parse_value()
  File "C:\Users\nickl\Anaconda3\envs\altius\lib\site-packages\astropy\io\fits\card.py", line 767, in _parse_value
    ".verify('fix').".format(self.keyword))
astropy.io.fits.verify.VerifyError: Unparsable card (CRDER1), fix it first with .verify('fix').
hdul.close()
```

```
hdul=fits.open(filename)
hdul.verify('silentfix+warn')
data = hdul[1].data

```
I was unable to find any function or method within 10-20 minutes of browsing the documentation that would convert the CompImageHDU  to a ImageHDU so I could not write any code to look at this solution.

It strikes me that, as presented through the Astropy FITS documentation, a novice user must go through `.data` to get to the data they want, which implies they will read all the header fields and values. If this fails, which it did for me, then if no solution is forthcoming after reading the primary documentation, there is a good chance they will either abandon astropy.io.fits as "not working" or too difficult to use or they will post questions to astropy support (which is what I did).  None of these scenarios are particularly desirable for anyone.

I think my intent with this pull request is to provide a short paragraph on the primary FITS page that gives a short practical account of what to do if someone encounters a verification exception when reading a FITS file.  I think my few lines describing the `hdul.verify()` option do a reasonable job of meeting that goal and direct the interested reader to the verification page for more detailed info. 

I would be happy to change the text in the paragraph if you feel it is inaccurate, misleading or insufficient but I have not yet seen or found anything to suggest this.  I can also accept that the pull request is not implemented if the circumstances I have encountered are considered too specific to be of use to the general FITS user. I would personally disagree with that decision but I would accept the review process.

dhomeier:
> It strikes me that, as presented through the Astropy FITS documentation, a novice user must go through `.data` to get to the data they want, which implies they will read all the header fields and values. If this fails, which it did for me, then if no solution is forthcoming after reading the primary documentation, there is a good chance they will either abandon astropy.io.fits as "not working" or too difficult to use or they will post questions to astropy support (which is what I did). None of these scenarios are particularly desirable for anyone.

Since `data` is the hdu property containing the actual data, it is quite inevitable to go through it, at least indirectly. But if I understand #10171 correctly, that PR should already fix avoidable crashes on accessing `data`, as long as all _required_ cards for interpreting it have valid values, is that right @saimn?
I still think _this_ PR adds valuable documentation on how to use `.verify` even once #10171 is merged, but then it probably needs to be clarified what cases remain where the data cannot be read with the default settings, but `verify` would resolve them easily – e.g. could `silentfix` also change required, but missing or invalid cards, to a default value?

saimn:
Yes, thanks @dhomeier , with #10171 you would be able to open the file and access the data without crashing. Because we need to parse only the *structural* keywords, the ones useful to read the data. The same is already true for image data:
```
In [15]: hdul = fits.open('test.fits')                                                         

In [17]: hdul[1].data                                                                          
Out[17]: array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])

In [18]: hdul[1].header                                                                        
Out[18]: WARNING: VerifyWarning: Verification reported errors: [astropy.io.fits.verify]
WARNING: VerifyWarning: Card 'FOO' is not FITS standard (invalid value string: 'NAN.0').  Fixed 'FOO' card to meet the FITS standard. [astropy.io.fits.verify]
WARNING: VerifyWarning: Note: astropy.io.fits uses zero-based indexing.
 [astropy.io.fits.verify]

XTENSION= 'IMAGE   '           / Image extension                                
BITPIX  =                  -64 / array data type                                
NAXIS   =                    1 / number of array dimensions                     
NAXIS1  =                   10                                                  
PCOUNT  =                    0 / number of parameters                           
FOO     = 'NAN.0   '                                 
```
By default `io.fits` makes it hard to write such an invalid file, but the idea is that you should be able to read existing files even if they are invalid.

Then I also agree that the documentation addition is useful, I just wanted to explain a bit more the current behavior. And the verification documentation (http://docs.astropy.org/en/latest/io/fits/usage/verification.html) states clearly that *When astropy reads a FITS file which does not conform to FITS standard, it will not raise an error and exit*.

docs/io/fits/index.rst
```diff
@@ -178,6 +178,22 @@ When writing a file (e.g., with the :func:`writeto` function), compression will
 be determined based on the filename extension given, or the compression used in
 a pre-existing file that is being written to.
 
+
+Working with non-standard files
+"""""""""""""""""""""""""""""""
+`astropy.io.fits` adheres to the FITS standard and verifies that all fields within a FITS file conform
+to the standard. It may raise ``VerifyError`` if a field is deemed to be non-compliant. This action can be modified
+using method ``HDUList.verify``. It is typically invoked after opening the file but before reading any headers or data::
```
saimn, line 186: Suggestion:

When astropy reads a FITS file which does not conform to FITS standard, it will not raise an error and exit. It will try to make the best educated interpretation and only gives up when the offending data is accessed and no unambiguous interpretation can be reached. If the file contains some invalid keyword, it may trigger warnings later when printing the header, or an exception when trying to write the file.
The file can be verified and fixed with `HDUList.verify`, and this can also be controlled when writing the file with the `output_verify` parameter.

docs/io/fits/index.rst
```diff
@@ -178,6 +178,22 @@ When writing a file (e.g., with the :func:`writeto` function), compression will
...
+
+    >>> hdul = fits.open(fits_image_filename)
+    >>> hdul.verify('silentfix+warn')
+    >>> data = hdul.data
```
saimn, line 190: Please use the context manager here, `with fits.open(fits_image_filename) as hdul: ...`. Also for the educational purpose of the example I would prefer `fix` instead of `silentfix` ;).

docs/io/fits/index.rst
ndl303: I am very happy with this overall direction as it looks like we are converging on a sensible addition to the FITS documentation.

I am confused in the first sentence when when you say "it will not raise an error and exit". This is the exact opposite of my experience: astropy does raise an error and exit when reading  non-compliant FITS files generated by the solar data observatory HMI instrument, (unless you include the "verify" option).  This is the reason why I decided to write the extra paragraph in the FITS documentation, as it took me some time and false attempts to find the `verify` solution, e.g I tried using `output_verify` which is documented in the `open` method but that did not work for me when reading files.

It seems from the previous messages that you are basing your statement _"it will not raise an error and exit"_ upon information on the verification page (http://docs.astropy.org/en/latest/io/fits/usage/verification.html). I believe this sentence is slightly misleading. I think it should say _when astropy reads a FITS file which does not conform to FITS standard, it will not IMMEDIATELY raise an error and exit_. 

The current version of Astropy does eventually raise an error if it cannot decode a value when reading a file. For example, I raise an error near line 766 of card.py in method `_parse_value` when I try to read my FITS file. If I look at this code I see,

```
        m = self._value_NFSC_RE.match(self._split()[1])

        if m is None:
            raise VerifyError("Unparsable card ({}), fix it first with "
                              ".verify('fix').".format(self.keyword))

```
Clearly a `VerifyError` is raised if `m is None` when `self._split()[1]` fails to match the regular expression. 

Am I making some mistake here and just missing the point? I apologize if that is the case.
I am new to making modifications to Astropy and I am still feeling my way through the system.

I think if we can get this issue clarified the pull request will sort itself out quite quickly. I am happy to run some test cases if that would help clarify any questions.

docs/io/fits/index.rst
ndl303: OK. Sounds good to me.  I was following the format of earlier examples on the documentation page which do not use the "with" statement. I'm ok with 'fix'

docs/io/fits/index.rst
ndl303: OK. I think I am slowly coming up to speed. Are you suggesting that we have a paragraph in the documentation that is consistent with the changes applied by @dhomeier  with #10171 ? This would then make your suggested paragraph make sense to me.

saimn: #10171 will fix the crash with your file when reading the data. So then the sentence "it will not raise an error and exit" will be true (hopefully, there may still be cases where it could crash, if a required keyword is invalid, but hopefully this should not be so frequent).

docs/io/fits/index.rst
```diff
@@ -178,6 +178,27 @@ When writing a file (e.g., with the :func:`writeto` function), compression will
...
+to make an educated interpretation of non-compliant fields. This may not always
+succeed and may trigger warnings when accessing headers or exceptions when writing
+to file. Verification of fields written to an output file can be controlled with
+the ``output_verify`` parameter of :func:`open`. Files opened for reading can be
+verified and fixed with method ``HDUList.verify``. This method is typically invoked
+after opening the file but before accessing any headers or data::
+
+    >>> with fits.open(fits_image_filename) as hdul:
+    >>>    hdul.verify('fix')
+    >>>    data = hdul[1].data
```
saimn, line 194: To fix the doctest:
```suggestion
    ...    hdul.verify('fix')
    ...    data = hdul[1].data
```

saimn:
I removed the merge commit and my commit which was conflicting with yours, so this is now good to go. Thanks for the contribution @ndl303 !

PR #10169, opened 2020-04-18
Support Table fuzzy or cross-match join with SkyCoord, Quantity, Column

taldcroft:
### Description

This PR adds the ability to specify a custom matching function for table joins.  In particular this makes it possible to do cross-match table joins on ``SkyCoord``,  ``Quantity``, or standard columns, where column entries within a specified  distance are considered to be matched. 

The change to `join` is to add a new kwarg `join_funcs` that is a dict of functions which are responsible for doing the actual cross-matching for a key column.

With sufficient ingenuity one could doing fuzzy text joins by supplying an appropriate join function.

To do:
- [x] Narrative docs describing using this and also how to write your own `join_func`.
- [x] Add `join_funcs` to `join` docstring.
- [x] Check for issues this might close.

### DIFF
--- a/astropy/table/operations.py
+++ b/astropy/table/operations.py
@@ -26,7 +26,7 @@ from . import _np_utils
 from .np_utils import fix_column_name, TableMergeError
 
 __all__ = ['join', 'setdiff', 'hstack', 'vstack', 'unique',
-           'skycoord_join', 'distance_join']
+           'join_skycoord', 'join_distance']
 
 
 def _merge_table_meta(out, tables, metadata_conflicts='warn'):
@@ -90,7 +90,7 @@ def _get_out_class(objs):
     return out_class
 
 
-def skycoord_join(distance, distance_kind='sky'):
+def join_skycoord(distance, distance_kind='sky'):
     """Helper function to join on SkyCoord columns using distance matching.
 
     This function is intended for use in ``table.join()`` to allow performing a
@@ -124,19 +124,19 @@ def skycoord_join(distance, distance_kind='sky'):
 
       >>> from astropy.coordinates import SkyCoord
       >>> import astropy.units as u
-      >>> from astropy.table import Table, skycoord_join
+      >>> from astropy.table import Table, join_skycoord
       >>> from astropy import table
 
       >>> sc1 = SkyCoord([0, 1, 1.1, 2], [0, 0, 0, 0], unit='deg')
       >>> sc2 = SkyCoord([0.5, 1.05, 2.1], [0, 0, 0], unit='deg')
 
-      >>> join_func = skycoord_join(0.2 * u.deg)
+      >>> join_func = join_skycoord(0.2 * u.deg)
       >>> join_func(sc1, sc2)  # Associate each coordinate with unique source ID
       (array([3, 1, 1, 2]), array([4, 1, 2]))
 
       >>> t1 = Table([sc1], names=['sc'])
       >>> t2 = Table([sc2], names=['sc'])
-      >>> t12 = table.join(t1, t2, join_funcs={'sc': skycoord_join(0.2 * u.deg)})
+      >>> t12 = table.join(t1, t2, join_funcs={'sc': join_skycoord(0.2 * u.deg)})
       >>> print(t12)  # Note new `sc_id` column with the IDs from join_func()
       sc_id   sc_1    sc_2
             deg,deg deg,deg
@@ -190,7 +190,7 @@ def skycoord_join(distance, distance_kind='sky'):
     return join_func
 
 
-def distance_join(distance, kdtree_args=None, query_args=None):
+def join_distance(distance, kdtree_args=None, query_args=None):
     """Helper function to join table columns using distance matching.
 
     This function is intended for use in ``table.join()`` to allow performing
@@ -221,7 +221,7 @@ def distance_join(distance, kdtree_args=None, query_args=None):
     Examples
     --------
 
-      >>> from astropy.table import Table, distance_join
+      >>> from astropy.table import Table, join_distance
       >>> from astropy import table
 
       >>> c1 = [0, 1, 1.1, 2]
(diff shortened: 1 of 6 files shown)


taldcroft:
@mhvk - ping to make sure this is on your radar for 4.1.

mhvk:
Definitely on the radar, but trying to get #10210 wrapped up first... Unfortuantely, timing of feature freeze is bad in Spring, with lots of student projects/committee meetings.

astropy/table/operations.py
```diff
@@ -87,9 +88,238 @@ def _get_out_class(objs):
...
+    The distance cross-matching is done using either
+    `~astropy.coordinates.SkyCoord.search_around_sky` or
+    `~astropy.coordinates.SkyCoord.search_around_3d`, depending on the value of
+    ``distance_kind``.
+
+    Parameters
+    ----------
+    distance : Quantity (angle or length)
+        Maximum distance between points to be considered a join match
+    distance_kind : str
```
mhvk, line 108: In principle, this could be inferred based on the unit of `distance`.

astropy/table/operations.py
```diff
@@ -87,9 +88,238 @@ def _get_out_class(objs):
...
+                ids[idx] = id_
+                id_ += 1
+
+        # End of enclosure join_func()
+        return ids1, ids2
+
+    return join_func
+
+
+def join_distance(distance, kdtree_args=None, query_args=None):
```
mhvk, line 191: Might it make sense to re-purpose people's experience with `isclose` and have an `atol` and `rtol`? (in which case this might become `join_close`).

taldcroft: This would need to be a completely new function since this is explicitly applicable to N-dimensional points and uses KDTree to find the neighbors.  For N-dim there is no concept of `rtol`.

For the 1-d case, I actually don't know how to do a cross match with `atol` and `rtol` without just doing the N^2 cross-match.

There is another problem I thought of, which is that this whole concept is based on the cross-matching being symmetric.  If point A is "close" to point B, then it is required that B is "close" to A.  That is true for a distance (`atol`), but breaks down for `rtol`.
```
In [7]: np.isclose(1, 0, atol=0.8, rtol=0.8)                                                                                                                  
Out[7]: False

In [8]: np.isclose(0, 1, atol=0.8, rtol=0.8)                                                                                                                  
Out[8]: True
```

mhvk: Oh, yes, my brain was not quite there with the 2-d case.  Let's just stick with `distance` which is nicely defined. (and, yes, `isclose` is not symmetric unlike the `math` equivalent; https://github.com/numpy/numpy/issues/10161)

mhvk:
circle-ci failures try to import scipy...

astropy/table/operations.py
taldcroft: Good idea, will do.

taldcroft:
> circle-ci failures try to import scipy...

Ah right, so need to skip test if no scipy.

astropy/table/operations.py
taldcroft: Even then I much prefer one less kwarg (*simple is better than complex*), on further reflection I think this "guessing" may be ambiguous because of `u.dimensionless_angles()`. Because of this equivalency one can pass in a dimensionless Quantity and it could either be a 3d cartesian distance or a dimensionless angular distance.  I don't know of any rock solid way to disambiguate, and even if we come up with one it will be complicated to explain to users. So I think it might be best to just leave this be (*explicit is better than implicit*).

mhvk: Can dimensionless ever be a 3-d distance? I guess the test would be `...unit.is_equivalent(u.radian)`. 

But perhaps an alternative is to have the default be determined by the unit? It is good to have the user be able to pass in an explicit `distance_kind` so that things will simply fail if `distance` has the wrong units.

docs/table/operations.rst
```diff
@@ -1075,6 +1075,96 @@ The rules for merging are the same as for `Merging metadata`_, and the
...
+match distance via a function enclosure.  The requirement of the join function
+is that it accepts two arguments corresponding to the two key columns, and
+returns a tuple of ``(ids1, ids2)``.  These identifiers correspond to the
+identification of each column entry with a unique matched source.
+
+  >>> join_func = join_skycoord(0.2 * u.deg)
+  >>> join_func(sc1, sc2)  # Associate each coordinate with unique source ID
+  (array([3, 1, 1, 2]), array([1, 4, 2]))
+
+If you would like to write your own we suggest starting from the source code
```
adrn, line 1152:
```suggestion
If you would like to write your own fuzzy matching function, we suggest starting from the source code
```
?

astropy/table/operations.py
```diff
@@ -87,9 +88,238 @@ def _get_out_class(objs):
     return out_class
 
 
+def join_skycoord(distance, distance_kind='sky'):
```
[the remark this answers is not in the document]
taldcroft: I see your point, but passing in the function is harder for the user (extra import required).  We have had this philosophical difference going back to forever, but in the end we agreed that passing in strings should generally be allowed instead of requiring imported objects.

How about meeting in the middle and allowing a string which must correspond to a function in `coordinates`i.e. (try `getattr(astropy.coordinates, distance_func)`, or a function that has the same API as the current search_around functions.  So instead `'sky'` or `'3d'` it would be the full name.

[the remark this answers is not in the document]
taldcroft: And I'll say that I am usually against putting in functionality where we have not yet identified a realistic use case.  In this case allowing passing in a user-defined distance function that meets the search_around API exactly.  So a plan B is to not allow passing in a function now and wait for user to request that feature.  And in fact in that case the user could make their own `join_func`.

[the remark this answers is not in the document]
adrn: Yea, philosophy. For completeness: This doesn't require another import if `import astropy.coordinates as coord` is already done (pass `coord.search_around_sky`). And the (weak) argument against string values is that it requires reading to see what values are allowed - if the default was `coord.search_around_sky`, the user might think to try `coord.search_<TAB>` (in a notebook) to see what other options there are. But, yes, this is a weak argument and perhaps too "behavioral"...

In practice, though, I don't feel very strongly about this, so I leave it up to you!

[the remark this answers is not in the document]
taldcroft: Yes, in my world I just do `from astropy.coordinates import SkyCoord` and that is all I need!  :smile:  And in practice time is running out ... I wasn't totally happy with `distance_kind` and I like `distance_func` better.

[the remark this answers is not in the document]
mhvk: Also, if the default of the function is based on the unit, this will very rarely be needed!

PR #10171, opened 2020-04-19
Avoid crash when reading data

saimn:
To avoid crash like in #10153, this invalid should trigger a validation warning or error later but not when reading the data.
See also https://github.com/astropy/astropy/pull/10165#issuecomment-616018021 and my commit comment below, hopefully this is clear enough (but it's late :)).

Reading data for a table uses a loop on all header keywords and
values, so if a value is invalid this triggers an exception. This can be
avoided by looping only on keywords, and reading values only for the
structural keywords. This makes the behavior more consistent with other
HDU classes, and delay the verification errors.

### DIFF
--- a/astropy/io/fits/tests/test_table.py
+++ b/astropy/io/fits/tests/test_table.py
@@ -3344,3 +3344,15 @@ def test_a3dtable(tmpdir):
             'Converted the XTENSION keyword to BINTABLE.')
 
         assert hdul[1].header['XTENSION'] == 'BINTABLE'
+
+
+def test_invalid_file(tmp_path):
+    hdu = fits.BinTableHDU()
+    # little trick to write an invalid card ...
+    hdu.header['FOO'] = None
+    hdu.header.cards['FOO']._value = np.nan
+
+    testfile = tmp_path / 'test.fits'
+    hdu.writeto(testfile, output_verify='ignore')
+    hdul = fits.open(testfile)
+    assert hdul[1].data is not None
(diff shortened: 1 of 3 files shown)


saimn:
Added a test, before the other commit to make sure that this was crashing before ;).
Anyone interested in reviewing ?

pllim:
cc @ndl303

astropy/io/fits/tests/test_table.py
```diff
@@ -3344,3 +3344,15 @@ def test_a3dtable(tmpdir):
...
+def test_invalid_file(tmp_path):
+    hdu = fits.BinTableHDU()
+    # little trick to write an invalid card ...
+    hdu.header['FOO'] = None
+    hdu.header.cards['FOO']._value = np.nan
+
+    testfile = tmp_path / 'test.fits'
+    hdu.writeto(testfile, output_verify='ignore')
+    hdul = fits.open(testfile)
+    assert hdul[1].data is not None
```
pllim, line 3358: Don't we want to close it gracefully?
```suggestion
    with fits.open(testfile) as hdul:
        assert hdul[1].data is not None
```

saimn: Yep, good point :)

pllim:
Note to whoever merging this: Remember to check the remote data job that is allowed to fail, just in case.

saimn:
All green and the dev job failure is unrelated. Thanks for the review @pllim ! (If you want more fitsy stuff, #10124 is also waiting for a charitable soul ;)).

PR #10173, opened 2020-04-20
Remove include line related to helpers

migueldvb:
### Description

`ah_bootstrap.py` has been removed from the repo and should not be
included in `MANIFEST.in`.


pllim:
Good catch. Thanks!

PR #10175, opened 2020-04-21
Fix simple_norm keyword name

larrybradley:
The `log` parameter is actually `log_a` and it should go at the end of the parameters list to match the function signature.


pllim:
I think the doc build test should still run. Can you please change that commit with `[skip ci]` to `[skip travis]` and let CircleCI run? Thanks!

larrybradley:
Repushed with [skip-travis].

PR #8213, opened 2018-12-03
Units interaction with unyt

mhvk:
As a proof of concept following #8210, a quick attempt at accepting `unyt` units and quantities in our `Unit` and `Quantity` constructors. With this:
```
import astropy.units as u
import unyt
q = 1.*unyt.m/unyt.s
u.Unit(q.units)
# Unit("m / s")
u.Quantity(q)
# <Quantity 1. m / s>
```
But note that this is far from good enough:
```
q * (10.*u.minute)
# OOPS: unyt_quantity(10., 'm/s')
```
The reason is that we look for a `unit` attribute for our multiplication, and `unyt` looks for `units`. Obviously, it is possible to look for both but that it is not that easy to do so without slowing other things down (such as multiplication with a plain array).

PR #10196, opened 2020-04-23
Add method to convert to AMUSE type quantity

### LINKED ISSUE #10194: Add conversion method for other unit frameworks, e.g. AMUSE
### Description

It would be useful to have a method to convert astropy quantities to quantities in a different unit framework, e.g. that of AMUSE. 
Quantities could be converted in a way along the lines of the example below:
```
import astropy.constants as c

G = c.G.to_amuse()
# or, more general,
G = c.G.to_framework("amuse")
```
From AMUSE, we could have a similar method that would do the reverse.

### Additional context

See also http://github.com/amusecode/amuse/issues/170 for a discussion on this from the AMUSE side.

rieder:
### Description

This pull request is to address issue #10194, adding a .to_amuse method to Astropy quantities.
This method returns an AMUSE unit (if AMUSE is installed, raises an exception otherwise).

Fixes #10194

### DIFF
--- a/astropy/units/quantity.py
+++ b/astropy/units/quantity.py
@@ -752,15 +752,15 @@ class Quantity(np.ndarray):
             raise ModuleNotFoundError(
                 "AMUSE does not seem to be installed"
             )
-    
+
         # Find SI bases of the unit
         si_bases = self.si.unit.bases
         si_powers = self.si.unit.powers
         si_units = list(zip(si_powers, si_bases))
-    
+
         # Find the quantity's value in base units
         si_value = self.si.value
-    
+
         # Reconstruct the quantity in AMUSE units
         amuse_quantity = si_value
         for base_unit in si_units:
@@ -778,7 +778,7 @@ class Quantity(np.ndarray):
                 amuse_quantity = amuse_quantity * (1 | amuse_units.mol**base_unit[0])
             elif base_unit[1].name == "cd":
                 amuse_quantity = amuse_quantity * (1 | amuse_units.cd**base_unit[0])
-    
+
         return amuse_quantity
 
     value = property(to_value,
--- a/astropy/units/quantity.py
+++ b/astropy/units/quantity.py
@@ -737,6 +737,50 @@ class Quantity(np.ndarray):
         # Index with empty tuple to decay array scalars in to numpy scalars.
         return value[()]
 
+    def to_amuse(self):
+        """
+        Convert a unit from Astropy to AMUSE.
+
+        Returns
+        -------
+        quantity : AMUSE quantity
+            The equivalent of the Astropy quantity in the AMUSE unit framework.
+        """
+        try:
+            import amuse.units.units as amuse_units
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "AMUSE does not seem to be installed"
+            )
+    
+        # Find SI bases of the unit
+        si_bases = self.si.unit.bases
+        si_powers = self.si.unit.powers
+        si_units = list(zip(si_powers, si_bases))
+    
+        # Find the quantity's value in base units
+        si_value = self.si.value
+    
+        # Reconstruct the quantity in AMUSE units
+        amuse_quantity = si_value
+        for base_unit in si_units:
+            if base_unit[1].name == "m":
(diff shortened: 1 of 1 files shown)


pllim:
Thank you for your PR! I think this needs more discussions as it introduces optional dependency. Is there a reason why this cannot be done from the AMUSE side alone; e.g., `amuse.to_astropy_unit` or `amuse.from_astropy_unit` or something like that?

cc @mhvk @adrn @astropy/coordinators

rieder:
Thanks for your comments!

Of course it could be done from the AMUSE side as well, but I think this is simpler - it directly allows conversion from an existing unit, rather than importing an additional conversion function. Likewise, on the AMUSE side I will implement a .to_astropy method for quantities to do the same but in reverse.

Re AMUSE as an optional dependency: I cannot see a way around this. To convert to AMUSE, AMUSE needs to be present. Likewise, converting from AMUSE to Astropy needs Astropy to be present. I think that with a clearly named method it should not be a problem (though perhaps the exception needs work).

mhvk:
@rieder - I'm very much in favour of making things round-trip, but to me, this feels the wrong way around: we should make sure `Unit` can interpret amuse units, and amuse should make sure it can interpret astropy units. The big advantage is that then we can just rely on an instance's methods, without any need to import the other package, and we minimize the knowledge we need of the other package. Ideally, we then extend that to quantities in a follow-up PR. What do you think?

p.s. My guess is that the simplest way to do the actual round-trip is `u.Unit(str(amuse_unit))`, or perhaps some `.si` version... See also #8213, my aborted attempt to integrate with `unyt`, in particular [the change I made for `Unit`](https://github.com/astropy/astropy/pull/8213/files#diff-4c3555e3125721f440d2dc5d9c9d4e2b), which did exactly that (I am much less keen on what I did for `Quantity`).

rieder:
@mhvk I think it would be great if the units were fully exchangeable, but it seems like a much more major effort to make it so to me. I would be happy if we can make this work, but it will take more effort and I will not be able to do it just by myself.

The advantage of this PR is that it is both straightforward and light-weight, while it solves most of the problems converting between the two systems by using SI base units rather than derived units (which unfortunately are not always equal in the two unit frameworks, making going through string representation tricky at best and wrong at worst).

Below some examples of AMUSE units and quantities to demonstrate how they differ from each other:
```python3
from amuse.units import units

distance = units.km  # distance is a unit of 1 km
time = units.s  # time is a unit of 1 s
speed = 1 * units.kms  # speed is a unit of 1 km / s.
speed = distance / time  # equivalent
speed.base_unit()  # will return unit<m * s**-1>
speed.factor  # will return 1000 (speed in base_unit())

vx = 1 | distance / time  # vx is a quantity with value 1 km / s. AMUSE overloads the pipe operator to assign quantities.
vx.in_(speed)  # print vx using the unit speed
vx.in_base()  # print vx in base units
vx.value_in(speed)  # print the value of vx in speed
vx.in_(units.au * units.yr**-1)  # print vx using the unit au/yr
velocity = vx.to_unit()  # convert from quantity to unit
```

mhvk:
@rieder - I guess my basic point is that I'm much more comfortable with a "from_amuse" than a "to_amuse" - the former properly puts the burden on astropy to get something sensible out of an amuse unit, rather than try to produce one. Plus, we can have the interpretation be part of the `Unit` initializer, while the latter will always need a special method.

On the string representation, is there any way in `amuse` to, e.g., get a fits/cds/si-compliant unit string out?

p.s. I have nothing against going via SI!

adrn:
Thanks @rieder for making the connection here! Just chiming in to say that I very much agree with @mhvk here and have the same questions.

mhvk:
@rieder - thanks for the examples! I have an alternative in #10205, which is based on string representations - I do think those *have to be* unique, but I'm happy to hear about examples where we cannot parse an amuse unit string (likely, this would be a bug in one or the other package!).

Note that right now this is very specifically for units only, where it is relatively easy to add this in a way that has no impact on speed, etc., for astropy itself. For interaction with `unyt`, I found that quantities were trickier and I would prefer to do those in a follow-up PR.

mhvk:
closing since I don't think this is the direction we'll end up going into.

PR #10204, opened 2020-04-24
add build_ext option to test command hints

eteq:
This just adds the ``python setup.py build_ext --inplace`` option that's already mentioned in the testing guidelines to the hint that ``python setup.py test`` provides. Hopefully this isn't too controversial since it's already in the testing guidelines as an option...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,11 @@ directly with::
     pip install -e .[test]
     pytest
 
+or::
+
+    python setup.py build_ext --inplace
+    pytest
+
 For more information, see:
 
   http://docs.astropy.org/en/latest/development/testguide.html#running-tests


setup.py
```diff
@@ -29,6 +29,11 @@
     pip install -e .[test]
     pytest
 
+or::
+
+    python setup.py build_ext --inplace
```
pllim, line 34: This is not completely equivalent to `pip install -e .[test]`, is it? Does it install the dependencies listed under `test` section in `setup.cfg`?

saimn: This does not install the package, it just builds extensions. `setup.py develop` is closer (and run `build_ext`) but it will not install test dependencies.

pllim: Might be confusing to just "or" without any explanation here. When do I know which command I should use?

setup.py
```diff
@@ -29,6 +29,11 @@
     pip install -e .[test]
     pytest
 
+or::
```
astrofrog, line 32:
```suggestion
or if you prefer to avoid installing the developer version::
```

mhvk:
Since it is documentation, not release critical, though I'd count it as a "bug fix", so remilestoned to 4.1.1

embray:
I don't think it's controversial.  It's definitely something one can do.  But is there any strong motivation for it, other than avoiding creating a virtualenv for development?

mhvk:
Don't feel strongly either way, but I think it is somewhat useful if one has a nice system installation anyway, to avoid reinstalling numpy, etc. So, +0.5 to merging it, but with @astrofrog's suggestion to give a sense of why one would do this.

setup.py
astrojuanlu: Shouldn't we have fewer invocations of `setup.py`, and not more? I'm wondering why a user would avoid installing the development version (the advice could be: "if you don't want to overwrite your perfectly stable Astropy version, create a new environment for development").

pllim: I am hoping that by the time they resort to installing from source, they already have a dev workflow and know what they are doing. If we start to be specific, then we have to cover all the different environment managers and dev workflows that one might be using when they encounter this text. It is a slippery slope.

pllim:
That was surprisingly a lot of argument for `+5 −0` change. 😹  What was the consensus again?

saimn:
Hmm , `setup.py` commands are being gradually replaced by `pip`, setuptools will stay for long (I guess) so using it directly is still possible but I think we should not advertise it (people who know what they do can still use it).

astrofrog:
I'm in favor of closing

PR #10229, opened 2020-04-29
Tidy up SlicedLowLevelWCS to facilitate more development

Cadair:
Following discussion with @nden, @astrofrog and @perrygreenfield we decided that #9736 requires more time in the oven (and experimentation outside of core) so that development will likely move to ndcube for the next dev cycle or two.

Therefore, this replaces #9736 for 4.1 and is to facilitate further development of those classes outside of core.

This does the following main things:

* Moves `SlicedLowLevelWCS` to `wrappers/sliced_wcs.py`
* Implements a `BaseWrapper` class, to reduce duplication of APE 14 wrappers.
* Moves the implementation of `array_index_to_world_values` and `world_to_array_index_values` to `BaseLowLevelWCS` as they can be implemented based on the `pixel_` versions.
* Adds a bunch of useful pytest fixtures which can be use inside and outside core.


astropy/wcs/wcsapi/high_level_api.py
```diff
@@ -91,6 +90,10 @@ def world_to_array_index(self, *world_objects):
         pixel indexing and ordering conventions. The indices should be returned
         as rounded integers.
         """
+        if self.pixel_n_dim == 1:
+            return np.round(self.world_to_pixel(*world_objects)).astype(int)
+        else:
+            return tuple(np.round(self.world_to_pixel(*world_objects)[::-1]).astype(int).tolist())
```
nden, line 96: I'd prefer to use something like  https://github.com/spacetelescope/gwcs/blob/master/gwcs/utils.py#L52 instead of `round`, to get consistent results of converting to the center of the pixel.

nden: I see what I suggested is implemented in the high-level class. What is the reason for these to be different?

astropy/wcs/wcsapi/wrappers/base.py
```diff
@@ -0,0 +1,77 @@
...
+
+class BaseWCSWrapper(BaseLowLevelWCS, metaclass=abc.ABCMeta):
+    """
+    A base wrapper class for things that modify Low Level WCSes.
+
+    This wrapper implements a transparent wrapper to many of the properties,
+    with the idea that not all of them would need to be overridden in your
+    wrapper, but some probably will.
+    """
+    def __init__(self, wcs, *args, **kwargs):
```
nden, line 14: There's no `Parameters` section. In different examples I've seen `wcs` as a WCS object or a list of WCSs.

astropy/wcs/wcsapi/wrappers/sliced_wcs.py
```diff
@@ -0,0 +1,252 @@
...
+
+from .base import BaseWCSWrapper
+from astropy.utils import isiterable
+
+__all__ = ['sanitize_slices', 'SlicedLowLevelWCS']
+
+
+def sanitize_slices(slices, ndim):
+    """
+    Given a set of input
```
nden, line 13: Is this docstring unfinished?

astropy/wcs/wcsapi/wrappers/sliced_wcs.py
```diff
@@ -0,0 +1,252 @@
...
+            else:
+                stop = slice2.stop + slice1.start
+            if slice1.stop is not None:
+                stop = min(slice1.stop, stop)
+    return slice(start, stop)
+
+
+class SlicedLowLevelWCS(BaseWCSWrapper):
+
+    def __init__(self, wcs, slices):
```
nden, line 101: As a public class this should have docstrings.

mhvk:
Bit less confident here, but removing milestone as it would seem that one can develop against master and that, therefore, this is not release-critical for 4.1.

Cadair:
I appreciate I didn't get this one across the line so if it slips I will live with it, but a word for its inclusion in 4.1. 

This PR came out of #9736 and discussions with interested people around it. The idea is for us to bundle the contents of #9736 in the upcoming release of ndcube, to give them some time out in the wild and see how well they work. Therefore, I would *really* like this in 4.1 so that NDCube 2.0 can depend upon 4.1 and make my life a whole lot easier.

Also the contents of this PR is mostly refactoring and docstring improvements at this point, it's just a nice no API change cleanup PR.

pllim:
Did this break a test? See #10299

PR #10236, opened 2020-04-30
fixed bug in constant model tag

cshanahan1:
I noticed a bug that wasn't being picked up in the tests when writing out constant type models. In the round trip, parameter units weren't preserved because `_parameter_to_value` was not being used in `to_tree`. 

This never caused the asdf roundtrip test to fail, so that might need to be looked into.


nden:
It was not caught  because there's no test initializing the model with a quantity. In the future we should add tests with quantities for all models.

PR #10263, opened 2020-05-02
Add spectral flux equivalency property to SpectralQuantity

eteq:
This is a follow-on from #10185 - in https://github.com/astropy/astropy/pull/10185#pullrequestreview-404383900 I suggested this idea, issued it as a PR to @astrofrog's #10185 branch via astrofrog/astropy#100, but then in https://github.com/astropy/astropy/pull/10185#issuecomment-622614974 @mhvk rightly brought up some questions, so I've re-worked it as a PR against master since #10185 was ready to merge. So this has already gotten some review from @astrofrog, but it would be good to hear what @mhvk thinks.

(I've labeled in `no-changelog-entry-needed` on the theory we might get it in for 4.1, but that will have to be revised if it gets pushed to 4.2).

### DIFF
--- a/astropy/coordinates/tests/test_spectral_quantity.py
+++ b/astropy/coordinates/tests/test_spectral_quantity.py
@@ -242,3 +242,12 @@ class TestSpectralQuantity:
         assert isinstance(q1, u.Quantity) and not isinstance(q1, SpectralQuantity)
         assert q1.value == np.sum(sq1.value)
         assert q1.unit == u.AA
+
+    def test_flux_density(self):
+        sq1 = SpectralQuantity([10, 20, 30] * u.AA)
+
+        f_lamb = [1, 1, 1]  * u.erg / u.cm**2 / u.s / u.AA
+
+        to_fnu = f_lamb.to(u.erg / u.cm**2 / u.s / u.Hz,
+                           equivalencies=sq1.spectral_density_equivalency)
+        assert_quantity_allclose(to_fnu, [1, 2**2, 3**2]*(3.33e-17 * u.erg / u.cm**2 / u.s / u.Hz), rtol=.01)
--- a/astropy/coordinates/spectral_quantity.py
+++ b/astropy/coordinates/spectral_quantity.py
@@ -295,7 +295,11 @@ class SpectralQuantity(SpecificTypeQuantity):
     def to_value(self, *args, **kwargs):
         return self.to(*args, **kwargs).value
 
-
     @property
     def spectral_density_equivalency(self):
+        """
+        A `~astropy.units.spectral_density` equivalency for this spectral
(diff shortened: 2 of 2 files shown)


astropy/coordinates/tests/test_spectral_quantity.py
```diff
@@ -242,3 +242,12 @@ def test_functions_std(self):
         assert isinstance(q1, u.Quantity) and not isinstance(q1, SpectralQuantity)
         assert q1.value == np.sum(sq1.value)
         assert q1.unit == u.AA
+
+    def test_flux_density(self):
+        sq1 = SpectralQuantity([10, 20, 30] * u.AA)
+
+        f_lamb = [1, 1, 1]  * u.erg / u.cm**2 / u.s / u.AA
```
pllim, line 249:
```suggestion
        f_lamb = [1, 1, 1]  * (u.erg / u.cm**2 / u.s / u.AA)
```

astropy/coordinates/tests/test_spectral_quantity.py
```diff
@@ -242,3 +242,12 @@ def test_functions_std(self):
...
+
+        to_fnu = f_lamb.to(u.erg / u.cm**2 / u.s / u.Hz,
+                           equivalencies=sq1.spectral_density_equivalency)
+        assert_quantity_allclose(to_fnu, [1, 2**2, 3**2]*(3.33e-17 * u.erg / u.cm**2 / u.s / u.Hz), rtol=.01)
```
pllim, line 253: Nitpick:
```suggestion
        assert_quantity_allclose(to_fnu, [1, 2**2, 3**2] * (3.33e-17 * (u.erg / u.cm**2 / u.s / u.Hz)), rtol=.01)
```

Also, why is the `rtol` so high?

mhvk:
Seeing it again, on its own, I'm still not really convinced, as it is so easy to do separately and would seem to below in `specutils`. On the other hand, arguably this way is more discoverable than via the equivalencies, and of course a large advantage of `SpectralQuantity` is already that one does not have to know about equivalencies. The two more or less balance out to a +0.1. If we go ahead, though, I think it should be documented briefly in the new page on spectral coordinates. (Since `SpectralCoord` will inherit it, the example can use that.)

mhvk:
Removing the milestone as it is not urgent.

nstarman:
> arguably this way is more discoverable than via the equivalencies, and of course a large advantage of `SpectralQuantity` is already that one does not have to know about equivalencies

I agree. Having relevant equivalencies easily accessible seems a nice feature. Something similar for dimensionless angles might be worthwhile on ``Angle`` and for distances in https://github.com/astropy/astropy/issues/11776.

PR #10266, opened 2020-05-02
Fix asdf minversion in install doc

pllim:
### Description

This pull request is to address outdated minversion of `asdf` documented in install doc. Noticed this while working on #10257.


pllim:
Rebased. Thanks for review!

Issue #6509, opened 2017-09-05
`NaN` equivalent for `astropy.time.Time`


MaxNoe:
Is there an equivalent of a `NaN` for `astropy.time.Time`? Like pandas `NaT` (Not a Time)?

pllim:
`not isinstance(var, astropy.time.Time)`?

MaxNoe:
Wat? I think you are missing the point.

It should be an instance of astropy.time.Time! Just like `np.nan` is an instance of `float`.
An instance conveying the meaning that the time is not valid.

pllim:
Not a direct answer to your question but currently `NaT` becomes a masked entry in `Table` (you probably already knows this but I am stating it here for others' future reference):
```python
>>> import pandas as pd
>>> from astropy.table import Table
>>> s = pd.to_datetime(pd.Series(['Jul 31, 2009', '2010-01-10', None]))
>>> df = pd.DataFrame(data={'time': s})
>>> df
        time
0 2009-07-31
1 2010-01-10
2        NaT
>>> t = Table.from_pandas(df)
>>> t
<Table masked=True length=3>
             time            
        datetime64[ns]       
-----------------------------
2009-07-31T00:00:00.000000000
2010-01-10T00:00:00.000000000
                           --
```
p.s. `Series` values taken from https://pandas.pydata.org/pandas-docs/stable/timeseries.html

taldcroft:
@MaxNoe - there is no `NaT`-equivalent in astropy `Time` at this moment, but there is a mostly-complete PR #6028 which adds support.  This allows for masked Time arrays much like `MaskedColumn` in `Table` or `MaskedArray` in numpy.  It also lets you do all the table operations like join or stacking that can generate missing values.

So unfortunately there is no good news for the current 2.0 or even master, but this should definitely be in the 3.0 release at the end of this year.

MaxNoe:
@taldcroft While the PR you mentioned enables support for masking array time values, there is still now support for a scalar time value that signals missing data or invalid values.

We just ran into this again. Why can we not simply allow `np.nan` as value for the time attributes?
Why explicitly fail?

Missing values occur all the time (no pun intended) and everywhere in the real world. Prohibiting this seems very unreasonable to me.

taldcroft:
@MaxNoe - can you give some examples of how you would like to use this?  I'm not entirely clear on your use case so examples would help.

MaxNoe:
[...]

PR #10272, opened 2020-05-02
Add support for 0d NaN / NaT Times

maxnoe:
# Description

For further discussion relevant to #6509, I here add test cases for possible `nan` / `NaT` values for `astropy.time` objects.

I have also started to look into implementing this.

### DIFF
--- a/astropy/time/tests/test_nan.py
+++ b/astropy/time/tests/test_nan.py
@@ -1,0 +1,98 @@
+import numpy as np
+from astropy.time import Time, TimeDelta
+import astropy.units as u
+import pytest
+
+
+FLOAT_FORMATS = [
+    'byear',
+    'cxcsec',
+    'decimalyear',
+    'gps',
+    'jd',
+    'jyear',
+    'mjd',
+    'plot_date',
+    'stardate',
+    'unix',
+    'unix_tai',
+]
+
+STRING_FORMATS = {
+    'byear_str': 'B1950.0',
+    'jyear_str': 'J2000.0',
+    'iso': '2020-01-01 20:00:00.5',
+    'isot': '2020-01-01T20:00:00.5',
+    'yday': '2020:300:20:00:05',
+}
+
+
+def test_time_creation():
+    '''Test a time scalar can be created as nan'''
+    t = Time(np.nan, format='jd')  # for now, in principle, it shouldn't matter
+    assert np.isnan(t)
+    assert np.isnat(t)
+
+    ts = Time([0, 1, np.nan, 2, 3], format='jd')
+    assert np.all(np.isnan(ts) == [False, False, True, False, False])
+
+
+def test_timedelta_creation():
+    '''Test a timedelta scalar can be created as nan'''
+    t = TimeDelta(np.nan)
+
+    t = TimeDelta(np.nan * u.s)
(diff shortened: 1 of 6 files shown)


astropy/time/tests/test_nan.py
```diff
@@ -0,0 +1,101 @@
...
+    'iso': '2020-01-01 20:00:00.5',
+    'isot': '2020-01-01T20:00:00.5',
+    'yday': '2020:300:20:00:05',
+}
+
+
+def test_time_creation():
+    '''Test a time scalar can be created as nan'''
+    t = Time(np.nan, format='jd')  # for now, in principle, it shouldn't matter
+    assert np.isnan(t._time.jd2)
```
taldcroft, line 33: Storing NaN in jd2 is considered an internal implementation detail of masking and could (in theory) change.  So I'd recommend not testing this here.

taldcroft:
I haven't looked much in detail but this is a great idea to define the API with respect to Time missing values and scalars.  You may want to put `[skip ci]` in each commit until tests are passing locally.

mhvk:
General discussion in #6509 - here just a note that if you want `np.isna[nt]` to work on `Time`, you'll need to define an `__array_ufunc__` to tell how it should treat `Time` instances. If you'd also want things like `np.nanmean` to work, one would likely also need to define `__array_function__` (which is planned a bit anyway, for `np.linspace` - see #8610.

All that said, it would be good if all these tests passed! (Except as @taldcroft noted, remove tests on private properties - those should remain implementation details - possibly including how to get `isnan` to work!)

MaxNoe:
@mhvk yes, thanks for the pointers.

This is what I started with, but I have not yet a working version

MaxNoe:
@mhvk I started working on this again. But defining `array_ufunc` is tricky. I defined it only for `isnan` and `isnat` but now stuff starts failing because what used to work, like:

`np.equal(time, 0)` now also calls ufunc.

Do I now have to implement every single numpy ufunc in `Time.__array_ufunc`. There has to be a better way.

mhvk:
@MaxNoe - indeed, every ufunc has to be dealt with. There is a mixin to let all operators go via the same path, but I'm not sure that is too helpful in this case. 

That said, I actually hadn't appreciated any ufuncs worked at all on `Time`! Are there any tests that are actually failing?

But on reflection/inspection, I see what happens: the ufunc creates an array of Time scalars and let those subtract:
```
In [9]: t = Time(np.arange(50000., 50003.), format='mjd')                                                                                                                

In [10]: np.subtract(t, t)                                                                                                                                   

Out[10]: 
array([<TimeDelta object: scale='tai' format='jd' value=0.0>,
       <TimeDelta object: scale='tai' format='jd' value=0.0>,
       <TimeDelta object: scale='tai' format='jd' value=0.0>],
      dtype=object)
```

So, dealing with those will make things better...  (right now, for 10 elements `t==t` is 100 times faster than `np.equal(t, t)`).

In any case, my sense would be to first of all think of ufuncs where we're adding new functionality, and then worry about those that we've broken.

pllim:

pllim:
Sorry for the duplicate bot posts. I fixed the bug upstream, hopefully.

PR #10274, opened 2020-05-02
Fix new failure of spectral coord test because of quantity_input changes

adrn:
Because of two merged PRs (SpectralCoord and changes to quantity_input_) that weren't consistent, travis is now failing. This should fix it!


mhvk:
p.s. We probably should stop matching whole error messages... (though obviously this was an unfortunate coincidence!)

mhvk:
p.s. Thanks for the very quick fix, @adrn!

PR #10290, opened 2020-05-04
Fixed a bug with frame attributes that are coordinates when checking for frame equivalency

### LINKED ISSUE #10287: Changes to SkyCoord equality incompatible with a SkyCoord frame attribute
In sunpy one of the frame attributes we set on some of our frames is an `observer` attribute, which is itself a `SkyCoord`. The recent changes to coordinate equality checking in https://github.com/astropy/astropy/pull/10154 break some functionality however, because `sc1 != sc2` now errors instead of returning `False` if `sc1` is in a different frame to `sc2`.

In particular, a traceback where we are now seeing errors is
```
>       assert_quantity_allclose(implicit.separation_3d(explicit2), 0*u.AU, atol=1e-10*u.AU)

sunpy/coordinates/tests/test_transformations.py:500: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../astropy/astropy/coordinates/sky_coordinate.py:1054: in separation_3d
    if not self.is_equivalent_frame(other):
../astropy/astropy/coordinates/sky_coordinate.py:970: in is_equivalent_frame
    if np.any(getattr(self, fattrnm) != getattr(other, fattrnm)):
../astropy/astropy/coordinates/baseframe.py:1656: in __ne__
    return np.logical_not(self == value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <HeliographicStonyhurst Coordinate (obstime=2019-06-01T00:00:00.000): (lon, lat, radius) in (deg, deg, AU)
    (0., -0.72605048, 1.01396311)>
value = <HeliographicStonyhurst Coordinate (obstime=2019-12-01T00:00:00.000): (lon, lat, radius) in (deg, deg, AU)
    (0., 0.94857499, 0.98618233)>

    def __eq__(self, value):
        """Equality operator for frame.
    
        This implements strict equality and requires that the frames are
        equivalent and that the representation data are exactly equal.
        """
        is_equiv = self.is_equivalent_frame(value)
    
        if self._data is None and value._data is None:
            # For Frame with no data, == compare is same as is_equivalent_frame()
            return is_equiv
    
        if not is_equiv:
>           raise TypeError(f'cannot compare: objects must have equivalent frames: '
                            f'{self.replicate_without_data()} vs. '
                            f'{value.replicate_without_data()}')
E           TypeError: cannot compare: objects must have equivalent frames: <HeliographicStonyhurst Frame (obstime=2019-06-01T00:00:00.000)> vs. <HeliographicStonyhurst Frame (obstime=2019-12-01T00:00:00.000)>
```

Because `is_equivalent_frame`  checks all the attrs in this line:
```
../astropy/astropy/coordinates/sky_coordinate.py:970: in is_equivalent_frame
    if np.any(getattr(self, fattrnm) != getattr(other, fattrnm)):
```
it will check our observer attribute, but now the `!=` call raises an error instead of just returning `False`, when `self` and `other` are not in the same frame.

I'm not entirely sure what the best way to solve this is. One the one hand, I see conceptually why an error is now raised when trying to compare coordinates in two different frames. On the other hand, I'm not sure how to practically solve this new issue, without making `!=` return `True` when the two frames are different.

ayshih:
Fixes #10287 

* `BaseCoordinateFrame`'s version of `is_equivalent_frame()` already had a private function for special handling of representation attributes (which is possibly removable in light of #10154, but I did not touch that), so I added special handling for coordinate attributes.  Now, if the coordinate attributes do not have equivalent frames, it avoids checking for equality (which would raise an error).
* I changed `SkyCoord`'s version of `is_equivalent_frame()` to call the private function under `BaseCoordinateFrame`.

### DIFF
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -1267,18 +1267,21 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
         staticmethod mainly as a convenient location, althought conceivable it
         might be desirable for subclasses to override this behavior.
 
-        Primary purpose is to check for equality of representations, since by
-        default representation equality is only "is it the same object", which
-        is too strict for frame comparisons.
+        Primary purpose is to check for equality of representations.  This
+        aspect can actually be simplified/removed now that representations have
+        equality defined.
 
-        Note: this method may be removed when/if representations have an
-        appropriate equality defined.
+        Secondary purpose is to check for equality of coordinate attributes,
+        which first checks whether they themselves are in equivalent frames
+        before checking for equality in the normal fashion.  This is because
+        checking for equality with non-equivalent frames raises an error.
         """
+        if left_fattr is right_fattr:
+            # shortcut if it's exactly the same object
+            return True
+
         if isinstance(left_fattr, r.BaseRepresentationOrDifferential):
-            if left_fattr is right_fattr:
-                # shortcut if it's exactly the same object
-                return True
-            elif isinstance(right_fattr, r.BaseRepresentationOrDifferential):
+            if isinstance(right_fattr, r.BaseRepresentationOrDifferential):
                 # both are representations.
                 if (getattr(left_fattr, 'differentials', False) or
                         getattr(right_fattr, 'differentials', False)):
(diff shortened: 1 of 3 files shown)


taldcroft:
Milestoned for 4.1 as this fixes a problem in SunPy introduced by #10154.

Issue #10291, opened 2020-05-04
Account for distance properly in aticq and atciqz


StuartLittlefair:
Both of these routines include a call to `erfa.ld` to account for light deflection from the Sun. However, in both routines we ignore any distance information and therefore treat objects in front of the Sun as if they were at large distances.

These routines need updating to fix this problem, and as a result the GCRS-ICRS transforms may need re-working, since the first step in this chain is to drop distance information and treat all objects as if they are at large distances, handling parallax separately.

cc @mhvk @mkbrewer

mhvk:
Thanks for raising this to keep track! We may have to start to try to keep `UnitSpherical` a bit more consistently in place, so that we can go for simpler transformations as needed. Though I guess that is really a separate issue - #10294.

mkbrewer:
I'm not sure what you are getting at here. The only input to ld() that contains a distance is em which you have. p, q and e are all unit vectors. If this is about the difference between p and q, then you should be able to handle that. Taking the example of atciqz(), you are passing it the topocentric position (p), but you also have astrom_eb available to calculate the barycentric position (q). This could be done inside atciqz() or you could modify atciqz() to pass both positions to it. The only issues are that q is actually supposed to be the heliocentic position and the position of the Sun is supposed to be light time delayed to the time that the light ray from the source makes its closest approach. These are really very minor issues however. The Sun is quite close to the solar system barycenter and it doesn't move much in the time that the light ray passes the Sun and arrives at the position of the observer. If you want to be rigorous about this, you'd need to also get the barycentric position of the Sun in order to calculate the correct heliocentric position of the source, but that wouldn't be very difficult either.

mkbrewer:
Oh never mind. I do see your problem. You have dropped the distance before calling atciqz() so you would need to alter it to take both positions. If you are going to do that, you might as well calculate the proper heliocentric position for the source. It might make a difference for Mercury and Venus. It would most certainly make a difference for the Parker Solar Probe.

mhvk:
Seeing @mkbrewer's comments, it might make sense to try to be as rigorous as we can at first (something I feel we've been attempting generally in astropy) and see how much worse in performance that is than the quicker route (which perhaps we can special-case).

mkbrewer:
Just copying this result from issue #10292.

Difference in apparent position of the Moon between my program and astropy without correcting light deflection from the Sun:

[...]

Issue #10239, opened 2020-04-30
Requesting local earth rotation angle and one additional built-in frame


mkbrewer:
Just as I find it frustrating that the Time package provides local sidereal time, but the coordinate package doesn't currently include a built-in frame providing topocentric position with respect to the true equator and equinox of date (issue #10230), I find it equally frustrating that the Time package doesn't provide the local earth rotation angle to complement the coordinate package's built-in CIRS frame. When working with an equatorial mounted telescope both a set of equatorial coordinates and its complementary time or angle are required in order to calculate the hour angle necessary to point the telescope. The Time package should be enhanced to provide the local earth rotation angle. 

In this regard, it would also be nice to provide a built-in frame to transform CIRS coordinates to observed equatorial coordinates to complement the AltAz built-in frame. This should provide both declination and hour angle (or right ascension if the local earth rotation angle is available) adjusted for refraction and polar motion. This would be very simple as the same function that is used by AltAz (atioq()) provides these also.

mkbrewer:
Observed declination and hour angle would probably be best here. These are the direct analog to azimuth and altitude and are independent of the origin of right ascension in use. It is quite easy to derive the observed right ascension in either frame from these. Subtracting the hour angle from the local sidereal time yields the observed right ascension referenced to the true equinox of date, while subtracting the hour angle from the local earth rotation angle yields the observed right ascension referenced to the CIO.

mhvk:
@mkbrewer - rephrasing your request just so that I'm sure I understand them correctly.

1. Is the request to have a method, say, `Time.earth_rotation_angle(longitude)` that uses SOFA's `era00` and adds the longitude? I guess, like `sidereal_time`, it would return the result as a `Longitude` with units of hour-angle. This would be easy to do (and after `sidereal_time`, which I implemented, there will hopefully be less discussion on whether this belongs on `EarthLocation` or on `Time`!). Should there be any arguments beyond `longitude` (which would default to the `location` stored on the time)?

2. And for the coordinate frame, you are thinking of an "HADec" exactly analogous to "AltAz", but for polar mount? As you note, all that would seem to take is replicating `cirs_to_altaz` but catching different outputs in
 https://github.com/astropy/astropy/blob/35085f6d4db8e67b6794012b84e598eb94119212/astropy/coordinates/builtin_frames/cirs_observed_transforms.py#L66
The most difficult part here may be the name...

[...]

Issue #10230, opened 2020-04-29
Requesting two additional built-in frames. 


mkbrewer:
Issue #1: The positions of the bodies are returned in the GCRS and there appears to be no way to easily transform them to topocentric astrometric positions. By this I mean simply the difference in the ICRS position of the body compensated for light time and the ICRS position of the observer. This is rather important if one wishes to place the body on a background map in the ICRS. It is also the only way to compare the output of AstroPy's ephemerides with that of JPL Horizons or Brandon Rhodes' Skyfield. Why isn't there a built-in frame for doing this?

Issue #2: Currently, there is also no built-in frame for transforming the returned positions to classical apparent place. By this I mean the topocentric position with respect to the true equator and equinox of date. This is quite important to people such as myself who still like their origin of right ascension to be an actual location on the sky rather than a convenient mathematical construct. It allows one to point an equatorial mounted telescope using the local sidereal time to calculate the hour angle. And again, this is the only way to compare the output of AstroPy's ephemerides with that of JPL Horizons or Skyfield. It is also quite simple to do. Just adjust the CIRS right ascension by subtracting the equation of the equinoxes. Note: I did find a function for doing this in solar_system.py, _apparent_position_in_true_coordinates(), but it feels sort of kludgy to use this. There should be a built-in frame for this.

mkbrewer:
Oh, I forgot to mention that this is in regard to the solar system ephemerides.

mkbrewer:
I looked into the first one and it's exactly the same as icrs_to_gcrs() without the call to atciqz().

bhazelton:
+1

The frame suggested in #2 would also be extremely helpful for radio astronomers who are used to thinking about source locations in hour angle.

mkbrewer:
Radio astronomer since 1975 here.

mhvk:
> Why isn't there a built-in frame for doing this? 

Probably simply because those who contributed to astropy have not yet needed it. Most of us are volunteers and few work on solar-system bodies (e.g., I'm not sure all transformations are safe for nearby objects).

But we very much appreciate contributions. Indeed, in #10149 we have just added a true equator, mean equinox frame following a similar request - hence, cc @StuartLittlefair. 

From what you write, it sounds like the frames that might be useful for you may not be too difficult to add - though I'll readily admit that I'm not completely sure what exactly you mean. (So far, much of what we have follows the IAU/SOFA documentation; it might help to point to a formal definition.)

mkbrewer:
[...]

Issue #9362, opened 2019-10-14
Improve coordinate transformations for Solar System objects


StuartLittlefair:
[This question](https://stackoverflow.com/questions/58332586/astropy-issue-converting-heliocenric-cartesian-to-ra-dec?noredirect=1) on stack overflow comparing transformations between coordinate systems for asteroids and planets got me thinking that the coordinate transformations astropy assume the motion of the SkyCoord object is negligible.

For example, in the ICRS->GCRS [transform](https://github.com/astropy/astropy/blob/master/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py#L121) we drop any velocity information in the SkyCoord, and use [erfa.atciqz](https://github.com/astropy/astropy/blob/master/cextern/erfa/atciqz.c) instead of [erfa.atciq](https://github.com/astropy/astropy/blob/master/cextern/erfa/atciq.c).

As the question on SO shows, the results can be out by a few degrees for nearby and fast moving objects...

StuartLittlefair:
Thinking about this issue, I'm not sure we can do the "right thing" for fast moving solar system objects. 

What is missing in our transformations is the planetary aberration term, or correcting for the fact that the light that reaches us at `obstime` was actually emitted earlier, when the body had a different position. 

For solar system objects we have ephemerides for we correct for this since we can predict their location at an arbitrary time. For other fast moving objects, even if the user provides instantaneous velocity information we can't correct unless we assume motion in a straight line.

See https://github.com/astropy/astropy/issues/10230#issuecomment-622664451 for the same point being raised in a similar location. Since that is being actively discussed I am flagging this as a duplicate and closing.

PR #10292, opened 2020-05-04
Improve get_body by using location of observer in light travel time correction

### LINKED ISSUE #10271: Adjust get_body to take into account location on Earth
In https://github.com/astropy/astropy/issues/10230#issuecomment-622655253, @mkbrewer noted that our `solar_system._get_apparent_body_position`, which is used in `get_body`, corrects for travel time to the center of the Earth rather than the actual location of the observer. While in general the resulting differences of a ms is not going to affect any of the locations of the planets at a level one likely cares about, the actual travel time is iterated on until it is less than `10 ns`, which is suggesting better precision. And the [docstring for `get_body`](https://docs.astropy.org/en/latest/api/astropy.coordinates.get_body.html#astropy.coordinates.get_body) also seems to suggest the actual location of the observer is used. cc @StuartLittlefair 

Also, crossref https://github.com/astropy/astropy/issues/10230#issuecomment-622659492, in which @ayshih mentions a sunpy routine that calculates light-travel times for an arbitrary observer.

StuartLittlefair:
### Description

There are a series of issues raised to help improve the utility of the coordinates package for solar system objects. These include #10271, #10239, #10230, #10291 and #9362. This is a minor PR to solve the least of those issues - a correct treatment of light travel time for non-geocentric observers. 

This is a very minor correction (around 0.1m for the ICRS location of the Sun in the test example!), but also a very small change to the code.

Fixes #10271

### DIFF
--- a/astropy/coordinates/solar_system.py
+++ b/astropy/coordinates/solar_system.py
@@ -374,7 +374,7 @@ def get_body_barycentric(body, time, ephemeris=None):
 get_body_barycentric.__doc__ += indent(_EPHEMERIS_NOTE)[4:]
 
 
-def _get_apparent_body_position(body, time, ephemeris, observer_loc):
+def _get_apparent_body_position(body, time, ephemeris, obsgeoloc=None):
     """Calculate the apparent position of body ``body`` relative to Earth.
 
     This corrects for the light-travel time to the object.
@@ -390,7 +390,7 @@ def _get_apparent_body_position(body, time, ephemeris, observer_loc):
     ephemeris : str, optional
         Ephemeris to use.  By default, use the one set with
         ``~astropy.coordinates.solar_system_ephemeris.set``
-    observer_loc : `~astropy.coordinates.CartesianRepresentation`, optional
+    obsgeoloc : `~astropy.coordinates.CartesianRepresentation`, optional
         The GCRS position of the observer
     Returns
     -------
@@ -410,8 +410,8 @@ def _get_apparent_body_position(body, time, ephemeris, observer_loc):
     emitted_time = time
     light_travel_time = 0. * u.s
     earth_loc = get_body_barycentric('earth', time, ephemeris)
-    if observer_loc is not None:
-        earth_loc += observer_loc
+    if obsgeoloc is not None:
+        earth_loc += obsgeoloc
     while np.any(np.fabs(delta_light_travel_time) > 1.0e-8*u.s):
         body_loc = get_body_barycentric(body, emitted_time, ephemeris)
         earth_distance = (body_loc - earth_loc).norm()
--- a/astropy/coordinates/solar_system.py
+++ b/astropy/coordinates/solar_system.py
@@ -374,7 +374,7 @@ def get_body_barycentric(body, time, ephemeris=None):
 get_body_barycentric.__doc__ += indent(_EPHEMERIS_NOTE)[4:]
 
 
-def _get_apparent_body_position(body, time, ephemeris):
+def _get_apparent_body_position(body, time, ephemeris, observer_loc):
     """Calculate the apparent position of body ``body`` relative to Earth.
 
     This corrects for the light-travel time to the object.
@@ -390,7 +390,8 @@ def _get_apparent_body_position(body, time, ephemeris):
     ephemeris : str, optional
         Ephemeris to use.  By default, use the one set with
         ``~astropy.coordinates.solar_system_ephemeris.set``
-
+    observer_loc : `~astropy.coordinates.CartesianRepresentation`, optional
+        The GCRS position of the observer
     Returns
     -------
     cartesian_position : `~astropy.coordinates.CartesianRepresentation`
@@ -409,6 +410,8 @@ def _get_apparent_body_position(body, time, ephemeris):
     emitted_time = time
     light_travel_time = 0. * u.s
     earth_loc = get_body_barycentric('earth', time, ephemeris)
+    if observer_loc is not None:
+        earth_loc += observer_loc
     while np.any(np.fabs(delta_light_travel_time) > 1.0e-8*u.s):
         body_loc = get_body_barycentric(body, emitted_time, ephemeris)
         earth_distance = (body_loc - earth_loc).norm()
@@ -456,15 +459,17 @@ def get_body(body, time, location=None, ephemeris=None):
     if location is None:
         location = time.location
 
-    cartrep = _get_apparent_body_position(body, time, ephemeris)
-    icrs = ICRS(cartrep)
     if location is not None:
         obsgeoloc, obsgeovel = location.get_gcrs_posvel(time)
-        gcrs = icrs.transform_to(GCRS(obstime=time,
-                                      obsgeoloc=obsgeoloc,
-                                      obsgeovel=obsgeovel))
     else:
-        gcrs = icrs.transform_to(GCRS(obstime=time))
+        obsgeoloc, obsgeovel = None, None
+
+    cartrep = _get_apparent_body_position(body, time, ephemeris, obsgeoloc)
+    icrs = ICRS(cartrep)
+    gcrs = icrs.transform_to(GCRS(obstime=time,
+                                  obsgeoloc=obsgeoloc,
+                                  obsgeovel=obsgeovel))
+
     return SkyCoord(gcrs)
 
 
--- a/astropy/coordinates/tests/test_solar_system.py
+++ b/astropy/coordinates/tests/test_solar_system.py
@@ -431,4 +431,4 @@ def test_regression_10271():
     icrs_sun_from_geocentre = _get_apparent_body_position('sun', t, 'builtin', geocentre)
 
     difference = (icrs_sun_from_alma - icrs_sun_from_geocentre).norm()
-    assert_quantity_allclose(difference, 0.13046941*u.m)
+    assert_quantity_allclose(difference, 0.13046941*u.m, atol=1*u.mm)
--- a/astropy/coordinates/tests/test_solar_system.py
+++ b/astropy/coordinates/tests/test_solar_system.py
@@ -10,9 +10,11 @@ from astropy.constants import c
 from astropy.coordinates.builtin_frames import GCRS
 from astropy.coordinates.earth import EarthLocation
 from astropy.coordinates.sky_coordinate import SkyCoord
+from astropy.coordinates.representation import CartesianRepresentation
 from astropy.coordinates.solar_system import (get_body, get_moon, BODY_NAME_TO_KERNEL_SPEC,
-                            _apparent_position_in_true_coordinates,
-                            get_body_barycentric, get_body_barycentric_posvel)
+                                              _apparent_position_in_true_coordinates,
+                                              _get_apparent_body_position,
+                                              get_body_barycentric, get_body_barycentric_posvel)
 from astropy.coordinates.funcs import get_sun
 from astropy.tests.helper import assert_quantity_allclose
 from astropy.units import allclose as quantity_allclose
@@ -417,3 +419,16 @@ def test_file_ephemeris_wrong_input():
     # Try loading a file that does exist, but is not an ephemeris file:
     with pytest.raises(ValueError):
         get_body('earth', time, ephemeris=__file__)
+
+
+def test_regression_10271():
+    t = Time(58973.534052125986, format='mjd')
+    # GCRS position of ALMA at this time
+    obs_p = CartesianRepresentation(5724535.74068625, -1311071.58985697, -2492738.93017009, u.m)
+
+    geocentre = CartesianRepresentation(0, 0, 0, u.m)
+    icrs_sun_from_alma = _get_apparent_body_position('sun', t, 'builtin', obs_p)
+    icrs_sun_from_geocentre = _get_apparent_body_position('sun', t, 'builtin', geocentre)
+
+    difference = (icrs_sun_from_alma - icrs_sun_from_geocentre).norm()
+    assert_quantity_allclose(difference, 0.13046941*u.m)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -501,6 +501,10 @@ astropy.coordinates
   ``*TrueEcliptic`` coordinate frames. The error would primarily result in an
   inaccuracy in the ecliptic latitude on the order of arcseconds. [#10129]
 
+- Fixed an error in the computation of the location of solar system bodies where the
+  Earth location of the observer was ignored during the correction for light travel
+  time. [#10292]
+
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 


astropy/coordinates/solar_system.py
```diff
@@ -374,7 +374,7 @@ def get_body_barycentric(body, time, ephemeris=None):
 get_body_barycentric.__doc__ += indent(_EPHEMERIS_NOTE)[4:]
 
 
-def _get_apparent_body_position(body, time, ephemeris):
+def _get_apparent_body_position(body, time, ephemeris, observer_loc):
```
mhvk, line 377: I'd just call it `obsgeoloc=None` as you do in the non-private function (and add a default; even though this is private, it may be used and we might as well not break such uses)

astropy/coordinates/tests/test_solar_system.py
```diff
@@ -417,3 +419,16 @@ def test_file_ephemeris_wrong_input():
...
+    t = Time(58973.534052125986, format='mjd')
+    # GCRS position of ALMA at this time
+    obs_p = CartesianRepresentation(5724535.74068625, -1311071.58985697, -2492738.93017009, u.m)
+
+    geocentre = CartesianRepresentation(0, 0, 0, u.m)
+    icrs_sun_from_alma = _get_apparent_body_position('sun', t, 'builtin', obs_p)
+    icrs_sun_from_geocentre = _get_apparent_body_position('sun', t, 'builtin', geocentre)
+
+    difference = (icrs_sun_from_alma - icrs_sun_from_geocentre).norm()
+    assert_quantity_allclose(difference, 0.13046941*u.m)
```
mhvk, line 434: Ah, I just noticed the CI 32-bit failure is real - if you put, say `atol=1*u.mm` here, it should be OK.

mkbrewer:
Just letting you know. The Sun isn't a very good test for this. It's much better to use the Moon. I'm currently seeing +/-200 mas error in RA/Dec and +/-500 meters in distance over the course of a day using an older version of astropy.

mhvk:
@mkbrewer - you're right, it would probably be good to have a test for the Moon. Might you be able to make a PR?

Independently, thanks again for pointing out these problems!

mkbrewer:
Since I haven't looked into your test procedures, I'm not sure what you would want. What I did was simply compare the Moon's position in RA/DEC, distance and Alt/Az  in one minute intervals for 24 hours with my own solar system ephemeris program based on SOFA. I tested the output of my program for this period on the Moon against both Skyfield and JPL Horizons and got agreement to 0.5 mas in Declination, 1.5 mas in Right Ascension and 0.7 meters in distance against Horizons and similar results against Skyfield so long as I disabled polar motion. Skyfield seems to have a problem with that, but it's only about 12 mas in Declination and 4 meters in distance. As does SOFA, btw, but it only affects Alt/Az. That error is a rather huge at +/-1 arcsecond! (It affects astropy as well. If you are interested, please see the issue I opened on the Skyfield repository). Horizons also has an issue with apparent Right Ascension, being 53 mas off from SOFA. I pointed this out to them last September and, of course, they pointed their fingers at SOFA. At least they are now acknowledging the issue in their print out though. Of course, these errors are all very small compared to the errors I got testing the older version of astropy. 

I could certainly provide you with my test data and the script I used to generate the astropy data to compare against it if that would be helpful.

StuartLittlefair:
@mkbrewer that would be fantastic, thank you! 

Perhaps open a new issue describing the Moon problems and provide the data there?

mkbrewer:
[astropy_tests.zip](https://github.com/astropy/astropy/files/4600206/astropy_tests.zip)
I was just informing you of the issues you might have comparing astropy Moon data to Skyfield or Horizons. There may be a problem with polar motion in SOFA that affects astropy, but I'm still looking into that. I will certainly raise an issue on that if I determine that there is, in fact, a problem, but I'm not there yet.

The file I attached contains a script for printing topocentric astrometric, apparent postions wrt the true equator and equinox, distance and alt/az that can be directly compared to equivalent Skyfield and Horizons output if you so choose. I also included the output from my program that you can compare to output from the script for astropy both before and after the modification in this merge to see how much of an improvement resulted from the change you made. I used the same EOPs (dut1,xp,yp) that astropy used on the  test date in the script for the input to my program, so my data should be completely compatible with astropy in that respect.

mkbrewer:
@StuartLittlefair please let me know if this works for you.

StuartLittlefair:
Thanks @mkbrewer. This helped me a lot dig into what is happening. The short story is that with this PR, the astropy positions for the Moon are pretty damn good! For the long story, see below.

 I used @mkbrewer's scripts together with my additions in this PR to get the following:

```python
from astropy.time import Time
from astropy.coordinates import solar_system_ephemeris, EarthLocation, get_body, get_body_barycentric, SkyCoord
from astropy.coordinates.solar_system import _apparent_position_in_true_coordinates

# For comparison with JPL Horizons as this is what they use
solar_system_ephemeris.set('de430')

class_lat = -22.959748*u.deg
class_lon = -67.787260*u.deg
class_elev = 5186*u.m
class_loc = EarthLocation.from_geodetic(class_lon, class_lat, class_elev)

t = Time('2017-04-06 00:00', scale='utc')
moon_barycentric = _get_apparent_body_position('moon', t, 'de430', 
                                    class_loc.get_gcrs_posvel(t)[0]) 
obs = get_body_barycentric('earth', t, 'de430') + class_loc.get_gcrs_posvel(t)[0]

# ASTROMETRIC ICRF
SkyCoord(coord.GCRS(moon_barycentric - obs, obstime=t)).to_string(style='hmsdms')
'09h05m23.649s +15d36m28.9186s'
```
c.f JPL Horizons "09 05 23.65 +15 36 28.9" and SkyField "09 05 23.64 +15 36 28.5"

What about JPL's True Equator, True Equinox of date ("apparent") coordinates? Let's look at what the result of calling astropy's ```get_body``` is with the changes in this PR:

```python
In [165]: _apparent_position_in_true_coordinates(get_body('moon', t, location=class_loc)).to_string(style='hmsdms')                                                
Out[165]: '09h06m21.1377s +15d32m12.2391s'
```

c.f JPL "09 06 21.13 +15 32 12.2" and SkyField "09 05 24.31 +15 36 25.3". 

You can see there is now excellent agreement with astropy and JPL Horizons for the moon! In fact, we're quite a bit more accurate than Skyfield for this one example. 

This also gives us confidence that implementing the transform between GCRS and TETE requested in #10230 will make it easier for people to use JPL Horizon query results within astropy and keep precision.

pllim:
Re: https://github.com/astropy/astropy/pull/10292#issuecomment-626728598

@StuartLittlefair , so are we expecting a follow-up PR to this? Or is this like a astropy-tutorial material?

mhvk:
Very good news! And particularly nice that it works so well with what really is a small bug fix! Would be nice to have the Moon example as a test case and/or as an example for `get_body` in the documentation.

StuartLittlefair:
I think what I will do is work on #10230 when I get the time. Including lunar test cases and a tutorial about the different frames, how they relate and when to use each might fit well in a PR for that issue.

mkbrewer:
Here are the comparisons I'm seeing from the coordinates that you posted with my own program based on SOFA, JPL Horizons and Skyfield. 

```
                       RA (deg)      Delta RA (mas)     Dec (deg)  Delta Dec(mas)
mkb       astrometric 136.348537482     0.0648        15.608032955    -0.0380
          apparent    136.588073175     2.0700        15.536733269    -0.6684

jpl       astrometric 136.348537637    -0.4932        15.608032907     0.1348
           apparent    136.588058954    53.2656        15.536733166   -0.2976

skyfield  astrometric 136.348537209     1.0476        15.608034242    -4.6712
             apparent 136.588072886     3.1105        15.536734528    -5.2008
```
For my program, the astrometric positions are good to the uas level. The discrepancy in apparent positions is a bit perplexing. I am using the EOPs used by astropy 3.2.2 here. Could they be slightly different?

The JPL results are slightly different due to the fact that they have their own set of EOPs, plus the 53 mas offset in apparent RA. Hint: When comparing with Horizons, it's best to choose decimal degrees for the output with "extra precision" checked.

The Skyfield results do not include polar motion, so the discrepancy there is not surprising. 

@StuartLittlefair could you please run the script In the `astropy_tests.zip` file that I posted above and upload the output as a `.txt` file? I'd like to compare your new positions over a 24 hour period that are generated by that script.

mkbrewer:
@pllim I was responding to @mhvk's request for a PR for testing on the Moon. 

> @mkbrewer - you're right, it would probably be good to have a test for the Moon. Might you be able to make a PR?
> 
> Independently, thanks again for pointing out these problems!

Since I don't know how to submit a PR for tests, that file `astropy_tests.zip` was my response. It contains a script to output Moon data from astropy over a 24 hour period and output from my own program based on SOFA as comparison data..

pllim:
@mkbrewer , to be credited properly for your contribution, it is still best for you to submit the PR yourself. Perhaps you might find https://docs.astropy.org/en/latest/development/workflow/development_workflow.html useful. Thanks!

mkbrewer:
One possibility for the discrepancy in apparent position between astropy and my program is the issue with gravitational light deflection #10291 which I see is still open.

StuartLittlefair:
I doubt it. For the example I looked at it made a negligible difference. 

Is there a reason you don’t want to upgrade astropy btw? Not saying you should but there may other bug fixes you are missing. 

I’ll post your example run on astropy dev as soon as I can. Lots to do and it’s early evening my time.

StuartLittlefair:
Hi @mkbrewer 

I've attached the full output from your code to this post. Here's a snippet. First the file of positions from your SOFA program that you attached above:

```
UT,LST,ICRS_RA,ICRS_DEC,RA,DEC,AZ,EL,DIST,DUT1,XP,YP 
2017-04-06 00:00:00.000000,8.445790414,136.348537301,15.608033005,136.588072995,15.536733318,15.032679644,50.303109080,376252883.251,0.462380,0.009587,0.384603
2017-04-06 00:01:00.000000,8.462502712,136.354077845,15.606277514,136.593609981,15.534971532,14.674616403,50.362682839,376251255.709,0.462379,0.009588,0.384604
```

And what we get on astropy master:

```
UTC,LST,ICRS_RA,ICRS_DEC,RA,DEC,AZ,EL,DIST,DUT1,XP,YP
2017-04-06 00:00,8.445791832,136.348537485,15.608032954,136.588073934,15.536733085,15.032729808,50.303166721,376252883.195,0.463626,0.009587,0.384603
2017-04-06 00:01,8.462504130,136.354078025,15.606277464,136.593610917,15.534971300,14.674666644,50.362740603,376251255.697,0.463625,0.009588,0.384604
```

Note that amongst the differences between my version and yours are discrepancies in LST and UTC-UT1.

Here's the output from the JPL web form for comparison:

```
*******************************************************************************************************************************************
 Date__(UT)__HR:MN     R.A._______(ICRF)_______DEC R.A.__(airless-appar)__DEC. Azi_(a-appr)_Elev            delta      deldot 1-way_down_LT
*******************************************************************************************************************************************
$$SOE
 2017-Apr-06 00:00  m  136.348537637  15.608032907 136.588058954  15.536733166  15.0327  50.3030 3.7625288339E+05  -0.0279606    0.02091741
 2017-Apr-06 01:00  m  136.678112032  15.494425098 136.917408421  15.422751330 352.5663  51.3299 3.7633487126E+05   0.0737328    0.02092197
$$EOE
```

The astropy coordinates I get are 0.5 mas (astrometric) and 52 mas (apparent) different from JPL. 
[moon_topo_astropy_master.txt](https://github.com/astropy/astropy/files/4611398/moon_topo_astropy_master.txt)

mkbrewer:
Thank You. That is very helpful. Here my results:
ra_dec_polar
distance_polar
az_el_sky_polar

The average difference in DUT1 is 250 ns, so that is negligible. The difference in LST is due to the  polar motion correction to longitude from equation (6.20) in USNO Circular 179. I am deriving my LST from equation (6.19) there. The hour angles are supposed to be the same and SOFA doesn't appear to add that correction to the local earth rotation angle, so I added it to LST. This is still under investigation on my part, so I'm not sure it is correct yet. In any case, it has no effect on anything else as I'm not using LST in any of my calculations.  

The discrepancy in apparent position is quite evident. @StuartLittlefair, you say that you looked at an example after correcting the light deflection problem then? By "negligible", do you mean much less than the 2.5 mas I'm seeing? If so, this could be due to the less than rigorous way that SOFA handles diurnal aberration when transforming from ICRS to observed. I am still investigating this and will post further comments about that under issue #10230.

There is a bit of noise in your distance calculation which could very well be due to the 1.0e-8 seconds (3 meters at the speed of light) that you are iterating down to.

There is a fairly large discrepancy in Alt/Az, I believe that this is due to some inconsistency regarding polar motion, but I'm still investigating that. This data will be very helpful in that regard. 

As you can see, this test against the Moon is very good for finding small problems. I will look into converting it into a pass/fail test that I could submit as a PR if I have the time. I'm more interested in actually solving the problems I see, rather than getting credit for a PR containing this test though, so it won't be a high priority for me. In the meantime, if one of you experts want to go ahead and implement this as a test, please feel free to use my script and data to do so.

mkbrewer:
On second thought, the discrepancy in apparent position is unlikely to be due to diurnal aberration as it's not diurnal. Instead there is gradient that appears to be related to the Moon's position wrt the Earth. This makes light deflection a better candidate for this discrepancy.

mkbrewer:
Yes! The discrepancy is due to light deflection. I reverted to using the _**incorrect**_ deflection that SOFA uses in Ldsun() in my program and the discrepancy is reduced by 90%. 

ra_dec_polar_sofa_standard_light_deflection

mhvk:
Thanks for the sleuthing!  And thanks for cross-posting at #10291 - indeed, let's move further discussion there since people will not easily find it here.

brandon-rhodes:
@StuartLittlefair — Could you share the Skyfield code that you used to generate the coordinates that do not quite match the other libraries?

I just wrote a script to see the results myself and try fixing Skyfield:
```
from skyfield.api import Topos, load
ts = load.timescale(builtin=True)
t = ts.utc(2017, 4, 6)
planets = load('de430t.bsp')
topos = Topos('22.959748 S', '67.787260 W', elevation_m=5186.0)
pos = (planets['earth'] + topos).at(t).observe(planets['moon'])
ra, dec, distance = pos.radec()
print('Astrometric  Right ascension {}  Declination {}'.format(ra, dec))
ra, dec, distance = pos.apparent().radec(epoch='date')
print('Apparent     Right ascension {}  Declination {}'.format(ra, dec))
```
But, alas, the output — at least here on my laptop — agrees perfectly with the HORIZONS output and Astropy output that you quote above:
```
Astrometric  Right ascension 09h 05m 23.65s  Declination +15deg 36' 28.9"
Apparent     Right ascension 09h 06m 21.14s  Declination +15deg 32' 12.2"
```
To be more exact: Skyfield here agrees most closely with AstroPy, while we both dissent very slightly from the HORIZONS result with respect to the arcseconds of the apparent RA, where HORIZONS has 21.13 but both Skyfield and AstroPy prefer 21.14.

If you’ll share your code, I'll be happy to improve Skyfield's docs to better help programmers avoid whatever snag your code ran into that hurt its accuracy. Thanks!

StuartLittlefair:
@brandon-rhodes do you have an alert for mentions of Skyfield ;) Thanks very much for following up...

You code matches mine exactly apart from the following differences:

1) I did not specify ```builtin=True``` for the timescale, but I suspect this can't hurt.
2) I used the de421 ephemeris (didn't know about de430t and de430 did not work.
3) I did not specify ```epoch='date'``` for the apparent position.

I suspect the last one explains most of the difference, since the astrometric coordinates are very close. The use of ```radec``` is *very* clearly documented in SkyField, but I was rushing and didn't read the docs! My apologies if I've caused you work as a result...

brandon-rhodes:
> @brandon-rhodes do you have an alert for mentions of Skyfield ;)

Alas, I do not! That’s not a bad idea, though. In this case it was an issue opened by @mkbrewer that mentioned a similar effort to reach agreement between his library and Astropy that led me to cross-reference with his issues over here.

I’m glad you found the docs clear! All I will do, then, is try to make sure that my example code is rigorous in using of-date RA and dec every time it calls `apparent()`, so that whichever example of that call a programmer’s eyes might fall on is likely to include the pattern. Thanks for your quick response!

brandon-rhodes:
As you’ll see in the commit just mentioned, upon review I think some blame might be placed on Skyfield’s `radec()` docs.

Instead of showing how to use `radec()` with real coordinates, the example — in a misguided effort to make the example “simple” and save CPU — instantiated a meaningless `ICRS([1, 2, 3])` vector and then asked its RA and declination. First, as I was reminded recently when responding to another user question, users are far better served by seeing examples where methods are called in realistic situations; many users never instantiate an `ICRS` manually, and so are confused about when they would be using one. Second, if the coordinate is a fake one like `[1, 2, 3]`, then there's no reason to prefer ICRS or epoch-of-date, so the example can’t illustrate the difference.

So instead I’ve pivoted the `radec()` docstring to a full realistic example of asking Skyfield for coordinates, and per best practices generate ICRS for my astrometric but epoch-of-date for my apparent coordinates. Hopefully that will make it a bit more likely that user code makes the same choice except in cases where the user really deliberately wants the idiosyncratic combination of apparent with ICRS.

PR #10295, opened 2020-05-04
Changed size property for BaseCoordinateFrame to return 0 if there is no data

ayshih:
Currently, the `size` property of `BaseCoordinateFrame` errors if the frame has no data.  This PR changes the property to return 0 instead.

The reason that I want this change is because I'd like to carry a data-less frame as a frame attribute.  The frame-attribute machinery checks the size of each attribute to see whether it is broadcast-able to the same size as overarching frame's data.  Thus, the erroring for the size of a data-less frame is problematic.  Rather than adding any special handling to the frame-attribute machinery, it seems cleaner to have `size` return 0 for a data-less frame.

### DIFF
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -705,7 +705,10 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
 
     @property
     def size(self):
-        return self.data.size
+        """
+        Returns the size of the underlying data if it exists, else returns 0.
+        """
+        return self.data.size if self.has_data else 0
 
     @property
     def isscalar(self):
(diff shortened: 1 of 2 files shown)


pllim:
Let's hold off on adding a change log for now until after 4.1 is released. Thanks!

mhvk:
Conceptually, I feel this is problematic: I dont think a frame with zero-sized data - e.g., the result of  `my_frame[:0]` - should be the same as a frame that holds no data.

But what actually goes wrong? I looked at `coordinates.attributes.Attribute` and am unsure why the code there wouldn't work for data-less frames. It suggests an underlying bug...

ayshih:
Ah!  The problem doesn't actually exist on master.  The line in 3.2:
https://github.com/astropy/astropy/blob/b70ce8748b03f60a14c2fe9fdd17b88aa89851e5/astropy/coordinates/baseframe.py#L619
was changed in 4.0 (by you in PR #10113) to:
https://github.com/astropy/astropy/blob/65e918cd13ae557d1ff4e20a72638615296df624/astropy/coordinates/baseframe.py#L618

SunPy still supports 3.2, but I guess the problem will go away once we bump our dependency.  I'll close this PR.

mhvk:
Ah, yes, that was me... Happy to hear it actually solved problems for sunpy too!

PR #10296, opened 2020-05-04
Use pmsafe instead of starpm for applying space motion to SkyCoords

### LINKED ISSUE #10092: Behaviour of apply_space_motion without distance without a distance
### Description

I'm confused about the behaviour of `apply_space_motion` when the `SkyCoord` to which it applies hasn't defined a distance. Just following the example from the documentation, if you remove the distance kwarg you get

```python
>>> c = SkyCoord(l=10*u.degree, b=45*u.degree, distance=100*u.pc,
                 pm_l_cosb=34*u.mas/u.yr, pm_b=-117*u.mas/u.yr,
                 frame='galactic',
                 obstime=Time('1988-12-18 05:11:23.5'))

>>> c.apply_space_motion(dt=10. * u.year) 
WARNING: ErfaWarning: ERFA function "starpm" yielded 1 of "binary logical OR of the above warnings" [astropy._erfa.core]
Out[77]:
<SkyCoord (Galactic): (l, b, distance) in (deg, deg, pc)
    (10., 45., 10000000.)
 (pm_l_cosb, pm_b, radial_velocity) in (mas / yr, mas / yr, km / s)
    (0., 0., 0.)>
```

which produces a very confusing result in which the resulting coordinates are located at a distance of `10000000 pc` and the proper motions are zero, but the original coordinates have not changed.

I would expect that if the parallax is zero, only the proper motion would be applies (i.e., the equivalent of assuming the object is at an infinite distance). Am I missing something?

StuartLittlefair:
### Description

This pull request fixes `SkyCoord.apply_space_motion` for the case where no distance is present on the `SkyCoord` and the proper motion is relatively high.  

Fixes #10092


StuartLittlefair:
Thanks @mhvk. Really useful comments, as per usual!

astropy/coordinates/sky_coordinate.py
```diff
@@ -713,26 +713,33 @@ def apply_space_motion(self, new_obstime=None, dt=None):
         icrsrep = self.icrs.represent_as(SphericalRepresentation, SphericalDifferential)
         icrsvel = icrsrep.differentials['s']
 
+        override_plx = False
```
eteq, line 716: (very minor)
```suggestion
        parallax_zero = False
```

astropy/coordinates/sky_coordinate.py
```diff
@@ -713,26 +713,33 @@ def apply_space_motion(self, new_obstime=None, dt=None):
...
         except u.UnitConversionError:  # No RV
             rv = 0.
 
-        starpm = erfa.starpm(icrsrep.lon.radian, icrsrep.lat.radian,
+        starpm = erfa.pmsafe(icrsrep.lon.radian, icrsrep.lat.radian,
                              icrsvel.d_lon.to_value(u.radian/u.yr),
                              icrsvel.d_lat.to_value(u.radian/u.yr),
                              plx, rv, t1.jd1, t1.jd2, t2.jd1, t2.jd2)
 
+        if override_plx:
```
eteq, line 733:
```suggestion
        if parallax_zero:
```

astropy/coordinates/sky_coordinate.py
```diff
@@ -713,26 +713,33 @@ def apply_space_motion(self, new_obstime=None, dt=None):
         icrsrep = self.icrs.represent_as(SphericalRepresentation, SphericalDifferential)
         icrsvel = icrsrep.differentials['s']
 
+        override_plx = False
         try:
             plx = icrsrep.distance.to_value(u.arcsecond, u.parallax())
         except u.UnitConversionError:  # No distance: set to 0 by convention
             plx = 0.
+            override_plx = True
```
eteq, line 721:
```suggestion
            parallax_zero = True
```

pllim:
Note: A rebase will be needed to get rid of that merge commit. FYI.

StuartLittlefair:
@pllim I think the circleci failure is unrelated...

PR #10306, opened 2020-05-05
Fix fast_reader parsing of RDB headers with names specified and performance issues

dhomeier:
#### for 4.2
Proposed fix for a `fast_reader` bug with `format='rdb'` identified in #9939 that occurs however unrelated to duplicate column names, whenever any of the `names`, `include_names` or `exclude_names` arguments are used. E.g.
```
>>> lines = ['a_0\ta_1\ta_2', 'N\tN\tN', '1\t2\t3', '10\t20\t30']
>>> ascii.read(lines, fast_reader='force', include_names=['a_1', 'a_2'])
<Table length=3>
a_1  a_2 
str2 str2
---- ----
   N    N
   2    3
  20   30
>>> ascii.read(lines, fast_reader='force', include_names=['a_1', 'a_2'], format='rdb')
  File "/sw/lib/python3.8/site-packages/astropy/io/ascii/ui.py", line 304, in read
    dat = fast_reader_rdr.read(table)
  File "/sw/lib/python3.8/site-packages/astropy/io/ascii/fastbasic.py", line 118, in read
    conversion_info = self._read_header()
  File "/sw/lib/python3.8/site-packages/astropy/io/ascii/fastbasic.py", line 341, in _read_header
    raise core.InconsistentTableError('RDB header mismatch between number of '
astropy.io.ascii.core.InconsistentTableError: RDB header mismatch between number of column names and column types
```

#### `names` fixes
The reason is `rdb` is using a two-line header format with line 2 storing the type information, both of which are parsed using `CParser.read_header` that has applied to the 2nd line the full set of renaming and selection operations intended for the 1st line with the actual column names.

The fix follows the discussion in https://github.com/astropy/astropy/pull/9939#issuecomment-622574610 by always setting the `types` entries from `header_names`, saving custom `names` once set, and disabling the `in|exclude_names` filtering on the 2nd header line.

#### Timings
I have made some performance tests with a 4096-column table (otherwise equivalent to the test `lines` from #9939 with two rows of integer data):

with unique column names - 4.0.1:
`ascii.read(lines, fast_reader='force')`: 230 ms
`ascii.read(lines, fast_reader=False)`: 260 ms
`ascii.read(lines, fast_reader=False, names=alt_names)`: 35 s

master with #9939 merged:
`ascii.read(lines, fast_reader='force')`: 240 ms
`ascii.read(lines, fast_reader=False)`: 7 s
`ascii.read(lines, fast_reader=False, names=alt_names)`: 17 s
with duplicate names:
`ascii.read(lines, fast_reader=False)`: 9 s

`fast_reader` fixes from this PR:
`ascii.read(lines, fast_reader='force')`: 245 ms
`ascii.read(lines, fast_reader=False)`: 7 s
`ascii.read(lines, fast_reader='force', names=alt_names)`: 316 ms
with duplicate names:
`ascii.read(lines, fast_reader='force')`: 2.7 s
`ascii.read(lines, fast_reader='force', names=alt_names)`: 320 ms

Timings for `include_names` range from ca. 40 ms for up to 2 % of columns selected, to about the same times as with `names` for most of the columns selected.

The primary motivation for these timings were to check if the additional header reading in the fix impacted performance (which it does at the ~1 % level), but it also revealed some changes with  the Python parser:

1. The column renaming (for `names=alt_names`) has significantly gained in performance thanks to the new `rename_columns` implementation in #9939.

2. The standard parsing has nonetheless taken a 20x performance hit over the `fast_reader`.

#### Additional performance tuning
edec8f7 was an attempt to address some of the performance penalties in constructing unique names for `CParser`:

1. by using its own instance of `_deduplicate_names` pulled into `cparser.pyx`; this does not add any string operations on the C level, but since it's giving about 30 % speedup basically for free, it's perhaps worth the slight increase in code complexity.

2. skipping the check for duplicate names entirely if `names` have been passed as input argument, since in that case `header_names` will never be used anyway.

#### Checks and tests for invalid `names` arguments
In the 2nd case above it is of course the user's responsibility to provide unique column names, but invalid `names` have raised different exceptions in various different parts of the code. 4334b08 adds checks directly on calling `ascii.read` both for the Python and fast readers and ensures consistent exceptions with those raised in `Table` plus tests. With the default settings with `guess` enabled the exceptions still tend to be buried among the output on all the different formats tried, but they should at least be a bit easier to identify.

#### Performance fixes for Python parser
The reason for the dramatic slowdown on regular input (with valid, unique column names) turned out to be the check for `colnames_uniq != table.colnames`; since the former is a list and the latter a tuple at that point, the comparison evaluated always `False` enforcing a call to `rename_columns`.
In addition, with `names` specified, `table.colnames` would first be renamed to `colnames_uniq` and then to `names` from input. Changed analogously to the handling in `CParser`. This improves the timings to
unique column names:
`ascii.read(lines, fast_reader=False)`: 280 ms
`ascii.read(lines, fast_reader=False, names=alt_names)`: 5.8 s
duplicate names:
`ascii.read(lines, fast_reader=False)`: 9 s
`ascii.read(lines, fast_reader=False, names=alt_names)`: 6.5 s

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -171,6 +171,11 @@ astropy.extern
 astropy.io.ascii
 ^^^^^^^^^^^^^^^^
 
+- Fix failure of ASCII ``fast_reader`` to handle ``names``, ``include_names``,
+  ``exclude_names`` arguments for ``RDB`` formatted tables; homogenised checks
+  and exceptions for invalif ``names`` arguments; improved performance when
+  parsing tables with duplicate column names. [#10306]
+
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -171,11 +171,6 @@ astropy.extern
 astropy.io.ascii
 ^^^^^^^^^^^^^^^^
 
-- Fix failure of ASCII ``fast_reader`` to handle ``names``, ``include_names``,
-  ``exclude_names`` arguments for ``RDB`` formatted tables; homogenised checks
-  and exceptions for invalif ``names`` arguments; improved performance when
-  parsing tables with duplicate column names. [#10306]
-
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
@@ -295,6 +290,11 @@ astropy.io.ascii
 - Functional Units can now be processed in CDS-tables. [#9971]
 
 - Allow reading in ASCII tables which have duplicate column names. [#9939]
+
+- Fixed failure of ASCII ``fast_reader`` to handle ``names``, ``include_names``,
+  ``exclude_names`` arguments for ``RDB`` formatted tables. Homogenised checks
+  and exceptions for invalid ``names`` arguments. Improved performance when
+  parsing "wide" tables with many columns. [#10306]
 
 - Added type validation of key arguments in calls to ``io.ascii.read()`` and
   ``io.ascii.write()`` functions. [#10005]
--- a/astropy/io/ascii/core.py
+++ b/astropy/io/ascii/core.py
@@ -1039,11 +1039,11 @@ def _deduplicate_names(names):
     existing_names = set()
 
     for name in names:
-        orig_name = name
+        base_name = name + '_'
         i = 1
         while name in existing_names:
             # Iterate until a unique name is found
-            name = orig_name + '_' + str(i)
+            name = base_name + str(i)
             i += 1
         new_names.append(name)
         existing_names.add(name)
@@ -1154,12 +1154,12 @@ def _apply_include_exclude_names(table, names, include_names, exclude_names):
         for ii, name in enumerate(names):
             table.rename_column(xxxs + str(ii), name)
 
-    colnames_uniq = _deduplicate_names(table.colnames)
-    if colnames_uniq != table.colnames:
-        rename_columns(table, colnames_uniq)
-
     if names is not None:
         rename_columns(table, names)
+    else:
+        colnames_uniq = _deduplicate_names(table.colnames)
+        if colnames_uniq != list(table.colnames):
+            rename_columns(table, colnames_uniq)
 
     names_set = set(table.colnames)
 
--- a/astropy/io/ascii/core.py
+++ b/astropy/io/ascii/core.py
@@ -1556,6 +1556,10 @@ def _get_reader(Reader, Inputter=None, Outputter=None, **kwargs):
         reader.header.splitter = kwargs['header_Splitter']()
     if 'names' in kwargs:
         reader.names = kwargs['names']
+        if None in reader.names:
+            raise TypeError('Cannot have None for column name')
+        if len(set(reader.names)) != len(reader.names):
+            raise ValueError('Duplicate column names')
     if 'include_names' in kwargs:
         reader.include_names = kwargs['include_names']
     if 'exclude_names' in kwargs:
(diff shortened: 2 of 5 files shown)


pllim:
Thanks! You can hold off on adding a change log until after 4.1 is released.

Does this need new benchmarks at https://github.com/astropy/astropy-benchmarks ?

dhomeier:
> Does this need new benchmarks at https://github.com/astropy/astropy-benchmarks ?

I had only noticed that repo in reviewing `CONTRIBUTING.md` right before this PR, but never added anything there. I think there are some benchmarks that could be added, will look into that once I have checked how those benchmarks are organised and executed.

pllim:
It was more of a question for the subpackage maintainers, but feel free to look anyway. Thanks! 😄

dhomeier:
Well I had been wondering if any of the `io.ascii` benchmarks from https://github.com/astropy/astropy-notebooks had made it there, so I am certainly going to have a closer look.

hamogu:
While I'm the deputy sub-package maintainer, I can't say that I have more than a vague memory of hearing about astropy benchmarks a few years ago. I certainly did not know where they live, how they work, or what is being tested there, so please don't expect input from me @dhomeier ;-)

dhomeier:
Rebased without [ci skip] to get tests with the ~400 commits added to master since.

taldcroft:
@dhomeier - sorry.. the title of this made me think the scope was basically the fast RDB parsing, which as you know I consider to be a low priority so I didn't have a look until now. (In fact if you had not spent so much work on this I might consider just dropping RDB support entirely from the fast reader - we are thinking about *reducing* some functionality for 4.2 to simplify code and reduce "technical debt").

On the other hand, the performance regression and bug you found in #9939 is critical and urgent.  Great catch! A separate PR or issue like "Serious performance regression from duplicate column handling" would have gotten a response 16 days ago. My bad, but I have many things going on and have to prioritize.

So... @bsipocz @pllim - what's happening with the 4.1 release? Can we get a bug fix / performance regression fix in for 4.1?  This can be literally changing `table.colnames` to `list(table.colnames)` in one line. This impacts performance for every call to the Python reader. This could go in a 4.1.1 release of course.

taldcroft:
About the benchmarking, see: https://www.astropy.org/astropy-benchmarks/#/

There are no obvious regressions there, so clearly we need to add tests and then pay attention (or have some alerts sent out for new regressions).

taldcroft:
A simple thing now is just pulling out 4362f199f into a separate performance-regression-fix PR.

pllim:
> what's happening with the 4.1 release?

I am not sure. Maybe @astrofrog knows.

astrofrog:
The release hasn't happened yet, so we can still backport bug/performance fixes

taldcroft:
@astrofrog - OK.  I'll actually review this whole PR which is a bugfix + performance fix. If there are issues that can't be resolved quickly then I'll just do that one commit.

astrofrog:
@taldcroft - I'm going to try and push ahead with the RC today, how soon can you review this? :grin:

astrofrog:
Note that the changelog entry will need to be moved to the v4.1 section

taldcroft:
See: https://github.com/astropy/astropy-benchmarks/issues/93

astrofrog:
I think the 32-bit failure can be ignored and has been fixed in master in the mean time

dhomeier:
Thanks all for getting this speedily through the final review. Perhaps I should have marked the performance impact more prominently, but I thought there was consensus not to let this interfere with the feature freeze for 4.1, which seemed stressful enough as it was. In any case it did not return wrong results, and only hit when `fast_reader=False` was explicitly selected or needed to handle duplicate columns or `*names`.
But 4362f19 would have only helped with the specific case where custom `names` have been provided, so I am glad it made it in as a whole fix!

Issue #10310, opened 2020-05-05
TST: Why so many conftest.py files?


pllim:
### Description

I see many copies of `conftest.py` files buried in the package. The most troubling are `astropy/conftest.py` and `conftest.py` that contain duplicated code.

1. `astropy/coordinates/tests/conftest.py`
2. `astropy/table/tests/conftest.py`
3. `astropy/wcs/tests/conftest.py`
4. `astropy/wcs/wcsapi/conftest.py`
5. `astropy/conftest.py`
6. `docs/conftest.py`
7. `conftest.py`

### Expected behavior

One and only one `conftest.py` at root directory.

### Proposed solution

* Move stuff from lower level `conftest.py` files to the root level one. Eliminated duplicate code.
* Move subpackage specfici test helper functions into `subpackage/tests/utils.py` if they are not needed at the root `conftest.py` level.

### Need to know

* Is there a reason why proposed solution above wouldn't work?

Cadair:
Having lower level conftest files which provide subpackage specific pytest fixtures is the recommended way of doing it. [ref](https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions)

I agree it can be a little magic as you have to look for conftest files to see where your fixtures are coming from, but I think it would cause more issues to try and work around pytest (pytest is sufficiently magic you are better off accepting the magic and becoming one with it than fighting against it imo).

I do wonder if the very top most level `conftest.py` file should be removed, I know pytest has been deprecating stuff in there because the behaviour is ill defined.

pllim:
Really? I thought it is the other way around? I remember `pytest` chastising me in another package for having lower level `conftest.py` and implored me to use root-level one.

Besides, we need the root-level one to run `pytest ...` from root level during dev testing locally.

Cadair:
:shrug:

pllim:
I think @astrofrog said something about the test runner. Ah well.

PR #10311, opened 2020-05-05
TST: Disable IERS download for testing

### LINKED ISSUE #10255: Add TEME tests that are not remote_data
#10149 added the TEME frame, and included necessary unit tests.  However, the tests are `remote_data` because that makes sense for the obvious uses cases for this frame (satellites).  It's not a huge priority, but it would be a bit more robust (and make our coverage stats look better) if there were some tests for this frame that didn't require remote data.

cc @StuartLittlefair

pllim:
### Description

This pull request is to follow-up on the idea in https://github.com/astropy/astropy/pull/10304#discussion_r420123116 that we can disable auto-downloading of IERS data completely for testing purpose.

The proof-of-concept part of this is in `docs/coordinates/satellites.rst` in that this fixes #10255 albeit slight changes in expected results. (This builds upon #10304 to disable `of_site` download.)

This also moves stuff from `astropy/conftest.py` up to root level `conftest.py`, as stated in #10310. 

~I took advantage of a set but unused `builtins._pytest_running` variable I saw in `astropy/conftest.py`.~ If people are interested to disable it outside of testing environment, say, for computing clusters or multiprocessing, we might need to resort to setting system environment variable instead though I am not sure if even that would solve the problem for everyone because it solves the download problem but not if you still need to download something but has race condition in caching.

**TODO**

- [x] Are we really doing this?
- [x] Fix failing tests -- most of them will warn about something is out of range and need to be handled accordingly
- [x] Not sure how to identify the rest of the tests that no longer need really need remote data that are not failing. (Battle for another day.)

Bonus:

* Minor PEP 8 fixes.
* Address some of #10344 where I see them.
* Use `pathlib` to handle file URIs.
* Attempt to fix flaky `test_positions_skyfield` failing CI.

### DIFF
--- a/astropy/coordinates/tests/test_solar_system.py
+++ b/astropy/coordinates/tests/test_solar_system.py
@@ -55,13 +55,13 @@ def test_positions_skyfield(tmpdir):
     # skyfield ephemeris
     try:
         planets = load('de421.bsp')
+        ts = load.timescale()
     except OSError as e:
         if os.environ.get('CI', False) and 'timed out' in str(e):
             pytest.xfail('Timed out in CI')
         else:
             raise
 
-    ts = load.timescale()
     mercury, jupiter, moon = planets['mercury'], planets['jupiter barycenter'], planets['moon']
     earth = planets['earth']
 
--- a/astropy/conftest.py
+++ b/astropy/conftest.py
@@ -48,6 +48,11 @@ def ignore_matplotlibrc():
 
 
 def pytest_configure(config):
+    from astropy.utils.iers import conf as iers_conf
+
+    # Disable IERS auto download for testing
+    iers_conf.auto_download = False
+
     builtins._pytest_running = True
     # do not assign to matplotlibrc_cache in function scope
     if HAS_MATPLOTLIB:
@@ -78,6 +83,11 @@ def pytest_configure(config):
 
 
 def pytest_unconfigure(config):
+    from astropy.utils.iers import conf as iers_conf
+
+    # Undo IERS auto download setting for testing
+    iers_conf.reset('auto_download')
+
     builtins._pytest_running = False
     # do not assign to matplotlibrc_cache in function scope
     if HAS_MATPLOTLIB:
--- a/astropy/time/tests/test_ut1.py
+++ b/astropy/time/tests/test_ut1.py
@@ -20,10 +20,36 @@ else:
     HAS_IERS_A = True
 
 
-class TestTimeUT1:
-    """Test Time.ut1 using IERS tables"""
+def do_ut1_prediction_tst(iers_type):
+    tnow = Time.now()
+    iers_tab = iers_type.open()
+    tnow.delta_ut1_utc, status = iers_tab.ut1_utc(tnow, return_status=True)
+    assert status == iers.FROM_IERS_A_PREDICTION
+    tnow_ut1_jd = tnow.ut1.jd
+    assert tnow_ut1_jd != tnow.jd
 
-    @pytest.mark.remote_data
+    delta_ut1_utc = tnow.delta_ut1_utc
+    with iers.earth_orientation_table.set(iers_type.open()):
+        delta2, status2 = tnow.get_delta_ut1_utc(return_status=True)
+        assert status2 == status
+        assert delta2.to_value('s') == delta_ut1_utc
+
+        tnow_ut1 = tnow.ut1
+        assert tnow_ut1._delta_ut1_utc == delta_ut1_utc
+        assert tnow_ut1.jd != tnow.jd
+
+
+@pytest.mark.remote_data
+class TestTimeUT1Remote:
+    def setup_class(cls):
+        # Need auto_download so that IERS_B won't be loaded and cause tests to
+        # fail.
+        iers_conf.auto_download = True
+
+    def teardown_class(cls):
+        # This setting is to be consistent with astropy/conftest.py
+        iers_conf.auto_download = False
+
     def test_utc_to_ut1(self):
         "Test conversion of UTC to UT1, making sure to include a leap second"""
         t = Time(['2012-06-30 12:00:00', '2012-06-30 23:59:59',
@@ -42,6 +68,13 @@ class TestTimeUT1:
         tnow = Time.now()
 
         tnow.ut1
+
+    def test_ut1_iers_auto(self):
+        do_ut1_prediction_tst(iers.IERS_Auto)
+
+
+class TestTimeUT1:
+    """Test Time.ut1 using IERS tables"""
 
     def test_ut1_to_utc(self):
         """Also test the reverse, around the leap second
@@ -73,31 +106,9 @@ class TestTimeUT1:
 
 
 class TestTimeUT1SpecificIERSTable:
-    def do_ut1_prediction_test(self, iers_type):
-        tnow = Time.now()
-        iers_tab = iers_type.open()
-        tnow.delta_ut1_utc, status = iers_tab.ut1_utc(tnow, return_status=True)
-        assert status == iers.FROM_IERS_A_PREDICTION
-        tnow_ut1_jd = tnow.ut1.jd
-        assert tnow_ut1_jd != tnow.jd
-
-        delta_ut1_utc = tnow.delta_ut1_utc
-        with iers.earth_orientation_table.set(iers_type.open()):
-            delta2, status2 = tnow.get_delta_ut1_utc(return_status=True)
-            assert status2 == status
-            assert delta2.to_value('s') == delta_ut1_utc
-
-            tnow_ut1 = tnow.ut1
-            assert tnow_ut1._delta_ut1_utc == delta_ut1_utc
-            assert tnow_ut1.jd != tnow.jd
-
     @pytest.mark.skipif('not HAS_IERS_A')
     def test_ut1_iers_A(self):
-        self.do_ut1_prediction_test(iers.IERS_A)
-
-    @pytest.mark.remote_data
-    def test_ut1_iers_auto(self):
-        self.do_ut1_prediction_test(iers.IERS_Auto)
+        do_ut1_prediction_tst(iers.IERS_A)
 
     def test_ut1_iers_B(self):
         tnow = Time.now()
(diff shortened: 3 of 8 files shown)


mhvk:
Nice find! But would it be possible to go one step further in the pytest configuration in `conftest.py`, and simply set `iers.confi.auto_download=False` there (and reset at the end)? If that works, we do not have to touch `iers.py` at all! Even better if it is possible to only do this for `remote_data!='any'`...

pllim:
Setting `conf` didn't work in `astropy/conftest.py`, and then I got distracted with `builtins`. Let me try it again but at root level `conftest.py`.

mhvk:
I hope this works. Yet another option might be to have a doctest marking `+OFFLINE`, which disables temporarily disables any internet connection. But to be honest that seems even trickier.

pllim:
> Even better if it is possible to only do this for `remote_data!='any'`

That seems like a lot more work, as then the setting needs to be set at a per-test level.

mhvk:
Ah, yes, and the `remote_data` should not really be about testing IERS download anyway.

This does look nice!  But it seems the test failures are real: I guess for some tests we need to explicitly enable downloading now. (Can look later.)

pllim:
> test failures are real

Yes. I added that as part of my todo in the original comment above. But before I spend the time fixing the tests, I need to know if we're really doing this. 😄

mhvk:
Fair enough to fix only once we all agree! FWIW, I'm in favour. Only tests that explicitly check for the updating should be allowed to try to get new versions.

Perhaps the main question is *how* to do it. E.g., an alternative might be to use a configuration file in tests that turns downloading off. On balance, though, I think that what you have is better. Anyway, let's also bring in @astrofrog and @bsipocz...

mhvk:
Looking at this again, I still like the solution. But I do worry about deleting `astropy/conftest.py` - I think that is needed for tests run from within python (via `astropy.test`).

astrofrog:
> Looking at this again, I still like the solution. But I do worry about deleting astropy/conftest.py - I think that is needed for tests run from within python (via astropy.test).

Indeed if we want astropy.test() to work we can't move the ``conftest.py`` file out of the package unfortunately.

pllim:
Okay, I'll reduce the scope when I get the chance. Thanks for the review!

pllim:
If this fails with the settings in `astropy/conftest.py`, which I think it did based on what Past Self said in https://github.com/astropy/astropy/pull/10311#issuecomment-624193711 , will setting another `pytest_configure`/`pytest_unconfigure` in root level `conftest.py` without touching `astropy/conftest.py` the correct way to go?

pllim:
Huh... The error I thought I would get didn't happen. I guess I'll fix the test failures and see what happens next.

pllim:
Note: There are some remote data failures despite all green here.

pllim:
Remote data job passed.

astropy/coordinates/tests/test_solar_system.py
```diff
@@ -55,13 +55,13 @@ def test_positions_skyfield(tmpdir):
     # skyfield ephemeris
     try:
         planets = load('de421.bsp')
+        ts = load.timescale()
```
brandon-rhodes, line 58: It might go against the intention here, but `load.timescale(builtin=True)` will suppress file download and use built-in copies of the leap second files instead.

pllim: I thought about that but I don't know what is really being tested here and the test is already marked as a remote data test. Maybe @adrn or @eteq can clarify.

mhvk: This was there already, just moved up so that if, somehow, `skyfield` tries to access the internet, we'll fail on jobs where we do not allow that. I think it is OK.

Also: the tests here are for specific times in the past, so should be OK not to update the leap-second list!

mhvk:
I'm merging since I think I can remove at least one `remote_data` already in a PR I'm doing!

astropy/coordinates/tests/test_solar_system.py
pllim: @mhvk , did you mean that I should have used `ts = load.timescale(builtin=True)` and not try to ignore its timeout error in CI? Can you please clarify? If so, I can open a follow-up PR.

pllim:
Issue opened at #10457 . Thanks, everyone!

astropy/coordinates/tests/test_solar_system.py
mhvk: No, sorry, I misunderstood here - this remains a remote-date test and what you did is exactly the right thing!

mhvk:
@pllim - I am confused: I tried over in #10337 to remove a `remote_data` check on a test that doesn't use any times in the far future, yet it failed miserably:
https://travis-ci.org/github/astropy/astropy/jobs/695168048

Now the build log is quite hard to parse, but if I do it locally I get a traceback which suggests the config change is not picked up:
```
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/tests/test_shape_manipulation.py:61: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/earth.py:695: in get_gcrs_posvel
    gcrs_data = self.get_gcrs(obstime).data
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/earth.py:675: in get_gcrs
    return itrs.transform_to(GCRS(obstime=obstime))
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/baseframe.py:1220: in transform_to
    return trans(self, new_frame)
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/transformations.py:1392: in __call__
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/transformations.py:928: in __call__
    reprwithoutdiff = supcall(from_diffless, toframe)
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py:107: in itrs_to_cirs
    pmat = cirs_to_itrs_mat(itrs_coo.obstime)
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py:51: in cirs_to_itrs_mat
    xp, yp = get_polar_motion(time)
../../.tox/test/lib/python3.8/site-packages/astropy/coordinates/builtin_frames/utils.py:41: in get_polar_motion
    iers_table = iers.earth_orientation_table.get()
../../.tox/test/lib/python3.8/site-packages/astropy/utils/state.py:40: in get
    return cls.validate(cls._value)
../../.tox/test/lib/python3.8/site-packages/astropy/utils/iers/iers.py:866: in validate
    value = IERS_Auto.open()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'astropy.utils.iers.iers.IERS_Auto'>

    @classmethod
    def open(cls):
        """If the configuration setting ``astropy.utils.iers.conf.auto_download``
        is set to True (default), then open a recent version of the IERS-A
        table with predictions for UT1-UTC and polar motion out to
        approximately one year from now.  If the available version of this file
        is older than ``astropy.utils.iers.conf.auto_max_age`` days old
        (or non-existent) then it will be downloaded over the network and cached.
    
        If the configuration setting ``astropy.utils.iers.conf.auto_download``
        is set to False then ``astropy.utils.iers.IERS()`` is returned.  This
        is normally the IERS-B table that is supplied with astropy.
    
        On the first call in a session, the table will be memoized (in the
        ``iers_table`` class attribute), and further calls to ``open`` will
        return this stored table.
    
        Returns
        -------
        `~astropy.table.QTable` instance with IERS (Earth rotation) data columns
    
        """
        if not conf.auto_download:
            cls.iers_table = IERS_B.open()
            return cls.iers_table
    
        all_urls = (conf.iers_auto_url, conf.iers_auto_url_mirror)
    
        if cls.iers_table is not None:
    
            # If the URL has changed, we need to redownload the file, so we
            # should ignore the internally cached version.
    
            if cls.iers_table.meta.get('data_url') in all_urls:
                return cls.iers_table
    
        try:
            filename = download_file(all_urls[0], sources=all_urls, cache=True)
        except Exception as err:
            # Issue a warning here, perhaps user is offline.  An exception
            # will be raised downstream when actually trying to interpolate
            # predictive values.
>           warn(AstropyWarning(
                f'failed to download {" and ".join(all_urls)}, '
                f'using local IERS-B: {err}'))
E           astropy.utils.exceptions.AstropyWarning: failed to download ftp://anonymous:mail%40astropy.org@gdc.cddis.eosdis.nasa.gov/pub/products/iers/finals2000A.all and https://datacenter.iers.org/data/9/finals2000A.all, using local IERS-B: <urlopen error Unable to open any source! Exceptions were {'ftp://anonymous:mail%40astropy.org@gdc.cddis.eosdis.nasa.gov/pub/products/iers/finals2000A.all': URLError("ftp error: OSError('An attempt was made to connect to the internet by a test that was not marked `remote_data`. The requested host was: 198.118.242.43')"), 'https://datacenter.iers.org/data/9/finals2000A.all': URLError(OSError('An attempt was made to connect to the internet by a test that was not marked `remote_data`. The requested host was: datacenter.iers.org'))}>
```

mhvk:
p.s. For now, I've just put the `remote_data` back on #10337, but I remain very confused! Did we need to do the home `conftest.py` as well after all?

pllim:
@mhvk , I am confused too. I have opened an investigation over at #10469.

astrofrog:
This is needed in LTS so I have updated the milestone and backported it.

PR #10313, opened 2020-05-05
EXP: EarthLocation context manager to force builtin sites

pllim:
### Description

This pull request is to follow-up on the idea at https://github.com/astropy/astropy/pull/10304/files#r420003642 to use context manager to force usage of builtin `sites.json` for testing.

The proof-of-concept part is in `astropy/io/misc/asdf/tags/coordinates/tests/test_earthlocation.py`. You can see how it is more elegant here compared to the implementation over at #10304 without a context manager.

### DIFF
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+from contextlib import contextmanager
 from warnings import warn
 import collections
 import socket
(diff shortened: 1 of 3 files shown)


mhvk:
Could this also be moved to `conftest`? I like that approach very much and am less sure we really need it beyond testing (or, if so, whether one isn't better off defining a configuration item or so); my tendency would be to stick with your `try/finally` otherwise.

pllim:
@mhvk , I only see parent classes and fixtures in `conftest.py` files. Do you mean I need to turn this into a `pytest` fixture? Can you please elaborate? Thanks!

mhvk:
Sorry, I'm being obscure: I meant could one, inside the general `conftest.py` of #10311, also set the sites to `builtin` and then reset them at the end? (But probably best to decide on #10311 overall before worrying about extending it!)

pllim:
I don't remember what I was trying to fix anymore.

Issue #10273, opened 2020-05-02
Issue with custom equivalency involving use of a constant


cosimoNigro:
### Description

I obtain an incorrect conversion when involving `astropy.constants` (h) in a custom equivalency

### Snippet to reproduce
```python
import astropy.units as u
import numpy as np
from astropy.constants import m_e, h

mec2 = m_e.to("erg", equivalencies=u.mass_energy())
epsilon_equivalencies = [
    (u.Hz, u.Unit(""), lambda x: h * x / mec2, lambda x: x * mec2 / h)
]
nu = np.logspace(23, 28, 10) * u.Hz
# dimensionless energy in electron rest mass energy units
epsilon = (h * nu / mec2).to("")
print("manual conversion: ", epsilon)

epsilon = nu.to("", equivalencies=epsilon_equivalencies)
print("equivalencies conversion: ", epsilon)
```
The output reads
```bash
manual conversion:  [8.09329979e+02 2.90858114e+03 1.04528986e+04 3.75657700e+04
 1.35004377e+05 4.85180576e+05 1.74364858e+06 6.26634810e+06
 2.25200874e+07 8.09329979e+07]
equivalencies conversion:  [8.09329979e-05 2.90858114e-04 1.04528986e-03 3.75657700e-03
 1.35004377e-02 4.85180576e-02 1.74364858e-01 6.26634810e-01
 2.25200874e+00 8.09329979e+00]
```
The difference is a factor `1e-7`, which is exactly the factor between `h` in SI and cgs units.
The issue can be avoided by explicitly converting h in cgs inside the equivalency
```python
epsilon_equivalencies = [
    (u.Hz, u.Unit(""), lambda x: h.cgs * x / mec2, lambda x: x * mec2 / h.cgs)
]
```
Why the equivalency cannot handle the simplification correctly?

### System Details

Darwin-19.4.0-x86_64-i386-64bit
Python 3.7.6 (default, Jan  8 2020, 13:42:34) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
Numpy 1.18.1
astropy 4.0.1.post1
Scipy 1.4.1

Have I misunderstood the use of custom equivalencies?
Thank you!

pllim:
I cannot explain why it is this way, but if you look inside

https://github.com/astropy/astropy/blob/8e19d39293261d9fbc1ff8cc173fc95f99c1ccca/astropy/units/equivalencies.py#L105

[...]

PR #10320, opened 2020-05-06
DOC: Clarify that equivalency cannot take Quantity

pllim:
### Description

This pull request is to address https://github.com/astropy/astropy/issues/10273#issuecomment-624514576 .

cc @cosimoNigro

### DIFF
--- a/docs/units/equivalencies.rst
+++ b/docs/units/equivalencies.rst
@@ -466,6 +466,9 @@ elements::
 ``backward`` are functions that convert values between those units. ``forward``
 and ``backward`` are optional, and if omitted such an equivalency simply
 declares that the two units should be taken as equivalent.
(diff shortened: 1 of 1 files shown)


docs/units/equivalencies.rst
```diff
@@ -479,6 +479,13 @@ for them::
   >>> u.l.to(u.kg, 1, equivalencies=liters_water)
   1.0
 
+.. note::
+
+    To avoid infinite recursion of using ``units`` within ``units``
```
mhvk, line 484: I think I'd just add a sentence to the end of l.468: "The functions must take and return non-Quantity, to avoid infinite recursion of using ``units`` within ``units``."

pllim: What about now?

pllim:
Thanks for the review, @mhvk !

PR #10322, opened 2020-05-07
Fix some backwards compatibility and conftest issues in WCS wrappers

Cadair:
This fixes a deprecated import I missed in #10229 and corrects some issues with the fixtures I just noticed.

(This is a bugfix that needs to make it into 4.1 I assume I have milestoned it right?)

### DIFF
--- a/astropy/wcs/wcsapi/conftest.py
+++ b/astropy/wcs/wcsapi/conftest.py
@@ -23,10 +23,11 @@ def spectral_1d_fitswcs():
 def time_1d_fitswcs():
     wcs = WCS(naxis=1)
     wcs.wcs.ctype = 'TIME',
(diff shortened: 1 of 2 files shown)


pllim:
4.1 or 4.1.1, I'll let @bsipocz decide. If this is pushed to 4.1.1, it will need a change log.

nden:
@Cadair [The change in this line](https://github.com/astropy/astropy/pull/10229/files#diff-1feb0db2986994e665778cc3af3782b7R79) actually changed the `repr` of the Sliced WCS object. It broke tests in gwcs. Was this the desired behavior? If so I can fix the tests.

Cadair:
Yeah, I changed it on purpose because generally I am of the opinion that reprs should show what type they are (and what instance via id).

PR #10324, opened 2020-05-07
MNT: PEP 8 clean-ups of some io.misc.asdf

pllim:
### Description

This pull request is to move over uncontroversial PEP 8 clean-ups done in #10260 .

* Apply #10189 to spectral coordinates added in #10185.
* Removed unused imports.
* Fixed other PEP 8 warnings.

@eslavich already successfully tests #10260 with `jwst`, so I don't see how this will break anything.

Not sure of the proper milestone for this, so feel free to re-milestone if needed.

### DIFF
--- a/astropy/io/misc/asdf/tags/transform/compound.py
+++ b/astropy/io/misc/asdf/tags/transform/compound.py
@@ -11,26 +11,26 @@ __all__ = ['CompoundType', 'RemapAxesType']
 
 
 _operator_to_tag_mapping = {
-    '+': 'add',
-    '-': 'subtract',
-    '*': 'multiply',
-    '/': 'divide',
+    '+':  'add',
+    '-':  'subtract',
+    '*':  'multiply',
+    '/':  'divide',
     '**': 'power',
-    '|': 'compose',
-    '&': 'concatenate',
+    '|':  'compose',
+    '&':  'concatenate',
     'fix_inputs': 'fix_inputs'
 }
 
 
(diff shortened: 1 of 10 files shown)


astropy/io/misc/asdf/tags/transform/compound.py
```diff
@@ -11,26 +11,26 @@
...
-    'concatenate' : '__and__',
-    'fix_inputs'  : 'fix_inputs'
+    'add': '__add__',
+    'subtract': '__sub__',
+    'multiply': '__mul__',
+    'divide': '__truediv__',
+    'power': '__pow__',
+    'compose': '__or__',
+    'concatenate': '__and__',
+    'fix_inputs': 'fix_inputs'
```
nden, line 33: I find this more difficult to read but whatever ...

pllim: Do you want me to revert this? I was just following the `flake8` suggestions.

nden: Not that it's terrible this way, it's just easier to read so if you don't mind.  Thanks for cleaning up.

pllim: What about now? Turns out `flake8` doesn't care for spacing after the `:` but not before.

nden: it's good now 👍

PR #10325, opened 2020-05-07
TST: Globally ignore import error during test collection for doctest

pllim:
### Description

This pull request is to globally ignore import errors during doctest collection by `pytest`. This also moves `importorskip` out from module level to test level.

Closes #10302 (cc @lpsinger)

Also discussed in https://github.com/astropy/pytest-doctestplus/pull/103#discussion_r418118783 and https://github.com/astropy/astropy/pull/10260#issuecomment-624106359 .


lpsinger:
This would give a false negative test result for actual bugs due to unintentional imports.

pllim:
@lpsinger , can you please clarify? With this patch, `from astropy.visualization import wcsaxes` does not raise error, although people wouldn't try to use it without `matplotlib` anyway? This is no different from how `astropy.io.misc.asdf` is being handled right now in `master`.

lpsinger:
Let's say that someone _accidentally_ added an unintended dependency on package `foo`. You would want the line `import foo` to generate an error in the doctests. This would silence that error.

pllim:
> Let's say that someone *accidentally* added an unintended dependency

It would be ignored during collection, but you would think the tests that try to use that dependency would error out during run time, no?

Still, if unwanted dependency is added but somehow not used in testing, then yeah, I guess that would sneak pass. I am hoping in that case, we would catch it in code review, because the coverage would report that the LOC is not tested. 🤷

pllim:
In addition, the name `doctest-ignore-import-errors` implies that this setting only affects `doctest`, and not regular tests? @jdavies-st recommended that we add this explicitly so we don't need to workaround a backward compatibility breakage over at `pytest-doctestplus`.

https://github.com/astropy/pytest-doctestplus/blob/dfc9e9f05f8c63f003d23fb82deacec11c6e3a11/pytest_doctestplus/plugin.py#L168-L176

pllim:
Okay, I realized that my PR title and original text was misleading. Sorry! Hopefully, I clarified them.

astrofrog:
I think it is safe to close this

pllim:
@astrofrog , any idea for #10302 then?

Issue #10323, opened 2020-05-07
sunpy wcs animator tests failing with astropy 4.1


dstansby:
This is the upstream version of https://github.com/sunpy/sunpy/issues/4096. 

The block here (introduced in https://github.com/astropy/astropy/pull/10185) is being triggered by sunpy tests:
https://github.com/astropy/astropy/blob/441486c8cf1d80409520ea756cf006932cb57c77/astropy/wcs/wcsapi/fitswcs.py#L439-L447

where `frame='helioprojective'`, a built in `sunpy` frame. The helioprojective frame doesn't support a `radial_velocity` kwarg though:
```
SkyCoord(0, 0, frame='helioprojective', unit='rad', radial_velocity=0*u.km/u.s)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dstansby/github/astropy/astropy/coordinates/sky_coordinate.py", line 313, in __init__
    skycoord_kwargs, components, info = _parse_coordinate_data(
  File "/Users/dstansby/github/astropy/astropy/coordinates/sky_coordinate_parsers.py", line 268, in _parse_coordinate_data
    raise ValueError('Unrecognized keyword argument(s) {}{}'
ValueError: Unrecognized keyword argument(s) 'radial_velocity'
```
causing the CI error. I'm struggling to make a self contained example to trigger the block at the moment.

pllim:
cc @astrofrog

Cadair:
A minimal way to reproduce is this:
```python
def test_debug_sunpy():
    from textwrap import dedent
    from astropy.wcs import WCS
    from astropy.io.fits import Header
    from astropy.wcs.wcsapi.wrappers.sliced_wcs import SlicedLowLevelWCS
    import sunpy.coordinates

    header = dedent("""\
        WCSAXES =                    4 / Number of coordinate axes
        CRPIX1  =                  0.0 / Pixel coordinate of reference point
        CRPIX2  =                  0.0 / Pixel coordinate of reference point
        CRPIX3  =                  0.0 / Pixel coordinate of reference point
        CRPIX4  =                  5.0 / Pixel coordinate of reference point
        CDELT1  =                  0.4 / [min] Coordinate increment at reference point
        CDELT2  =                2E-11 / [m] Coordinate increment at reference point
        CDELT3  =   0.0027777777777778 / [deg] Coordinate increment at reference point
        CDELT4  =   0.0013888888888889 / [deg] Coordinate increment at reference point
[...]

PR #10326, opened 2020-05-07
Don't assume coordinate frame supports radial_velocity in FITS WCS

astrofrog:
Fixes https://github.com/astropy/astropy/issues/10323 (I think?)

@dstansby @Cadair - can you confirm?


astrofrog:
Investigating issues

astrofrog:
This should now work and is ready for review. The fact there is a default distance defined is not ideal but also shouldn't stop things from working, so I'll open an issue to discuss this.

dstansby:
I can confirm that this fixes the downstream sunpy problems.

PR #10330, opened 2020-05-08
Missing f for an f-string in a coordinates error message

ayshih:
In a coordinates error message, an f-string is missing its `f`.


taldcroft:
Thanks @ayshih !

PR #10336, opened 2020-05-09
Add docs for Wtbarr

### LINKED ISSUE #10335: Add documentation for wtbarr structure
Document `Wtbarr` structure and its members. This was missed in #9641. I will work on this.

EDIT: #9641 was milestoned to 4.1

mcara:
This PR adds documentation for the `Wtbarr` structure that was not added in https://github.com/astropy/astropy/pull/9641.

In addition, this PR enables `tp_traverse` and `tp_clear` slots for `Wtbarr` structure. 

Fixes #10335

### DIFF
--- a/astropy/wcs/docstrings.py
+++ b/astropy/wcs/docstrings.py
@@ -688,15 +688,11 @@ An undefined value is represented by NaN.
 """
 
 extlev = """
-``int`` (read-only)
-
-``EXTLEV`` identifying the binary table extension.
+``int`` (read-only) ``EXTLEV`` identifying the binary table extension.
 """
 
 extnam = """
-``str`` (read-only)
-
-``EXTNAME`` identifying the binary table extension.
+``str`` (read-only) ``EXTNAME`` identifying the binary table extension.
 """
 
 extrema = """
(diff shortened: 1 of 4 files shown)


mcara:
This PR does not need a changelog.

astropy/wcs/docstrings.py
```diff
@@ -666,15 +666,11 @@ def RA_DEC_ORDER(indent=0):
 """
 
 extlev = """
-``int`` (read-only)
-
-``EXTLEV`` identifying the binary table extension.
+``int`` (read-only) ``EXTLEV`` identifying the binary table extension.
```
pllim, line 669: What is the reason to remove the blank line separation here?

mcara: Because with that line, only `int (read-only)` would appear in the table listing attributes of the class. Compare to `Tabprm` - same thing is there - "short description" is on a single line.

mcara: Because with that line, only `int (read-only)` would appear in the table listing attributes of the class. Compare to `Tabprm` - same thing is there - "short description" is on a single line.

In other words, if `Tabprm` were documented with a line between `int` (or whatever type) and description, then in ['Attributes Summary'](https://docs.astropy.org/en/stable/api/astropy.wcs.Tabprm.html) table, each row would have had only type.

PR #10337, opened 2020-05-09
Allow numpy shape functions on ShapedNDArray subclasses

mhvk:
Partially addresses #8610, though for now only the simplest numpy functions are supported on all `ShapedNDArray` subclasses (representations, coordinates, and times):
```
        np.moveaxis, np.rollaxis,
        np.atleast_1d, np.atleast_2d, np.atleast_3d, np.expand_dims,
        np.broadcast_to, np.flip, np.fliplr, np.flipud, np.rot90,
        np.roll, np.delete,
```

For instance,
```
t = Time(np.arange(50000, 50005), format='mjd')
np.roll(t, 2)
# <Time object: scale='utc' format='mjd' value=[50003. 50004. 50000. 50001. 50002.]>
```

EDIT: while testing, noticed that some functions already worked; e.g., `np.moveaxis` and `np.rollaxis` (and `np.where` though it returns an object `ndarray`).


mhvk:
Wow, astrobot is good with adding labels! Note that I also moved `ShapedLikeNDArray` out of `misc.py`, as I felt it made much more sense to have it in its own module (as it is becoming quite large, and probably could use its own documentation). If need be, I can separate that out.

mhvk:
Looking at the PR again, I felt one commit made it harder to review than necessary, so I moved that to its own PR (#10425). For this one, the only real change is in `ShapedLikeNDArray` - the rest is all tests to make sure it actually works. It complements the array-like shape methods provided with support for numpy functions that affect the shape.

One thing to do in follow-up is to properly implement `np.where` - right now, I kept it as it was, which means it works but just returns an object array.

@eteq, @taldcroft, @astrofrog - review would be most approciated, it is not as bad as it looks!

astropy/utils/misc.py
```diff
@@ -19,19 +19,21 @@
 import threading
 import re
 
-from itertools import zip_longest
 from contextlib import contextmanager
 from collections import defaultdict, OrderedDict
 
 from astropy.utils.decorators import deprecated
+# For backwards compatibility, import the shape-related stuff.
```
taldcroft, line 26: What about these symbol definitions for back-compatibility?  Are there other uses of `ShapedLikeNDArray` outside astropy core?

CHANGES.rst
```diff
@@ -71,6 +78,9 @@ astropy.units
 astropy.utils
 ^^^^^^^^^^^^^
 
+- ``ShapedLikeNDArray`` has gained the capability to use numpy functions
```
taldcroft, line 81: And moved into a new module, so API change?

astropy/utils/shapes.py
```diff
@@ -0,0 +1,329 @@
...
+    # Functions that change shape or essentially do indexing.
+    _APPLICABLE_FUNCTIONS = {
+        np.moveaxis, np.rollaxis,
+        np.atleast_1d, np.atleast_2d, np.atleast_3d, np.expand_dims,
+        np.broadcast_to, np.flip, np.fliplr, np.flipud, np.rot90,
+        np.roll, np.delete,
+        }
+
+    # Functions that for now should just pass through since they
+    # (sort of) work.  TODO: create a proper implementation.
```
taldcroft, line 215: For these (at least `np.where`) it isn't clear what a "proper implementation" should be.  Consider:
```
# tm, tm0 are Time objects
np.where.(tm > tm0, tm, 5)
```
An object array seems like the only option.

But if the second arg is a `Time` (`np.where(tm1 > tm0, tm1, tm2)`) and you want the output to be a single `Time` object (same for coordinates), I suppose you can use `info.new_like` and setitem functionality to make this work in the limited cases where it *can* work. You do get into a bit of API inconsistency, but maybe that's OK?

mhvk: Agreed it is not completely obvious, though precedence from the numpy side is that the alternatives for where are cast to a common type (see below). This argues that one should always get a `Time` out and your example should just raise an error. But partially because it is not obvious what the right behaviour is, I deferred it... (for `Time` itself, I wouldn't know what to do with one time that has a `location`, and one that has not; this is going to be a problem for `concatenate` as well).
```
In [2]: a = np.arange(5.)                                                       

In [3]: np.where(a>3, a, True)                                                  
Out[3]: array([1., 1., 1., 1., 4.])
```

CHANGES.rst
mhvk: Good point! Done

astropy/utils/misc.py
mhvk: But at least I shouldn't leave commented-out code - I've now simply remove it.

astropy/time/tests/test_methods.py
```diff
@@ -293,26 +251,96 @@ def test_broadcast(self, masked):
...
+        mjd = t0_reshape.mjd  # Creates a cache of the mjd attribute
+        t0_reshape.shape = (5, 2, 5)
+        assert t0_reshape.shape == (5, 2, 5)
+        assert mjd.shape != t0_reshape.mjd.shape  # Cache got cleared
+        assert np.all(t0_reshape.jd1 == self.t0._time.jd1.reshape(5, 2, 5))
+        assert np.all(t0_reshape.jd2 == self.t0._time.jd2.reshape(5, 2, 5))
+        assert t0_reshape.location is None
+        # But if the shape doesn't work, one should get an error.
+        t0_reshape_t = t0_reshape.T
+        with pytest.raises(AttributeError):
```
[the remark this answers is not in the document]
mhvk: Hmm, good question. It is actually a somewhat subtle bug, I now realize:
```
In [5]: np.ones((3, 2)).T.shape = 6,                                            
<ipython-input-5-384542aa1cc8> in <module>
----> 1 np.ones((3, 2)).T.shape = 6,

AttributeError: incompatible shape for a non-contiguous array

In [6]: np.ones((3, 2)).T.shape = 1, 2                                          
<ipython-input-6-570b08d46880> in <module>
----> 1 np.ones((3, 2)).T.shape = 1, 2

ValueError: cannot reshape array of size 6 into shape (1,2)
```
Since this bug was already present, I should solve it in a separate PR, so it can go into 4.0.x and 4.1.1. For now, I just made an issue, #10426

mhvk:
OK, comments addressed!  I added more tests using `Time` - they just mimic what was added in `test_representation_methods`, but it makes sense to have them for the different sub-packages.

mhvk:
Note: numpy-dev travis failure is unrelated; see #10432

mhvk:
@taldcroft - thanks! I realized I had not tested `SkyCoord` and also not added documentation, so decided I should remedy that. The latter is very short, let me know if you want to have a look (last commit).

docs/coordinates/skycoord.rst
```diff
@@ -332,7 +332,13 @@ To store arrays of coordinates in a |SkyCoord| object::
 
 In addition to vectorized transformations, you can do the usual array slicing,
 dicing, and selection using the same methods and attributes that you use for
-`~numpy.ndarray` instances::
+`~numpy.ndarray` instances.  Similarly, on ``numpy`` version 1.17 or later,
+corresponding functions as well as others that affect the shape, such as
+`~numpy.at_least1d` and `~numpy.rollaxis`, work as expected.
+(The relevant functions have to be explicitly enabled; let us know if a
```
adrn, line 338:
```suggestion
(The relevant functions have to be explicitly enabled in ``astropy``; let us know if a
```
To clarify that the user doesn't need to enable anything: something has to be changed in the astropy code itself!

mhvk: Good idea! I'll do it in all the places where I added this... (and fix the docs build at the same time)

mhvk:
Thanks, @taldcroft and @adrn for the reviews. Finally got all the documentation tests to pass (really need an .rst checker for emacs...), and since you both signed off on complementary parts, I'll merge.

pllim:
Does the error I see in #10471 look familiar to you?

pllim:
Hmm, looks like the tests in `TestManipulation` were skipped on Windows because they are marked as remote data. That would explain why CI was green and that maybe I am not crazy.

PR #10339, opened 2020-05-11
Allow pixel units in SpectralQuantity and SpectralCoord

astrofrog:
Specutils needs to subclass from ``SpectralCoord`` to represent spectral axes, but needs the ability to represent uncalibrated spectral units - as discussed in https://github.com/astropy/specutils/pull/674, this can be done by allowing ``SpectralCoord`` to take pixel units in addition to the usual spectral units. I think this is an important fix to get in for 4.1 so that specutils can start relying on astropy's SpectralCoord implementation.

### DIFF
--- a/astropy/coordinates/spectral_coordinate.py
+++ b/astropy/coordinates/spectral_coordinate.py
@@ -719,6 +719,9 @@ class SpectralCoord(SpectralQuantity):
             transformed based on the observer's new velocity frame.
         """
 
+        if self.unit is u.pixel:
+            raise u.UnitsError("Cannot transform spectral coordinates in pixel units")
+
         if self.observer is None or self.target is None:
             raise ValueError("This method can only be used if both observer "
                              "and target are defined on the SpectralCoord.")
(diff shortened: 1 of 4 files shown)


mhvk:
I don't like this much - the whole beauty of the class is that it can convert between all those representations - it feels a bit like allowing `mm` in `Angle` just because one might have a plate scale! I think it makes sense only if the user can also pass in or set an equivalency between pixels and another unit. But that seems rather out of place here, and needs machinery that would seem to go well beyond a bug fix. 

Is there any reason this cannot simply be done on the subclass in specutils? In particular, just override `_equivalent_units` on that class? (Otherwise, it would just mean slightly less clear error messages, it would seem, or overriding two methods to do the unit check before calling `super`).

mhvk:
p.s. There actually is a funny "bug" that could perhaps become a feature:
```
from astropy.coordinates import Angle
with u.add_enabled_equivalencies(u.plate_scale(140*u.mm/u.arcsec)): 
    a = Angle(10*u.mm)

a
# <Angle 10. mm>
```
Possibly we should have a way of storing that equivalency on `Angle` (or its unit)...

astrofrog:
As I was implementing this I realized I also wasn't a real fan of this. Great idea to overload ``_equivalent_unit``, I'll try that.

astrofrog:
And of course that works fine... sorry for the noise! I agree that adding this here wouldn't make any more sense than adding it to ``Angle`` and so forth.

Issue #6280, opened 2017-06-26
Decide if obstime transformations of coordinates with velocities should result in motion


eteq:
This issues is a follow-up discussion to an unresolved question from #6219 (in that PR we are punting on the question by raising an error).

The starting point for the topic is https://github.com/astropy/astropy/pull/6219#issuecomment-309399689 - @StuartLittlefair's position is that if you transform to a frame with a different `obstime`, a coordinate with velocities should be moved to a different position based on its velocity vector.  @StuartLittlefair added more in https://github.com/astropy/astropy/pull/6219#issuecomment-310907959 and https://github.com/astropy/astropy/pull/6219#issuecomment-310908691 and @mhvk agreed in https://github.com/astropy/astropy/pull/6219#issuecomment-310911108

By contrast, I think this is *not* what we want and will be extremely confusing and inconsistent with the other coordinate machinery - see https://github.com/astropy/astropy/pull/6219#issuecomment-310881737.  In https://github.com/astropy/astropy/pull/6219#issuecomment-310906282 and https://github.com/astropy/astropy/pull/6219#issuecomment-310908439 @adrn agreed.

This issue is hence a placeholder, because this is a somewhat significant philosophical question for `coordinates`, so we should resolve it for the next version.

eteq:
(not that this issue is explicitly referenced in a `NotImplementedError` I am proposing to add in #6219)

astrofrog:
I would be in favor of *not* updating the position based on the velocity. The velocity vector may be that of a flow/stream rather than a point object that is moving for example. This is not to say that there couldn't be a keyword argument or a separate method to update the positions though.

mhvk:
If I input Barnard's Star with its Hipparcos position and proper motion, and try to determine where it will be at a give time, then surely by default it should account for the time difference.... More generally, one thing I do very often is to register images to, say, UCAC4 or other position + proper motion catalogues. For this to work generally, one again does want to apply proper motions.

I must admit I cannot see the general use case for not taking into account `obstime` - it seems this is sensible only for coordinate frames in which `obstime` is not needed in the first place.

astrofrog:
What about the case of a binary star? If I have the velocity and proper motion of a star in a binary, any linear extrapolation would be very wrong.

astrofrog:
We could have a required keyword argument for cases like this to avoid any ambiguity?

adrn:
Also solar system objects

mhvk:
[...]

PR #10341, opened 2020-05-11
Added a context manager to set all of the finite-difference time steps in a TransformGraph

ayshih:
I've been needing to set the finite-difference time step (`finite_difference_dt`) in coordinate transformations involving velocities to get good accuracy in some situations (similarly described [here in the Astropy documentation](https://docs.astropy.org/en/stable/coordinates/velocities.html#astropy-coordinate-finite-difference-velocities)), but it's a bit of a pain to change it for one transformation, and it's a lot of a pain to change it if one isn't sure which transformations are going to be called.

This PR provides a context manager as a method in the `TransformGraph` class to temporarily set the `finite_difference_dt` attribute for *all* transforms in the graph that have that attribute.  The original settings are restored after the context manager.

For example, with this PR, the example in the documentation:
```python
>>> from astropy.coordinates import frame_transform_graph, AltAz, CIRS
>>> trans = frame_transform_graph.get_transform(AltAz, CIRS).transforms[0]
>>> trans.finite_difference_dt = 1*u.year
>>> gcrs = aa.transform_to(GCRS(obstime=time))  
>>> trans.finite_difference_dt = 1*u.second  # return to default
```
could be rewritten as:
```python
>>> from astropy.coordinates import frame_transform_graph
>>> with frame_transform_graph.impose_finite_difference_dt(1*u.year):
...     gcrs = aa.transform_to(GCRS(obstime=time))
```

### DIFF
--- a/astropy/coordinates/tests/test_transformations.py
+++ b/astropy/coordinates/tests/test_transformations.py
@@ -7,7 +7,7 @@ import pytest
 
 from astropy import units as u
 from astropy.coordinates import transformations as t
-from astropy.coordinates.builtin_frames import ICRS, FK5, FK4, FK4NoETerms, Galactic, AltAz
+from astropy.coordinates.builtin_frames import ICRS, FK5, FK4, FK4NoETerms, Galactic, AltAz, HCRS
 from astropy.coordinates import representation as r
 from astropy.coordinates.baseframe import frame_transform_graph
 from astropy.tests.helper import assert_quantity_allclose as assert_allclose
@@ -586,3 +586,38 @@ def test_remove_transform_errors():
 
     with pytest.raises(ValueError):
         graph.remove_transform(TCoo1, TCoo1, 1)
+
+
+def test_impose_finite_difference_dt():
+    class H1(HCRS):
+        pass
+
+    class H2(HCRS):
+        pass
+
+    class H3(HCRS):
+        pass
+
+    graph = t.TransformGraph()
+    tfun = lambda c, f: f.__class__(ra=c.ra, dec=c.dec)
+
+    # Set up a number of transforms with different time steps
+    old_dt = 1*u.min
+    transform1 = t.FunctionTransformWithFiniteDifference(tfun, H1, H1, register_graph=graph,
+                                                         finite_difference_dt=old_dt)
+    transform2 = t.FunctionTransformWithFiniteDifference(tfun, H2, H2, register_graph=graph,
+                                                         finite_difference_dt=old_dt * 2)
+    transform3 = t.FunctionTransformWithFiniteDifference(tfun, H2, H3, register_graph=graph,
+                                                         finite_difference_dt=old_dt * 3)
+
+    # Check that all of the transforms have the same new time step
+    new_dt = 1*u.yr
+    with graph.impose_finite_difference_dt(new_dt):
+        assert transform1.finite_difference_dt == new_dt
+        assert transform2.finite_difference_dt == new_dt
+        assert transform3.finite_difference_dt == new_dt
+
+    # Check that all of the original time steps have been restored
+    assert transform1.finite_difference_dt == old_dt
+    assert transform2.finite_difference_dt == old_dt * 2
+    assert transform3.finite_difference_dt == old_dt * 3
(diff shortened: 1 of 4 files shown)


mhvk:
@ayshih - this looks very promising indeed! I'll try to bring in @eteq and @adrn again, as I realize now we did in fact have a discussion about some of this when the proper motions were first implemented, and how to handle transformations to a different obstime was explicitly left as a to-do item - see #6280. Not dealing with it has left us in a fine mess...

Your approach of having a context manager may well be the most sensible, and I certainly like in this PR that it seems to properly identify the cause by making explicit the previously implicit ICRS->BCRS transformation.

astropy/coordinates/tests/test_transformations.py
```diff
@@ -579,3 +579,38 @@ def test_remove_transform_errors():
...
+    new_dt = 1*u.yr
+    with graph.impose_finite_difference_dt(new_dt):
+        assert transform1.finite_difference_dt == new_dt
+        assert transform2.finite_difference_dt == new_dt
+        assert transform3.finite_difference_dt == new_dt
+
+    # Check that all of the original time steps have been restored
+    assert transform1.finite_difference_dt == old_dt
+    assert transform2.finite_difference_dt == old_dt * 2
+    assert transform3.finite_difference_dt == old_dt * 3
```
mhvk, line 616: From other tests, it looks like you should unregister the transformations at the end of the test (really, this should be done with a module-level fixture or so!!)

astropy/coordinates/transformations.py
```diff
@@ -675,6 +675,38 @@ def deco(func):
...
+        ----------
+        dt : `~astropy.units.Quantity` or callable
+            If a quantity, this is the size of the differential used to do the finite difference.
+            If a callable, should accept ``(fromcoord, toframe)`` and return the ``dt`` value.
+        """
+        key = 'finite_difference_dt'
+        saved_settings = []
+
+        try:
+            for from_frame in self._graph:
```
mhvk, line 698: since `self._graph` is just a dict, you can iterate over `.values()`
```
for to_frames in self._graph.values():
    for transform in to_frames.values():
```

astropy/coordinates/tests/test_transformations.py
ayshih: These transformations are registered only to the locally created `graph`, not the global `frame_transform_graph`, so there's no need for any explicit cleanup.  Everything goes poof at the end of the test.

astropy/coordinates/transformations.py
ayshih: Ah, good point!

astropy/coordinates/tests/test_transformations.py
mhvk: Ah, yes, you make your own `TransformGraph()` - sorry for the noise!

adrn:
Thanks @ayshih - this is cool! 

I'm curious to hear your thoughts on this approach versus adding a coordinates-level config item to handle this. Or maybe it would make sense to have both?

ayshih:
@adrn I've been working a bunch with both cosmic objects and solar-system bodies in the same session, so I want to impose a long `finite_difference_dt` when I'm transforming star coordinates, but have an orders-of-magnitude shorter `finite_difference_dt` when I'm transforming planet coordinates.  Having only a config setting wouldn't cover my use case, although it may be fine for most use cases.

mhvk:
I think the context manager also has the advantage that one can easily check whether the finite differences are what is causing a problem. Also, the explicitness is good - less risk of one user finding problems and another not because of different configurations.

astropy/coordinates/transformations.py
ayshih: Done

ayshih:
With feature freeze for 4.2 imminent, can this get reviewed?  Thanks!

adrn:
Will get to this today - thanks for the reminder, @ayshih !

mhvk (approved):
Had a quick look again and one suggestion inside - but it doesn't depend on it and so I'll approve to ensure this gets merged for 4.2 (though still would be good for @adrn to check too, as someone more likely to actually use this!)

adrn (approved):

adrn:
I'm happy with this - thanks @ayshih for the addition (and for the patience...)

Issue #10344, opened 2020-05-12
Ideally should avoid hardcoded file path in tests


pllim:
### Description

I see things in the tests like:
```python
ROOT = os.path.abspath(os.path.dirname(__file__))
t_comp = read(os.path.join(ROOT, filename))
```

Is there a reason why this is not used instead?
```python
t_comp = read(get_pkg_data_filename(filename))
```

There is also a variation using `DATA` global variable.

One of the subpackage where I saw this was `io.ascii`, so cc @taldcroft and @hamogu

hamogu:
It's mostly because the io.ascii code predates astropy. `get_pkg_data_filename` was not around in those days. We've never had problems with this, so I don't think it's worth anybodies time to change all the tests for negligible benefit unless you are aware of a situation where the current version will fail for some legal way of running the tests (e.g. some tox setting, headless servers, etc. - I don't think any of these apply here though).

pllim:
I like consistency but I have no strong opinion here since I think this only affects some tests and they are passing.

dhomeier:
Could be useful to have a comment to this effect somewhere, so people writing new tests are reminded to use `get_pkg_data_filename` where possible. But these cases are scattered over at least have a dozen files, so not sure where to best place it.

pllim:
It is already documented in https://docs.astropy.org/en/latest/development/testguide.html#working-with-data-files but I don't know who reads it.

taldcroft:
I definitely like consistency and bringing code to a certain standard within the project. If anybody feels motivated then I will happily approve!  I lot of new code gets created on the basis of existing code, not the documentation, (speaking from personal experience!) so there is value in making the changes.

mhvk:
I'm with @taldcroft here and for `io.fits` in particular feel that the fact that the code has a somewhat deviating standard is one of the things it makes it a bit more difficult to work on. (Absolutely no criticism implied, just that consistency is very helpful!)

Since I think nobody is arguing *against* changing it to make it more standard, I removed labels to leave this as a low-priority issue.

pllim:
Re: `io.fits` -- That one has a lot of history and was basically copy-pasted from PyFITS. 😬

Iamsoto:
Hey guys, 
I got some free time for now, mind if I try and tackle this?

pllim:
Go for it, @Iamsoto ! Hopefully some of them are already fixed on `master` since, but I am sure there are some left.

Iamsoto:
[...]

Issue #4630, opened 2016-02-23
Refactor optional dependencies flags into utils module


taldcroft:
#4606 has many instances of:

```
try:
    import pathlib
except ImportError:
    HAS_PATHLIB = False
else:
    HAS_PATHLIB = True
```

This idiom gets repeated in many other places for other optional dependencies.  Might it make sense to refactor all these into a single module (perhaps opt_deps.py) in utils?  Then application code would be:

```
from ...utils.opt_deps import HAS_PATHLIB
```

hamogu:
That would also make is easier to find and change if the optional dependency changes, is removed (e.g. we decide in the future that scipy becomes a hard dependency).

astrofrog:
In Glue, I also made decorators that can be used directly on the test functions, e.g. ``@requires_pathlib``, so we could do that here too optionally?

taldcroft:
@astrofrog - is this `@requires_X` mapped to `pytest.mark.skipif` so pytest is doing the right thing?  It seems like `@requires` could take an argument of one or more optional deps, e.g. `@test_requires('pathlib', 'scipy', 'beautifulsoup')`.  I think it makes sense to have `test` in the decorator name if it is specifically targeted at tests.

astrofrog:
@taldcroft - yes, you can see what I have here:

https://github.com/glue-viz/glue/blob/master/glue/tests/helpers.py

However, it's hard to do boolean logic with that, so it might be over-engineering, and the constants you are suggesting might be enough.

saimn:
Glue's decorator seems a good way to go. 
https://github.com/glue-viz/glue/blob/master/glue/tests/helpers.py#L14
It provides both a decorator which wraps pytest's skipif, and a `...._INSTALLED` variable to check if a module is available. (Should be renamed to `HAS_...` ?)

pllim:
`pathlib` is now provided by the minversion of Python that we support, so that `if ... else ...` can be taken out altogether. For testing, `pytest` has `importorskip`, so can this issue be closed?

astrofrog:
I think it might still be worth considering centralizing the ``HAS_...`` variables for other optional dependencies?

pllim:
I started working on this and then realized that we cannot really lazy import until minversion of Python is 3.7. I think something like this might work based on PEP 562, otherwise we either over-engineer or incur a penalty of checking for everything at import time.

I called this file `astropy/utils/compat/optional_deps.py` but nothing is set in stone yet:

```python
[...]

PR #10345, opened 2020-05-12
MNT: Remove outdated dependency checks

pllim:
### Description

This pull request is to address part of #4630, where a dependency check is no longer needed because:

* Minimum supported version for `scipy` is beyond 0.12 (0.18 when PR is opened), so removing `OLDER_SCIPY` checks.
* `bz2` and `lzma` are now part of Python standard library since 3.3 and we only support Python 3.6 or later (when PR is opened), but Python can be built without those modules, so adding a note about that ~removing `HAS_BZ2` and `HAS_XZ` checks~.

Bonus:

* Might as well address some of #10344 that I came across.
* Minor PEP 8 clean-ups.

### DIFF
--- a/astropy/coordinates/tests/test_matching.py
+++ b/astropy/coordinates/tests/test_matching.py
@@ -159,7 +159,7 @@ def test_matching_method():
     assert len(idx1) == len(d2d1) == len(d3d1) == 20
 
 
-@pytest.mark.skipif(not HAS_SCIPY, reason="Requires scipy > 0.12.0 ")
+@pytest.mark.skipif(not HAS_SCIPY, reason="Requires scipy")
 def test_search_around():
     from astropy.coordinates import ICRS, SkyCoord
     from astropy.coordinates.matching import search_around_sky, search_around_3d
@@ -241,7 +241,7 @@ def test_search_around():
     assert d3d.unit == u.dimensionless_unscaled
 
 
-@pytest.mark.skipif(not HAS_SCIPY, reason="Requires scipy > 0.12.0 ")
+@pytest.mark.skipif(not HAS_SCIPY, reason="Requires scipy")
 def test_search_around_scalar():
     from astropy.coordinates import SkyCoord, Angle
 
@@ -260,7 +260,7 @@ def test_search_around_scalar():
     assert 'search_around_3d' in str(excinfo.value)
 
 
-@pytest.mark.skipif(not HAS_SCIPY, reason="Requires scipy > 0.12.0 ")
+@pytest.mark.skipif(not HAS_SCIPY, reason="Requires scipy")
 def test_match_catalog_empty():
     from astropy.coordinates import SkyCoord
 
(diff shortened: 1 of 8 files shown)


pllim:
Not sure what's going on with the `flake8` check failure in `html-docs` job:
```
codestyle installdeps: flake8
 
codestyle installed: flake8==3.8.1,importlib-metadata==1.6.0,mccabe==0.6.1,pycodestyle==2.6.0,pyflakes==2.2.0,zipp==3.1.0
 
codestyle run-test-pre: PYTHONHASHSEED='887351196'
 
codestyle run-test: commands[0] | flake8 astropy --count --select=E101,W191,W291,W292,W293,W391,E111,E112,E113,E30,E502,E722,E901,E902,E999,F822,F823
 
astropy:0:1: E902 FileNotFoundError: [Errno 2] No such file or directory: 'astropy'
1
```

Same error on https://travis-ci.org/github/astropy/astropy/jobs/686214819 , which is not even reporting back to the PR.

saimn:
For `bz2` and `lzma`, the issue is that Python can be build without those, if the libraries are not installed. This lead to #9652 and #9761 (at least) to make those optional.

pllim:
Who would build Python without `bz2` and `lzma` -- `conda`? PyPI?

saimn:
You can build Python from source, pyenv does that. And you can build it without the bzip2 and xz libs (there are quite a few optional libs, I didn't find a list but there is some info here: https://devguide.python.org/setup/#unix)

pllim:
It sounds like a lot of extra code to cater to such specific edge case, but whatever, I'll remove it. Thanks for the clarification.

pllim:
@hamogu , you might want to look again.

@saimn , I reverted the `bz2` and `lzma` checks removal.

hamogu:
@pllim I did not know about issues this would create, so I'm clearly not a competent reviewer for this change and I'll defer to more experienced developers (as I should have done from the beginning).

astropy/io/ascii/tests/test_read.py
```diff
@@ -28,14 +27,15 @@
...
+# NOTE: Python can be built without bz2.
 try:
     import bz2  # noqa
 except ImportError:
     HAS_BZ2 = False
 else:
     HAS_BZ2 = True
 
-asciiIO = lambda x: BytesIO(x.encode('ascii'))
+asciiIO = lambda x: BytesIO(x.encode('ascii'))  # noqa
```
mhvk, line 38: Maybe just fix it?
```
def asciiIO(x):
    return BytesIO(x.encode('ascii'))
```

astropy/coordinates/tests/test_matching.py
```diff
@@ -165,8 +159,7 @@ def test_matching_method():
     assert len(idx1) == len(d2d1) == len(d3d1) == 20
 
 
-@pytest.mark.skipif(not HAS_SCIPY or OLDER_SCIPY,
-                    reason="Requires scipy > 0.12.0 ")
+@pytest.mark.skipif(not HAS_SCIPY, reason="Requires scipy > 0.12.0 ")
```
mhvk, line 162: If you're still up for it, I'd do `reason="Requires scipy"`, since the version has now become irrelevant.

astropy/io/ascii/tests/test_read.py
pllim: OK, done.

astropy/coordinates/tests/test_matching.py
pllim: Done. Thanks for review!

pllim:
We can ignore the coverage failure. Thanks, everyone!

PR #10346, opened 2020-05-12
Bugfix: Adding bounds to wcs fitting

ceb8:
### Description

This pull request addresses three small bugs in `wcs.utils.fit_wcs_from_points`:
- Added bounds to the least square fit requiring that the fit CRPIX value be within the bounds of the input coordinate/pixel pairs. This prevents getting physically meaningless CRPIX fits like "-44811.23089933425  -40821.30343222443" .
- Corrected the pixel shape by adding 1 to each dimension because it is a subtraction of indices, not lengths.
- Moved the xp,xy assignment of the try/except so that they get assigned regardless.

The first two bugs are illustrated in the following code snippet:
Screen Shot 2020-05-12 at 3 34 01 PM

And the corrected result is:
Screen Shot 2020-05-12 at 3 36 50 PM

@cshanahan1 will want to look at this.

### DIFF
--- a/astropy/wcs/utils.py
+++ b/astropy/wcs/utils.py
@@ -1015,8 +1015,8 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
     from .wcs import Sip
     from scipy.optimize import least_squares
 
+    xp, yp = xy
     try:
-        xp, yp = xy
         lon, lat = world_coords.data.lon.deg, world_coords.data.lat.deg
     except AttributeError:
         unit_sph =  world_coords.unit_spherical
@@ -1056,7 +1056,7 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
         raise ValueError("sip_degree must be None, or integer.")
 
     # set pixel_shape to span of input points
-    wcs.pixel_shape = (xp.max()-xp.min(), yp.max()-yp.min())
+    wcs.pixel_shape = (xp.max()+1-xp.min(), yp.max()+1-yp.min())
 
     # determine CRVAL from input
     close = lambda l, p: p[np.argmin(np.abs(l))]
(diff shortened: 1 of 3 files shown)


pllim:
Need test(s) and change log. Thanks!

ceb8:
@pllim Since you put this in the 4.1.1 milestone, I put my changelog entry in the 4.1 section, should I have put it in the 4.2 section instead?

I've added a test, and confirmed that it fails before the bugfix and passes after.

I'm also confused about the html-docs test error, it seems to be an error in running the codedocstyle command at all? Do you have any insite? Nothing I did should be affecting the docs.

pllim:
Rebase to get rid of the error.

As for change log, I am not sure. This PR is opened at a weird window of a release.

ceb8:
@pllim Ok, I'll leave the changelog entry failing, and when this gets reviewed and is ready for merge I can put it in whatever the correct section is then.

astropy/wcs/utils.py
```diff
@@ -1077,9 +1077,16 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
     # fit linear terms, assign to wcs
     # use (1, 0, 0, 1) as initial guess, in case input wcs was passed in
     # and cd terms are way off.
+    # Use bounds to require that the fit center pixel is on the input image
+    xpmin, xpmax, ypmin, ypmax = xp.min(), xp.max(), yp.min(), yp.max()
+    if xpmin==xpmax: xpmin, xpmax = xpmin-0.5, xpmax+0.5
+    if ypmin==ypmax: ypmin, ypmax = ypmin-0.5, ypmax+0.5
```
nden, line 1083: Please split the above two lines so it's consistent with the rest of the package.

nden (changes requested):
Looks good!
I have a few style related requests and then this can be merged.
In addition to the comments, please go over the changes and add spaces between operators and after commas, basically conform to PEP8 style.
Thanks!

ceb8:
@nden I believe I've addressed all your comments. And I moved the changelog entry to match the milestone.

nden (approved):

Issue #10233, opened 2020-04-29
astropy.table saves incorrectly FITS tables with a column “a: 0”. Loading this file leads to a ValueError exception.


mboquien:
### Description
astropy.table fails to save correctly FITS tables containing a column named “a: 0” and cannot load the generated file. This occurs when the column name starts with a letter, then contains a colon, then a space, then a number. In this case the TTYPE entry appears to be duplicated. For instance with the above example the FITS header generated by astropy.table is the following:

XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                    8 / length of dimension 1                          
NAXIS2  =                    1 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
GCOUNT  =                    1 / number of groups                               
TFIELDS =                    1 / number of table fields                         
TTYPE1  = 'a: 0'                                                                
TTYPE1  = 'a: 0'                                                                
TFORM1  = 'D       '                                                            
TTYPE1  = 'a: 0'                                                                

Attempting to read this file with astropy.table leads to a ValueError exception:

```
ValueError                                Traceback (most recent call last)
<ipython-input-2-b76ff7c17fe6> in <module>
----> 1 t = Table.read('8d4584eb-4f4b-4215-b138-5070614b49f7.fits')

/usr/lib/python3.8/site-packages/astropy/table/connect.py in __call__(self, *args, **kwargs)
     50     def __call__(self, *args, **kwargs):
     51         cls = self._cls
---> 52         out = registry.read(cls, *args, **kwargs)
     53 
     54         # For some readers (e.g., ascii.ecsv), the returned `out` class is not

/usr/lib/python3.8/site-packages/astropy/io/registry.py in read(cls, format, *args, **kwargs)
    521 
    522         reader = get_reader(format, cls)
--> 523         data = reader(*args, **kwargs)
    524 
    525         if not isinstance(data, cls):

/usr/lib/python3.8/site-packages/astropy/io/fits/connect.py in read_table_fits(input, hdu, astropy_native, memmap, character_as_bytes)
    213 
    214         try:
--> 215             return read_table_fits(hdulist, hdu=hdu,
    216                                    astropy_native=astropy_native)
    217         finally:

/usr/lib/python3.8/site-packages/astropy/io/fits/connect.py in read_table_fits(input, hdu, astropy_native, memmap, character_as_bytes)
[...]

PR #10357, opened 2020-05-16
Check and restrict keyword names for FITS RVKC header entries

### LINKED ISSUE #13273: HDUList.header returns incorrect values for .keys() and .items()
### Description

HDUList.header returns incorrect values for .keys() and .items()

In the example below, `hdulist[0].header`, shows a key/value pair of

  OBJECT  = 'TILEID: 1106'

`hdulist[0].header['OBJECT']` returns the value from the key/value pair.

However, both `hdulist[0].header.keys()` and `hdulist[0].header.items()` indicate a key of 'OBJECT.TILEID'. Using the returned key from .keys() or .items() returns a bogus value:

`hdulist[0].header['OBJECT.TILEID']  => 1106.0`

### Expected behavior

Expect 'OBJECT' but not 'OBJECT.TILEID' to be listed in header.keys() and header.items().

### Actual behavior

Got bad values for a key in hdulist[0].header.  Seems that the value of a key/value became part of the key. See "Reproduce"

### Steps to Reproduce

```python
>>> import astropy.io.fits
>>> hdulist = astropy.io.fits.open("https://astroarchive.noirlab.edu/api/retrieve/00b228b9384ab0e358fd12b3aa1586d0/")
>>> 'OBJECT' in hdulist[0].header        # => True
>>> 'OBJECT' in hdulist[0].header.keys() # => False  # but used as key next
>>> hdulist[0].header['OBJECT']          # => 'TILEID: 1106'
>>> hdulist[0].header['OBJECT.TILEID']   # => 1106.0   # BOGUS key and value
>>> hdulist[0].header                    # Shows 'OBJECT' but not 'OBJECT.TILEID'
```

### System Details

Linux-5.4.0-113-generic-x86_64-with-glibc2.31
Python 3.9.7 (default, Sep  6 2021, 18:25:07) [GCC 9.3.0]
Numpy 1.22.4
pyerfa 2.0.0.1
astropy 5.1
Scipy 1.8.1
Matplotlib 3.5.2

dhomeier:
### Description
#10233 reported side effects of parsing FITS header cards as record-valued keyword cards: any key value following the RVKC pattern of `name: <int or float>` triggers the creation of a RVKC with an unexpected card value.
For reserved keywords that are (against the recommendations) given values like `'A: 0'` this results in errors as the keyword is inserted multiple times, apparently trying to create a regular card.

In https://github.com/astropy/astropy/issues/10233#issuecomment-621589224 two limiting cases for restricting the allowed names for RVKCs were discussed:

1. Allow **only** names closely following the patterns defined in https://fits.gsfc.nasa.gov/wcs/dcs_20040422.pdf

2. Reject **only** mandatory or reserved keywords (Appendix C of https://fits.gsfc.nasa.gov/fits_standard.html)

Since the full scope in which RVKCs are used is not exactly known, this PR adopts a somewhat relaxed version of the 1. option. As it turned out, some of the existing tests expect names like `D2IM1` to be accepted, so the pattern was made a bit more general than proposed in https://github.com/astropy/astropy/issues/10233#issuecomment-621919377; this in turn required to explicitly exclude some reserved keywords which would otherwise match the pattern.

Fix #13273

### DIFF
--- a/astropy/io/fits/card.py
+++ b/astropy/io/fits/card.py
@@ -27,6 +27,13 @@ VALUE_INDICATOR = '= '  # The standard FITS value indicator
 VALUE_INDICATOR_LEN = len(VALUE_INDICATOR)
 HIERARCH_VALUE_INDICATOR = '='  # HIERARCH cards may use a shortened indicator
 
+# Reserved keyword names. Some of these match the allowed pattern for record-valued keyword cards
+# (_keywd_RVKC_RE), so they have to be checked against explicitly (first four letters only).
+KEYWORD_RSRV = ('CDEL', 'CROT', 'CD1_', 'CD2_', 'CDIS', 'CRPI', 'CRVA', 'CTYP', 'CUNI', 'CHEC',
+                'DATA', 'DATE', 'GCOU', 'PCOU', 'PSCA', 'PTYP', 'PZER', 'TDIM', 'TELE', 'TTYP',
+                'TFOR', 'TUNI', 'TNUL', 'TSCA', 'TZER', 'TDIS', 'TBCO', 'TCTY', 'TCUN', 'TCRP',
+                'TCRV', 'TCDL', 'TRPO', 'THEA', 'TDMA', 'TDMI', 'TLMA', 'TLMI')
+
 
 class Undefined:
     """Undefined value."""
@@ -46,9 +53,10 @@ class Card(_Verify):
 
     # String for a FITS standard compliant (FSC) keyword.
     _keywd_FSC_RE = re.compile(r'^[A-Z0-9_-]{0,%d}$' % KEYWORD_LENGTH)
-    # This will match any printable ASCII character excluding '='
-    _keywd_hierarch_RE = re.compile(r'^(?:HIERARCH +)?(?:^[ -<>-~]+ ?)+$',
-                                    re.I)
+    # This will match any printable ASCII character excluding '='.
+    _keywd_hierarch_RE = re.compile(r'^(?:HIERARCH +)?(?:^[ -<>-~]+ ?)+$', re.I)
+    # Format for keywords containing record-valued keyword cards.
+    _keywd_RVKC_RE = re.compile(r'^[T0-9]?[CDP][A-Z0-9]{1,5} *$')
 
     # A number sub-string, either an integer or a float in fixed or
     # scientific notation.  One for FSC and one for non-FSC (NFSC) format:
@@ -611,7 +619,8 @@ class Card(_Verify):
             return self._check_if_rvkc_image(*args)
         elif len(args) == 2:
             keyword, value = args
-            if not isinstance(keyword, str):
+            if not (isinstance(keyword, str) and keyword[:4] not in KEYWORD_RSRV and
+                    self._keywd_RVKC_RE.match(keyword.split('.')[0])):
                 return False
             if keyword in self._commentary_keywords:
                 return False
@@ -655,6 +664,9 @@ class Card(_Verify):
         # do not contain RVKC field-specifiers; it's very much a
         # micro-optimization but it does make a measurable difference
         if not rest or rest[0] != "'" or rest.find(': ') < 2:
+            return False
+        if not (isinstance(keyword, str) and keyword[:4] not in KEYWORD_RSRV and
+                self._keywd_RVKC_RE.match(keyword.split('.')[0])):
             return False
 
         match = self._rvkc_keyword_val_comm_RE.match(rest)
--- a/astropy/io/fits/column.py
+++ b/astropy/io/fits/column.py
@@ -162,7 +162,7 @@ TFORMAT_ASCII_RE = re.compile(r'(?:(?P<format>[AIJ])(?P<width>[0-9]+)?)|'
                               r'(?:(?P<widthf>[0-9]+)\.'
                               r'(?P<precision>[0-9]+))?)')
 
-TTYPE_RE = re.compile(r'[0-9a-zA-Z_]+')
+TTYPE_RE = re.compile(r'[0-9a-zA-Z_]+$')
 """
 Regular expression for valid table column names.  See FITS Standard v3.0 section
 7.2.2.
--- a/astropy/io/fits/tests/test_core.py
+++ b/astropy/io/fits/tests/test_core.py
@@ -145,6 +145,26 @@ class TestCore(FitsTestCase):
             assert table.data.dtype.names == ('c2', 'c4', 'foo')
             assert table.columns.names == ['c2', 'c4', 'foo']
 
+    def test_disapproved_colnames(self):
+        """Check that non-recommended column names raise a warning, but are accepted."""
+
+        w = (r"It is strongly recommended that column names contain only upper and lower-case "
+             r"ASCII letters, digits, or underscores for maximum compatibility ")
+        wn = r"with other software \(got '{:s}'\)."
+        coldefs = {'a: 0': (0, 'I', 'i2'), '2.1': (1.0, 'E', 'f4'), 'C3.1': (2, 'D', 'f8')}
+        columns = []
+
+        for k, v in coldefs.items():
+            with pytest.warns(fits.verify.VerifyWarning, match=w+wn.format(k)):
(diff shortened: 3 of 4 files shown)


pllim:
Does this completely address #10233 ? Also, if this better qualifies as API change rather than bug fix, please remilestone to 4.2. Thanks!

dhomeier:
How to deal with the load of VerifyWarnings now popping up that the column name check works? I am tempted to add it to filterwarnings, as we might in general assume that the test writers are aware they are using non-standard names.
A bit less sure about the `HIERARCH` warning, as the name conforms to the standard here; but then it using the implicit `HIERARCH` card creation, so a warning still seems appropriate.

dhomeier:
> Does this completely address #10233 ?

It fixes the original bug in allowing tables with column names like `a: 0` to be written (after giving a due VerifyWarning).
It is not clear whether the now allowed range of RVKC names covers all existing downstream use cases out there.

pllim:
FWIW IMHO, if a decision is made to do partial support and designate the rest as "won't fix," we can still close that issue but with clarifications. It is better than leaving it open indefinitely. I'll let y'all make that call. Thanks again!

dhomeier:
I'd agree that the issue as reported is fixed; how to handle RVKC support beyond that can be discussed outside of it.

saimn:
Why the choice of matching `[CDP]*` keywords ?

dhomeier:
Why those 3 letters? The combinations `DP` and `DQ`are explicitly presented in the WCS paper. `test_update_rvkc_2` uses a `D2IM1`, but that test seems to date back to pyfits SVN, so I am pessimistic finding out the background of it. The paper also mentions `CPDIS`, `CQDIS`, `PV`_i_ma_ keywords as well as the standard `CD` and `PC` coefficients, though it is not clear which, if any of them, would make it into RVKCs – I just went for the cautious/more generous option by including all three initials and then making no restrictions on the following digit-letter sequence.

nden:
I haven't followed this from the beginning. Just want to mention that the `D2IM`, `CPDIS*` and `DP` keywords are used currently to describe HST imaging distortion. The distortion paper is now fully supported in wcslib, though some parts, like `DP`, `CPDIS` type distortion continue to use the implementation in the python wrappers. But having a support in wcslib means there might be other files out there that use any of the RVK in the paper.

saimn:
From what I can see in wcslib only `DP`/`DQ` are used as RVK, and maybe `TPV` I'm not sure.

I don't know what's the best choice... the safe option (only excluding sepcific keywords) could go as a bugfix, but I think the "more aggressive" option (currently implemented here) should go to the next major release to give time for people to test. And with `[CDP]` you're allowing a pretty broad range of keywords already, so hopefully this should be fine.

dhomeier:
> From what I can see in wcslib only `DP`/`DQ` are used as RVK, and maybe `TPV` I'm not sure.

From @nden's comment, and also what's included in the tests, `D2IM` is also in use, and perhaps `CPDIS` as well.

saimn:
Right, `D2IM` is used as RVK but in the Python code only, I missed that.

saimn:
About the verify warnings, the problem is that the serialization of mixin columns is using a `.` as separator, and dot is not part of the "recommended" characters... It is a bit unfortunate that files written by astropy trigger warnings, especially if the user can do nothing about this. It's not clear to me exactly how the serialization works and if we could change the separator (but there are not many choices, only `_` which is not ideal). So maybe the solution would be to allow `.` in the TTYPE regex, it would be an exception to the recommendation but I guess we can live with that and I don't see many other choices.

saimn:
@dhomeier - Your other PR about RVKC reminds me about this one ;). Also cc @embray, maybe you have a better idea about what to do here ?

dhomeier:
> @dhomeier - Your other PR about RVKC reminds me about this one ;). Also cc @embray, maybe you have a better idea about what to do here ?

Hah, I recalled we had been discussing valid RVKC patterns before when working on that, but was not sure this was still open. But I just checked locally, that after rebasing this is going along well with #11301.
Also the headers I've seen in the various WCS formats seem to confirm the choice of keyword names, but it's probably best to rebase again once #11260 is merged and see if there are any issues (direct conflicts are unlikely, since WCSLIB is doing the actual parsing internally).

embray:
I agree that RVKC should (almost) only allow keywords described in the distortions paper, plus possibly other WCS keywords as discussed above.  I think I originally allowed them *very broadly* because it was a potentially useful feature for representing other hierarchical data in FITS.  But the format has never been described outside the distortions paper, so I don't imagine it has any other use cases in the wild.  It would probably simplify the code and speed it up to restrict it only to relevant keywords, as well as fix unintentional use of the convention as in the original issue.

It would have been nice if the convention had a built-in way to explicitly specify which keywords are using the convention, but that ship has sailed, so maybe the best approach is to implicitly assume that the convention only applies to keywords described in the distortions paper.

I think this regexp is trying to be too clever:

```python
_keywd_RVKC_RE = re.compile(r'^[T0-9]?[CDP][A-Z0-9]{1,5} *$')
```

You could make a more explicit regexp that doesn't also capture unrelated keywords, and thus eliminate the need for `KEYWORD_RSRV`, like:

```python
_keywd_RVKC = re.compile(r'('
    r'([T0-9](DP|DQ)[0-9]{1,3})|'  # keywords accepting prefix w/ table index
    r'((CPDIS|DQDIS|DP|DQ|CPERR|CQERR|D2IM)[0-9]{1,2})|'  # keywords accepting coord index
    r'(DVERR)  # keywords with no index
    r')[A-Z]? *$'  # optional coord version and trailing blanks
)
```

It's much less dense as a regexp, but makes it clearer exactly what's being matched, and easier to extend if we discover other keywords that require RVKC support.

dhomeier:
I like that pattern much better, definitely makes it clearer how the keywords should look like.
Actually `C[PQ]DIS`_j_ and `*ERR` are syntactically plain string/float-valued keywords as listed in Table 2 of the Distortion Paper, so they will not need special RVKC parsing. That would leave the RE at the `DP|DQ` variations plus `D2IM` and `DF`_nn_ which is tested in `test_overly_aggressive_rvkc_lookup`.
```python
_keywd_RVKC = re.compile(r'('
        r'([T0-9](DP|DQ)[0-9]{1,3})|'    # keywords accepting pixel list prefix or table index
        r'((DP|DQ|DF|D2IM)[0-9]{1,2})|'  # keywords accepting coord index
        r') *$'                          # trailing blanks
        )
```
From the Distortion Paper I am quite sure that the `*D[PQ]` keywords always end in a single number, but in principle that number may of course have two or even three digits (no idea about `DF`), so the 2-3 digit patterns here are probably sufficiently specific and general.

pllim:
Sorry for the duplicate bot posts. I fixed the bug upstream, hopefully.

dhomeier:
> @dhomeier - Do you think you could rebase and update the regex as discussed above ?

@saimn is that the version from https://github.com/astropy/astropy/pull/10357#issuecomment-775204793 now?

dhomeier:
> About the verify warnings, the problem is that the serialization of mixin columns is using a `.` as separator, and dot is not part of the "recommended" characters... It is a bit unfortunate that files written by astropy trigger warnings, especially if the user can do nothing about this. It's not clear to me exactly how the serialization works and if we could change the separator (but there are not many choices, only `_` which is not ideal). So maybe the solution would be to allow `.` in the TTYPE regex, it would be an exception to the recommendation but I guess we can live with that and I don't see many other choices.

@saimn, @embray how do you think about checking for
```python
TTYPE_RE = re.compile(r'[0-9a-zA-Z_]+[0-9a-zA-Z_.]*$')
```
for rather generally permitting the mixin column style, i.e. additionally allow the `.` anywhere but in the first position (or should we exclude trailing periods as well)?

saimn:
> @saimn is that the version from https://github.com/astropy/astropy/pull/10357#issuecomment-775204793 now, or the original https://github.com/astropy/astropy/pull/10357#issuecomment-774694287?

I think the version you proposed in https://github.com/astropy/astropy/pull/10357#issuecomment-775204793

saimn:
> additionally allow the . anywhere but in the first position

Sounds good to me :+1:

Issue #10342, opened 2020-05-11
Of required and optional dependencies


pllim:
## Description

As `astropy` grows over the years, we now have different kinds of dependencies:

* Absolutely required and documented as such: `numpy`
* Required for certain jobs, e.g., building docs and testing.
* Declared optional but really required for some subpackages, such as but not limited to:
    * `matplotlib` for `visualization.wcsaxes`
    * `asdf` for `io.misc.asdf` (this one is even cyclic, i.e., we also depend on it for schemas version or something)
    * `PyYAML` for `io.misc.yaml` and maybe even `io.fits` (see #7783) 
* Truly optional, i.e., there is a fallback alternative if the dependency is missing -- Do we really have this? Need to double check.

All the dependencies are declared in [the installation page](https://docs.astropy.org/en/latest/install.html) but unfortunately this page is purely informational and does not enforce anything code-wise; i.e., it will not automatically downloads `asdf` for you if you use `astropy.io.misc.asdf`. (But do we even want that?)

The complexity of these dependencies occasionally cause problems when they are missing or incompatible, resulting in but not limited to:

* Unable to use part of the code directly.
* Unable to use code in downstream package that uses our affected code.
* Unable to test the code effectively, causing bugs to slip through during development process. (Either forget to test with or without the dependency.)

Some workaround to make the dependency optional also caused performance issue; e.g., #10302 .

## Things we could do to improve the situation

* Import-only test job as discussed in #7939 . Such a job will ensure things like, but not limited to:
   * `pytest` not required at runtime.
  * Optional dependencies don't cross-pollinate; e.g., `io.fits` would work even if `PyYAML` is not present, or `visualization.wcsaxes` is not affected by the presence or lack of `asdf`.
  * Importing `astropy` without any optional dependencies will not cause `ImportError`.
  * Defining special cases where `ImportError` is acceptable; e.g., `io.misc.asdf` can throw `ImportError` if `asdf` is not installed.
    * However, will this accidentally let real problems through?
    * Can we overload `__doctest_requires__` for this purpose or is subpackage `conftest.py` incantation sufficient?
    * Perhaps even a new `__doctest_subpackage_requires__` directive in `pytest-doctestplus` to avoid recursion into a subpackage.
* Skip test collection in `setup_package.py` to avoid problem with requiring `extension-helpers`; e.g. #10246 .
* Do not use `pytest.importorskip` in non-test code; e.g., #10302 vs #10325.
[...]

PR #10358, opened 2020-05-16
Skip doctest collection for subpackages depending on available dependencies

astrofrog:
This is an experiment to try out the proposed fix for https://github.com/astropy/pytest-doctestplus/pull/112 which allows doctest collection to be skipped per subpackage based on requirements.

The first commit's CI should fail before I actually define any options in ``setup.cfg`` (this is a deliberate test). Once I've checked that it fails, I'll push the fixes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -114,6 +114,10 @@ filterwarnings =
     ignore:PY_SSIZE_T_CLEAN will be required for '#' formats
     ignore:::astropy.tests.plugins.display
     ignore:::astropy.tests.disable_internet
+doctest_norecursedirs =
+    */setup_package.py
+doctest_subpackage_requires =
+        astropy/io/misc/asdf/*:asdf
 
 [bdist_wininst]
 bitmap = static/wininst_background.bmp
--- a/setup.cfg
+++ b/setup.cfg
@@ -117,7 +117,7 @@ filterwarnings =
 doctest_norecursedirs =
     */setup_package.py
 doctest_subpackage_requires =
-        astropy/io/misc/asdf/*:asdf
+    astropy/io/misc/asdf/*=asdf
 
 [bdist_wininst]
 bitmap = static/wininst_background.bmp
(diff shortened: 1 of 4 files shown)


astrofrog:
Ok so that fails nicely. Now pushing the skipping configuration!

astrofrog:
Ok so let's see if that works. Note to self that I should find a way to avoid the following warning during collection though:

```
collecting ... The 'asdf' distribution was not found and is required by the application
collecting 8561 items / 22 skipped / 8539 selected
```

astrofrog:
It works!

setup.cfg
```diff
@@ -114,6 +114,10 @@ filterwarnings =
     ignore:PY_SSIZE_T_CLEAN will be required for '#' formats
     ignore:::astropy.tests.plugins.display
     ignore:::astropy.tests.disable_internet
+doctest_norecursedirs =
+    */setup_package.py
+doctest_subpackage_requires =
+    astropy/io/misc/asdf/*=asdf
```
pllim, line 120: Spaces around `=` is even more consistent! 😉 
```suggestion
    astropy/io/misc/asdf/* = asdf
```

astrofrog:
> Hmm, doesn't appear we have a job that runs without optional dependencies here but that is out of scope (see #10342).

Good point, I'll add a job to the Travis matrix with minimal dependencies, in the initial stage of testing

astrofrog:
@pllim - to test this more, I've also (for now) reverted some of the changes in https://github.com/astropy/astropy/pull/10257 since we can now more easily skip the doctest collection, and I think the current structure is more convoluted with checks and exceptions all the time. I think it's reasonable to expect that importing astropy.io.misc.yaml should fail if PyYAML is not installed.

astrofrog:
(still experimental btw, do not merge!)

pllim:
Re: `yaml` -- Yes, sound fine. Thanks for cleaning up my mess!

mhvk:
Belated :+1: to this change - very nice to have the `yaml.py` file cleaned up (even though it was made messy only out of necessity).

PR #10359, opened 2020-05-16
BugFix: Bug 10319 fix input_units_equivalencies key error

### LINKED ISSUE #10319: issue with fix_inputs and default equivalencies
`Model.input_units_equivalencies` crashes when the model is the result of `fix_inputs` operation and the constituent models have no equivalencies set.

```
p = Polynomial1D(1, c0=1, c1=2)
m = p & p
mf = fix_inputs(m, {'x0': 1})
mf.input_units_equivalencies                                                 
<ipython-input-59-92566a4c98dd> in <module>
----> 1 mf.input_units_equivalencies

~/dev/astropy/astropy/modeling/core.py in input_units_equivalencies(self)
   3128         inputs_map = self.inputs_map()
   3129         return {key: inputs_map[key][0].input_units_equivalencies[orig_key]
-> 3130                 for key, (mod, orig_key) in inputs_map.items()
   3131                 if inputs_map[key][0].input_units_equivalencies is not None}
   3132 

~/dev/astropy/astropy/modeling/core.py in <dictcomp>(.0)
   3129         return {key: inputs_map[key][0].input_units_equivalencies[orig_key]
   3130                 for key, (mod, orig_key) in inputs_map.items()
-> 3131                 if inputs_map[key][0].input_units_equivalencies is not None}
   3132 
   3133     @property

KeyError: 'x1'

```
It looks like `fix_inputs` changes the default value of `input_units_equivalencies` of the constituent models from `None` to `{}`. This results in a crash when trying to access the property.
From the above example, accessing the property on the child models gives different results.
```
m.inputs_map()['x1'][0].input_units_equivalencies
None

mf.inputs_map()['x1'][0].input_units_equivalencies
{}
```

Iamsoto:
### Description
As mentioned on issue #10319 
_Model.input_units_equivalencies crashes when the model is the result of fix_inputs operation and the constituent models have no equivalencies set._

I believe this to be a bug within the input_units_equivalencies method in modeling.core. Where the method will recursively call a key that does not exist. I have implemented a simple one-line fix for this problem with the hope this will resolve all problems similar to the one stated. 

This will be my first pull request to the project, and I hope it helps. 

**UPDATE:** 

This current pull request will fix the key error as mentioned in #10319, but there is still another discrepencey. Specifically this one:  

`m.inputs_map()['x1'][0].input_units_equivalencies`
(returns)
None

`mf.inputs_map()['x1'][0].input_units_equivalencies`
(returns)
{}

So I did some digging and this is what I've found out.
`CompoundModel `has the property function `input_units_equivalencies` but the Base `Model ` class simply has the attribute `input_units_equivalencies = None` .

Why does this matter? Let's look at the recursive logic for
`mf.inputs_map()['x1'][0].input_units_equivalencies`. Bare with me here.
.

`Polynomial1D.input_unit_equivalencies` (returns None)--> `CompoundModel.input_unit_equivalencies` (returns {}) --> output

Where` CompoundModel.input_unit_equivalencies` returns {} due to python's list comprehension.

And if we add one more layer of recursion, as we do when we call `fix_input( ... ).input_unit_equivalencies` we get a key error. Note `{}['x1']`.
But this current pull request will take care of the key error.

If we don't want inconsistent default values, I BELIEVE we can change
`input_unit_equivalencies = None` in astropy/core.py Model to
to:
`input_unit_equivalencies = {}`
But I'd like someone else's input before I change default values.

I hope this all makes sense and will be helpful.

Fixes #10319

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -3236,9 +3236,12 @@ class CompoundModel(Model):
     @property
     def input_units_equivalencies(self):
         inputs_map = self.inputs_map()
-        return {key: inputs_map[key][0].input_units_equivalencies[orig_key]
+        input_units_equivalencies_dict = {key: inputs_map[key][0].input_units_equivalencies[orig_key]
                 for key, (mod, orig_key) in inputs_map.items()
-                if inputs_map[key][0].input_units_equivalencies is not None}
+                if inputs_map[key][0].input_units_equivalencies is not None }
+        if not input_units_equivalencies_dict:
+            return None
+        return input_units_equivalencies_dict
 
     @property
     def input_units_allow_dimensionless(self):
--- a/astropy/modeling/tests/test_quantities_evaluation.py
+++ b/astropy/modeling/tests/test_quantities_evaluation.py
@@ -9,7 +9,6 @@ import numpy as np
 import pytest
 from numpy.testing import assert_allclose
 
-
 from astropy.modeling.core import Model
 from astropy.modeling.models import Gaussian1D, Shift, Scale, Pix2Sky_TAN
 from astropy import units as u
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -721,6 +721,10 @@ astropy.io.votable
 astropy.modeling
 ^^^^^^^^^^^^^^^^
 
+- bugfix. ``Model.input_units_equivalencies`` would throw keyerror on fix_inputs
+  if models have no set unit equivalencies.
+  [#10359]
+
 astropy.nddata
 ^^^^^^^^^^^^^^
 
--- a/astropy/modeling/tests/test_models_quantities.py
+++ b/astropy/modeling/tests/test_models_quantities.py
@@ -7,6 +7,7 @@ import numpy as np
 
 from astropy import units as u
 from astropy.tests.helper import assert_quantity_allclose
+from astropy.modeling.core import fix_inputs, CompoundModel
 
 from astropy.modeling.functional_models import (
     Gaussian1D,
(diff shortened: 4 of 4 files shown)


nden:
@lamsoto Thanks for looking into this.
I am hesitant to change the default value of `Model.input_units_equivalencies` because there might be code checking if it's None.
I can look at it from a different angle and say that `CompoundModel.input_units_equivalencies` needs to return `None` if the constituent models don't have equivalencies set. So the fix to this problem would be to check if the dictionary is empty and if so return `None`. What do you think?

Iamsoto:
@nden In my head a default value of {} made more sense. But Icompletely understand the arguement for making it None. 

The fix shouldn't be too hard. So I'll make the necessary changes + tests then push them! 

Cheers!

Iamsoto:
Hey, the appropriate changes should be here now, I apologize for not making that clear

nden:
It's my bad, I was swamped with other things. I will review it today.

Iamsoto:
No worries! Take your time

astropy/modeling/tests/test_quantities_evaluation.py
```diff
@@ -317,6 +318,40 @@ def test_compound_input_units_equivalencies():
...
+                assert fixed_input_model.inputs_map()['x1'][0].input_units_equivalencies == None
+
+            elif op == '|':
+                test_compound_model = test_model | test_model
+                assert test_compound_model.inputs_map()['x'][0].input_units_equivalencies == None
+                fixed_input_model = fix_inputs(test_compound_model, {'x':1})
+
+            assert fixed_input_model.input_units_equivalencies == None
+
+
```
nden, line 354: Please write this test as a parametrized pytest test. Here's an [example](https://github.com/astropy/astropy/blob/1b8b4d17864223678003e8867852c19369ea28ce/astropy/modeling/tests/test_models_quantities.py#L213). 

```
@pytest.mark.parametrize('model', [Shift(10 * u.deg), Polynomial1D(1, c0=1, c1=2), Scale(2)])
def test_compound_model_input_units_equivalencies_defaults(model):
```
This will pass one model at a time.
Don't loop over operators, simply write the expression for the compound model using each operator.

nden:
And please rebase and squash commits at the end.

Iamsoto:
Thanks for taking a look at this, I'll get to work on what you've written

astropy/modeling/tests/test_quantities_evaluation.py
Iamsoto: Done.

After seeing the models in `test_models_quantities.py` I decided the move the test their and use the same models as parameters. I hope it still makes sense for the test to be there

astropy/modeling/core.py
```diff
@@ -3107,9 +3107,10 @@ def input_units(self):
     @property
     def input_units_equivalencies(self):
         inputs_map = self.inputs_map()
-        return {key: inputs_map[key][0].input_units_equivalencies[orig_key]
+        results = {key: inputs_map[key][0].input_units_equivalencies[orig_key]
                 for key, (mod, orig_key) in inputs_map.items()
-                if inputs_map[key][0].input_units_equivalencies is not None}
+                if inputs_map[key][0].input_units_equivalencies is not None }
+        return results if results != {} else None
```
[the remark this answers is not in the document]
Iamsoto: Done!

Iamsoto:
Apologies. I forgot about this PR for a while

pllim:

WilliamJamieson:
@Iamsoto this PR seems to fix the described bug and does not appear to cause any issues with other modeling functionalities. 

So PR should be reopened and rebased. The changlog will need to be updated using the new [changelog fragments](https://github.com/astropy/astropy/blob/main/docs/changes/README.rst). Once all the CI has passed I think we can merge these changes. I can assist you with doing these tasks or do them for you if you would like.

nden:
I think I let this fall through the cracks. Sorry about this.

pllim:
Ops... You should remove the "Close?" label too. I'll remove it.

pllim:
I think the bot will give you 30 more days. If you rebase, you get 6 more months. Or if the original author is no longer interested, a maintainer can take over.

PR #10360, opened 2020-05-17
Faster parsing of fixed-format time strings in Time

taldcroft:
### Description

This is a PR to greatly speed up parsing of year, mon, day-of-month (or day-of-year), hour minute second string inputs.  On my machine for large input arrays (e.g. 100000) of ISO or ISOT times it speeds up creation of the time object by a factor of 25. For year day-of-year ('yday') format the speed-up is closer to a factor of 50.

This PR adds a time config option `use_fast_parser` which can take values of `'True'`, `'False'`, or `'force'`. The latter is mostly for testing to force use of the C parser and raise an exception if it fails. With a value of `'True'` then the fast parser will be tried but falls through to the Python parser if it fails.

This includes a somewhat-flexible format specification protocol that can be used for variations on the theme. For instance Chandra ops uses an odd format `'YYYYDDD.hhmmssfff'` (where `fff` is fractional part of the second).

To do:

- [x] Tests
- [x] Docs


pllim:
If this moves forward, might need a benchmark over at https://github.com/astropy/astropy-benchmarks/

mhvk:
Had a very brief look only. Just one note: on a somewhat similar chase of the ultimate in precision, I found it was quite hard to beat the python built-in `int` and `float` with anything in numpy even for substantial arrays. A loop that replaces the regex might be quite fast. That said, here we already have numpy's U or S format...

Another solution, once you have wrapped up how to do it reliably here, would be to write a C routine that does the conversion for a single element and then create a ufunc from it (or perhaps a gufunc acting on the uint8 view, as you have here). This is not as hard as it may sound, and would be fairly easily extended to the other time formats.

taldcroft:
@mhvk - this PR now includes a proof-of-concept for a pure C parser. I just stuck it in the top-level directory for now.

My testing is basically:
```
gcc parse_time.c; ./a.out "2000-01-01 12:23:45.5665"
```

I am very rusty at C, but it seems to work and is blindingly fast, around 70 ns per time. This compares to around 700-800 ns for the original numpy/Python solution (which is a factor of 9 faster than the current code in master). 

Any thoughts on the best direction forward for gluing this into astropy?  I would go for ctypes, but just because I have used that before and it is pretty simple. My expertise here is quite limited though, any advice appreciated.

taldcroft:
For now I'm working on using ctypes to glue this in so I can measure end-to-end performance. I am working on an idea for a slightly more generalized interface to support different formats (in particular yday).

taldcroft:
I pushed up more POC files, a Makefile and a demonstration with Python.  From the top-level astropy directory:
```
make all
python parse_time.py
parse time = 168 nsec
total time = 276 nsec
```

Using astropy 4.0.1 the time to create a Time object (per ISO time for n_times > 1e6) is 7800 nsec, so this is a speedup of at least 25. Yay.

One cheat at the moment is that it assumes the UTF-32 numpy array is plain ASCII and does no checking that the other 3 bytes in each character are zero. Surprisingly doing this check in numpy was extremely slow, so I took it out. Hopefully this can be done in C without a huge penalty.
```
if val1.dtype.kind == 'U':
    chars.shape = (-1, 4)
    assert np.all(chars[:, 1:4] == 0)  # Took over 250 nsec
    chars = chars[:, 0]
```

mhvk:
I think the logical thing would be to wrap it in a ufunc, which is easy in principle but a bit tricky for strings in practice. But that can obviously be a second step (unless you're having problems with multidimensional arrays - that is the one thing ufuncs bring for free!). The nice thing with having a C/ufunc framework is that we can extend it easily to other conversions where we currently use the very slow `nditer` approach.

taldcroft:
Looking at https://numpy.org/devdocs/user/c-info.ufunc-tutorial.html I'm trying to understand what making this into a ufunc really wins vs. expected performance degradation due all the fancy wrapping. At least for me the top priority is performance rather than generalizability for other applications.

I might be limiting the scope of ideas, but basically I only see this being useful (in practice for performance critical applications) for parsing ISO-like times and times that are like YYYY:DDD:hh:mm:ss.fff, and variants.  For Chandra ops there is a bizarre but widely used format YYYYDDD.hhmmssfff that could also be parsed in this framework (and performance here does matter). Where else might this basic idea be utilized?

mhvk:
The win for the current would be mostly for arbitrarily shaped arrays, where one avoids a possible copy in flattening. But perhaps the bigger win would be that it can be extended easily: it would get rid of all the `nditer` constructs in `formats`. Anyway, as noted, it can easily be done after the fact, so better first to get the basic code working! 

p.s. Might be good to move reasonably quickly to `setup_package.py` instead of `Makefile`, though... the one in `convolution` may be a nice one to start from.

taldcroft:
@mhvk - I think this is getting pretty close to being ready for review in terms of the actual code. Of course I still need to do tests. As you can see this is pretty focused on a narrow but important use case.  I don't actually think that ufunc generalization of what I've done here with the new C parser would help to get rid of other `nditer` calls or speed those cases up (though to be honest I only gave it a 30 second scan so maybe wrong there).

pllim:

astropy/time/formats.py
```diff
@@ -1405,6 +1551,28 @@ class TimeYearDayTime(TimeISO):
...
+    #   012345678901234567890
+    #   yyyy:ddd:hh:mm:ss.fff
+    # Parsed as ('yyyy', ':ddd', ':hh', ':mm', ':ss', '.fff')
+    #
+    # delims: character at corresponding `starts` position (0 => no character)
+    # starts: position where component starts (including delimiter if present)
+    # stops: position where component ends (-1 => continue to end of string)
+
+    # Before: yr mon  doy     hour      minute    second    frac
+    use_fast_parser = True
```
mhvk, line 1568: Would it be an idea to abstract this out a bit more, so that the class perhaps defines the parser proper instead of the pieces needed for it? I think it might be a bit more future proof. I.e., something like
```
fast_parser = CParserWrapper(delims=..., starts=..., stops=..., ...)
```
and then the `CParserWrapper.__call__` method would have most of what is now in  `set_jds_fast`.

mhvk: Another option (closer to what you have) would be to here define `fast_parser_pars = dict(delims=..., starts=..., ...)`.

But I think I'd prefer the option of a user-defined parser.

Though I can also see the argument that really the `TimeFormat` class *is* the parser so this just belongs here...

mhvk:
@taldcroft - sorry for having been so slow in reviewing. Even now I fear I don't really have the time to give this a proper review, and my off-the-cuff comments may be less than useful. 

I should be clear that I think the approach here is good. So, if someone else has time to review and approve, that would be totally fine by me (this is not that `astropy.time` specific, so I don't think it needs a "specialist"!). If not, I'll try to get to this in a week or so.

astropy/time/formats.py
taldcroft: I see the benefit of encapsulating the parameters in a single attribute `fast_parser_pars`. The downside is that you cannot easily use attribute inheritance ala `TimeISO` and `TimeISOT`. But repeating those pars in `TimeISOT` is not the end of the world, so I'll plan on that new single container.

About abstracting this out, I've become much more resistant to making abstractions to solve a problem that we don't have yet. This functionality can be fully customized by a subclass that overrides `set_jds_fast`. This seems simple and gives subclasses the ability to do whatever is needed if they invent a different fast parser.

mhvk: OK, makes sense not to abstract out. One thought to make it easier to extend in future if there ever is a need: how about if the python and fast versions not not set jd1, jd2 themselves but rather return jd1, jd2 and those are then stored on the class inside `set_jds`?

astropy/time/formats.py
```diff
@@ -1261,6 +1336,58 @@ def set_jds(self, val1, val2):
                               *iterator.operands[1:])
         self.jd1, self.jd2 = day_frac(jd1, jd2)
```
mhvk, line 1337: Concretely, the suggestion would be to `return day_frac(jd1, jd2)` here

astropy/time/formats.py
```diff
@@ -1261,6 +1336,58 @@ def set_jds(self, val1, val2):
...
+            chars, n_times, val1_str_len, self.has_day_of_year,
+            delims, starts, stops, break_allowed,
+            year, month, day, hour, minute, second)
+        if status == 0:
+            # All went well, finish the job
+            jd1, jd2 = erfa.dtf2d(self.scale.upper().encode('ascii'),
+                                  year, month, day, hour, minute, second)
+            jd1.shape = val1.shape
+            jd2.shape = val1.shape
+            self.jd1, self.jd2 = day_frac(jd1, jd2)
```
mhvk, line 1382: And here.

astropy/time/formats.py
```diff
@@ -1241,6 +1297,25 @@ def parse_string(self, timestr, subfmts):
...
 
     def set_jds(self, val1, val2):
+        """Parse the time strings contained in val1 and set jd1, jd2"""
+        # If specific input subformat is required then use the Python parser.
+        # Also do this if Time format class does not define `use_fast_parser`
+        # or if the fast parser is entirely disabled.
+        if (self.in_subfmt != '*'
+                or not self.__class__.__dict__.get('use_fast_parser')
+                or conf.use_fast_parser == 'False'):
+            self.set_jds_python(val1, val2)
```
mhvk, line 1307: And then obviously assign to `self.jd1, self.jd2` here and below. Advantage is that it separates the tasks of calculating jd1 and jd2 from setting.

astropy/time/formats.py
taldcroft: Done

taldcroft:
@mhvk @dhomeier - FYI from my perspective this is ready for review. All CI was passing in c526f0f except for the change log entry. I plan to squash out the first few commits which were the original numpy-based version of this, but will wait on that pending review.

@dhomeier - if you can look at the C code and see if there are any obvious problems that would be great. My C is extremely rusty.

taldcroft:
@mhvk - gentle reminder, 4.2 feature freeze in 8 days Friday Oct 23.

astropy/time/formats.py
```diff
@@ -1259,7 +1346,64 @@ def set_jds(self, val1, val2):
...
+            if status != 0:
+                raise ValueError('input is not pure ASCII')
+            # It might be possible to avoid this copy with cleverness in
+            # parse_times.c but leave that for another day.
+            chars = chars[::4]
+        chars = np.ascontiguousarray(chars)
+
+        # Pre-allocate output components
+        n_times = len(chars) // val1_str_len
+        year = np.zeros(n_times, dtype=np.intc)
```
mhvk, line 1372: Maybe just `np.empty`? (Faster and we know that these will be filled.)

astropy/time/formats.py
```diff
@@ -1259,7 +1346,64 @@ def set_jds(self, val1, val2):
...
+        month = np.zeros(n_times, dtype=np.intc)
+        day = np.zeros(n_times, dtype=np.intc)
+        hour = np.zeros(n_times, dtype=np.intc)
+        minute = np.zeros(n_times, dtype=np.intc)
+        second = np.zeros(n_times, dtype=np.double)
+
+        # Set up parser parameters as numpy arrays for passing to C parser
```
mhvk, line 1379: Can be later, but in principle this could be done ahead of time (at class creation time, like `subfmts`).

astropy/time/formats.py
```diff
@@ -1259,7 +1346,64 @@ def set_jds(self, val1, val2):
...
+        pars = self.fast_parser_pars
+        delims = np.array(pars['delims'], dtype=np.uint8)
```
mhvk, line 1381: Would it be an idea to create a structured array? Those may to C-side structs, so are easy to use and it would mean only passing one item on.

If one goes that route, one could also create a structured output array.

astropy/time/formats.py
```diff
@@ -1187,7 +1220,38 @@ class TimeString(TimeUnique):
...
+
+    - ``stops`` (tuple of int): position where component ends. Use -1 to
+      continue to end of string, or for the month component for formats that use
+      day of year.
+
+    - ``break_allowed`` (tuple of int): if true (1) then the time string can
+          legally end just before the corresponding component (e.g. "2000-01-01"
+          is a valid time but "2000-01-01 12" is not).
+
+    - ``has_day_of_year`` (int): 0 if dates have year, month, day; 1 if year,
```
mhvk, line 1253: Could one just infer this based on the month being absent?

astropy/time/formats.py
```diff
@@ -1242,6 +1306,29 @@ def parse_string(self, timestr, subfmts):
 
     def set_jds(self, val1, val2):
         """Parse the time strings contained in val1 and set jd1, jd2"""
+        # If specific input subformat is required then use the Python parser.
+        # Also do this if Time format class does not define `use_fast_parser` or
+        # if the fast parser is entirely disabled. Note that `use_fast_parser`
+        # is ignored for format classes that don't have a fast parser.
+        if (self.in_subfmt != '*'
```
mhvk, line 1313: Could one reverse the logic?
```
if (self.in_subfmt == '*'
        and 'fast_parser_pars' in self.__class__.__dict__
        and conf.use_fast_parser != 'False'):
    try:
        self.jd1, self.jd2 = self.get_jds_fast(val1, val2)
        return
    except Exception:
        # Fall through to the Python parser unless fast is forced.
        if conf.use_fast_parser == 'force':
            raise

self.jd1, self.jd2 = self.get_jds_python()
return
```

astropy/time/formats.py
```diff
@@ -1345,6 +1489,23 @@ class TimeISO(TimeString):
...
 
+    # Define positions and starting delimiter for year, month, day, hour,
+    # minute, seconds components of an ISO time. This is used by the fast
+    # C-parser parse_ymdhms_times()
+    #
+    #  "2000-01-12 13:14:15.678"
+    #   01234567890123456789012
+    #   yyyy-mm-dd hh:mm:ss.fff
+    # Parsed as ('yyyy', '-mm', '-dd', ' hh', ':mm', ':ss', '.fff')
+    fast_parser_pars = dict(
```
mhvk, line 1500: Related to the question about a structured array above: would this be clearer if defined as one? I.e., something like
```
fast_parser_pars = np.array([
    (0, 0, 3, 0),
    (ord('-'), 4, 6, 0),
    ...],
    dtype=[('delims', np.intc), ('starts', np.intc), ('stops', np.intc), 'break_allowed', np.intc)])
```

mhvk (approved):
Overall, this looks very nice!  I have a number of in-line comments, but I don't feel any of them is essential to address; I think the benefit for the existing time formats amply justifies the code.

My only worry is baking in the API in for possible subclasses. For instance, I think the code will become cleaner with structured dtypes (structs on the C side), and I think it would be fairly easy to use the C code as the guts of a `ufunc` that would do the shaping, etc., automatically (making `get_jds_fast` much simpler); one could also imagine calling the erfa C function from the inside. Given this, my tendency would be to mark the implementation as experimental (and/or make the new attribute and class property private), leaving us the freedom to change it (e.g., for a string based class I have myself, I would like to use something similar but I cannot as it organizes seconds as an int and a fraction).

In the meantime, I will approve this, as also as is it is fine to go in (and many of the things above can be done without changing anything in the API).

p.s. Somewhat sadly, the whole ISO string parsing is implemented in the C implementation of `datetime.fromisostring` in CPython, but unfortunately it is baked in to return a `datetime` object.

taldcroft:
@mhvk - thanks. Your comments look reasonable. I like the idea of structs / structured arrays, so I'll see if I can get that going. Also, you make a good point about keeping the seconds as the int part and float fractional part for flexibility, so I will make the C part leave those separate and then the Python code can do that addition (of 2 numpy arrays) at minimal cost.

If I do all that would you still want this to be labeled experimental?

mhvk:
@taldcroft - hope the structured dtype indeed makes it (even) better!

As for my request for experimental: it is probably not really needed regardless, since right now it is advertised that one can use it only in the docstring. And your current API already allows one just to define `fast_parser_args` to whatever is handy and then override `set_jds_fast()` (or `set_jds()` if one has a fast parser that does accept `subfmt`). 

That said, I could see the case for moving the call to the erfa function also up to the C code, which means the code itself might return something different...

Anyway, overall, it would seem quite possible to change things through the usual deprecation mechanisms, so let's not bother with experimental warnings.

mhvk:
p.s. Realize I forgot something that I had wondered about a few days ago: are there good benchmarks already that show the benefit of the new code (and which will help ensure that any new implementation doesn't regress)?

astropy/time/formats.py
taldcroft: I think this representation is significantly less clear so I'm staying with the original. Basically transposing the original structure makes it nearly impossible to correlate the parser elements to year, month, day, hour, minute, seconds. In the current representation this is pretty clear.

astropy/time/formats.py
taldcroft: I agree this *looks* cleaner.  But I dropped this in and it failed most tests. Since the original is working I'll leave it.

astropy/time/formats.py
taldcroft: Maybe. OTOH explicit is better than implicit. Leaving this.

astropy/time/formats.py
taldcroft: I have the structured output array working and this is much nicer. For inputs I'm leaving them separate (see later comment).

astropy/time/formats.py
taldcroft: In the best case (just one time) this would reduce runtime by about 4 us out of 145 us.

It is certainly possible to put this code up into TimeFormatMeta as a special case `if 'fast_parser_pars' in members:`, but it increases code complexity by moving code into a somewhat unrelated/random place. Given only the 2-3% performance improvement I'm going to leave this be.

taldcroft:
>  are there good benchmarks already that show the benefit of the new code

Sadly, no.

astropy/time/formats.py
taldcroft: > not set jd1, jd2 themselves but rather return jd1, jd2

Done.

astropy/time/formats.py
taldcroft: Done. For some reason I have a memory that `empty` was not especially faster than `zeros`, but I checked and that is clearly not the case. I changed the code and my memory.

taldcroft:
@mhvk - I have marked all the comments as "resolved". I think the important bits that somewhat lock in the API have changed for the better, in particular the structured array for outputs and keeping integer and fractional seconds separate.  (I hope you can use this now for your custom format!). I also realized the new class properties should not be on the base `TimeFormat` but just on `TimeString`.

p.s. that C code `#pragma pack(4)` took me about an hour to figure out. The compiler was inserting 4 empty bytes within the structure (presumably to make it fit into a multiple of 8 bytes). This was giving SEGV and nothing making sense, and I was nearly to the point of giving up.

mhvk:
Oops, I should have thought about this.  I remeber `np.dtype` has a relevant option...

Ah, yes:
```
align : bool, optional
    Add padding to the fields to match what a C compiler would output
    for a similar C-struct. Can be ``True`` only if `obj` is a dictionary
    or a comma-separated string. If a struct dtype is being created,
    this also sets a sticky alignment flag ``isalignedstruct``.
```

taldcroft:
@mvhk - are you still good with this for merge?

mhvk:
Yes, good to merge!

mhvk:
Damn! This is not big-endian proof... See https://travis-ci.com/github/astropy/astropy/jobs/405392253
I'll try to make a fix as part of #10908.

PR #10362, opened 2020-05-17
WIP: fix input broadcasting in modeling

### LINKED ISSUE #9953: Modeling issues with input shape validation

### LINKED ISSUE #10170: Model evaluation crashes if some inputs are scalars

nden:
This PR is functionally complete but I marked it as WIP to get opinions on the approach. I will add documentation later.

This takes a slightly different approach to validating inputs in modeling. 
The previous behavior was:

- Input shapes were compared (in _validate_input_shapes) and *if not the same* an error was raised.
- If `Model.standard_broadcasting=True` parameters and inputs were broadcasted to a common shape.
- There was no way to specify that inputs should not be broadcast among themselves or with parameters.

The new approach:

- `Model.standard_broadcasting=True` means that broadcasting rules apply to inputs and parameters. `prepare_inputs` is not broadcasting the inputs before passing them to `Model.evaluate` (although it computes the broadcasted shape ) but let's the function implementation deal with this following numpy broadcasting rules.
- If `Model.standard_broadcasting=False` it is not assumed that inputs should broadcast nor that inputs and parameters should broadcast. Dealing with input and output shapes is left to the writer of the model. This is a rare use case but is necessary.
- A special treatment is the case of `n_outputs > n_inputs` with `standard_broadcasting=True`. In this case the shape fo the extra outputs is assumed to be the precomputed broadcasted shape.

Fixes #10170 
Fixes #9953

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -1724,7 +1724,6 @@ class Model(metaclass=_ModelMeta):
 
     def prepare_outputs(self, format_info, *outputs, **kwargs):
         model_set_axis = kwargs.get('model_set_axis', None)
-
         if len(self) == 1:
             return _prepare_outputs_single_model(outputs, format_info)
         else:
@@ -3724,8 +3723,9 @@ def render_model(model, arr=None, coords=None):
 
 
 def _prepare_inputs_single_model(model, params, inputs, broadcasted_shape, **kwargs):
-    #broadcasts = []
     shapes_of_inputs = [inp.shape for inp in inputs]
+    if model.standard_broadcasting:
+        shapes_of_inputs = [check_broadcast(shape, broadcasted_shape) for shape in shapes_of_inputs]
     param_broadcast = ()
 
     for idx, _input in enumerate(inputs):
@@ -3738,14 +3738,11 @@ def _prepare_inputs_single_model(model, params, inputs, broadcasted_shape, **kwa
             inputs[idx] = _input.reshape((1,))
 
     if params and model.standard_broadcasting:
-        #print('params', params)
         param_shapes = [param.shape for param in params]
 
         param_broadcast = check_broadcast(*param_shapes)
-        #print('param_broadcast', param_broadcast)
         if broadcasted_shape is not None:
             broadcasted_shape = check_broadcast(broadcasted_shape, param_broadcast)
-        #print('broadcasted_shape', broadcasted_shape)
         shapes_of_inputs = [check_broadcast(shape, param_broadcast) for shape in shapes_of_inputs]
 
     if model.n_outputs > model.n_inputs:
@@ -3756,7 +3753,6 @@ def _prepare_inputs_single_model(model, params, inputs, broadcasted_shape, **kwa
             # inputs necessary (see _prepare_outputs_single_model)
             shapes_of_inputs.append(None)
         shapes_of_inputs.extend([broadcasted_shape] * extra_outputs)
-
     return inputs, (shapes_of_inputs,)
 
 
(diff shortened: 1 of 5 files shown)


nden:
@chris-simpson

perrygreenfield:
It would be good to give an example of `Model.standard_broacasting=False` to make it clear (e.g., the model is doing some weird shape changing; what do they call that? Oh yeah, a shapeshifter :-).

That is, point to such a model that does that.

pllim:
> shapeshifter :-)

Does it only happen during full moon? :D

astropy/modeling/core.py
```diff
@@ -2657,7 +2658,7 @@ def __call__(self, *args, **kw):
         if with_bbox and bbox is not None:
             # first check inputs are consistent in shape
             input_shape = _validate_input_shapes(args, (), self._n_models,
-                                                 self.model_set_axis, self.standard_broadcasting)
+                                                 self.model_set_axis)
```
perrygreenfield, line 2661: Will this cause problems if self.standard_broadcasting is False (well, it looks like this doesn't change past behavior, but still, is it not a problem?)

astropy/modeling/tests/test_input.py
```diff
@@ -961,3 +961,113 @@ def test_call_keyword_mappings(model):
...
+    result = c([1, 1], 1)
+    expected = [[1., 1.], [0., 0.], 2.]
+    for r, e in zip (result, expected):
+        assert_allclose(r, e)
+
+    c = M(1)
+    result = c(1, [1, 1])
+    expected = [1., 0., [2., 2.]]
+    for r, e in zip (result, expected):
+        assert_allclose(r, e)
```
perrygreenfield, line 1073: Need a test case where inputs are not shape compatible for a `standard_broadcasting=False`

PR #10363, opened 2020-05-17
Add fallback option when image file is not in local directory

ADiegoCAlonso:
### Description

This pull request is to address the case where the image is not found in IMAGE_PATH, it will retrieve the image binary from repo.

Motivation: This code is published in other sites (e.g. [here](https://docs.astropy.org/en/stable/generated/examples/io/split-jpeg-to-fits.html)) without a link to the image file. By adding this fallback option, it increases the robustness of the code.

Fixes #<Issue Number>

### DIFF
--- a/examples/io/split-jpeg-to-fits.py
+++ b/examples/io/split-jpeg-to-fits.py
@@ -30,9 +30,18 @@ from astropy.visualization import astropy_mpl_style
 plt.style.use(astropy_mpl_style)
 
 ##############################################################################
-# Load and display the original 3-color jpeg image:
+# Load and display the original 3-color jpeg image.
+# Download image if not found locally:
+import os.path
+import urllib.request
 
-image = Image.open('Hs-2009-14-a-web.jpg')
+IMAGE_PATH = "Hs-2009-14-a-web.jpg"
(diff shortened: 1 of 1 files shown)


pllim:
Thanks, @ADiegoCAlonso ! However, I am a little confused. Under what circumstances would the script not find the input image file? It is version-controlled in the same directory as the script.

Making this example access the internet might complicate our doctest.

ADiegoCAlonso:
@pllim I see your point. Indeed the code and data in the repo is self-contained and the fallback is not needed. The reason I proposed this change is because there is an [external website](https://docs.astropy.org/en/stable/generated/examples/io/split-jpeg-to-fits.html) that includes only the code of this script without a link to the data. An alternative option would be to modify that  website, which belongs to astropy.org.

adrn:
@ADiegoCAlonso Thanks for the pull request! I agree with @pllim here that we should avoid adding a web request in here. I would instead opt for adding a comment above the `Image.open()` command with a link to the data file. However, note that this may all change significantly in the coming months as we are in the process of overhauling the example gallery.

ADiegoCAlonso:
Thank you for the review. 
Based on your comments, I proceed to close this PR.

PR #10373, opened 2020-05-19
Set up Hypothesis and add property-based tests for `Time` precision

Zac-HD:
This pull request follows issue #9059.  Its major contribution here by @aarchiba, who wrote the excellent tests for `Time` precision in #9532 and found bugs #9533 and #9579.

I have rebased on master and clarified a few Hypothesis idioms, added a section to the testing guide, and simplified the install and CI configuration (as `pytest-astropy` now depends on Hypothesis :tada:).

Reviewers please note that I am *not* an astrophysicist - I'm a Hypothesis core developer and want to support trustworthy and open scientific computing, but you'll have to bring the domain expertise :grin:

EDIT: Close #9532

### DIFF
--- a/astropy/time/tests/test_precision.py
+++ b/astropy/time/tests/test_precision.py
@@ -1,28 +1,163 @@
-import functools
-
 import decimal
-import pytest
-import numpy as np
+import warnings
+import functools
+import contextlib
 from decimal import Decimal
-from datetime import datetime
+from datetime import datetime, timedelta
 
-import astropy.units as u
+import pytest
+from hypothesis import assume, example, given, target
+from hypothesis.extra.numpy import array_shapes, arrays
+from hypothesis.strategies import (composite, datetimes, floats, integers,
+                                   one_of, sampled_from, timedeltas, tuples)
+
+import numpy as np
+
 import astropy._erfa as erfa
-from astropy.time import Time, TimeDelta
+import astropy.units as u
+from astropy.tests.helper import assert_quantity_allclose
+from astropy.time import STANDARD_TIME_SCALES, Time, TimeDelta
 from astropy.time.utils import day_frac, two_sum
 from astropy.utils import iers
-from astropy.utils.exceptions import ErfaWarning
+from astropy.utils.exceptions import ErfaError, ErfaWarning
 
-allclose_jd = functools.partial(np.allclose, rtol=2. ** -52, atol=0)
-allclose_jd2 = functools.partial(np.allclose, rtol=2. ** -52,
-                                 atol=2. ** -52)  # 20 ps atol
-allclose_sec = functools.partial(np.allclose, rtol=2. ** -52,
-                                 atol=2. ** -52 * 24 * 3600)  # 20 ps atol
+allclose_jd = functools.partial(np.allclose, rtol=np.finfo(float).eps, atol=0)
+allclose_jd2 = functools.partial(np.allclose, rtol=np.finfo(float).eps,
+                                 atol=np.finfo(float).eps)  # 20 ps atol
+allclose_sec = functools.partial(np.allclose, rtol=np.finfo(float).eps,
+                                 atol=np.finfo(float).eps * 24 * 3600)
 
-tiny = 2. ** -52
+tiny = np.finfo(float).eps
 dt_tiny = TimeDelta(tiny, format='jd')
 
 
+@pytest.fixture(scope='module')
+def iers_b():
+    """This is an expensive operation, so we share it between tests using a
+    module-scoped fixture instead of using the context manager form.  This
+    is particularly important for Hypothesis, which invokes the decorated
+    test function many times (100 by default; see conftest.py for details).
+    """
+    with iers.earth_orientation_table.set(iers.IERS_B.open(iers.IERS_B_FILE)):
+        yield "<using IERS-B orientation table>"
+
+
+@contextlib.contextmanager
+def quiet_erfa():
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=ErfaWarning)
+        yield
+
+
+def assert_almost_equal(a, b, *, rtol=None, atol=None, label=''):
+    """Assert numbers are almost equal.
+
+    This version also lets hypothesis know how far apart the inputs are, so
+    that it can work towards a failure and present the worst failure ever seen
+    as well as the simplest, which often just barely exceeds the threshold.
+    """
+    __tracebackhide__ = True
+    if rtol is None or rtol == 0:
+        thresh = atol
+    elif atol is None:
+        thresh = rtol * (abs(a) + abs(b)) / 2
+    else:
+        thresh = atol + rtol * (abs(a) + abs(b)) / 2
+
+    amb = (a - b)
+    if isinstance(amb, TimeDelta):
+        ambv = amb.to_value(u.s)
+        target(ambv, label=label + " (a-b).to_value(u.s), from TimeDelta")
+        target(-ambv, label=label + " (b-a).to_value(u.s), from TimeDelta")
+        if isinstance(thresh, u.Quantity):
+            amb = amb.to(thresh.unit)
+    else:
+        try:
+            target_value = float(amb)
+        except TypeError:
+            pass
+        else:
+            target(target_value, label=label + " float(a-b)")
+            target(-target_value, label=label + " float(b-a)")
+
+    assert abs(amb) < thresh
+
+
+# Days that end with leap seconds
+# Some time scales use a so-called "leap smear" to cope with these, others
+# have times they can't represent or can represent two different ways.
+# In any case these days are liable to cause trouble in time conversions.
+# Note that from_erfa includes some weird non-integer steps before 1970.
+leap_second_table = iers.LeapSeconds.from_iers_leap_seconds()
+# Days that contain leap_seconds
+leap_second_days = leap_second_table["mjd"] - 1
+leap_second_deltas = list(zip(leap_second_days[1:],
+                              np.diff(leap_second_table["tai_utc"])))
+
+today = Time.now()
+mjd0 = Time(0, format="mjd")
+
+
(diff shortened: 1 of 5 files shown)


astropy/time/tests/test_precision.py
```diff
@@ -1,28 +1,158 @@
...
+
+
+@contextlib.contextmanager
+def quiet_erfa():
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=ErfaWarning)
+        yield
+
+
+def assert_almost_equal(a, b, *, rtol=None, atol=None, label=''):
```
pllim, line 47: This looks like it gotta move upstream to testing utils or something.

Disclaimer: I did not follow original discussions in the other PR.

pllim:
Thanks for reviving this, @Zac-HD !

p.s. Looks like tests are failing with `HypothesisDeprecationWarning`.

Zac-HD:
> p.s. Looks like tests are failing with `HypothesisDeprecationWarning`.

Ah, bad interaction with function-scoped fixtures.  Just adding `(scope='module')` is enough to fix that, and makes things *marginally* more efficient to boot.

pllim:
Oh, looks like a real failure now. Maybe @taldcroft or @mhvk can make sense of it.

```
Falsifying example: test_conversion_never_loses_precision(
    iers_b=None, scale1='tai', scale2='tcb', jds=(-31738.500000000346, 0.0),
)
```

Zac-HD:
OK, I've marked that one test as `xfail` and everything else is working.

My preference would be to merge this PR as-is if it's otherwise acceptable, and have someone with more astro background than me investigate it in a follow-up issue - 'has an xfail test' seems to me like an improvement on the no-property-tests status quo :stuck_out_tongue_winking_eye:

pllim:
Seems a little heavy-handed to xfail the whole test when only one of the inputs failed, but then again this has dragged on long enough, so I can see pros/cons on both sides. 🤔 

Please be patient with us while we wait for the subpackage maintainers to review this thoroughly. Thanks again!

Zac-HD:
> Seems a little heavy-handed to xfail the whole test when only one of the inputs failed

That's the thing - *many* inputs failed, but Hypothesis reports a single minimal reproducing example.  It makes debugging much easier, though it *can* also make errors look less serious than they really are.

mhvk:
@Zac-HD - the failure is almost certainly due to `erfa`, in an issue that is being addressed (for overly long discussion, see #6970). I think it makes sense to `xfail` it for now, so we actually notice if things improve in the future. That said, should there be a test with double the tolerance now that does not fail, so we know we do not regress?

Zac-HD:
It doesn't seem to be a tolerance issue - I get `astropy.utils.exceptions.ErfaError: ERFA function "taiutc" yielded 1 of "unacceptable date"` with `2*dt_tiny` or `4*dt_tiny`.

astropy/time/tests/test_precision.py
```diff
@@ -142,69 +293,368 @@ def test_leap_seconds_rounded_correctly():
...
-    t2 = getattr(t, scale2)
-    assert t2.jd1 % 1 == 0
-    assert abs(t2.jd2) <= 0.5
-    assert abs(t2.jd2) < 0.5 or t2.jd1 % 2 == 0
+    # If the rates ever differ, might lose one ULP
+    with quiet_erfa():
+        t2 = t + 2*dt_tiny
+        try:
+            assert getattr(t, scale2) < getattr(t2, scale2)
+        except iers.IERSRangeError:  # UT1 conversion needs IERS data
```
mhvk, line 426: I guess we'd need a separate `except ERFAError` for `utc` here to be able to remove the `xfail`. This happens for years so early that JD cannot be computed:  `jd1+jd2 < -31738.5`.

Zac-HD:
```python
Falsifying example: test_conversion_never_loses_precision(
    iers_b='<using IERS-B orientation table>',
    scale1='tai',
    scale2='tcb',
    jds=(-31738.500000000346, 0.0),
)
```

Not just UTC either :confused:

mhvk: Ah, yes, to convert to `tdb` or `tcb`, one needs a `delta_tdb_tt` which cannot be calculated for those early times, as it needs UTC.  See figure in https://docs.astropy.org/en/latest/time/index.html#time-scale

I guess we should step back and think about the purpose of this test. Presumably just to check accuracy for allowed transformations, which for the broad range covered here is only between TAI, TT, and TCG and between TDB and TCB.

Zac-HD:
https://github.com/astropy/astropy/blob/818a24173e0ba3021a7e0b9f2f2297a1fa9a4543/astropy/time/tests/test_precision.py#L429-L437

After clarifying that scale conversion is allowed to fail for very early times *unless* both time scales are either geocentric or barycentric, we're all good - and upholding our desired precision budget :grin: 

...except on the python3.6-oldest-dependencies job :cry:

astropy/time/tests/test_precision.py
Zac-HD: Ah, that's done it - thanks @mhvk!

astropy/time/tests/test_precision.py
Zac-HD: https://github.com/HypothesisWorks/hypothesis/pull/2393 summarises Hypothesis' position concisely: `target()`ing the loss of precision enables a guided search mode, and reporting the highest score makes serious problems obvious even after shrinking to a minimal example.

This specific function leverages that upstream functionality, and also integrates with Astropy's time and units functionality tightly enough that I'm not sure I could make it generic without introducing precision errors.  I'm happy to move it elsewhere in Astropy if you have a suggestion, though personally I'd be inclined to leave it as-is until an equivalent helper is needed elsewhere ("YAGNI").

astropy/time/tests/test_precision.py
```diff
@@ -142,69 +293,376 @@ def test_leap_seconds_rounded_correctly():
...
     t = Time(jd1, jd2, scale=scale1, format="jd")
-    t2 = getattr(t, scale2)
-    assert t2.jd1 % 1 == 0
-    assert abs(t2.jd2) <= 0.5
-    assert abs(t2.jd2) < 0.5 or t2.jd1 % 2 == 0
+    try:
+        # If the rates ever differ, might lose one ULP
+        with quiet_erfa():
+            t2 = t + 2*dt_tiny
+            assert getattr(t, scale2) < getattr(t2, scale2)
```
mhvk, line 424: So, I guess somehow with the oldest dependencies an extra ULP was lost, and it is not surprising this is for UTC since that involves converting the time scale to TAI and back. My sense would be to add `3*dt_tiny` if both scales are UTC.

astropy/time/tests/test_precision.py
```diff
@@ -142,69 +293,376 @@ def test_leap_seconds_rounded_correctly():
...
+         jds=(0.0, 2.2204460492503136e-13),
+         delta=6.661338147750941e-13)
+@example(scale='utc',
+         jds=(2441682.5, 2.2204460492503136e-16),
+         delta=7.327471962526035e-12)
+@example(scale='utc', jds=(0.0, 5.787592627370942e-13), delta=0.0)
+def test_jd_add_subtract_round_trip(scale, jds, delta):
+    jd1, jd2 = jds
+    if scale == 'utc' and 1 > abs(jd1+jd2):
+        # near-zero UTC times degrade accuracy; why?
```
mhvk, line 464: This is related to the point above.  near-zero UTC leads to having a jd2 near 0.5, and then the conversion to TAI can push it over 0.5 which loses a bit of precision. Not sure why the threshold would not be something like `4*dt_tiny`, though...

astropy/time/tests/test_precision.py
Zac-HD: I'm not *entirely* sure why this happens, but after a few attempts I gave up and slapped an "xfail under Numpy 1.16" decorator on it.

Zac-HD:
OK, I've done all the investigation I can and the tests are all passing now :grin: 

Happy to help out with the Hypothesis parts going forward, but further work really needs more Astropy expertise than I have at the moment.

astropy/time/tests/test_precision.py
mhvk: Hmm, looks like decorator is on wrong function?  I'll fix...

mhvk:
OK, I tried to directly fix the failing test for numpy 1.16 instead. If that fails, one can go back to the `xfail` (but now on the right function :smile:)

mhvk:
And of course I did it wrong at least twice, but I think I now found that the failing test was just a case already found before, of a UTC JD close to 0, and had nothing to do with the numpy version (that is the risk of hypothesis, I guess, that the different travis runs do not necessarily try the same tests - good in many ways, but tricky here).

astropy/time/tests/test_precision.py
Zac-HD: :man_facepalming: haha, thanks...

Zac-HD:
Thanks @mhvk!  This is definitely getting into territory where the limit is my knowledge of Time code, not Hypothesis :slightly_smiling_face:

mhvk:
OK, tests now pass, so I'll merge. Thanks so much @Zac-HD and @aarchiba for getting this going!!!

Zac-HD:
You're welcome!  

I've updated the list of suggested tests in #9017 - the next set should be faster now that we've sorted out all the dependency and config issues, and upgraded Hypothesis a bit :grin:

mhvk:
@bsipocz - I think I did the milestone wrong here. This is really a new type of testing that I think might be better done in 4.2. Is it too late to change it? If you have already included it in 4.1 or 4.0.2, then probably not quite worth reverting either.

pllim:
Re: milestone -- cc @astrofrog and @eteq

astrofrog:
I think this is already in 4.1rc1, no harm in including it and certainly not much point removing it I think

PR #10374, opened 2020-05-19
Fix continuous integration for v4.0.x

astrofrog:
Work in progress! Just experimenting at this point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -179,6 +179,7 @@ jobs:
           stage: Final tests
           env: SETUP_CMD='build_docs -b linkcheck'
                MATPLOTLIB_VERSION="<3.1"
+               SPHINX_VERSION="<3"
                PIP_DEPENDENCIES="" CONDA_DEPENDENCIES=""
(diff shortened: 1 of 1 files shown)


astrofrog:
The linkcheck build now goes a lot further but there are broken links which are seen on master too, so I think this is ready to merge.

pllim:
I thought linkcheck is only supposed to run in cron?

astrofrog:
@pllim - yes but I think in v4.0.x we don't use cron:

https://github.com/astropy/astropy/commit/586ea15bb3de37de1b8eefb86b51c18a213a1aa3#diff-354f30a63fb0907d4ad57269548329e3

because that CI doesn't run much anyway.

astrofrog:
Basically we have to fix the permanent redirects that are happening in https://travis-ci.org/github/astropy/astropy/jobs/688893381 for this to turn green. I think we should maybe fix them in master then backport the fixes though, so maybe we can merge this for now?

PR #10375, opened 2020-05-20
Fitting uncertainties on model / parameters

cshanahan1:
This adds attributes related to fitting uncertainties on Models and Parameters. I implemented it for LevMarLSQFitter to begin with, but if possible all fitters should set uncertainties on models and their parameters. 

-`model.cov_matrix` is the parameter covariance matrix, and is set by the fitter. 
-`model.stds` is an array that contains the uncertainties on the individual parameters, in the same order as `model.param_names` for all non-fixed parameters. These values are calculated from `model.cov_matrix`, and therefore will only be set if the covariance matrix is set by the fitter.
-If available, each individual `parameter` in the model has a `parameter.std` attribute.

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -690,6 +690,12 @@ class Model(metaclass=_ModelMeta):
     # model inputs. Only has an effect if input_units is defined.
     input_units_equivalencies = None
 
+    # Covarience matrix can be set by fitter if available.
+    # If cov_matrix is availble, then std will set as well
+    _cov_matrix = None
+    _stds = None
+
+
     def __init__(self, *args, meta=None, name=None, **kwargs):
         super().__init__()
         self._default_inputs_outputs()
@@ -717,6 +723,7 @@ class Model(metaclass=_ModelMeta):
         # Raise DeprecationWarning on classes with class attributes
         # ``inputs`` and ``outputs``.
         self._inputs_deprecation()
+
 
     def _inputs_deprecation(self):
         if hasattr(self.__class__, 'inputs') and isinstance(self.__class__.inputs, tuple):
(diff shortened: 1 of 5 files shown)


perrygreenfield:
Would it be better to have the `.stds` as a dictionary instead of an array. I can imagine it being a bit of a pain to figure out which one is which. An alternative is to have some machinery allow one to index both the covariance matrix and standard deviations by the parameter name. I'm just thinking off the top of my head so don't take it too seriously (yet anyway, I wonder what others think).

nden:
@eteq @karllark @crawfordsm Do you have any comments on this?

perrygreenfield:
To elaborate a little more, if both of these attributes were objects, where the array was one of its attributes/properties, the print and repr could give a much more informative display of which parameters the values were associated with, as well as allowing indexing by name.

hcferguson:
Or a table? That way you can use all Astropy table machinery for filtering, etc. (Seems fine as long as it's 2D. Crazy if higher dimensionality).

karllark:
This looks reasonable to me.  Great to see the uncertainties starting to be implemented for model fitting.

astropy/modeling/core.py
```diff
@@ -1284,6 +1291,34 @@ def has_user_bounding_box(self):
...
+        """
+        Fitter should set covarience matrix, if available.
+        """
+        return self._cov_matrix
+
+
+    @cov_matrix.setter
+    def cov_matrix(self, value):
+
+        self._cov_matrix = value
```
nden, line 1305: This should have some checks for the dimensionality and type of `value`.

nden:
Suoerseded by #10552

astropy/modeling/fitting.py
```diff
@@ -556,6 +557,17 @@ def __call__(self, model, x, y, z=None, weights=None, rcond=None):
                           AstropyUserWarning)
 
         _fitter_to_model_params(model_copy, lacoef.flatten())
+
+        # calculate and set covariance matrix
+        a = a * scl
+        x_dot_x_prime = np.dot(a.T, a)
+        x_dot_x_prime = x_dot_x_prime
```
perrygreenfield, line 564: Not sure I understand the point of this statemetn

PR #10380, opened 2020-05-20
Fix unwanted behavior of CompoundModel inverse from my PR#10176

chris-simpson:
### Description

This pull request is to address unwanted behaviour in the `CompoundModel.replace_submodel()` method regarding inverses. I had flagged this up in the original PR #10176 but hadn't fully appreciated the consequences.

Suppose I have a model
```
m = A & (B | C)
```
and `C` has a user inverse defined when `m` is created. This means that a user inverse is defined for the `CompoundModel` `(B | C)` when `m` in built. The behaviour in #10176 is to not change user inverses, but this means that if I replace `C` with `C2` then the inverse of `(B | C2)` remains as the inverse of `(B | C)`. I think this is unwanted, and new user inverses should be created as `m` is rebuilt up the tree (from actually using this method in such a case). This is easily fixed.

I note that the fact that inverses are defined when the `CompoundModel` instance is created leads to other slightly odd (but well-defined) behaviour. For examples, if I define `m` as above, and _then_ define an inverse for `C` by setting `m[2].inverse`, `m.inverse` is not changed.

I've added a test to confirm the new behaviour.

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -3356,8 +3356,7 @@ class CompoundModel(Model):
                     branch = getattr(branch, node)
                 setattr(branch, tree[-1], model)
                 model = CompoundModel(branch.op, branch.left, branch.right,
-                                      name=branch.name,
-                                      inverse=branch._user_inverse)
+                                      name=branch.name)
                 tree = tree[:-1]
             return model
 
--- a/astropy/modeling/tests/test_compound.py
+++ b/astropy/modeling/tests/test_compound.py
@@ -701,9 +701,12 @@ def test_replace_submodel():
     S1 = Shift(1)
     S2 = Shift(2)
     S3 = Shift(3, name='S3')
-    S3.inverse = Shift(-2.9)
 
-    m = S1 & (S2 | S3)
+    S23 = S2 | S3
+    S23.inverse = Shift(-4.9)
+    m = S1 & S23
+
+    # This should delete the S23._user_inverse
     m2 = m.replace_submodel('S3', Shift(4))
     assert m2(1, 2) == (2, 8)
     assert m2.inverse(2, 8) == (1, 2)
--- a/astropy/modeling/tests/test_compound.py
+++ b/astropy/modeling/tests/test_compound.py
@@ -696,3 +696,14 @@ def test_replace_submodel():
     m2 = m.replace_submodel('poly', Polynomial1D(degree=1, c0=[1, 2],
                                                  n_models=2))
     assert_array_equal(m2([0, 1]), (2, 4))
+
+    # Ensure previous _user_inverse doesn't stick around
+    S1 = Shift(1)
+    S2 = Shift(2)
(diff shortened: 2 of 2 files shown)


nden:
@chris-simpson  #10384 deprecates passing an inverse in the CompoundModel constructor. 
The codestyle job is complaining. It'd be good to get this in the 4.1 release.

eslavich:
@chris-simpson we just merged a related PR (#10384) that moves construction of the CompoundModel inverse to the property instead of the initializer, and creates a new inverse every time the property is accessed.  I tried adding your new test code to master and it passes.

chris-simpson:
I missed that. #10384 sounds very useful and will help enormously with manipulating CompoundModel instances. So I'll close this. Thanks for the heads-up.

chris-simpson:
I misunderstood the earlier comments. #10384 doesn't fix this problem, and this PR needs to be accepted.

chris-simpson:
There's quite a bit of urgency here. I'm not sure what your release plan is but this needs to be fixed asap in master (behaviour is broken) and definitely before you release whichever version has #10384 (because the current code is raising DeprecationWarnings). It's literally one parameter that needs to be removed from a function call.

pllim:
@chris-simpson , while I understand your concerns, this PR has unrelated commits and needs some cleaning up.

It seems that #10384 is milestoned to 4.1. There is definitely still time to get this into 4.1rc2, I think? cc @eslavich @nden @eteq

chris-simpson:
i'm working on it

astropy/modeling/core.py
```diff
@@ -3356,8 +3356,7 @@ def replace_submodel(self, name, model):
                     branch = getattr(branch, node)
                 setattr(branch, tree[-1], model)
                 model = CompoundModel(branch.op, branch.left, branch.right,
```
eslavich, line 3358: Is it worth emitting a warning when an inverse is lost?  Just to prompt the user to choose between setting a new inverse or deliberately continuing without it.

Incidentally, is it intentional that we're also discarding any user-defined bounding box?

chris-simpson: I think `bounding_box`es will be propagated up the nodes as the model is rebuilt, right? If there's a user-defined one on a higher node, then it will be removed, just like a user-defined inverse. So this appears consistent.

I'm genuinely not sure about the answer to the first question.

nden: In my mind a model and its inverse go together. If one is replacing a model they are probably thinking about the inverse as well. I think there's no need for a warning here. However, we can reconsider in the future if we find this is causing confusion.
That said, it may be good to mention what happens to the inverse in the documentation.
Other than that it looks good to me. I agree it should go in 4.1.

chris-simpson:
Any possibility of this getting reviewed and merged soon please?

nden:
Sorry, I thought it was merged. Will look it over now.

nden:
too late :-)

Issue #10382, opened 2020-05-22
Test failure in io.votable when building wheels on MacPython


astrofrog:
I'm seeing the following test failure on Travis in the wheel building machinery, and am not quite sure why it hasn't happened in our CI. Does anyone have any ideas what could cause this?

```
_____________________________ test_unicode_as_char _____________________________

    def test_unicode_as_char():

        config = {'verify': 'exception'}

        field = tree.Field(

            None, name='unicode_in_char', datatype='char',

            arraysize='*', config=config)

        c = converters.get_converter(field, config=config)

    

        # Test parsing.

        c.parse('XYZ')  # ASCII succeeds

        with pytest.raises(

                exceptions.W55,

                match=r'FIELD \(unicode_in_char\) has datatype="char" but contains non-ASCII value'):

>           c.parse("zła")  # non-ASCII

E           Failed: DID NOT RAISE <class 'astropy.io.votable.exceptions.W55'>

../venv/lib/python3.6/site-packages/astropy/io/votable/tests/converter_test.py:81: Failed

=============================== warnings summary ===============================
```

I'll try and investigate some more, but putting this up here in case anyone is able to reproduce this and/or knows how to fix it :)

astrofrog:
@dhomeier - looks like you saw this in https://github.com/astropy/astropy/pull/9929#issuecomment-604540866 - where exactly did you see it, and are you able to reproduce it locally?

astrofrog:
Fix in https://github.com/astropy/astropy/pull/10383

Issue #8775, opened 2019-05-29
pedantic/verify does not control all the warnings in votable parsing


pllim:
As pointed out by @astrofrog in #8756 , `pedantic` (renamed to `verify` in #8715) does not universally control *all* the warnings emitted when parsing VO table. One example:

https://github.com/astropy/astropy/blob/62f1ac1726e57facf922eab9fa3ee36b03aa8a2d/astropy/io/votable/converters.py#L301-L303

In this example, it explicitly called `vo_warn`, which would give warning (or not), but it will never promote the warning to error even if `config['verify']` is set to `'error'`. But should it ever be an error, given that some default is assigned anyway? This is a API design question.

pllim:
Somewhat related... https://github.com/astropy/astropy/pull/11677#pullrequestreview-649564146

PR #10383, opened 2020-05-22
Fix test_unicode_as_char test

astrofrog:
This fixes the ``test_unicode_as_char`` test - we didn't see any failures in CI because we automatically upgrade warnings to exceptions, but if ``error`` is commented out from the pytest warning control in ``setup.cfg``, this test fails. This is manifesting itself in the wheel building since the warning control options aren't getting picked up there. In any case tests should definitely pass without ``error`` set (regardless of why the warning control options aren't picked up in the wheel building).

The code in the tests seems to trigger warnings not exceptions, hence the change from ``raises`` to ``warns``.

This is currently a release blocker for 4.1, so I'll merge it as soon as CI passes, but tagging @tomdonaldson @pllim @tboch in case anyone has time for a quick review, and in case any follow up PR is needed to actually change the main code to raise exceptions rather than warnings (for now the priority is to fix the test).

Fixes https://github.com/astropy/astropy/issues/10382

In future we'll need to think about having a CI job that *doesn't* raise warnings to errors to better catch cases like this.

### DIFF
--- a/astropy/io/votable/tests/converter_test.py
+++ b/astropy/io/votable/tests/converter_test.py
@@ -75,7 +75,7 @@ def test_unicode_as_char():
 
     # Test parsing.
     c.parse('XYZ')  # ASCII succeeds
-    with pytest.raises(
+    with pytest.warns(
             exceptions.W55,
             match=r'FIELD \(unicode_in_char\) has datatype="char" but contains non-ASCII value'):
         c.parse("zła")  # non-ASCII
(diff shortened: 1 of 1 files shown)


astrofrog:
As indicated above, I'll go ahead and merge this - if anyone wants to open up a follow up PR to change the behavior of the code itself being tested, please feel free to!

pllim:
Indeed, looks like the converter for this portion of the logic only calls `warn` instead of `warn_or_raise`. Good catch! xref #9505

dhomeier:
Cannot see if I mentioned it in https://github.com/astropy/astropy/pull/9929#issuecomment-604540866, but it did fail locally – was not clear if it was an issue with my `locale` settings. But it's passing now, thanks!

tomdonaldson:
Thanks for fixing this @astrofrog.  

I was not fully aware of the `error` setting in `setup.cfg`, and probably conflated its effects with those of the `config = {'verify': 'exception'}` in the test.  It's not obvious to me what the `error` setting is for, but that's for another day.

I verified that the current main code behavior is fine, although it could be made clearer by addressing #10389.

pllim:
Hah. I also forgot I opened #8775 .

PR #10384, opened 2020-05-22
Construct CompoundModel inverse lazily

eslavich:
### Description

This pull request addresses a serious bug caused by recent changes to the ASDF serializer for Model instances.  In #10189 we changed TransformType to yield the partially constructed model before setting all the fields, to support user inverses that form a reference loop.  That creates problems for CompoundModel because it accesses some of those fields in its initializer.

This fix has two parts: the first is to set all fields but inverse before yielding the model from TransformType, since inverse is the only one that might lead to a reference loop.  The second is to change CompoundModel to not construct its inverse in the initializer.  Instead, it will create the inverse only when the inverse property is accessed.

### DIFF
--- a/astropy/io/misc/asdf/tags/transform/basic.py
+++ b/astropy/io/misc/asdf/tags/transform/basic.py
@@ -19,9 +19,6 @@ class TransformType(AstropyAsdfType):
 
     @classmethod
     def _from_tree_base_transform_members(cls, model, node, ctx):
-        if 'inverse' in node:
-            model.inverse = node['inverse']
-
         if 'name' in node:
             model.name = node['name']
 
@@ -46,7 +43,10 @@ class TransformType(AstropyAsdfType):
                 param_and_model_constraints[constraint] = node[constraint]
         model._initialize_constraints(param_and_model_constraints)
 
-        return model
+        yield model
+
+        if 'inverse' in node:
+            model.inverse = node['inverse']
 
     @classmethod
     def from_tree_transform(cls, node, ctx):
@@ -56,8 +56,7 @@ class TransformType(AstropyAsdfType):
     @classmethod
     def from_tree(cls, node, ctx):
         model = cls.from_tree_transform(node, ctx)
-        yield model
-        cls._from_tree_base_transform_members(model, node, ctx)
+        return cls._from_tree_base_transform_members(model, node, ctx)
 
     @classmethod
     def _to_tree_base_transform_members(cls, model, node, ctx):
@@ -91,6 +90,8 @@ class TransformType(AstropyAsdfType):
             if bounds_nondefaults:
                 node['bounds'] = bounds_nondefaults
 
+        return node
+
     @classmethod
     def to_tree_transform(cls, model, ctx):
         raise NotImplementedError("Must be implemented in TransformType subclasses")
@@ -98,8 +99,7 @@ class TransformType(AstropyAsdfType):
     @classmethod
     def to_tree(cls, model, ctx):
         node = cls.to_tree_transform(model, ctx)
-        cls._to_tree_base_transform_members(model, node, ctx)
-        return node
+        return cls._to_tree_base_transform_members(model, node, ctx)
 
     @classmethod
     def assert_equal(cls, a, b):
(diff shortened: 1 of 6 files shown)


nden:
I would wait for @perrygreenfield to review before merging.

pllim:
Note: Travis CI is not run! I think the commit message needs to change to ensure full test suite runs and passes before merge. Thanks.

pllim:
And maybe squash while you are at it. 😉

astropy/modeling/core.py
```diff
@@ -2943,29 +2939,14 @@ def rename(self, name):
     def isleaf(self):
         return False
 
-    def has_inverse(self):
-        return self._has_inverse
-
     @property
     def inverse(self):
```
perrygreenfield, line 2943: Is this handling user inverses properly? The` Model `class property for inverse does, but I assume this supersedes this. I don't see a test for `CompoundModel` user inverses in `test_compound.py`. But remember, I'm dim.

astropy/io/misc/asdf/tags/transform/basic.py
```diff
@@ -91,15 +90,16 @@ def _to_tree_base_transform_members(cls, model, node, ctx):
...
     @classmethod
     def to_tree_transform(cls, model, ctx):
         raise NotImplementedError("Must be implemented in TransformType subclasses")
 
     @classmethod
     def to_tree(cls, model, ctx):
         node = cls.to_tree_transform(model, ctx)
-        cls._to_tree_base_transform_members(model, node, ctx)
-        return node
+        return cls._to_tree_base_transform_members(model, node, ctx)
```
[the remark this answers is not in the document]
eslavich: Yeah, I only changed this one because I thought it would be confusing if `_to_tree_base_transform_members` and `_from_tree_base_transform_members` worked differently, and I needed to return the latter to make the generator work.

astropy/modeling/core.py
```diff
@@ -1111,7 +1122,9 @@ def inverse(self):
         if self._user_inverse is not None:
             return self._user_inverse
         elif self._inverse is not None:
-            return self._inverse()
+            result = self._inverse()
+            if result is not NotImplemented:
+                return result
 
         raise NotImplementedError("An analytical inverse transform has not "
                                   "been implemented for this model.")
```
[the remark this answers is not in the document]
eslavich: That's a good point I'll change that message while I have this thing up on the lift

astropy/modeling/core.py
eslavich: It is handling user inverses, but the way it accomplishes that is a little... creative.  The `Model` metaclass pops the `CompoundModel` definition of `inverse` out of the class attributes and assigns it to `_inverse` instead.  So we're not actually overriding `Model`'s definition of `inverse`, even though it seems that way.

perrygreenfield: OK, thanks, but still, I wonder if a test is needed if it doesn't exist. Not urgent of course.

eslavich: I think this might cover it:

https://github.com/astropy/astropy/blob/master/astropy/io/misc/asdf/tags/transform/tests/test_transform.py#L30

m1 is a `CompoundModel`, and we round trip that thing through an asdf file and compare the result.

perrygreenfield: OK, good.

eslavich:
> Please check the remote data job to see if failures are related or not.

The failures in the remote data job don't seem related, and they also occurred in the last build on master, so I think we're all set there.

pllim:
Sorry, I mean the remote data job on Travis, just in case. Thanks for your patience!

astropy/modeling/core.py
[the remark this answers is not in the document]
eslavich: (this change is included in the squashed commit)

astrofrog:
Yeah there are real failures in the remote/dev job on Travis that are likely unrelated to this PR.

pllim:
From what I can understand in [4.1rc1 wiki](https://github.com/astropy/astropy/wiki/astropy-v4.1-release-candidate-(RC)-Testing), this PR did not get backported in time for `4.1rc1`, is that true?

Issue #10386, opened 2020-05-22
TST: Remote data job failed


pllim:
### Description

Example log: https://travis-ci.org/github/astropy/astropy/jobs/690064397

Possible cause: #10376 

### Expected behavior

Tests pass.

### Actual behavior

<details>
  <summary>Click to see traceback.</summary>

```
__________________________ TestIERS_Auto.test_simple ___________________________
self = <astropy.utils.iers.tests.test_iers.TestIERS_Auto object at 0x7f6b186ee630>
    @pytest.mark.remote_data
    def test_simple(self):
    
        with iers.conf.set_temp('iers_auto_url', self.iers_a_url_1):
    
            dat = iers.IERS_Auto.open()
            assert dat['MJD'][0] == 57359.0 * u.d
            assert dat['MJD'][-1] == 57539.0 * u.d
    
            # Pretend we are accessing at a time 7 days after start of predictive data
            predictive_mjd = dat.meta['predictive_mjd']
            dat._time_now = Time(predictive_mjd, format='mjd') + 7 * u.d
    
            # Look at times before and after the test file begins.  0.1292905 is
            # the IERS-B value from MJD=57359.  The value in
            # finals2000A-2016-02-30-test has been replaced at this point.
>           assert np.allclose(dat.ut1_utc(Time(50000, format='mjd').jd).value, 0.1292905)
E           AssertionError: assert False
E            +  where False = <function allclose at 0x7f6b41651268>(0.1293286, 0.1292905)
E            +    where <function allclose at 0x7f6b41651268> = np.allclose
E            +    and   0.1293286 = <Quantity 0.1293286 s>.value
E            +      where <Quantity 0.1293286 s> = <bound method IERS.ut1_utc of <IERS_Auto length=181>\n year month  day  ...        dX_2000A              dY_2000A      ...           nan                   nan       0\n   16     7    30 ...                   nan                   nan       0>(2450000.5)
E            +        where <bound method IERS.ut1_utc of <IERS_Auto length=181>\n year month  day  ...        dX_2000A              dY_2000A      ...           nan                   nan       0\n   16     7    30 ...                   nan                   nan       0> = <IERS_Auto length=181>\n year month  day  ...        dX_2000A              dY_2000A       NutFlag\n                  ......            nan                   nan       0\n   16     7    30 ...                   nan                   nan       0.ut1_utc
E            +        and   2450000.5 = <Time object: scale='utc' format='mjd' value=50000.0>.jd
[...]

PR #10387, opened 2020-05-22
Fix remote test failures related to IERS update

astrofrog:
This fixes two of the failures mentioned by @pllim in https://github.com/astropy/astropy/issues/10386. However, I'm not sure how to deal with the following:

```
________________ test_IERS_B_parameters_loading_into_IERS_Auto _________________
    @pytest.mark.remote_data
    def test_IERS_B_parameters_loading_into_IERS_Auto():
        A = iers.IERS_Auto.open()
        B = iers.IERS_B.open()
    
        ok_A = A["MJD"] <= B["MJD"][-1]
        assert not np.all(ok_A), "IERS B covers all of IERS A: should not happen"
    
        i_B = np.searchsorted(B["MJD"], A["MJD"][ok_A])
    
        assert np.all(np.diff(i_B) == 1), "Valid region not contiguous"
        assert np.all(A["MJD"][ok_A] == B["MJD"][i_B])
        # Check that values are copied correctly.  Since units are not
        # necessarily the same, we use allclose with very strict tolerance.
        for name in ("UT1_UTC", "PM_x", "PM_y", "dX_2000A", "dY_2000A"):
            assert_quantity_allclose(
                A[name][ok_A], B[name][i_B], rtol=1e-15,
                err_msg=("Bug #9206 IERS B parameter {} not copied over "
>                        "correctly to IERS Auto".format(name)))
../../.tox/py37-test-devdeps/lib/python3.7/site-packages/astropy/utils/iers/tests/test_iers.py:330: 
actual = <Quantity [ 0.8078584,  0.8051525,  0.8024015, ..., -0.2382365, -0.2385745,
           -0.2390142] s>
desired = <Quantity [ 0.8078584,  0.8051525,  0.8024015, ..., -0.2382558, -0.2385857,
           -0.239021 ] s>
rtol = 1e-15, atol = None
kwargs = {'err_msg': 'Bug #9206 IERS B parameter UT1_UTC not copied over correctly to IERS Auto'}
np = <module 'numpy' from '/home/travis/build/astropy/astropy/.tox/py37-test-devdeps/lib/python3.7/site-packages/numpy/__init__.py'>
_unquantify_allclose_arguments = <function _unquantify_allclose_arguments at 0x7f6b3dd97a60>
    def assert_quantity_allclose(actual, desired, rtol=1.e-7, atol=None,
                                 **kwargs):
        """
        Raise an assertion if two objects are not equal up to desired tolerance.
    
        This is a :class:`~astropy.units.Quantity`-aware version of
        :func:`numpy.testing.assert_allclose`.
        """
        import numpy as np
        from astropy.units.quantity import _unquantify_allclose_arguments
        np.testing.assert_allclose(*_unquantify_allclose_arguments(
>           actual, desired, rtol, atol), **kwargs)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-15, atol=0
E       Bug #9206 IERS B parameter UT1_UTC not copied over correctly to IERS Auto
E       Mismatched elements: 18 / 17275 (0.104%)
E       Max absolute difference: 2.43e-05
E       Max relative difference: 0.00010208
E        x: array([ 0.807858,  0.805153,  0.802401, ..., -0.238236, -0.238574,
E              -0.239014])
E        y: array([ 0.807858,  0.805153,  0.802401, ..., -0.238256, -0.238586,
E              -0.239021])
../../.tox/py37-test-devdeps/lib/python3.7/site-packages/astropy/tests/helper.py:483: AssertionError
_____________________________ [doctest] frames.rst _____________________________
```

I re-ran the update script hence the small updates in this PR, but that didn't help - the failures still happen.


eteq:
Hmm. I'll have to do some local tests, but I wonder if this is due to the recent changes in how the IERS files are delivered, such that they were out-of-sync at the time the update happened?

astrofrog:
I updated them a few days ago, and then again today, and in both cases the remote failures have happened - very strange!

eteq:
Hmm.  The mystery deepens: try as I might I *cannot* reproduce this locally - the test just keeps passing.  I wonder if this is a location problem? I.e., @astrofrog (and travis) might be getting European servers but I'm getting a us one?

eteq:
Oh wait, nevermind - it's even more insidious: apparently this test passes on py3.8 but *not* 3.7.  I can reproduce now though so perhaps I can make comparisons between the two envs and work it out.

astrofrog:
I've tracked the issue down to the following line:

```
        mjd_b = table['MJD'][np.isfinite(table['UT1_UTC_B'])]
```

in ``_substitute_iers_b`` - if I remove the finite value filtering, everything works, so I think there is a mismatch between whether non-finite values are ignored or not in the test too

eteq:
Hmm, so maybe it's just that past IERS B/A combinations didn't have empty UT1-UTC entries? I think is what gives non-finite answers there as it maps onto nans in the resulting table.  So it's possible this was wrong all along but we just didn't realize it because there weren't problematic tables?

I'm still struggling to find any good matches by actually looking at the files directly.  I guess that was the purpose of the substitution in the first place but it may be this is due for a more complicated clean-up in 4.2 if we can band-aid this for now...

(And to continue my more prosaic line of inquiry above: in a new py38 environment it's also now failing so it was probably a red herring due to something stuck in my initial environment)

astrofrog:
I've pushed a commit that resolves the issue for me locally, though I'm not sure if it's the *right* fix. But conceptually, doing the filtering to me seems odd since later on it seems we don't pre-filter when assigning values to the table.

@mhvk - I think we really need your expertise here to determine whether this is the right fix thought! (assuming CI actually passes)

astrofrog:
While this seems to fix the problematic test, other tests are now failing :cry:

pllim:
`isfinite` was added in #9226

astrofrog:
If anyone wants to push commits here please feel free to!

mhvk:
OK, I at least now understand the problem: IERS_A includes also IERS_B values, but those are not always identical to the actual IERS_B (since, surprisingly, those *change*). So, we wrote code to update with values from IERS_B; the `np.isfinite` is to restrict the copying to replacing those values it had in the first place. But this goes wrong in this case, since we now have a very new IERS_B which has values beyond the "iers_b" values included in IERS_A. And clearly the test did not take into account that this might happen.

I must admit that I'm not sure what the right behaviour really should be; one could argue for just replacing everything, but on the other hand the iers_b bundled in iers_a might be more up to date. My own sense would be that IERS_Auto should also download IERS_B and just make its own mix. But that's really for another PR, so I think here it makes most sense to simply adjust the test so that it actually tests what the current code is supposed to do.  I'll push a fix with that.  Then, we can open another issue to think through what the ultimate behaviour should be.

mhvk:
And there was an issue more or less about this (but without the `utils.iers` label, now corrected; maybe `iers` should just move up...). See #9227.

astrofrog:
Thanks for fixing this! I agree this is the right approach for now (just making sure the test matches what's actually done). I'll go ahead and merge this since the remote test build is now passing.

eteq:
Ohhh, that explains it @mhvk - it hadn't occurred to me that IERS B might be too *new*!

While investigating this (although not solving it without your input...) I also encountered a few other confusing things about IERS B but I guess  #9227 is the place to discuss that

PR #10388, opened 2020-05-22
TST: GitHub Actions as CI

pllim:
**Update:** Due to Travis CI changes in pricing, we decided to push ahead with this as soon as we have a minimal viable product. We will fix the rest of the jobs in follow-up PRs.

~This is a small proof-of-concept of using GitHub Actions as CI. Might expand if there is buy-in, or abandon if there isn't.~

**Pros**

- [Artifact support](https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts) but I haven't tried to use it yet.
- Better integration with GitHub UI.
  - The cancel button is easier to access than Azure Pipelines.
- Unlike Azure, no need for special Microsoft account or let Microsoft access GitHub credentials. (The latter is probably moot point, har har...).
- Performance on par with Azure (theoretically, considering they probably share the same servers or something...).
- Potentially useful for other things too, like bot work and delivery.

**Cons**

- No [anchors](https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/m-p/30336) or [templating](https://github.community/t5/GitHub-Actions/Yaml-templating-like-azure-pipeline/td-p/39636).
- No [allowed failures](https://github.community/t5/GitHub-Actions/continue-on-error-allow-failure-UI-indication/td-p/37033). Also see actions/toolkit#399
- No built in support for [skip ci](https://github.community/t5/GitHub-Actions/GitHub-Actions-does-not-respect-skip-ci/td-p/42834).
  - On the bright side, I can use `[skip ci]` to skip all the other CIs and only run Actions. (Har har...)
- No `ccache`? If there is one, it is not documented.
- Cannot [re-run only one job](https://github.community/t5/GitHub-Actions/Ability-to-rerun-just-a-single-job-in-a-workflow/td-p/41629).
- No extra architecture support (e.g., ARM64) without [hosting self-runners](https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/).
- No [re-run with SSH](https://github.community/t5/GitHub-Actions/Debugging-github-actions-locally-or-ssh-to-workspace/td-p/38949) like CircleCI.

**CI jobs and features to implement (please check off if done)**

From `.travis.yml`:

- [ ] Caching with `ccache` and `.hypothesis`.
- [ ] `pytest` conditions in test modules using `TRAVIS` env var.
- [ ] Slack notification (both on success and failure) for cron job.
- Initial tests stage:
  - [x] Python 3.7 with minimal dependencies
  - [x] Python 3.8 with all optional dependencies (`-v --develop`, `--durations=50`, `clang`)
  - [x] Code style checks
- Comprehensive tests stage:
  - [x] Python 3.7 with oldest supported version of all dependencies (`--open-files`)
  - [x] Python 3.7 with numpy 1.17 and full coverage (`--remote-data=astropy`, `LC_CTYPE=C.ascii`, `LC_ALL=C`) -- need `codecov` token
- Final tests stage:
  - [x] Python 3.8 with all optional dependencies
  - [x] Python 3.7 with remote data and dev version of key dependencies (`--remote-data=any`) -- allowed to fail
- Cron:
  - [x] Python 3.7 with all optional dependencies for OSX (`clang`, `--remote-data=astropy`)
  - [x] Python 3.7 with dev version of key dependencies
  - [x] Documentation link check
  - [x] Python dev with basic dependencies (`3.9-dev`, `--durations=50`, `clang`)
  - [ ] Big-endian s390x architecture with apt (`arch: s390x`, `ASTROPY_USE_SYSTEM_ALL=1`) (@astrofrog will do in a follow-up PR)
  - [ ] ARM64 architecture with apt (`arch: arm64`, `ASTROPY_USE_SYSTEM_ALL=1`) (@astrofrog will do in a follow-up PR)
  - [x] Bundling with pyinstaller

From `.circleci/config.yml`:

- [ ] 32-bit and parallel (custom `astropy.cfg`, Python 3.7 `-n=4 --durations=50`, Python 3.8 `py38-test-double`)
- [ ] `image-tests-mpl302`
- [ ] `image-tests-mpl310`
- [ ] `image-tests-mpldev`

From `azure-piplines.yml`:

- [ ] Build wheels and publish to PyPI for release.

**TODO**

- [ ] Fix failures. It kicked off at https://github.com/pllim/astropy/actions/runs but probably affected by the degraded GitHub service that was happening.
  - [ ] Once Actions reports to the PR, will adding branch filters back break it?
- [x] Discuss and compare to #8445 .
- [x] See what to do with the cron stuff -- separate or same YAML?
- [x] See how much more we can port to Actions from Travis CI and maybe even CircleCI. Should use matrix where possible.
- [x] Remove ported jobs from the other CIs.
- [x] Figure out the badge situations in README.
- [x] Determine if we are happy to make remote/numpy dev be not allowed failures since there isn't an equivalent feature
- [ ] See if we can add `--open-files` back to a newer job in future PR.
- [ ] See if codecov reports back to PR. If not, find solution.

**Possible links of interest**

Thanks to everyone who provided these resources:

- https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization
- https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action
- https://github.com/uraimo/run-on-arch-action 
- https://github.com/tepickering/package-template/blob/add_github_actions/%7B%7B%20cookiecutter.package_name%20%7D%7D/.github/workflows/tox-tests.yml

### DIFF
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -1,0 +1,68 @@
+name: Run tests (cron)
+
+on:
+  schedule:
+    # run every Monday at 6am UTC
+    - cron: '0 6 * * 1'
+
+jobs:
+  tests:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+
+          - name: Python 3.7 with all optional dependencies (MacOS X)
+            os: macos-latest
+            python: 3.7
+            toxenv: py37-test-devdeps
+            toxposargs: --remote-data=astropy
+            # compiler: clang
+
+          # We check numpy-dev also in a job that only runs from cron, so that
+          # we can spot issues sooner. We do not use remote data here, since
+          # that gives too many false positives due to URL timeouts. We also
+          # install all dependencies via pip here so we pick up the latest
+          # releases.
+          - name: Python 3.7 with dev version of key dependencies
+            os: ubuntu-latest
+            python: 3.7
+            toxenv: py37-test-devdeps
+
+          - name: Documentation link check
+            os: ubuntu-latest
+            python: 3.7
+            toxenv: linkcheck
+
+          - name: Bundling with pyinstaller
+            os: ubuntu-latest
+            python: 3.8
+            toxenv: pyinstaller
+
+          # Test against Python dev in cron job.
+          - name: Python dev with basic dependencies
+            os: ubuntu-latest
+            python: 3.10-dev
+            toxenv: pydev-test
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install language-pack-de and tzdata
+      if: ${{ matrix.os == 'ubuntu-latest' }}
+      run: sudo apt-get install language-pack-de tzdata
+    - name: Install graphviz
+      if: ${{ matrix.toxenv == 'linkcheck' }}
+      run: sudo apt-get install graphviz
+    - name: Install Python dependencies
+      run: python -m pip install --upgrade tox codecov
+    - name: Run tests
+      run: tox ${{ matrix.toxargs}} -e ${{ matrix.toxenv}} -- ${{ matrix.toxposargs}}
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -14,13 +14,6 @@ jobs:
       matrix:
         include:
 
-          - name: Python 3.7 with all optional dependencies (MacOS X)
-            os: macos-latest
-            python: 3.7
-            toxenv: py37-test-devdeps
-            toxposargs: --remote-data=astropy
-            # compiler: clang
-
           # We check numpy-dev also in a job that only runs from cron, so that
           # we can spot issues sooner. We do not use remote data here, since
           # that gives too many false positives due to URL timeouts. We also
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -4,7 +4,6 @@ on:
   push:
   pull_request:
 
-
 jobs:
   tests:
     name: ${{ matrix.name }}
@@ -38,7 +37,6 @@ jobs:
             os: ubuntu-16.04
             python: 3.7
             toxenv: py37-test-oldestdeps
-            toxposargs: --open-files
 
           - name: Python 3.7 with numpy 1.17 and full coverage
             os: ubuntu-latest
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -30,7 +30,6 @@ jobs:
             toxenv: py38-test-alldeps
             toxargs: -v --develop
             toxposargs: --durations=50
-            # compiler: clang
 
           # NOTE: In the build below we also check that tests do not open and
           # leave open any files. This has a performance impact on running the
@@ -51,6 +50,11 @@ jobs:
             os: windows-latest
             python: 3.8
             toxenv: py38-test-alldeps
+
+          - name: Python 3.7 with all optional dependencies (MacOS X)
+            os: macos-latest
+            python: 3.7
+            toxenv: py37-test-devdeps
 
           - name: Python 3.7 with remote data and dev version of key dependencies
             os: ubuntu-latest
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -80,8 +80,8 @@ jobs:
     - name: Run tests
       run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
     # TODO: Do we need --gcov-glob "*cextern*" ?
-    - name: Coverage report
-      uses: codecov/codecov-action@v1
+    - name: Upload coverage to codecov
       if: ${{ matrix.toxenv == 'py37-test-alldeps-numpy117-cov-clocale' }}
+      uses: codecov/codecov-action@v1
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
+        file: ./coverage.xml
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -3,14 +3,10 @@ name: Run tests
 # TODO: cron on a separate YAML?
 on:
   push:
-    # Sequence of patterns matched against refs/heads
-    branches:
-    - master
-    - stable
-    - LTS
   pull_request:
-    branches:
-    - master
+  schedule:
+    # run every Monday at 6am UTC
+    - cron: '0 6 * * 1'
 
 jobs:
   tests:
@@ -20,14 +16,19 @@ jobs:
       fail-fast: false
       matrix:
         include:
+
+          # Standard jobs
+
           - name: Code style checks
             os: ubuntu-latest
             python: 3.x
             toxenv: codestyle
+
           - name: Python 3.7 with minimal dependencies
             os: ubuntu-latest
             python: 3.7
             toxenv: py37-test
+
           - name: Python 3.8 with all optional dependencies
             os: ubuntu-latest
             python: 3.8
@@ -35,6 +36,7 @@ jobs:
             toxargs: -v --develop
             toxposargs: --durations=50
             # compiler: clang
+
           # NOTE: In the build below we also check that tests do not open and
           # leave open any files. This has a performance impact on running the
           # tests, hence why it is not enabled by default.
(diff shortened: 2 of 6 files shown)


Cadair:
> The cancel button is easier to access than Azure Pipelines.

Don't they both ingrate with the GH actions UI for this? You can restart pipelines from the actions tab as well.

image

pllim:
Re: https://github.com/astropy/astropy/pull/10388#issuecomment-633260141

Not sure. I'll have to check again next time one of my other project kicks off Azure. I vaguely remember having to jump through hoops with my Microsoft account to get to the cancel button but it was a while ago.

Cadair:
The GH account integration with pipelines has improved a lot in the last few months.

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,124 @@
...
+        echo "TOXENV="$TOXENV
+        echo "TOXARGS="$TOXARGS
+        tox ${TOXARGS}
+
+  # TODO: Maybe this can be part of a matrix too but I just want to show
+  #       stages without porting everything for now.
+  # Full tests with coverage checks.
+  coverage_test:
+    runs-on: ubuntu-latest
+    needs: [code_style, initial_tests]
```
astrofrog, line 83: I think we could try and keep things simple and not worry about replicating stages for now - we could have a single jobs section with an include matrix?

astrofrog: Instead of stages I think we could rely on Actions' fast-fail functionality (default) which means that all jobs stop if one fail. In practice this means if e.g. code style fails it won't waste time on longer builds for example.

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,124 @@
...
+
+# TODO: cron on a separate YAML?
+on:
+  push:
+  pull_request:
+
+# The following is needed to avoid issues if e.g. Matplotlib tries
+# to open a GUI window.
+env:
+  SETUP_XVFB: True
```
astrofrog, line 11: I think this was a ci-helpers thing

pllim: Yeah, this PR been open for a while and probably need upgrades.

.github/workflows/ci_workflows.yml
pllim: Up to you. I was just trying to recreate what we had at the time.

astrofrog: I think the issue is that the stages makes everything much more verbose. I think I would prefer if we could make things more concise and avoid repetition and then decide if adding stages back are worth it?

astrofrog:
@pllim - since there was a conflict, and to pick up the fact actions now works in PRs, I took the liberty of rebasing this and removing the temporary commit. I also added a commit which simplifies the configuration a lot by getting rid of the stages (this reduces a lot of configuration duplication). Once we are happy with the config, we could change ``fail-fast`` to ``true`` which means that if some of the faster builds (e.g. code style or minimal deps) fail the whole CI stops, which in effect reproduces what we were trying to accomplish with stages. We could add stages back as you had them of course, but I think we should aim to have a CI config file as simple as possible. Finally, I've removed jobs from the Travis configuration which are now in Actions. I have a backup of your branch if needed for any reason though. I haven't yet added the ability to use clang for one of the Linux jobs.

astrofrog:
One thing we should decide is how to handle allowed failures - since each job appears as one entry here maybe we don't actually need to replicate this functionality, and that way remote data and numpy dev failures won't be hidden away but are also easy to ignore.

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,135 @@
...
+    - name: Install Python dependencies
+      if: ${{ skip != true }}
+      run: python -m pip install --upgrade tox codecov
+    - name: Run tests
+      if: ${{ skip != true }}
+      run: |
+        echo "CC="$CC
+        echo "LC_CTYPE="$LC_CTYPE
+        echo "LC_ALL="$LC_ALL
+        tox ${{ matrix.toxargs}} -e ${{ matrix.toxenv}} -- ${{ matrix.toxposargs}}
```
Cadair, line 131:
```suggestion
        tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
```

astrofrog:
I've also now copied over most cron jobs, except the arm and s390x arch ones which I can do a follow up PR for later. I made a separate file for cron otherwise it gets a bit messy to skip some jobs and not others for cron and I haven't found a way to hide them in pull requests then. I've also added a TMP commit to enable cron builds here so we can check they run.

.github/workflows/ci_workflows.yml
astrofrog: Sorry I just saw this and force pushed in the mean time. Can you just push a commit directly to fix this in both files? (there is a cron file now)

astrofrog:
The main thing I haven't implemented is the use of clang which was specified in the Travis file for some jobs. Is that actually something we want/need? We could get rid of that and just make the MacOS X build be non-cron which would then by default use clang?

@Cadair - you could always push your changes here to move the image tests away from CircleCI?

Cadair:
As far as i can tell actions artifacts aren't up to it. I suggest we just leave circle for now and focus on killing off Travis.

astrofrog:
Ok so looking at the failures for now:

* ``test_update_parallel`` was skipped on Travis but also fails on Actions so just skipping on all CI
* **Python 3.7 with oldest supported version of all dependencies** fails because there are some open files after the tests (mainly FITS files I think). @saimn any idea what could be going on there?
* The link checker is failing because https://docs.astropy.org/en/v4.0/whatsnew/4.0.html is broken (any ideas why?)
* There are some real failures with Numpy dev I think (cc @mhvk)

I still need to look at the other failures.

saimn:
Re-clang: yes I think we had clang on linux because the macos queue on Travis was too slow. So if we can have a macos job by default, we don't need clang on linux.

astrofrog:
@saimn - ok let's do that. The MacOS X builds start up pretty quickly on Actions.

astrofrog:
@pllim - what was the motivation for branch filters btw? I think we can just leave the default and it just won't run on branches that don't have the config?

astrofrog:
Anyway I'm done for now, I think we are pretty close to having something mergeable though, as things like CircleCI and Azure can be left to later as well as ARM and s390x.

Cadair:
I don't think we want to or even should remove Azure and CircleCI. I think we should aim to get this to working for PRs and then we can worry about some of the other stuff afterwards.

saimn:
> Python 3.7 with oldest supported version of all dependencies fails because there are some open files after the tests (mainly FITS files I think). @saimn any idea what could be going on there?

Not sure, but it seems to be more than just fits. There are a lot of failure everywhere.

pllim:
Thank you so much for taking this over. I like this staggered time zone collaboration thing, waking up to an almost complete solution. You are an angel!

> how to handle allowed failures

It is not natively supported yet and for some reason GitHub/Microsoft is kind of resistant to implementing it. For now, we just have to live with ❌ for those jobs, which might unsettle contributors, so we need to document this new behavior somewhere, maybe the testing page?

> follow up PR

Yes, I agree. Also, having failing cron jobs is not a blocker for this PR. We can handle those separately too. In fact, I might make the linkcheck less frequent now that it is possible (as a separate PR)!

> open files

That does look weird. I'll have a look.

> linkcheck

I think based on Slack conversations, there is a path forward and we're waiting for @bsipocz to wake up? 😸 Okay to address in follow-up PR as well.

> Numpy dev

Again, can sort this out in a follow up PR.

> Python dev (now 3.10) 

Yeah, I tried adding it to Travis back in #10875 but no luck. I agree we can comment out or stick with 3.9.dev for now, if you really want some sort of dev.

> clang

Great, looks like it is solved in a conversation above.

> branch filters

I don't remember what I wanted to do with them. It's so long ago. Just ignore it for now and we can revisit later.

bsipocz:
Don't hold this up on my account. We can iterate in follow ups if needed. This week is super busy with big deadline, so I won't be able to tinker with any of these rapidly enough for the night off of Travis.

pllim:
p.s. @astrofrog suggested https://github.com/nektos/act

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,85 @@
...
+on:
+  push:
+  pull_request:
+
+jobs:
+  tests:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
pllim, line 12: Is this suppose to be changed before merge?
```suggestion
      fail-fast: true
```

bsipocz-2nd: I would go with true, but in fact am OK with either.

bsipocz-2nd: (hmm, wrong browser....)

.github/workflows/ci_workflows.yml
```diff
@@ -1,4 +1,4 @@
-name: Run tests
+name: CI
```
pllim, line 1: Note: This name has to match what is on README

.github/workflows/ci_workflows.yml
astrofrog: I would normally go with yes but actually the numpy dev job might cause the others to stop so maybe keep false until we discuss allowed failures etc?

pllim: Yeah, you are right, the OSX job got canceled.

.github/workflows/ci_cron.yml
```diff
@@ -0,0 +1,61 @@
+name: Run tests (cron)
+
+on:
+  schedule:
+    # run every Monday at 6am UTC
+    - cron: '0 6 * * 1'
```
bsipocz, line 6: is this frequent enough? We used to run cron jobs daily.

.github/workflows/ci_workflows.yml
bsipocz: since we already have multiple workflows, I'm getting on board of indeed separating the allowed failures one out. In a follow-up.

README.rst
```diff
@@ -2,7 +2,7 @@
 Astropy
 =======
 
-|Travis Status| |CircleCI Status| |Coverage Status| |PyPI Status| |Documentation Status|
+|Actions Status| |CircleCI Status| |Coverage Status| |PyPI Status| |Documentation Status|
```
bsipocz, line 5: travis isn't completely removed yet, so maybe keep both?

bsipocz (approved):
minor comments only, no need to address them here, they are just thought for the follow-up.

README.rst
pllim: It doesn't run no more. We can add it back if we decide to pay.

.github/workflows/ci_cron.yml
pllim: Good point. We can follow up as separate PR.

Cadair (approved):
:shipit:

astrofrog:
🎉

README.rst
bsipocz: ok, indeed. I only said it based on that the config remained.

mwcraig:
Nice work! 🎉

bsipocz:
I've added this to the backport as a first step of getting #11050 addressed, and to get some CI on the 4.2.x branch.  I'm certain that more backports and/or fine-tuning will be necessary (cleaning up cron as that is not run on non-master branches, etc., adding back 3.6 testing, etc.), but still getting this is superb for now.

PR #10395, opened 2020-05-25
Explicitly convert tmpdir in test to string

astrofrog:
On some platforms, ``check_free_space_in_dir`` doesn't work with non-string paths (in fact the docstring of ``check_free_space_in_dir`` explicitly asks for str), so fixing this. This is needed to avoid having to patch the repo during the wheels-building process.

### DIFF
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -1280,7 +1280,7 @@ def test_cache_contents_agrees_with_get_urls(temp_cache, valid_urls):
 
 def test_free_space_checker_huge(tmpdir):
     with pytest.raises(OSError):
-        check_free_space_in_dir(tmpdir, 1_000_000_000_000_000_000)
+        check_free_space_in_dir(str(tmpdir), 1_000_000_000_000_000_000)
 
 
 def test_get_free_space_file_directory(tmpdir):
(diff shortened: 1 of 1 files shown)


astropy/utils/tests/test_data.py
```diff
@@ -1289,7 +1289,7 @@ def test_get_free_space_file_directory(tmpdir):
         pass
     with pytest.raises(OSError):
         get_free_space_in_dir(fn)
```
saimn, line 1291: This one should be modified as well ?

astrofrog: Good catch thanks!

saimn:
I didn't know that `os.path` function could actually work with pytest's `tmpdir` object.
For the future it would be a good idea to recommend using teh `tmp_path` fixture instead, which is a `pathlib.Path` object.

astrofrog:
I agree with ``tmp_path`` going forward, and I also think it'd be good to make sure anything that can accept a string path in astropy can also accept a ``Path`` object. For this PR, my main concern is just making sure the tests work.

pllim:
Weird. Why didn't the Windows job catch this?

astrofrog:
@pllim - I'm not entirely sure.

pllim:
I think `tmpdir.strpath` would have worked too? But anyway, the ship has sailed. If you feel strongly that a refactor is necessary, please open a new issue. Thanks for fixing this!

Issue #10247, opened 2020-05-01
Warnings when accessing distance on SkyCoord at origin with RV defined


astrofrog:
This does not raise warnings:

```python
In [13]: SkyCoord(CartesianRepresentation([0*u.km, 0*u.km, 0*u.km]), 
    ...:                                 frame='icrs').distance                                                      
Out[13]: <Distance 0. km>
```

but this does:

```python
In [10]: from astropy.coordinates import SkyCoord, CartesianRepresentation                                           

In [11]: from astropy import units as u                                                                              

In [12]: SkyCoord(CartesianRepresentation([0*u.km, 0*u.km, 0*u.km]), 
    ...:                                 frame='icrs', radial_velocity=30 * u.km / u.s).distance                     
/home/tom/Code/astropy/astropy/units/quantity.py:477: RuntimeWarning: invalid value encountered in true_divide
  result = super().__array_ufunc__(function, method, *arrays, **kwargs)
/home/tom/Code/astropy/astropy/units/quantity.py:477: RuntimeWarning: divide by zero encountered in true_divide
  result = super().__array_ufunc__(function, method, *arrays, **kwargs)
/home/tom/Code/astropy/astropy/units/quantity.py:477: RuntimeWarning: invalid value encountered in pdp
  result = super().__array_ufunc__(function, method, *arrays, **kwargs)
Out[12]: <Distance 0. km>
```

I think the second case should probably also not emit warnings though - the result should be the same as in the first case?

mhvk:
Looking at this now, I see that the underlying reason is that the coordinates are converted to spherical in order to get the distance (inside `BaseFrame.__getattr__`; see below), and one gets divisions by 0 since the conversion of the differential needs to divide by the norm. I guess easiest is to guard against this in the representation/differential conversion.

astrofrog:
This has been fixed in the meantime

PR #10398, opened 2020-05-26
Simplify SpectralCoord implementation

astrofrog:
The current SpectralCoord implementation relies on an internal 'fake' observer and target when these are not specified explicitly, but I believe this actually makes the implementation more complicated than it needs to be (the original intent was to actually make it simpler but this is not the case in the end!).

In addition, some of the implementation is also overly complex because of an original use case (changing the actual observer location and taking into account parallax effects) which didn't make it into the final API since we determined that this might not be needed. This PR shows how we can simplify the implementation while still having the tests pass (at least they do for me locally).

Even if we do eventually implement more complex use cases, I think we should keep the implementation as simple as possible for the current use cases, for maintainability and speed.

The resulting code in this PR has fewer frame transformations and vector operations, and is therefore faster. For instance:

**Before**

```python
In [1]: from astropy.coordinates import SpectralCoord                                                                                                                            

In [2]: from astropy import units as u                                                                                                                                           

In [3]: velocity = 100 * u.km / u.s                                                                                                                                              

In [4]: %timeit sc = SpectralCoord(100, 'GHz')                                                                                                                                   
2.77 ms ± 121 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: sc = SpectralCoord(100, 'GHz')                                                                                                                                   

In [6]: %timeit res = sc.with_radial_velocity_shift(velocity)                                                                                                                    
7.11 ms ± 410 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

**After**

```python
In [1]: from astropy.coordinates import SpectralCoord                                                                                                                            
^[[A^[[A
In [2]: from astropy import units as u                                                                                                                                           

In [3]: velocity = 100 * u.km / u.s                                                                                                                                              

In [4]: %timeit sc = SpectralCoord(100, 'GHz')                                                                                                                                   
134 µs ± 2.37 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [5]: sc = SpectralCoord(100, 'GHz')                                                                                                                                           

In [6]: %timeit res = sc.with_radial_velocity_shift(velocity)                                                                                                                    
418 µs ± 34.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

So the initialization in this case is 20x faster and the call to ``with_radial_velocity_shift`` is 17x faster. This is just an example of course, but overall I think all operations will be the same or faster.

This will require a careful review to make sure I haven't actually broken anything. It might be possible to simplify things further but I haven't had a chance to work on this more as I'm not going to be working much on astropy for a couple of months. So if anyone would like to review this in detail and push commits to it, please feel free to do so!

(cc @mhvk who originally commented on the SpectralCoord PR that the internal coordinates seemed unnecessary)

### DIFF
--- a/astropy/coordinates/spectral_coordinate.py
+++ b/astropy/coordinates/spectral_coordinate.py
@@ -548,93 +548,6 @@ class SpectralCoord(SpectralQuantity):
 
         return pos_hat
 
-    def _change_observer_to(self, observer, target=None):
-        """
-        Moves the observer to the provided coordinate/frame.
-
-        Parameters
-        ----------
-        observer : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
-            The new observation frame or coordinate.
-        target : `~astropy.coordinates.SkyCoord`, optional
-            The `~astropy.coordinates.SkyCoord` object representing the target of the observation.
-            If none given, defaults to currently defined target.
-
-        Returns
-        -------
-        new_coord : `SpectralCoord`
-            The new coordinate object representing the spectral data
-            transformed to the new observer frame.
-        """
-
-        if self.observer is None:
-            raise ValueError("No observer has been set, cannot change "
-                             "observer.")
-
-        target = self.target if target is None else target
-
-        # Check velocities and distance values on the new frame. This is
-        #  handled in the frame validation.
-        observer = self._validate_coordinate(observer, label='observer')
-
-        # Calculate the initial and final los velocity
-        init_obs_vel = self._calculate_radial_velocity(self.observer, target)
-        fin_obs_vel = self._calculate_radial_velocity(observer, target)
-
-        line_of_sight_unit_vec = self._normalized_position_vector(observer, target)
-
-        new_data = self._project_velocity_and_shift(init_obs_vel, fin_obs_vel, line_of_sight_unit_vec)
-
-        new_coord = self.replicate(value=new_data,
-                                   observer=observer if observer is not self.observer else None,
-                                   target=target if target is not self.target else None)
-
-        return new_coord
-
-    def _project_velocity_and_shift(self, init_vel, fin_vel, line_of_sight_unit_vec):
-        """
-        Calculate the velocity projection given two vectors.
-
-        Parameters
-        ----------
-        init_vel :`BaseRepresentation`
-            Initial velocity vector.
-        fin_vel : `BaseRepresentation`
-            Final velocity vector.
-        line_of_sight_unit_vec : `BaseRepresentation`
-            Unit vector pointing from observer to target
-
-        Returns
-        -------
-        new_data : `u.Quantity`
-            Spectral axis data with velocity shift applied.
-        """
-
-        # Project the velocity shift vector onto the line-of-sight vector
-        # between the target and the new observation frame.
-        init_proj_vel = line_of_sight_unit_vec.dot(init_vel) * line_of_sight_unit_vec
-
-        # Calculate the magnitude of the velocity shift between the two vectors.
-        # The vectors are aligned but may be in opposite directions, so we use
-        # the dot product to determine this.
-
-        diff_vel = fin_vel - init_proj_vel
-        delta_vel = line_of_sight_unit_vec.dot(diff_vel)
-
-        # In the case where the projected velocity is nan, we can assume that
-        # the final velocity different is zero, and thus the actual velocity
-        # delta is equal to the original radial velocity.
-
-        # TODO: Due to lack of precision in some coordinate transformations,
-        # we may end up with a final velocity very close to, but not quite at,
-        # zero. In this case, set a tolerance for the final velocity; if it's
-        # below this tolerance, assume the delta velocity is essentially nan.
-
-        if np.isnan(delta_vel) or fin_vel.norm() < 1e-7 * fin_vel.xyz.unit:
-            delta_vel = -self.radial_velocity
-
-        return _apply_relativistic_doppler_shift(self, delta_vel)
-
     @u.quantity_input(velocity=u.km/u.s)
     def with_observer_stationary_relative_to(self, frame, velocity=None, preserve_observer_frame=False):
         """
@@ -689,7 +602,7 @@ class SpectralCoord(SpectralQuantity):
                 # otherwise frame is ready to go
             else:
                 if velocity is None:
-                    differentials = CartesianDifferential(0 * u.m / u.s, 0 * u.m / u.s, 0 * u.m / u.s)
+                    differentials = ZERO_VELOCITIES
                 else:
                     differentials = CartesianDifferential(*velocity)
                 frame = frame.realize_frame(frame.data.with_differentials(differentials))
@@ -711,7 +624,14 @@ class SpectralCoord(SpectralQuantity):
         observer = update_differentials_to_match(self.observer, frame,
                                                  preserve_observer_frame=preserve_observer_frame)
 
-        new_coord = self._change_observer_to(observer)
+        # Calculate the initial and final los velocity
+        init_obs_vel = self._calculate_radial_velocity(self.observer, self.target, as_scalar=True)
+        fin_obs_vel = self._calculate_radial_velocity(observer, self.target, as_scalar=True)
+
+        # Apply transformation to data
+        new_data = _apply_relativistic_doppler_shift(self, fin_obs_vel - init_obs_vel)
+
+        new_coord = self.replicate(value=new_data, observer=observer)
 
         return new_coord
 
@@ -788,17 +708,10 @@ class SpectralCoord(SpectralQuantity):
                         .realize_frame(observer_icrs.cartesian.with_differentials(observer_velocity))
                         .transform_to(self._observer))
 
-        init_obs_vel = self._calculate_radial_velocity(observer_icrs, target_icrs)
-        fin_obs_vel = self._calculate_radial_velocity(new_observer, new_target)
+        init_obs_vel = self._calculate_radial_velocity(observer_icrs, target_icrs, as_scalar=True)
+        fin_obs_vel = self._calculate_radial_velocity(new_observer, new_target, as_scalar=True)
 
-        line_of_sight_unit_vec = self._normalized_position_vector(observer_icrs, target_icrs)
-
-        new_data = self._project_velocity_and_shift(init_obs_vel, fin_obs_vel, line_of_sight_unit_vec)
-
-        if self.observer is None or self.target is None:
-            radial_velocity = line_of_sight_unit_vec.dot(fin_obs_vel)
-        else:
-            radial_velocity = None
+        new_data = _apply_relativistic_doppler_shift(self, fin_obs_vel - init_obs_vel)
 
         # If an observer/target pair were not defined already, we want to avoid
         # providing an explicit pair, so create a new SpectralCoord object
@@ -806,9 +719,8 @@ class SpectralCoord(SpectralQuantity):
         # will be created. Otherwise, use the available observer/target
         # instances to create the pair.
         return self.replicate(value=new_data,
-                              observer=new_observer if self.observer is not None else None,
-                              target=new_target if self.target is not None else None,
-                              radial_velocity=radial_velocity)
+                              observer=new_observer,
+                              target=new_target)
 
     def to_rest(self):
         """
--- a/astropy/coordinates/spectral_coordinate.py
+++ b/astropy/coordinates/spectral_coordinate.py
@@ -656,7 +656,8 @@ class SpectralCoord(SpectralQuantity):
         -------
         `SpectralCoord`
             New spectral coordinate with the target/observer velocity changed
-            to incorporate the shift.
+            to incorporate the shift. This is always a new object even if
+            ``target_shift`` and ``observer_shift`` are both `None`.
         """
 
         if observer_shift is not None and (self.target is None or
--- a/astropy/coordinates/spectral_coordinate.py
+++ b/astropy/coordinates/spectral_coordinate.py
@@ -294,7 +294,7 @@ class SpectralCoord(SpectralQuantity):
         if 's' not in coord.data.differentials:
             warnings.warn(
                 "No velocity defined on frame, assuming {}.".format(
-                    u.Quantity([0, 0, 0], unit=u.km/u.s)),
+                    ZERO_VELOCITIES),
                 NoVelocityWarning)
 
             coord = attach_zero_velocities(coord)
--- a/astropy/coordinates/spectral_coordinate.py
+++ b/astropy/coordinates/spectral_coordinate.py
@@ -680,7 +680,7 @@ class SpectralCoord(SpectralQuantity):
                 target_shift = _redshift_to_velocity(target_shift)
             if self._observer is None or self._target is None:
                 return self.replicate(value=_apply_relativistic_doppler_shift(self, target_shift),
-                                      radial_velocity=self._radial_velocity + target_shift)
+                                      radial_velocity=self.radial_velocity + target_shift)
 
         if observer_shift is None:
             observer_shift = 0 * KMS
--- a/astropy/coordinates/spectral_coordinate.py
+++ b/astropy/coordinates/spectral_coordinate.py
@@ -24,6 +24,7 @@ class NoDistanceWarning(AstropyUserWarning):
 
 
 KMS = u.km / u.s
+C_KMS = c.to(KMS)
 ZERO_VELOCITIES = CartesianDifferential([0, 0, 0] * KMS)
 
 # Default distance to use for target when none is provided
@@ -37,7 +38,7 @@ def _velocity_to_redshift(velocity):
     """
     Convert a velocity to a relativistic redshift.
     """
-    beta = velocity / c
+    beta = velocity / C_KMS
     return np.sqrt((1 + beta) / (1 - beta)) - 1
 
 
@@ -46,7 +47,7 @@ def _redshift_to_velocity(redshift):
     Convert a relativistic redshift to a velocity.
     """
     zponesq = (1 + redshift) ** 2
-    return (c * (zponesq - 1) / (zponesq + 1))
+    return (C_KMS * (zponesq - 1) / (zponesq + 1))
 
 
 def _relativistic_velocity_addition(vel1, vel2):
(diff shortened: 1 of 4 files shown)


astropy/wcs/wcsapi/sliced_low_level_wcs.py
```diff
@@ -3,8 +3,3 @@
 from .wrappers.sliced_wcs import SlicedLowLevelWCS, sanitize_slices
 from astropy.utils.exceptions import AstropyDeprecationWarning
 
-warnings.warn(
```
Cadair, line 6: why you take away my warning? :sob:

astrofrog: Oops sorry :)

astrofrog: Fixed

astropy/coordinates/spectral_coordinate.py
```diff
@@ -515,8 +460,14 @@ def radial_velocity(self):
         coordinate frame in that this calculates the radial velocity with
         respect to the *observer*, not the origin of the frame.
         """
-        return self._calculate_radial_velocity(self._observer, self._target,
-                                               as_scalar=True)
+        if self._observer is None or self._target is None:
+            if self._radial_velocity is None:
```
mhvk, line 464: Is it good to have this as a default? Or would `AttributeError` be more reasonable? Though I can see that if one wants there to be a (reasonable) default of 0, it is good to keep track of the fact that it is entered as a default, not an explicit value.

astropy/coordinates/spectral_coordinate.py
```diff
@@ -738,7 +602,7 @@ def with_observer_stationary_relative_to(self, frame, velocity=None, preserve_ob
                 # otherwise frame is ready to go
             else:
                 if velocity is None:
-                    differentials = CartesianDifferential(0 * u.m / u.s, 0 * u.m / u.s, 0 * u.m / u.s)
+                    differentials = ZERO_VELOCITIES
                 else:
                     differentials = CartesianDifferential(*velocity)
```
mhvk, line 607: This makes me wonder, why not allow the velocity to be specified as a `CartesianDifferential` in the first place? With a recent PR of mine, that will now be passed along, but you have to *not* expand the velocity (which is fine, a triplet can be used as input anyway). Anyway, bottom line, would suggest to remove the `*` here and update the docstring that a `CartesianDifferential` is allowed...

mhvk: And then below too work in terms of a differential... (since this is really rather off-topic, do feel free to ignore).

astropy/coordinates/tests/test_spectral_coordinate.py
```diff
@@ -520,6 +520,13 @@ def test_los_shift_radial_velocity():
     sc12 = sc10.with_radial_velocity_shift(-3 * u.km / u.s)
     assert_quantity_allclose(sc12.radial_velocity, -2 * u.km / u.s)
 
+    # Check that things work if radial_velocity wasn't specified at all
+
+    sc13 = SpectralCoord(500 * u.nm)
+    sc14 = sc13.with_radial_velocity_shift(1 * u.km / u.s)
```
mhvk, line 526: Probably should check what `radial_velocity` gives here too.

astropy/coordinates/spectral_coordinate.py
```diff
@@ -799,11 +671,16 @@ def with_radial_velocity_shift(self, target_shift=None, observer_shift=None):
         #  assume it's a redshift float value and convert to velocity
 
         if target_shift is None:
+            if self._observer is None or self._target is None:
+                return self.replicate()
```
nmearl, line 675: Is there a particular reason we still replicate instead of just returning `self`?

astropy/coordinates/spectral_coordinate.py
```diff
@@ -831,27 +708,19 @@ def with_radial_velocity_shift(self, target_shift=None, observer_shift=None):
...
-            radial_velocity = line_of_sight_unit_vec.dot(fin_obs_vel)
-        else:
-            radial_velocity = None
+        new_data = _apply_relativistic_doppler_shift(self, fin_obs_vel - init_obs_vel)
 
         # If an observer/target pair were not defined already, we want to avoid
         # providing an explicit pair, so create a new SpectralCoord object
         # with just the radial velocity set so new implicit observer/target
         # will be created. Otherwise, use the available observer/target
         # instances to create the pair.
```
nmearl, line 720:
```suggestion

```
No longer relevant, really.

astropy/coordinates/tests/test_spectral_coordinate.py
nmearl: Isn't that done in the line below?

astropy/coordinates/spectral_coordinate.py
pllim: Re: just returning self -- Then it won't be a copy anymore, like the docstring claims?

astropy/coordinates/spectral_coordinate.py
```diff
@@ -14,7 +14,17 @@
 
 __all__ = ['SpectralCoord']
 
+
+class NoVelocityWarning(AstropyUserWarning):
+    pass
+
+
+class NiDistanceWarning(AstropyUserWarning):
```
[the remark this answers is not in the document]
astrofrog: ni

astropy/coordinates/spectral_coordinate.py
```diff
@@ -334,13 +279,14 @@ def _validate_coordinate(coord, label=''):
         # for generating differentials
         # TODO: change this to not set the distance and yield a warning once
         # there's a good way to address this in astropy.coordinates
-        if hasattr(coord, 'distance') and \
-                coord.distance.unit.physical_type == 'dimensionless':
+        with np.errstate(all='ignore'):
+            distance = getattr(coord, 'distance', None)
```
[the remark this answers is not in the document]
astrofrog: See https://github.com/astropy/astropy/issues/10247 - I'll add a link to the issue in the comment above

astropy/coordinates/spectral_coordinate.py
astrofrog: Yes the idea was to differentiate between explicitly specified zero versus default though maybe that distinction is unimportant. I don't feel strongly about this, but I think raising an ``AttributeError`` would be too much.

astropy/coordinates/spectral_coordinate.py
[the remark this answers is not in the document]
astrofrog: Fixed :)

astropy/coordinates/spectral_coordinate.py
```diff
@@ -14,7 +14,17 @@
...
+class NoVelocityWarning(AstropyUserWarning):
+    pass
+
+
+class NiDistanceWarning(AstropyUserWarning):
+    pass
+
+
 KMS = u.km / u.s
+ZERO_VELOCITIES = CartesianDifferential(u.Quantity([0, 0, 0] * KMS))
```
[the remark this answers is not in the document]
astrofrog: Fixed

astropy/coordinates/spectral_coordinate.py
```diff
@@ -36,7 +46,7 @@ def _redshift_to_velocity(redshift):
     Convert a relativistic redshift to a velocity.
     """
     zponesq = (1 + redshift) ** 2
-    return c * (zponesq - 1) / (zponesq + 1)
+    return (c * (zponesq - 1) / (zponesq + 1)).to(u.km / u.s)
```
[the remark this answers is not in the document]
astrofrog: Removed

astropy/coordinates/spectral_coordinate.py
```diff
@@ -128,11 +138,18 @@ def attach_zero_velocities(coord):
     """
     Set the differentials to be stationary on a coordinate object.
     """
-    coord_diffs = CartesianDifferential(u.Quantity([0, 0, 0] * KMS))
+    coord_diffs = ZERO_VELOCITIES
```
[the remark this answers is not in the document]
astrofrog: Fixed

astropy/coordinates/spectral_coordinate.py
```diff
@@ -334,13 +279,14 @@ def _validate_coordinate(coord, label=''):
...
-                coord.distance.unit.physical_type == 'dimensionless':
+        with np.errstate(all='ignore'):
+            distance = getattr(coord, 'distance', None)
+        if distance is not None and distance.unit.physical_type == 'dimensionless':
             coord = SkyCoord(coord, distance=DEFAULT_DISTANCE)
             warnings.warn(
                 "Distance on coordinate object is dimensionless, an "
                 f"abritrary distance value of {DEFAULT_DISTANCE} will be set instead.",
-                AstropyUserWarning)
+                NiDistanceWarning)
```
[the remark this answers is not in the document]
astrofrog: Fixed

astropy/coordinates/spectral_coordinate.py
```diff
@@ -407,6 +353,12 @@ def replicate(self, value=None, unit=None,
             else:
                 value, unit = value.value, value.unit
 
+        if observer is not None:
```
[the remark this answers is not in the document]
astrofrog: Agree, fixed!

astropy/coordinates/spectral_coordinate.py
```diff
@@ -407,6 +353,12 @@ def replicate(self, value=None, unit=None,
...
+            observer = self._validate_coordinate(observer)
+
+        if target is not None:
+            target = self._validate_coordinate(target)
+
         value = value if value is not None else self.value
         unit = unit or self.unit
         observer = observer or self.observer
```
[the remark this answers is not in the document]
astrofrog: Done

astropy/coordinates/spectral_coordinate.py
```diff
@@ -324,6 +265,10 @@ def _validate_coordinate(coord, label=''):
         coord : `~astropy.coordinates.BaseCoordinateFrame`
             The new frame to be used for target or observer.
         """
```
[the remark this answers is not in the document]
astrofrog: Done

astropy/coordinates/spectral_coordinate.py
```diff
@@ -452,28 +406,19 @@ def observer(self):
...
-        # The default target is based off the observer frame. In the case
-        # where both observer/target are initialized to defaults, and then
-        # the user sets a new observer, we need to create a new target based
-        # on the input frame to maintain rv/redshift continuity.
-        if self.target is None:
-            self._target = self._create_coord_from_offset(
-                value, self.radial_velocity)
-
-        self._observer = value
+        # Switch to auto-computing radial velocity
```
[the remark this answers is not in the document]
astrofrog: Done

astropy/coordinates/spectral_coordinate.py
```diff
@@ -485,19 +430,19 @@ def target(self):
...
 
-        self._target_specified = value is not None
+        self._target = self._validate_coordinate(value, label='target')
 
-        value = self._validate_coordinate(value, label='target')
-
-        self._target = value
+        # Switch to auto-computing radial velocity
+        if self._observer is not None:
+            self._radial_velocity = None
```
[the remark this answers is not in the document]
astrofrog: Done

astropy/coordinates/spectral_coordinate.py
astrofrog: I've adjusted it to allow ``CartesianDifferential`` to be passed in

astropy/coordinates/tests/test_spectral_coordinate.py
astrofrog: Indeed this is done below

astropy/coordinates/spectral_coordinate.py
astrofrog: Yep - I would always expect replicate to return a copy

astrofrog:
@mhvk - thanks for the detailed review! I think I've addressed all your comments

@nmearl - thanks for the suggestions! I've included them here.

astropy/coordinates/spectral_coordinate.py
nmearl: Right, of course -- `replicate` _should_ return a copy. My main question is: why are we replicating in this case? If a user gets here, they're not quite using the method as intended _and_ they're duplicating data which gives the impression that ✨ something happened ✨, but really we internally deal with this as "no observer or target; can't do anything; returning". Perhaps a warning to make the user aware, or an error to make it clear that this method _does_ require arguments to work properly?

Overall, a minor quibble, so no need to address this with any earnest. 😁

pllim: If they not supposed to get to that part of the code, then throwing an exception would be the safest?

astrofrog:
I've rebased this - I'd really like to get this in v4.2 if possible.

@mhvk @nmearl - are you happy with this now?

Anyone else want to review this? (@eteq?)

astropy/coordinates/spectral_coordinate.py
mhvk: @astrofrog - no conclusion seems to have been reached here...

mhvk:
@astrofrog - I think there is just one outstanding comment... Otherwise looks good!

astropy/coordinates/spectral_coordinate.py
astrofrog: There could be programmatic reasons why this method gets called with no arguments (e.g if iterating over many objects and using some logic that means that sometimes it is called with no arguments) - in this case I would always want the results to be new objects, not self sometimes and a new object other times. But maybe I can make this explicit in the docstring.

astrofrog: See 3465d53abe7b66922cc55848d91419a3457e12fb for a clarification in the docstring. Is this ok now?

mhvk: Looks good to me

mhvk (approved):
I think it all looks good to me - agreed this is a nice simplification that should find its way in!

astrofrog:
Ok thanks! I'll look into fixing the CI today.

Cadair (approved):
Well it didn't break the ndcube tests...

Cadair:
@astrofrog I rebased to fix the merge conflict and fixed the flake8 issue :smile:

astrofrog:
@StuartLittlefair @adrn @astrojuanlu @eteq @mhvk - it would be best for this to be merged by a coordinates maintainer, so I'll leave it up to you to merge. I think it'd be a shame to miss v4.2 though so please merge today if possible! 🙏

StuartLittlefair:
Hi @astrofrog - since @mhvk has run his fine tooth comb over this PR I'm happy to merge in principle, but I notice from the coverage that the cases where, for example, ```self._target``` and ```self._observer``` are both None are not being tested. There are a number of these lines. 

Maybe worth looking to see if the code is well tested in these instances before hitting the button?

astrofrog:
@StuartLittlefair - I've tried to improve the test coverage, now 97% for the ``spectral_coordinate.py`` file. I won't have time to get to 100% today but hopefully this should be good enough?

pllim:
Tsk tsk... Should have checked the remote data job. Can you please fix the failing doctest?

```
_________________________ [doctest] spectralcoord.rst __________________________
310 method. This method can take the name of an existing coordinate/velocity frame,
311 a :class:`~astropy.coordinates.BaseCoordinateFrame` instance, or any arbitrary
312 3D position and velocity coordinate object defined either as a
313 :class:`~astropy.coordinates.BaseCoordinateFrame` or a |SkyCoord| object. Most
314 commonly-used frames are accessible using strings. For example to transform to a
315 velocity frame stationary with respect to the center of the Earth (so removing
316 the effect of the Earth's rotation), we can use the ``'gcrs'`` which stands for
317 *Geocentric Celestial Reference System* (GCRS)::
318 
319     >>> sc_ttau.with_observer_stationary_relative_to('gcrs')  # doctest: +FLOAT_CMP +REMOTE_DATA
    @@ -1,7 +1,7 @@
    -<SpectralCoord
    +<SpectralCoord
        (observer: <GCRS Coordinate (obstime=2019-04-24T02:32:10.000, obsgeoloc=(0., 0., 0.) m, obsgeovel=(0., 0., 0.) m / s): (x, y, z) in m
    -                  (-5878853.86171412, -192921.84773269, -2470794.19765021)
    +                  (-5878853.86172938, -192921.84773269, -2470794.19765021)
                    (v_x, v_y, v_z) in km / s
    -                  (4.33251262e-09, 8.96175625e-08, -1.49258412e-08)>
    +                  (-1.09244138e-08, 8.96161946e-08, -1.49267726e-08)>
         target: <ICRS Coordinate: (ra, dec, distance) in (deg, deg, pc)
                     (65.497625, 19.53511111, 144.321)
    @@ -9,6 +9,6 @@
                     (1.37949782e-15, 1.46375638e-15, 23.9)>
         observer to target (computed from above):
    -      radial_velocity=40.674086368345165 km / s
    -      redshift=0.00013568335316072044)
    +      radial_velocity=40.67408630725199 km / s
    +      redshift=0.0001356833529568835)
       [200.00024141, 210.00025348, 220.00026555, 230.00027762, 240.00028969,
        250.00030176, 260.00031383, 270.0003259 , 280.00033797, 290.00035004,
/home/travis/build/astropy/astropy/docs/coordinates/spectralcoord.rst:319: DocTestFailure
```

https://travis-ci.com/github/astropy/astropy/jobs/404743171 (This PR)

https://travis-ci.com/github/astropy/astropy/jobs/404850035 (On `master` after merge)

astrofrog:
Huh wait but I did check it locally and that's what led to the doctest updates in https://github.com/astropy/astropy/pull/10398/commits/26442d19cdfc87f3492f21cc638694145f525b03 - I wonder why the results are different on Travis?!

pllim:
> I wonder why the results are different on Travis?!

The same job also uses numpy dev, though not sure if that is the cause or not. If only we can use https://docs.travis-ci.com/user/running-build-in-debug-mode/

StuartLittlefair:
argh. I'm away from the computer until late tonight. Sorry for the premature merge!

mhvk:
Looks like some rounding error problem - these velocities are equal to 0 km/s within a few μm/s... Just skip the test until we have a `FLOAT_CMP` where we also can give absolute tolerances?

pllim:
Started attempt at #10928

PR #10400, opened 2020-05-26
Don't set observer or target on SpectralCoord in FITS WCS if they can't be converted to ICRS

### LINKED ISSUE #10399: Spectral Coord tries to convert target to ICRS (even when it can't)
When a spectral coord is constructed with a target (i.e. from the celestial axes of a WCS) but not an observer, and that target coordinate is unable to be transformed to ICRS it errors.

I discovered this in the NDCube test suite, and am currently unable to reproduce on its own. Opening this so I don't forget and so that it blocks 4.1.

astrofrog:
This is a fix (I think) for https://github.com/astropy/astropy/issues/10399 - basically SpectralCoord as it is written right now works properly only if observer and/or target can be converted to ICRS. This requirement meant that the APE 14 WCS API was broken compared to previously in some cases where the observer or target was not convertible to ICRS. To preserve backward-compatibility, this PR therefore discards observer/target information if it cannot be used in SpectralCoord - this is not a step backwards compared to 4.0 since there the WCS just returned a Quantity anyway.

Refactoring such as #10398 allows the ICRS restriction to be relaxed a bit but we should maybe leave refactoring that significant to 4.2 in which case this PR should I think provide a stop-gap measure to make sure we don't break some WCS transformations as described in #10399.

EDIT: Fix #10399

### DIFF
--- a/astropy/wcs/wcsapi/fitswcs.py
+++ b/astropy/wcs/wcsapi/fitswcs.py
@@ -8,7 +8,7 @@ import warnings
 import numpy as np
 
 from astropy import units as u
-from astropy.coordinates import SpectralCoord, Galactic
+from astropy.coordinates import SpectralCoord, Galactic, ICRS
 from astropy.coordinates.spectral_coordinate import update_differentials_to_match, attach_zero_velocities
 from astropy.utils.exceptions import AstropyUserWarning
 from astropy.constants import c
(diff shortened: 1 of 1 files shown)


pllim:
I think someone from `specutils` should also sign off on this but I am not sure who to ping for that.

Cadair:
I can't vouch for the fix, but it makes the ndcube tests pass so, lgtm.

nmearl:
The reason we try and convert everything to ICRS first is that generally from there most transformations don't use finite differencing which is preferable for precision sake. Instead of just discarding the observer/target information, we could raise a warning stating "Can't transform to ICRS, further transformations may use finite differencing and therefore might lose precision.", and allow the observer/target information to be added in whatever frame they were provided.

astrofrog:
@nmearl - just to be clear, the issue here is not just in the conversion to ICRS in the transformations but actually in the setting up of the virtual/hidden observer/target, which also converts to ICRS. I tried to fix that by avoiding the conversion to ICRS but ran into issues. I agree that in the long term that would be a better fix though.

I don't have time to work on this further for the 4.1 release, so I would still suggest for now that we merge this present PR to fix the backward-compatibility issue with FITS WCS, and then consider adding the fallback to no ICRS conversion with a warning as part of #10398 for 4.2?

pllim:
Two approvals, so merging. #yolo

PR #10419, opened 2020-05-29
Allow zero-length table from names and structured-array dtype

### LINKED ISSUE #10393: Initialising Table with structured dtype produces inconsistent/confusing results
### Description
A discussion in the Facebook group brought up that a new, empty table can be constructed from an existing table by passing its `dtype`:
```
>>> oldtable.dtype
dtype([('wavelength', '<f8'), ('flux', '<f4'), ('uncertainty', '<f4')])
>>> Table(dtype=oldtable.dtype)
<Table length=0>
wavelength   flux  uncertainty
 float64   float32   float32  
---------- ------- -----------
```
and it is equally possible to initialise it in the more pedestrian manner
```
>>> newtable = Table(names=oldtable.colnames, dtype=[dt for n, dt in oldtable.dtype.descr])
```
but directly using the old `dtype` if `names` are already specified throws these exceptions:
```
 >>> newtable = Table(names=oldtable.colnames, dtype=oldtable.dtype)
  File "/Users/derek/lib/python3.8/site-packages/astropy/table/table.py", line 621, in __init__
    self._check_names_dtype(names, dtype, n_cols)
  File "/Users/derek/lib/python3.8/site-packages/astropy/table/table.py", line 873, in _check_names_dtype
    raise ValueError(f'{inp_str} must be a list or None')
ValueError: dtype must be a list or None

>>> newtable = Table(names=oldtable.colnames, dtype=oldtable.dtype.descr)
  File "/Users/derek/lib/python3.8/site-packages/astropy/table/table.py", line 1088, in _convert_data_to_col
    col = col_cls(name=name, data=data, dtype=dtype,
  File "/Users/derek/lib/python3.8/site-packages/astropy/table/column.py", line 879, in __new__
    self = super().__new__(
  File "/Users/derek/lib/python3.8/site-packages/astropy/table/column.py", line 254, in __new__
    if np.dtype(dtype).char == 'S':
TypeError: data type 'wavelength' not understood
```
The same happens when the `Table` is initialised with a structured array.

### Expected behavior

`Table` should either raise a more meaningful error, or (perhaps preferably) accept the structured `dtype`, possibly with a warning that `dtype.names` will be replaced with the specified `names`.

### Actual behavior

See above. While the recommended method in the docs `newtable = oldtable[:0].copy()` works (and has the advantage of additionally copying any `unit` properties), the discrepant behaviour between specifying `dtype` alone, and together with `names` is potentially confusing, and the exceptions not very clear.

https://github.com/astropy/astropy/blob/eeb9b7ac0e18b262fae64cd2fa46d0e1cff1e4d5/astropy/table/table.py#L592-L594

"unpacks" the `dtype` automatically when no other `names` information is available, and the same might be done with `names` set (but discarding `dtype.names`, since `names` should certainly take priority).
 
On a slightly related note, it would be convenient to have a method to directly get all units for the table columns in a form that could be passed as `units=`. Of course one can return to the `newtable = oldtable[:0].copy()` method, but if modifying names or dtypes is desired, such an option would come in handy.

### System Details
Astropy version 4.2.dev112+g595484597.
Date: 2020-05-22T19:44:39
Platform: macOS-10.12.6-x86_64-i386-64bit
Executable: /sw/bin/python3.8
Full Python Version: 
3.8.3 (default, May 14 2020, 22:17:35) 

Package versions:
Numpy: 1.19.0rc1
Scipy: 1.4.1
Matplotlib: 3.2.1
h5py: 2.9.0
Pandas: not available
Cython: 0.29.16

taldcroft:
### Description

This pull request is to allow initializing a zero-length table using a list of names and a dtype from a structured array.

Fixes #10393

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -595,13 +595,6 @@ class Table:
                 except Exception:
                     raise ValueError('dtype was specified but could not be '
                                      'parsed for column names')
-            elif dtype is not None:
-                # Names supplied along with structured-array dtype. Convert to
-                # a list of dtypes for later use. If dtype is already a list of
-                # dtype-specifiers e.g. ['f4', 'i4'] then skip this.
-                if isinstance(dtype, np.dtype):
-                    dtype = [dtype[name] for name in dtype.names]
-
             # names is guaranteed to be set at this point
             init_func = self._init_from_list
             n_cols = len(names)
@@ -620,6 +613,9 @@ class Table:
             names = default_names or [None] * n_cols
         if dtype is None:
             dtype = [None] * n_cols
+        elif isinstance(dtype, np.dtype):
+            # Convert a numpy dtype input to a list of dtypes for later use.
+            dtype = [dtype[name] for name in dtype.names]
 
         # Numpy does not support bytes column names on Python 3, so fix them
         # up now.
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -595,6 +595,13 @@ class Table:
                 except Exception:
                     raise ValueError('dtype was specified but could not be '
                                      'parsed for column names')
+            elif dtype is not None:
+                # Names supplied along with structured-array dtype. Convert to
+                # a list of dtypes for later use. If dtype is already a list of
+                # dtype-specifiers e.g. ['f4', 'i4'] then skip this.
+                if isinstance(dtype, np.dtype):
+                    dtype = [dtype[name] for name in dtype.names]
+
             # names is guaranteed to be set at this point
             init_func = self._init_from_list
             n_cols = len(names)
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -609,13 +609,16 @@ class Table:
         # within the appropriate initialization routine, either from
         # existing specification or auto-generated.
 
-        if names is None:
-            names = default_names or [None] * n_cols
         if dtype is None:
             dtype = [None] * n_cols
         elif isinstance(dtype, np.dtype):
+            if default_names is None:
+                default_names = dtype.names
             # Convert a numpy dtype input to a list of dtypes for later use.
             dtype = [dtype[name] for name in dtype.names]
+
+        if names is None:
+            names = default_names or [None] * n_cols
 
         # Numpy does not support bytes column names on Python 3, so fix them
         # up now.
--- a/astropy/table/tests/test_table.py
+++ b/astropy/table/tests/test_table.py
@@ -2208,18 +2208,6 @@ def test_table_meta_copy():
     assert t2.meta[1] is not t.meta[1]  # Value is NOT the same list object
 
 
-@pytest.mark.parametrize('has_data', [True, False])
-def test_init_table_with_names_and_structured_dtype(has_data):
-    """Test fix for #10393"""
-    arr = np.ones(2, dtype=np.dtype([('a', 'i4'), ('b', 'f4')]))
-    data_args = [arr] if has_data else []
-    t = Table(*data_args, names=['x', 'y'], dtype=arr.dtype)
-    assert t.colnames == ['x', 'y']
-    assert str(t['x'].dtype) == 'int32'
-    assert str(t['y'].dtype) == 'float32'
-    assert len(t) == (2 if has_data else 0)
-
-
 def test_table_meta_copy_with_meta_arg():
     """
     Test no copy vs light (key) copy vs deep copy of table meta when meta is
--- a/astropy/table/tests/test_table.py
+++ b/astropy/table/tests/test_table.py
@@ -369,11 +369,11 @@ class TestRound():
             t.round(0.5)
 
     def test_round_kind(self, table_types):
-        for typecode in 'bBhHiIlLqQpPefdgFDG': # AllInteger, AllFloat
+        for typecode in 'bBhHiIlLqQpPefdgFDG':  # AllInteger, AllFloat
             arr = np.array([4, 16], dtype=typecode)
             t = Table([arr])
             col0 = t['col0']
-            t.round(decimals=-1) # Round to nearest 10
+            t.round(decimals=-1)  # Round to nearest 10
             assert np.all(t['col0'] == [0, 20])
             assert t['col0'] is col0
 
--- a/astropy/table/tests/test_table.py
+++ b/astropy/table/tests/test_table.py
@@ -2208,6 +2208,18 @@ def test_table_meta_copy():
     assert t2.meta[1] is not t.meta[1]  # Value is NOT the same list object
 
 
+def test_init_table_with_names_and_structured_dtype():
+    """Test fix for #10393"""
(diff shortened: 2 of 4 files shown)


astropy/table/table.py
```diff
@@ -595,6 +595,13 @@ def __init__(self, data=None, masked=False, names=None, dtype=None,
                 except Exception:
                     raise ValueError('dtype was specified but could not be '
                                      'parsed for column names')
+            elif dtype is not None:
```
dhomeier, line 598: My idea was actually to move this outside of the `data is None` block to
```
        if names is None:
            names = default_names or [None] * n_cols
        if dtype is None:
            dtype = [None] * n_cols
        elif isinstance(dtype, np.dtype):  
            # Names supplied along with structured-array dtype. Convert to
            # a list of dtypes for later use. If dtype is already a list of
            # dtype-specifiers e.g. ['f4', 'i4'] then skip this.
            dtype = [dtype[name] for name in dtype.names]
```
so this would also allow changing the original `data.dtype` from an `np.dtype` (rather than just a list of dtypes).

astropy/table/tests/test_table.py
```diff
@@ -2208,6 +2208,18 @@ def test_table_meta_copy():
...
+    """Test fix for #10393"""
+    arr = np.ones(2, dtype=np.dtype([('a', 'i4'), ('b', 'f4')]))
+    t1 = Table(arr)
+    for dtype in (arr.dtype, t1.dtype):
+        t2 = Table(names=['x', 'y'], dtype=arr.dtype)
+        assert t2.colnames == ['x', 'y']
+        assert str(t2['x'].dtype) == 'int32'
+        assert str(t2['y'].dtype) == 'float32'
+        assert len(t2) == 0
+
```
dhomeier, line 2221:
```suggestion

    dt2 = np.dtype([('a', 'i2'), ('b', 'f8')]))
    t3 = Table(data=arr, names=['x', 'y'], dtype=dt2)
    assert t3.colnames == ['x', 'y']
    assert str(t3['x'].dtype) == 'int16'
    assert str(t3['y'].dtype) == 'float64'
    assert len(t3) == 2


```
With the proposed modifications to `table.py` this should work as well.

astropy/table/tests/test_table.py
```diff
@@ -2208,6 +2208,18 @@ def test_table_meta_copy():
     assert t2.meta[1] is not t.meta[1]  # Value is NOT the same list object
 
 
+def test_init_table_with_names_and_structured_dtype():
+    """Test fix for #10393"""
+    arr = np.ones(2, dtype=np.dtype([('a', 'i4'), ('b', 'f4')]))
+    t1 = Table(arr)
+    for dtype in (arr.dtype, t1.dtype):
+        t2 = Table(names=['x', 'y'], dtype=arr.dtype)
```
dhomeier, line 2216:
```suggestion
        t2 = Table(names=['x', 'y'], dtype=dtype)
```
don't you want to iterate over both? ;-)

dhomeier:
Re. my suggestion the `t3 = Table(data=arr, names=['x', 'y'], dtype=dt2)` case was not part of the original issue, but since `t3 = Table(data=arr, names=['x', 'y'], dtype=['i2', 'f8'])` already works, there is no reason this should not work with an `np.dtype`.

The latter brings one additional corner case: with
```
dt2 = np.dtype([('c', 'i2'), ('d', 'f8')]))
```
should creating the new array without explicit naming
```
t3 = Table(data=arr, dtype=dt2)
```
let the `dtype` names `'c', 'd'` override the original ones? I tend to say yes, since they were more explicitly given, but that might give rise to unexpected side effects (although any such use would raise an exception until this PR).
This would need some reordering of the `names` and `dtype` assignments
```
        if dtype is None:
            dtype = [None] * n_cols
        elif isinstance(dtype, np.dtype):
            # Names supplied along with structured-array dtype. Convert to
            # a list of dtypes for later use. If dtype is already a list of
            # dtype-specifiers e.g. ['f4', 'i4'] then skip this.
            # This will rename original names from data.
            default_names = dtype.names
            dtype = [dtype[name] for name in default_names]
        if names is None:
            names = default_names or [None] * n_cols
```
so the `names=` argument would still take highest priority.

taldcroft:
@dhomeier - thanks for the suggestions and comments!  I have incorporated them into the current version, albeit with small mods.

About the corner case `t3 = Table(data=arr, dtype=dt2)`, I'm inclined to leave this as taking column names from `data`. 

I think the goal should be a clear and consistent precedence (which we could then document):
- `names`
- `data`
- `dtype`

So right now when `data` does not specify the column names then it will fall through to autonaming `col0` etc.  I'm not sure when I'll have a chance now to look at the code to ensure this consistency, but will plan to get to it (if you don't first :smile:).

dhomeier:
@taldcroft thanks for the update!

> I think the goal should be a clear and consistent precedence (which we could then document):
> 
> * `names`
> * `data`
> * `dtype`

Admittedly the more logical precedence to me seemed

* `names`
* `dtype`
* `data`

both from the rule "explicit over implicit" and the fact that `dtype` proper replaces `data.dtype`, so one might expect `dtype.names` to also take precedence.
But in fact I like your solution now better since it gives more flexibility; as mentioned above you can simply change the precedence by passing `names=dtype.names` as well.

Looks good as is, only for the tests my suggestion also included a check if the new table actually has its `dtype` changed to that of the `dtype` arg, if different. But `test_init_table.py` already has numerous tests for the `dtype=[...]` syntax, so maybe no separate test is needed.

I am not too worried about an unnamed `data` input keeping the default names, since anything but the simple case of all identically typed columns will have a structured `np.dtype` anyway.

mhvk:
@taldcroft, @dhomeier - I removed myself as reviewer as you seem to have converged here; let me know if that was wrong.

astropy/table/tests/test_table.py
taldcroft: Just getting back to this. l must be missing something, but I can't see how this is functionally different from the case of `has_data = True`. I'll go ahead and move this test function to `test_init_table.py`, but want to understand if your little snippet needs to go in as well.

dhomeier: > Just getting back to this. l must be missing something, but I can't see how this is functionally different from the case of `has_data = True`

I missed that this still was a loose end. The test differs in the different `dtype` - `dtype=arr.dtype` is redundant per the behaviour we have now agreed on, since it already defaults to the input `arr`'s dtype. But for `t3` it is changed to `('int16', 'float64')`. Functionally this is equivalent to what ` test_set_dtype` is already testing, but that would use the `dtype=('i2', 'f8')` syntax. So I think it might be worth testing again with an `np.dtype` object here (and verify that `names=[]` still takes precedence for the colnames, which is perhaps less obvious and more easily broken).

mhvk:
@dhomeier - OK, I'll approve on your behalf once the last bits are sorted out!

mhvk:
Failure is unrelated, so merging. Thanks, @taldcroft for the fix and @dhomeier for the review!

dhomeier:
Thanks both for bringing the fix in!

PR #10421, opened 2020-06-01
Used f-string to concatenate 'bibtexreference' and 'url'

jerobado:
### Description

This pull request is to address the [Coding Guideline](https://docs.astropy.org/en/latest/development/codeguide.html) wherein f-strings should be used when possible.

Fixes N/A

### DIFF
--- a/astropy/__init__.py
+++ b/astropy/__init__.py
@@ -326,7 +326,7 @@ def _get_bibtex():
     with open(citation_file, 'r') as citation:
         refs = citation.read().split('@ARTICLE')[1:]
         if len(refs) == 0: return ''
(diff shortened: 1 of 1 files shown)


taldcroft:
Looks OK to me.  There is a point when using f-strings with embedded code *decreases* code clarity by making it harder to read.  Probably the second change in this PR is slightly over that line (in other words, I would have left the second one as is), but is OK.

jerobado:
Thank you for the feedback, this is well noted.

pllim:
Thanks, @jerobado ! And thanks for the extra pairs of eyes, @taldcroft !

Issue #10119, opened 2020-04-06
Make documentation more clear on lack of support for masked quantities in tables


ehsteve:
### Description

The current documentation makes it seem like it should be possible to have a column with quantities that are masked. `QTable` for example allows for `masked=True`. There are many examples throughout the documentation (for example [here](https://astropy.readthedocs.io/en/latest/table/masking.html#table-creation)) which show a masked column in a `QTable`.

### Expected behavior
```python 
from astropy.table import QTable, MaskedColumn
import astropy.units as u

tbl = QTable(masked=True)
tbl.add_column(MaskedColumn([1, 2] * u.keV, mask=[True, False]))
print(tbl)

output

col0
keV 
----
 -
 2.0
```

### Actual behavior

```python 
from astropy.table import QTable, MaskedColumn
import astropy.units as u

tbl = QTable(masked=True)
tbl.add_column(MaskedColumn([1, 2] * u.keV, mask=[True, False]))
print(tbl)

output
col0
keV 
----
 1.0
 2.0
```

### Recommendation

Given that masked Quantities arrays are not a thing (see #1852), I would recommend that there be a note or warning adding to the documentation to make it crystal clear that it is currently not possible to make a `MaskedColumn` with quantities. I might even go so far as to issue a warning when a user tried to create a `MaskedColumn` with quantities.

### System Details

```
import astropy; print("astropy", astropy.__version__)
'4.0.1.post1'
```

pllim:
cc @taldcroft and @mhvk . We discussed this in Slack and I also find the documentation confusing. I thought there was an issue already about this but I could not find it. FYI.

taldcroft:
For the record, there is https://github.com/astropy/astropy/pull/6031.  This solution cannot support integer masked quantities and may have some problems in the details (as noted there, what to do with `np.sum()` of a masked quantity).  And there is https://github.com/astropy/astropy/pull/8133, which may be more robust but is a fair bit more tricky and hasn't yet seen a working implementation AFAIK. And there has been a philosophical difference about whether it is best to do masking with a dual `Quantity / MaskedQuantity` approach or one `Quantity` class with `NaN` as a marker ala Pandas.

I would like to get *something* done since since #1852 is over 6 years old and still relevant (and perfect is the enemy of good), but certainly improving documentation is better than zero.

mhvk:
[...]

PR #10423, opened 2020-06-01
Introduce a Masked class, apply to Quantity, attempt Column, Representation

mhvk:
EDIT: the base implementation is in #11127 - this one tries to use `Masked` for columns and gives a sense how it would work for representations.

@astrofrog - as per your request in #10119, here's what I have. I got things to working quite well with `Quantity` and all its subclasses; the first thing to look at may be the quite detailed tests (which should be useful for whatever we end up implementing).

I had started on `MaskedColumn` and had also tried to move the information up in representations (to go for coordinates) but that is not finished.  Part of the difficulty with `MaskedColumn` is that I made some design choices which are different from `MaskedArray`, so it would need a kind of compatibility layer if we want to keep exact one-to-one matches, and/or some way to decide which type one wants.

It may be worth pointing out two specific things:
- I very much on purpose decided to treat reductions different from `MaskedArray` - any time there is any items in the reduction masked, the result is masked as well. So, e.g., for `.sum()`, masked elements are not implicitly replaced by zero. The logic is that one cannot interpret the sum meaningfully without knowing how many elements actually contributed. Conversely, following that same logic, for methods where skipping a masked element does make sense (like `.mean()`, where one divides by the number of unmasked elements), I do use it.
- Currently, the implementation does not support structured arrays. The question for those, which I haven't really found a "best" answer for, is whether the mask should be per full item, or whether the mask itself should also be a structured array, i.e., allow masking individual pieces. `MaskedArray` does the latter and it is quite a hassle to keep track, but perhaps better.

Anyway, suggestions very welcome. Indeed, this is definitely at a point where some further design choices have to be made, and it would be good to step back a bit and decide what exactly we want masking to mean.

cc @taldcroft

EDIT: Masked column was very much a work in progress. Tests still passed without the 3 last commits (i.e., at 379a8812e87ad95995c42db45998eec147e5a57e)

### DIFF
--- a/astropy/coordinates/angles.py
+++ b/astropy/coordinates/angles.py
@@ -477,7 +477,7 @@ def _no_angle_subclass(obj):
     if isinstance(obj, tuple):
         return tuple(_no_angle_subclass(_obj) for _obj in obj)
 
-    return obj.view(Angle) if isinstance(obj, Angle) else obj
+    return obj.view(Angle) if isinstance(obj, (Latitude, Longitude)) else obj
 
 
 class Latitude(Angle):
--- a/astropy/coordinates/representation.py
+++ b/astropy/coordinates/representation.py
@@ -432,6 +432,16 @@ class BaseRepresentationOrDifferential(ShapedLikeNDArray):
                 else:
                     reshaped.append(val)
 
+    @property
+    def mask(self):
+        masks = [getattr(getattr(self, component), 'mask', False)
+                 for component in self.components]
+        masks = [mask for mask in masks if mask is not False]
+        if masks:
+            return functools.reduce(np.logical_or, masks)
+        else:
+            return np.broadcast_to(np.False_, self.shape)
+
     # Required to support multiplication and division, and defined by the base
     # representation and differential classes.
     @abc.abstractmethod
--- a/astropy/table/_column_mixins.pyx
+++ b/astropy/table/_column_mixins.pyx
@@ -24,6 +24,8 @@ Column is itself an array.
 
 import sys
 import numpy as np
+from astropy.utils.masked import MaskedNDArray
+
 
 cdef tuple INTEGER_TYPES = (int, np.integer)
 
@@ -51,10 +53,10 @@ ctypedef object (*item_getter)(object, object)
 
 
 cdef inline object base_getitem(object self, object item, item_getter getitem):
-    if (<ndarray>self).ndim > 1 and isinstance(item, INTEGER_TYPES):
-        return self.data[item]
-
-    value = getitem(self, item)
+    if isinstance(item, INTEGER_TYPES):
+        value = self.data[item]
+    else:
+        value = getitem(self, item)
 
     try:
         if value.dtype.char == 'S' and not value.shape:
@@ -74,14 +76,20 @@ cdef class _ColumnGetitemShim:
         return base_getitem(self, item, column_getitem)
 
 
-MaskedArray = np.ma.MaskedArray
-
-
-cdef inline object masked_column_getitem(object self, object item):
-    value = MaskedArray.__getitem__(self, item)
-    return self._copy_attrs_slice(value)
-
-
 cdef class _MaskedColumnGetitemShim(_ColumnGetitemShim):
     def __getitem__(self, item):
-        return base_getitem(self, item, masked_column_getitem)
+        if isinstance(item, INTEGER_TYPES):
+            value = self.unmasked[item]
+            mask = self.mask[item]
+            try:
+                if value.dtype.char == 'S' and not value.shape:
+                    value = value.decode('utf-8', errors='replace')
+            except AttributeError:
+                pass
+
+            if mask.shape or mask:
+                return MaskedNDArray(value, mask)
+            else:
+                return value
+
+        return MaskedNDArray.__getitem__(self, item)
--- a/astropy/table/column.py
+++ b/astropy/table/column.py
@@ -14,6 +14,7 @@ from astropy.utils.console import color_print
 from astropy.utils.metadata import MetaData
 from astropy.utils.data_info import BaseColumnInfo, dtype_info_name
 from astropy.utils.misc import dtype_bytes_or_chars
+from astropy.utils.masked import MaskedNDArray
 from . import groups
 from . import pprint
 from .np_utils import fix_column_name
@@ -918,8 +919,8 @@ class BaseColumn(_ColumnGetitemShim, np.ndarray):
             arr = np.asarray(value)
             if arr.dtype.char == 'U':
                 arr = np.char.encode(arr, encoding='utf-8')
-                if isinstance(value, np.ma.MaskedArray):
-                    arr = np.ma.array(arr, mask=value.mask, copy=False)
+                if hasattr(value, 'mask'):
+                    arr = MaskedNDArray(arr, mask=value.mask, copy=False)
             value = arr
 
         return value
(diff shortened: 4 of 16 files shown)


taldcroft:
@mhvk - without actually looking at this, can you talk to the rationale for trying to apply this to MaskedColumn in a single PR? Offhand this seems like a huge PR and maybe it would be best to try to break it into smaller bits starting with only Quantity.

mhvk:
Yes, I think the reason I pushed this here was just to show that it could be done, but would involve some API changes. But I kept separate branches even locally. I am in the process of rebasing and will push the quantity-only one separately.  There's a bit of a chicken-and-egg problem in that ideally there be more documentation than the tests, but to find the energy to write that, I'd like to get a sense that this is actually going to get used...

mhvk:
@taldcroft - see #11127. Most important would be to decide on the points noted there...

mhvk:
@taldcroft - I'm wondering if going the route of an APE would be good to get our thoughts properly organized and checked. Would you be willing to think along? We could perhaps start with a wiki page or a google docs or so.

E.g., another design decision I made was to keep masking close to `ndarray`, always *below* other classes like `Quantity`. Like with using `nan` for masked, this means a lot of things go automatic, and it means the masking code does not have to know about details of `Angle`, `Time`, `SkyCoord`, etc. (but unlike for `nan`, where one would need `np.nanmean`, etc., the performance cost for masking is always born just by the masked classes).

taldcroft:
@mhvk - looking again at this and #11127, I think there is a lot of value in this PR that shows both the base implementation and how it gets used in practice. That (with the tests) is the effective first version of documentation for developers, and it proves that the infrastructure supports the intended use cases within astropy.

Some points / questions:

At this point it seems the only way to make a MaskedQuantity is with something like `Masked([1,2]*u.m)`. With this the `MaskedQuantity` class is generated on the fly. While I see the nice generalizability of that idiom, it seems like for general usability there need to be fixed class initializers like `MaskedQuantity` (and indeed `Masked<mixin_class>` for all the mixin classes). This will allow `isinstance` tests and initializing with a `mask` kwarg.

If I look at the MRO of `Masked([1,2]*u.m)` it shows `astropy.utils.masked.core.MaskedQuantity` but that doesn't actually exist. This is confusing.

About `MaskedColumn` and the API changes from `MaskedArray`.  I think that given the long heritage of `MaskedColumn`, we are going to need to leave `MaskedColumn` untouched for now. Even if you can get a `Masked(Column)` version passing our tests, I'm not convinced those tests are covering all the use in the wild. People are pretty creative... 

My suggestion would be to get to the point where the new infrastructure is passing our tests (already there, I guess?), and then back out changes to `MaskedColumn` before merging. After a release then introduce an opt-in version of MaskedColumn based on `Masked` which people can use via a configuration option. The main reason for users to do this would be better consistency for reduction operations.  (Others drivers?)

For me the next step here would be to modify table/operations.py to support all operations for all mixins! If you are somewhat careful about commits and keep the "applications" separate from the infrastructure (basically the new `astropy.utils.masked` subpackage, then I think it would make sense to merge the infrastructure commits without the applications. (I guess you can keep on cherry-picking infrastructure commits into #11127.) Since this is a new subpackage we should definitely try to break into manageable chunks and merge as soon as we are convinced it will be accepted (even if incomplete and there are follow-on actions like docs). I'm OK with an new/experimental package in astropy not being perfect. 

There is of course the elephant in the room of whether this functionality is so widely useful that it should be released as a separate package. :smile:  Anyway, thanks for the amazing work.

taldcroft:
Can we have this work:
```
MaskedQuantity = Masked(Quantity)
```
where this accepts all the `Quantity` initializers and adds `mask`?

taldcroft:
One thing that isn't quite working now:
```
In [21]: sc                                                                     
Out[21]: 
    [(1., 3.), (2., 4.)]>

In [22]: Masked(sc)                                                             
Out[22]: 
MaskedNDArray([<SkyCoord (ICRS): (ra, dec) in deg
    (1., 3.)>,
    (2., 4.)>], dtype=object)
```

mhvk:
@taldcroft - thanks for the broad review and thoughts! I agree with the plan to start with just the infrastructure, so that initially nothing truly depends on it and we can experiment without feeling we've locked in the API (including e.g. structured `dtype`). With that, I think the first goal would then be to get #11127 ready for merger, without including any of the extra commits here.

We probably should discuss in #11127 what is still missing - looking at it after nearly half a year myself it would be good to have a basic design description for developers (can be just in the module docstring) as well as perhaps a very brief overview in the docs (which may include a short list of differences for people used to `MaskedArray`).

Explicit `MaskedQuantity`, etc., would definitely be good but probably for a follow-up PR, so that review can focus on how that actually works in practice (e.g., `info` might want to gain a mask attribute to generalize between options). An explicit goal would then be for all the tests that now work with unit-full `MaskedColumn` to work with the new `MaskedQuantity`....

I very much like the idea of `Masked(Quantity)` creating/returning the `MaskedQuantity` class, and have that have an initializer which just adds a `mask` keyword argument to the standard one! One might also consider creating some kind of `@maskable` class decorator which would check for a `mask` argument and, if present, return the `Masked` version while just passing through to the regular class if not. (This of course does not preclude the explicit version.) Note that it is already possible to get a `MaskedQuantity` by passing in masked array input, as in `u.Quantity(Masked(data, mask), u.m)` or `ma * u.m`.

I also agree with the suggestion to drop the idea of making a direct replacement for `MaskedColumn`, and instead aim for an configuration option to switch it out for a new `Masked(Column)` which behaves just like the other `Masked` instances.

On more complicated mixins: `Masked` as written only works on `ndarray`, since it inserts itself just above the `ndarray`. In a way, it is very much like your `nan-for-masked` version, which I realized had the enormous benefit that `Masked` only had to know how to deal with arrays, and that this was also all that was needed, since the behaviour in `Quantity` and other subclasses basically does not depend on whether items are masked. But it does mean that higher-order classes like `SkyCoord` will need some general machinery that collects masks from the underlying arrays, similar to what you implemented for `Time`. I think it may well be possible to generalize that, just like reshaping is generalized via `ShapedLikeNDArray`, e.g., by inheriting from a `MaskableNDArrayContainer`. Indeed, it may well be possible to make `Masked(SkyCoord)` work by recognizing not just subclasses of `ndarray` but also of `ShapedLikeNDArray`!

One nice thing is that since the `erfa` functions are all `ufuncs`, mask propagation for those is already working!

Madeintheshade:
I believe you're going about it all the wrong way.

From what I've read seems like the origin of the Node 1,a. may have been corrupted in a sense that part of the original Node's1,a-(b) has removed or missing data.
Possibly the concept of a double Node 1,a-(b) was born. A twin so to speak was originally created. Then somehow was separated before the sequencing of rejoining was completed.
Node 1,a. Wouldn't accept any other replica or imposter Node 1,b...as a additional parent.
The original node 1,a.m allows the rest of the tree branches to form because of there being associated or partly related to Node 1,a. Somehow.
Possibly Node 1,a. Needs the original  Node 1,b. That may have been indirectlyor even directly separated somehow Before the symbiosis synchronization of the 2 rare Node's re-joining as one.

(Hence) The missing data. and
Difference in time referencing, where 
Node 1,a is waiting for it's original associated missing part b, other in order to re-join as one.

To possibly correct what may have happened originally in the beginning before being separated indirectly, or directly.
 
There may have been 2 other early parts of missing associated sub data, that could have been indirectly separated or removed before the re-joining / sequencing happened.

mhvk:
Going to close this one, since now that the base implementation is merged, it is better to just slowly add pieces.

PR #10425, opened 2020-06-02
Simplify repetition of many pytest.xfail by making it a decorator.

mhvk:
Looking at #10337, I see that this commit just hinders review, so let me pull it out. It only makes the test look a little less clunky.


mhvk:
Thanks, @pllim! I think I'll just merge, since if @adrn has time, I'd rather he look at some of the other PRs! And really this is a simple PR, mostly one where I wished I had known how simple things could be earlier...

PR #10428, opened 2020-06-02
Relax precision test for scale UTC

mhvk:
This example caused an error in tests of #10337, and reflects an oversight in #10373, of assuming that UTC is like all other timescales in subtraction, when in reality it is converted to TAI first.

I fear this may be one problem with using hypothesis, that it will give us "random" errors that are unrelated to a PR. cc @Zac-HD for advise on how best to deal with things like this...

Leaving milestone as I just wrote in #10373 that that one should probably have been 4.2; this one should be the same as that of #10373 whatever the conclusion there.

### DIFF
--- a/astropy/time/tests/test_precision.py
+++ b/astropy/time/tests/test_precision.py
@@ -496,16 +496,25 @@ def test_jd_add_subtract_round_trip(scale, jds, delta):
 @example(scale='utc',
          jds_a=(2455000.0, 0.0),
          jds_b=(2443144.5, 0.5000462962962965))
+@example(scale='utc',
+         jds_a=(2459003.0, 0.267502885949074),
+         jds_b=(2454657.001045462, 0.49895453779026877))
 def test_timedelta_full_precision(scale, jds_a, jds_b):
     jd1_a, jd2_a = jds_a
     jd1_b, jd2_b = jds_b
     assume(scale != 'utc'
            or (2440000 < jd1_a+jd2_a < 2460000
                and 2440000 < jd1_b+jd2_b < 2460000))
+    if scale == 'utc':
+        # UTC subtraction implies a scale change, so possible rounding errors.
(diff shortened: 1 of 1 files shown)


Zac-HD:
> I fear this may be one problem with using hypothesis, that it will give us "random" errors that are unrelated to a PR. cc Zac for advice on how best to deal with things like this...

It turns out that this hinges on a fundamental question about your workflow: is the point of testing to *find bugs*, or to *prevent regressions*?  https://blog.nelhage.com/post/two-kinds-of-testing/ describes this really well.

My advice for the first case is basically "run property-based tests for longer to start with" - if they don't fail while looping overnight, they're much less likely to do so in CI.  This is how we do it for Hypothesis itself, and it *does* converge to no surprises remarkably quickly.

Alternatively (or as well!) you could set the `derandomize=True` setting for your CI and local profiles, and reserve `derandomize=False` for the fuzzing profile which can then be used to search for new bugs *outside* your development cycle.

https://github.com/astropy/astropy/blob/a45af44fdd406cd7b53f3e7a48ff4d22dca3b67e/conftest.py#L21-L25

I hope that's helpful!

mhvk:
Yes, very helpful! My sense would be to try to have a bit of the best of both worlds, to have no randomization for PRs, where the primary goal is to guard against regressions, and turn randomization on for our cron jobs, as a way to keep on poking. We'll also need good documentation for ourselves...

mhvk:
Raised a new issue for how to set up hypothesis further #10462.

This PR is ready to go in, I think. @aarchiba, would you be able to review?

mhvk:
For now setting milestone to 4.0.2, i.e., the same as #10373; if the latter's milestone is changed, this one should be too.

mhvk:
@taldcroft - would you be able to do a quick review? It is just adjusting a test slightly.

PR #10431, opened 2020-06-02
Ensure we catch all exceptions when trying to set shape.

### LINKED ISSUE #10426: Possible bug in setting shape of Time, Representation, etc.
Currently, setting the shape of a `Time` or `Representation` ~`SkyCoord`~ (edit: frames and sky coordinates cannot have their shape set) implies going through all attributes and trying to set their shape in turn, reverting if there is any `AttributeError` for items for which the shapes stride imply that the shape change is not possible. However, one can also get a `ValueError`, which should thus be checked as well. (Note that it is likely difficult to set up a state where this would actually cause a problem, so this does not have high priority)
```
In [5]: np.ones((3, 2)).T.shape = 6,                                            
<ipython-input-5-384542aa1cc8> in <module>
----> 1 np.ones((3, 2)).T.shape = 6,

AttributeError: incompatible shape for a non-contiguous array

In [6]: np.ones((3, 2)).T.shape = 1, 2                                          
<ipython-input-6-570b08d46880> in <module>
----> 1 np.ones((3, 2)).T.shape = 1, 2

ValueError: cannot reshape array of size 6 into shape (1,2)
```

mhvk:
The check was only for AttributeError, but as noted in https://github.com/astropy/astropy/pull/10337#discussion_r433752781, a ValueError is possible too. Since we will want to undo any half-processed shape change regardless of the exception, we just catch them all (they will be
re-raised anyway). 

Note that I put no-changelog-entry-required, since I could not think of any way it would be possible to trigger the `ValueError` and end up with a corrupted `Time` or `Representation` (i.e., one in which some parts had their shape changed and others not). The fix is really more a matter of principle, which makes the code and tests clearer.

fixes #10426

### DIFF
--- a/astropy/coordinates/representation.py
+++ b/astropy/coordinates/representation.py
@@ -422,6 +422,8 @@ class BaseRepresentationOrDifferential(ShapedLikeNDArray):
 
         Raises
         ------
+        ValueError
+            If the new shape has the wrong total number of elements.
         AttributeError
             If the shape of any of the components cannot be changed without the
             arrays being copied.  For these cases, use the ``reshape`` method
@@ -441,7 +443,7 @@ class BaseRepresentationOrDifferential(ShapedLikeNDArray):
             if val.size > 1:
                 try:
                     val.shape = shape
-                except AttributeError:
+                except Exception:
                     for val2 in reshaped:
                         val2.shape = oldshape
                     raise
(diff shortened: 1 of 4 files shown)


mhvk:
@adrn - since you just looked at a related PR, could I bother you with this one? It is really simple and ready to go in...

adrn:
@mhvk Doh, looks like you'll have to rebase...but LGTM!

mhvk:
Thanks! Rebased and will merge if tests pass.

PR #10432, opened 2020-06-02
Check values equal to zero in QuantityAttribute a bit more carefully.

mhvk:
In the process resolving a bug that we tried to create an array with shape=None, which is being deprecated, which is causing failures on the numpy-dev runs.


mhvk:
numpy-dev job passed.

PR #10436, opened 2020-06-03
Ignore pytest warnings due to asdf plugin

eslavich:
### Description

This pull request adds an ignore filter for pytest deprecation warnings caused by the asdf plugin.  We've fixed the warnings in asdf master but haven't released yet, and in the meantime this will prevent errors during test collection.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -114,6 +114,7 @@ filterwarnings =
     ignore:PY_SSIZE_T_CLEAN will be required for '#' formats
     ignore:::astropy.tests.plugins.display
     ignore:::astropy.tests.disable_internet
+    ignore:direct construction of Asdf:pytest.PytestDeprecationWarning
 doctest_norecursedirs =
     */setup_package.py
 doctest_subpackage_requires =


pllim:
But our CI is passing, so why is this ignore necessary? Can you please clarify? Thanks.

astrofrog:
Some users have reported seeing this. I haven't figured out why we don't see it in CI though...

eslavich:
I haven't figured out why yet, but it looks like the asdf schema tests aren't running on Travis right now.  For example, we should see this test when asdf is installed:
```
astropy/io/misc/asdf/data/schemas/astropy.org/astropy/coordinates/frames/fk4-1.0.0.yaml
```
but the Travis output doesn't contain the string `.yaml`.

eslavich:
I may be confused, but it looks like the tox environment invokes pytest with `--pyargs astropy` from an empty temp directory.  From there, setup.cfg isn't available, so aren't we losing all of the configuration under `[tool:pytest]`?

saimn:
The builds say `rootdir: /home/travis/build/astropy/astropy, inifile: setup.cfg`, so it seems that pytest finds the config file. (`cov-config` is for `pytest-cov`, but we could also specify the config file explicitely with `-c {toxinidir}/setup.cfg` to be sure).
But then `asdf_schema_root` is a relative path, so maybe the path is broken when running from another directory?

eslavich:
> The builds say `rootdir: /home/travis/build/astropy/astropy, inifile: setup.cfg`

So it does!  I wonder if all I need to do is add the astropy/io/misc/asdf/data directory similar to how the docs were added.  I'll give that a try.

pllim:
Re: https://github.com/astropy/astropy/pull/10436#issuecomment-639167101

Does that mean this is really a path problem and there is no need to ignore warning?

eslavich:
Unfortunately, looks like that isn't going to work -- pytest ends up loading astropy/conftest from both the installed package and the source tree and raises ImportMismatchError.

How about a second pytest command just for the schema tests?

eslavich:
> Does that mean this is really a path problem and there is no need to ignore warning?

I think we'll still need the ignore.  The path problem is just preventing tox from running all of the tests, which is why the warning doesn't show up and cause the Travis build to fail.

pllim:
> I think we'll still need the ignore. The path problem is just preventing tox from running all of the tests, which is why the warning doesn't show up and cause the Travis build to fail.

Ah, yes. Sorry, my brain was all mixed up jumping between issues.

However, I think this needs to be done in the following order:

1. Fix the path stuff and make sure the tests are called without ignoring anything.
2. If the warnings appear in CI, then only we ignore it.

What do you think? Thanks for your patience!

> How about a second pytest command just for the schema tests?

If it works, why not?

eslavich:
Tox isn't behaving the same on Travis as on my laptop, so I'm going to have to do some troubleshooting.  I'll comment again once things have stabilized.

eslavich:
Okay, this is sorted out.  The [initial commit](https://github.com/astropy/astropy/pull/10436/commits/a0ed2ec4b966c9d36fcfd1950726c7144a97d79c) adds the ASDF schema tests to the Travis build, but the job fails due to the deprecation warning.  The [next commit](https://github.com/astropy/astropy/pull/10436/commits/66969c5805df13deb43ccc8add04ff83c4e29ec9) ignores the warning which allows the new job to succeed.

pllim:
Weird that Travis not reporting back.

Job with warning: https://travis-ci.org/github/astropy/astropy/jobs/695151063
Job without warning: https://travis-ci.org/github/astropy/astropy/jobs/695160765

pllim:
@eslavich , I don't see any actual tests in the YAML files, so what does `pytest astropy/io/misc/asdf/data` really do here? 

Is there really no way to have a standard job collect them via [pytest discovery](https://docs.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery) so that we don't need a special job to just do this?

How often you think these tests will break due to `asdf` changes? I worry that it will fail for unrelated PRs.

eslavich:
> @eslavich , I don't see any actual tests in the YAML files, so what does `pytest astropy/io/misc/asdf/data` really do here?

The ASDF plugin generates test cases on-the-fly for relevant YAML files that are collected by pytest.  Tests include validating the each schema against the metaschema and validating examples listed in the schema against the schema itself.

> Is there really no way to have a standard job collect them via [pytest discovery](https://docs.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery) so that we don't need a special job to just do this?

It may be possible to make the plugin smarter, but not every YAML file is an ASDF schema, so the plugin needs to know which directories contain YAML files that should be tested as schemas.  Currently those directories are specified relative to the pytest root directory which means that paths within the installed astropy package are filtered out.

Since astropy packages its test code, it's not possible to run tests on the source tree unless astropy is installed in editable/develop mode.  That is what forced me to create a new tox job, since the existing job does not use an editable install.

> How often you think these tests will break due to `asdf` changes? I worry that it will fail for unrelated PRs.

Our release procedures for asdf include running the astropy schema tests, so if we follow those we'll notice and fix any test failures before making a new asdf version available.  If astropy wants to be as defensive as possible, it could pin the version of asdf used in this new tox job.

pllim:
> pin the version of asdf used in this new tox job

Hmm. I like this idea. If you are ready to bump it, you can open a PR here.

pllim:
One more thing... Does it really need to be in the initial stage? Can it be moved to a second stage or even cron job?

eslavich:
> One more thing... Does it really need to be in the initial stage? Can it be moved to a second stage or even cron job?

Ideally this would run on PRs to prevent incorrect schemas from being merged, but I don't have an opinion on the stage.

.travis.yml
```diff
@@ -134,6 +134,13 @@ matrix:
           stage: Initial tests
           env: TOXENV='codestyle'
 
+        # ASDF schema tests
+        - language: python
+          python: 3.7
+          name: ASDF schema tests
+          stage: Initial tests
```
pllim, line 141: Since you don't have a strong opinion, why don't we not run it unless initial stage passes?
```suggestion
          stage: Comprehensive tests
```

tox.ini
```diff
@@ -89,6 +89,17 @@ commands =
     cov-!double: pytest --pyargs astropy {toxinidir}/docs {env:MPLFLAGS} --cov astropy --cov-config={toxinidir}/setup.cfg {posargs}
     double: python -c 'import sys; from astropy import test; test(); sys.exit(test())'
 
+[testenv:schema_tests]
+changedir = {toxinidir}
+description = "run ASDF schema tests"
+extras =
+    test
+    all
```
pllim, line 97: An `asdf` pin would be nice.

.travis.yml
eslavich: This change is included in the latest commit

tox.ini
eslavich: Pinned at 2.6.0

saimn:
FWIW, having another look at the asdf plugin, it is able to find the yaml files when running from another directory with tox:
https://github.com/spacetelescope/asdf/blob/ad141da17bbb46d3ed73e00b41f401d49fdef391/pytest_asdf/plugin.py#L236-L237
But then it uses absolute paths to compare the collected file to the schema path.
https://github.com/spacetelescope/asdf/blob/ad141da17bbb46d3ed73e00b41f401d49fdef391
And because the plugin uses absolute paths, it can test only the yaml files in the source directory:
```
(Pdb) path
local('/home/sconseil/dev/astropy/.tox/test-alldeps/lib/python3.7/site-packages/astropy/io/misc/asdf/data/schemas/astropy.org/astropy/coordinates/angle-1.0.0.yaml')
(Pdb) schema_roots
['/home/sconseil/dev/astropy/astropy/io/misc/asdf/data/schemas']
```
One could argue that `asdf_schema_root` should be relative to the path currently tested.

pllim:
Given @saimn 's revelation of another possible way to fix this somewhere else, I am not going to merge this for now until we are sure that this is the way to go.

eslavich:
I backed out the changes that fix the schema tests, leaving only the new ignore filter.

We're moving the schemas to https://github.com/astropy/asdf-astropy so they'll start getting tested there soon enough.

Issue #10397, opened 2020-05-26
Issue with MJDREF and MJD-OBS in WCS


astrofrog:
The following example demonstrates some peculiar behavior with respect to dates in WCS headers:

```python
In [8]: from astropy.io import fits                                                                                                                                              
   ...:  
   ...:  
   ...: HEADER = fits.Header.fromstring("""  
   ...: WCSAXES =                    2 / Number of coordinate axes  
   ...: DATE-OBS= '2011-02-15T00:00:00.34' / ISO-8601 time of observation  
   ...: MJD-OBS =      55607.000003935 / [d] MJD at start of observation  
   ...: """, sep="\n") 
   ...:                                                                                                                                                                          

In [9]: WCS(HEADER).to_header()                                                                                                                                                  
Out[9]: 
WCSAXES =                    2 / Number of coordinate axes                      
CRPIX1  =                  0.0 / Pixel coordinate of reference point            
CRPIX2  =                  0.0 / Pixel coordinate of reference point            
CDELT1  =                  1.0 / Coordinate increment at reference point        
CDELT2  =                  1.0 / Coordinate increment at reference point        
CRVAL1  =                  0.0 / Coordinate value at reference point            
CRVAL2  =                  0.0 / Coordinate value at reference point            
LATPOLE =                 90.0 / [deg] Native latitude of celestial pole        
MJDREFI =                  0.0 / [d] MJD of fiducial time, integer part         
MJDREFF =                  0.0 / [d] MJD of fiducial time, fractional part      
DATE-OBS= '2011-02-15T00:00:00.34' / ISO-8601 time of observation               
MJD-OBS =      55607.000003935 / [d] MJD of observation                         
MJD-OBS =      55607.000003935 / [d] MJD at start of observation  
```

The resulting header has MJDREFI and MJDREFF set both to 0 **and** a duplicate ``MJD-OBS`` entry. Round tripping via ``WCS`` again results in a warning and yet another keyword ``DATEREF``:

```python
In [11]: WCS(WCS(HEADER).to_header()).to_header()                                                                                                                                
WARNING: FITSFixedWarning: 'datfix' made the change 'Set DATE-REF to '1858-11-17' from MJD-REF'. [astropy.wcs.wcs]
Out[11]: 
WCSAXES =                    2 / Number of coordinate axes                      
CRPIX1  =                  0.0 / Pixel coordinate of reference point            
CRPIX2  =                  0.0 / Pixel coordinate of reference point            
[...]

PR #10433, opened 2020-06-03
Update bundled WCSLIB to v7.3

mcara:
This PR updates bundled WCSLIB to version 7.3. This update partially fixes https://github.com/astropy/astropy/issues/10397 but the

> WARNING: FITSFixedWarning: 'datfix' made the change ...

is still there on the second pass. With a small workaround (a separate PR) this warning could be quieted.

CC: @nden @astrofrog

PR #10440, opened 2020-06-03
Do not issue datfix warning when MJDREF has default value

mcara:
This PR removes the `datfix` warning when setting `DATREF` to the default value from a default value of `MJDREF`. This PR, in part, addresses  https://github.com/astropy/astropy/issues/10397

CC: @nden


nden:
This PR will need to be included in the 4.1 release. It partially fixes an issue with `4.1rc1`.

pllim:
Note: Goes with #10433

PR #10442, opened 2020-06-04
Remove round-trip checking code in WCSAxes

### LINKED ISSUE #10441: x-axis missing on WCSAxes whole sphere plot
### Description
When plotting a full sphere plot (in this case it happens to be a map of the Sun), WCSAxes does not draw the x-axis when the whole map is shown. Example code is given below.

The issue is that the x-axis lies at -180deg, which is technically the same longitude as 180deg. When WCSAxes tries to work out if a given pixel location is within the valid bounds of the transformation
https://github.com/astropy/astropy/blob/64bc74b1cef62ddf9fe764282a18ade3e879028a/astropy/visualization/wcsaxes/wcsapi.py#L356-L358
, it round trips the transformation, ie. for the bottom left corner of the bottom left pixel:
```python
pixel = (-0.5, -0.5)
world = wcs.pixel_to_world_values(*pixel)
pixel2 = wcs.world_to_pixel_values(*world)
```
`pixel2` comes out as `(array(359.5), array(-0.5))` however, which is different to the input pixel, so WCSAxes treats this as invalid, and does not draw x-axis because it thinks the spine has invalid coordinates. I think it is actually on the edge of valid pixel locations, so should be drawn.

Downstream issue at https://github.com/sunpy/sunpy/issues/3611

### Expected behavior
x-axis is drawn.

### Actual behavior
x-axis isn't drawn:
Figure_1

### Steps to Reproduce

```python
import numpy as np
from astropy.wcs import WCS

shape = [180, 360]
data = np.random.rand(*shape)
header = {'wcsaxes': 2,
          'crpix1': 180.5,
          'crpix2': 90.5,
          'cdelt1': 1.0,
          'cdelt2': 1.0,
          'cunit1': 'deg',
          'cunit2': 'deg',
          'ctype1': 'CRLN-CAR',
          'ctype2': 'CRLT-CAR',
          'crval1': 0.0,
          'crval2': 0.0,
          'lonpole': 0.0,
          'latpole': 90.0,
          'dateref': '1858-11-17',
          'mjdrefi': 0.0,
          'mjdreff': 0.0,
          'date-obs': '2000-01-01T00:00:00.000'}

wcs = WCS(header)

pixel = (-0.5, -0.5)
world = wcs.pixel_to_world_values(*pixel)
pixel2 = wcs.world_to_pixel_values(*world)
print(pixel)
print(world)
print(pixel2)

ax = plt.subplot(projection=wcs)
ax.imshow(data, origin='lower')
plt.show()

```

### System Details

```python
macOS-10.15.4-x86_64-i386-64bit
Python 3.8.2 (default, Mar 26 2020, 10:43:30) 
Numpy 1.18.1
astropy 4.2.dev162+g6054cc78b
Scipy 1.4.1
Matplotlib 3.2.1
```

dstansby:
Fixes #10441.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -318,6 +318,8 @@ astropy.utils
 
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
+- Fixed an issue where ticks were sometimes not drawn at the edges of a spherical
+  projection on a WCSAxes. [#10442]
 
 astropy.wcs
 ^^^^^^^^^^^
--- a/astropy/visualization/wcsaxes/tests/test_wcsapi.py
+++ b/astropy/visualization/wcsaxes/tests/test_wcsapi.py
@@ -416,3 +416,30 @@ class TestWCSAPI:
         ax.set_xlim(-0.5, 148.5)
         ax.set_ylim(-0.5, 148.5)
         return fig
+
+
+def test_edge_axes():
+    # Check that axes on the edge of a spherical projection are shown properley
+    # (see https://github.com/astropy/astropy/issues/10441)
+    shape = [180, 360]
+    data = np.random.rand(*shape)
+    header = {'wcsaxes': 2,
+              'crpix1': 180.5, 'crpix2': 90.5,
+              'cdelt1': 1.0, 'cdelt2': 1.0,
+              'cunit1': 'deg', 'cunit2': 'deg',
+              'ctype1': 'CRLN-CAR', 'ctype2': 'CRLT-CAR',
+              'crval1': 0.0, 'crval2': 0.0,
+              'lonpole': 0.0, 'latpole': 90.0,
+              }
+    wcs = WCS(header)
+    fig = plt.figure()
+    ax = fig.add_axes([0.1, 0.1, 0.8, 0.8], projection=wcs)
+    ax.imshow(data, origin='lower')
+    # By default the x- and y- axes should be drawn
+    lon = ax.coords[0]
+    lat = ax.coords[1]
+    fig.canvas.draw()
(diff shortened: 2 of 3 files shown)


astrofrog:
This might break some solar examples, since the round tripping was added in part to avoid plotting grid lines behind the sun! 🌞

dstansby:
😬 okay, I'll run the sunpy tests through with this and see what happens!

astrofrog:
We should definitely have a test here though that exercises the round-tripping properly.

Also just out of curiosity, what issues did you have with running the tests locally?

dstansby:
For the tests I get `E   pytest.PytestDeprecationWarning: direct construction of DocTestModulePlus has been deprecated, please use DocTestModulePlus.from_parent`, I think everything is installed correctly, but it might just be a mangled conda environment.

dstansby:
The sunpy figure tests seem to work fine with this branch: https://github.com/sunpy/sunpy/pull/4247

ayshih:
> This might break some solar examples, since the round tripping was added in part to avoid plotting grid lines behind the sun! 🌞

The round-trip check here doesn't seem relevant for this purpose, because it's pixel->world->pixel, which wouldn't ever resolve to be the far side of the Sun.

I believe the world->pixel->world round-trip check for grid lines starts here:
https://github.com/astropy/astropy/blob/e20ebfdf687015e83a8d1a9ca79fd22ac942fe74/astropy/visualization/wcsaxes/coordinate_helpers.py#L911-L915

dstansby:
It looks like this is the way to go then - I've added a test to make sure the ticks are there as expected.

PR #10449, opened 2020-06-04
MNT: Only import in setup when needed

### LINKED ISSUE #10429: Change order of operations in setup.py
In the ``setup.py`` file we should make sure that the checks for the ``test`` and ``build_docs`` commands come *before* the extension-helpers imports, otherwise contributors who don't have extension-helpers installed will see an import error when trying to run ``python setup.py test`` rather than the help messages. This will also need to be fixed in the coordinated packages and the package template.

(cc @aarchiba since this affected you today!)

pllim:
### Description

This pull request is to address the problem of import error occurring before it should happen during installation process.

Fixes #10429

cc @aarchiba

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,8 +95,8 @@ del split_version  # clean up namespace.
 release = 'dev' not in version
 """.lstrip()
 
-# Only import these if the above checks are okay.
-# See https://github.com/astropy/astropy/issues/10429
+# Only import these if the above checks are okay
+# to avoid masking the real problem with import error.
 import os  # noqa
 from setuptools import setup  # noqa
 from extension_helpers import get_extensions  # noqa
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -99,6 +95,12 @@ def split_version(version):
 release = 'dev' not in version
 """.lstrip()
 
+# Only import these if the above checks are okay.
+# See https://github.com/astropy/astropy/issues/10429
```
astrofrog, line 99: Since this is not a temporary workaround, maybe just explain why it has to be here and don't include the URL?

pllim: Done. Thanks for the review!

saimn:
This should go to 4.1 as well, isn't it ?

pllim:
Is it that critical? 🤷

saimn:
It's the first release with the new helpers, so I guess some users will try to run the `setup.py` and get the wrong message (the `extension-helpers` exception).

pllim:
Okay, I remilestoned to 4.1 but @eteq should make the final call. Thanks!

PR #10452, opened 2020-06-05
Fix travis cron tests with apt

### LINKED ISSUE #10450: TST: Cron jobs for s390x and arm64 failed with hypothesis error
Example log: https://travis-ci.org/github/astropy/astropy/builds/694700963

```
____________ ERROR collecting astropy/time/tests/test_precision.py _____________
astropy/time/tests/test_precision.py:9: in <module>
    from hypothesis import assume, example, given, target
E   ImportError: cannot import name 'target' from 'hypothesis' (/usr/lib/python3/dist-packages/hypothesis/__init__.py)
____________ ERROR collecting astropy/time/tests/test_precision.py _____________
ImportError while importing test module '/home/travis/build/astropy/astropy/astropy/time/tests/test_precision.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
astropy/time/tests/test_precision.py:9: in <module>
    from hypothesis import assume, example, given, target
E   ImportError: cannot import name 'target' from 'hypothesis' (/usr/lib/python3/dist-packages/hypothesis/__init__.py)
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!
```

@mhvk or @Zac-HD , any ideas? Thanks.

mhvk:
fixes #10450

Note: will only run s390x and arm tests. If those pass, will revert the changes that made them run outside of cron.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ env:
 
         # For installing without tox, we can rely mostly on what Debian already
         # includes.  This is only used if SETUP_METHOD='apt'.
-        - APT_DEPENDENCIES="python3-pip python3-dev python3-venv python3-setuptools cython3 ipython3 python3-jinja2 python3-numpy python3-pytest-astropy python3-pytest-cov python3-pytest-xdist python3-pytest-filter-subpackage python3-objgraph python3-coverage tzdata"
+        - APT_DEPENDENCIES="python3-pip python3-dev python3-venv python3-setuptools cython3 ipython3 python3-jinja2 python3-numpy python3-pytest-astropy python3-pytest-cov python3-pytest-xdist python3-pytest-filter-subpackage python3-objgraph python3-coverage python3-attr tzdata"
 
         # The following is needed to avoid issues if e.g. Matplotlib tries
         # to open a GUI window.
--- a/astropy/table/pandas.py
+++ b/astropy/table/pandas.py
@@ -29,5 +29,5 @@ try:
     display.display(dhtml)
 except ImportError:
     print(ascii_uncoded)
-except UnicodeEncodeError:
+except (UnicodeEncodeError, SyntaxError):
     pass


.travis.yml
```diff
@@ -185,15 +185,22 @@ matrix:
           arch: s390x
           language: c
           dist: bionic
-          stage: Cron tests
+          stage: Final tests
```
pllim, line 188: For debugging purpose, why not move them to initial stage so they run first?

pllim:
I'll add an experimental label for now given that this PR has commits that should not be merged. Feel free to remove the label when you think it is ready. Thanks!

.travis.yml
mhvk: Duh, yes, of course!

mhvk:
I fear I don't understand what is happening exactly here, and I won't have more time for a while. Since it may well resolve itself as things get updated, I'm just going to let it rest.

mhvk:
A fresh day often helps! Especially if my own package [suffers the same problem](https://github.com/mhvk/baseband/pull/405) even though it is not even using `hypothesis`...  The problem, I think, ends up being that `bionic` includes some packages that are too old, and the update to Debian testing doesn't update them.

Specifically, Debian testing's `python3-hypothesis` depends on `python3-attr`, but does not set a minimum version even though upstream does depend on `attrs>=19.2.0`, and the ubuntu bionic has too old a version. @olebole - where would I report this? Is it easy to fix??

As another annoyance, `astropy/table/pandas.py` doesn't do its exception checking right, which with a sufficiently old version also causes problems - and all that for showing us a panda. 

Must admit I don't know how @eteq got https://github.com/astropy/astropy/commit/ffa3ffe8482083c9f99e05194ae7ebe7a23ab081 in master without review - I may be humourless but this easter egg really puzzled me and cost me time (given `git blame`, not only me), a bit like the review-less introduction of `TimeStardate`, which, as it was not reviewed, did not do things right and thus needed fixing in #9563.

Anyway, will now remove the commit that put `s390` and `arm` on `Initial tests` and then this will in fact fix #10450

mhvk:
coverage test failure is obviously nonsensical since no code was touched...

mhvk:
@pllim - In a less grumpy mood, today I do appreciate the easter egg!

olebole:
(better answering late then never ;-) )
> Specifically, Debian testing's python3-hypothesis depends on python3-attr, but does not set a minimum version even though upstream does depend on attrs>=19.2.0, and the ubuntu bionic has too old a version. @olebole - where would I report this? Is it easy to fix??

For Debian, you can report bugs and wishes to our Bugtracker, see https://www.debian.org/Bugs/Reporting . However, usually the package of a released Debian or Ubuntu release is not automatically kept up-to-date; it stays at the version when it was released. For Debian, there is a "backports" repository which can be enabled, but this needs someone who actually backports the package (incl. its dependencies). For Python packages, it is probably simplest to use pip when one needs newer versions.

mhvk:
@olebole - thanks!  I'm happy to report it, but before I do, what is the standard policy for python packages in Debian? Would you normally copy over version limits from `setup.cfg` to the `Depends` and `Recommends` sections in the control file, or is the practice to just make sure things work within a given distribution (say, `testing`)?

olebole:
This is the responsibility the maintainer. However, there is only one version of every package in the current testing Debian release, so its major use is backporting to the last stable release. So the number is omitted if the version in the stable release (currently "Buster") is sufficient (ir if the developer is just lazy).

However, when a package does not work with the distributed dependency, this is for sure a bug and should be reported.

mhvk:
OK, that makes sense. Since in this case the package didn't work with a Ubuntu distribution, I don't think it is a Debian problem (but I'll check whether Debian stable would have the right `python3-attr` just in case). Thanks!!

PR #10460, opened 2020-06-06
Fix Table.read/write.help when reader/writer has no docstring

saimn:
The `jsviewer` writer has no docstring, which breaks `Table.write.help`:
```
In [1]: %astropy                                                                                         
Numpy 1.19.0rc2
Astropy 4.2.dev222+gd20175a31.d20200605

In [2]: t = Table([[1,2],[3,4]])                                                                         

In [3]: t.write.help('jsviewer')                                                                         
<ipython-input-3-e3c3947780db> in <module>
----> 1 t.write.help('jsviewer')

~/dev/astropy/astropy/io/registry.py in help(self, format, out)
    675             reader_doc += re.sub('.', '=', header)
    676             reader_doc += os.linesep
--> 677             reader_doc += inspect.cleandoc(doc)
    678 
    679         if out is None:

~/.pyenv/versions/miniconda3-latest/envs/dragons38/lib/python3.8/inspect.py in cleandoc(doc)
    629     onwards is removed."""
    630     try:
--> 631         lines = doc.expandtabs().split('\n')
    632     except UnicodeError:
    633         return None

AttributeError: 'NoneType' object has no attribute 'expandtabs'
```

### DIFF
--- a/astropy/io/tests/test_registry_help.py
+++ b/astropy/io/tests/test_registry_help.py
@@ -138,3 +138,18 @@ def test_ccddata_read_help_fits():
     assert "CCDData.read(format='fits') documentation" in doc
     assert "Generate a CCDData object from a FITS file" in doc
     assert "hdu_uncertainty : str or None, optional" in doc
+
+
+def test_table_write_help_jsviewer():
+    """
+    Test dynamically created documentation help via the I/O registry for
+    'jsviewer'.
+    """
+    out = StringIO()
+    Table.write.help('jsviewer', out)
+    doc = out.getvalue()
+
+    # Check a smattering of expected content
+    assert "Table.write general documentation" not in doc
+    assert "The available built-in formats" not in doc
+    assert "Table.write(format='jsviewer') documentation" in doc
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -897,6 +897,8 @@ astropy.io.misc
 astropy.io.registry
 ^^^^^^^^^^^^^^^^^^^
 
+- Fix ``Table.(read|write).help`` when reader or writer has no docstring. [#10460]
+
 astropy.io.votable
 ^^^^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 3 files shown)


astropy/io/tests/test_registry_help.py
```diff
@@ -138,3 +147,48 @@ def test_ccddata_read_help_fits():
...
+
+    # Check a smattering of expected content
+    assert "Table.write general documentation" not in doc
+    assert "The available built-in formats" not in doc
+    assert "Table.write(format='jsviewer') documentation" in doc
+
+
+def test_table_read_no_help():
+    try:
+        io_registry.register_reader('test1', Table, empty_reader)
```
saimn, line 169: No sure what's the best way to register temporarily a reader/writer. I saw in other tests that another class was used (`TestData`) instead of `Table`, but `TestData.read.help` does not work...

taldcroft: I'd be OK on this with just the one test that demonstrates fixing the problem with jsviewer highlighted in the description.  I'd have to review the code and prior examples, but honestly this seems like too much effort for this small fix which is clearly correct.

saimn: Yes agreed :smile:, but if I or someone else update the jsviewer docstring in the future it may lead to an inconsistent test/docstring, so I didn't want to rely on the jsviewer test for this. As you prefer, I'm happy to remove the two extra tests if using the registry like this is overkill.

taldcroft: You are right that the test could in the future be a problem, but it is not highly likely. I have come to accept the reality of limited resources and that "perfect is the enemy of good". So I recommend to just get this in with an OK test and move on to the next one out of hundreds of issues to deal with. If it breaks later we'll deal.

saimn: Ok, I removed the tests and added the changelog entry. And squashed and rebased to get the fix for the docs build.

PR #10464, opened 2020-06-06
Split setup.cfg test and all into types; use in tox.ini

mhvk:
As a start to address #10342. Two commits just to have something concrete to discuss.
1. Split the `test` dependencies into those that are needed to *do* the testing, and a new `test_all` that includes packages used for comparison purposes.
2. Split off a new `rec` set of recommended packages from `all`; the objective criterion is that these are all listed as "key" dependencies in `devdeps` - though they also happen to match more or less what I would pick.

p.s. Cancelled all tests, though checked locally that things work. Can adapt `.travis.yml` if we find there is a use for it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,15 +50,19 @@ asdf_extensions =
     astropy-asdf = astropy.io.misc.asdf.extension:AstropyAsdfExtension
 
 [options.extras_require]
-test =
+test =  # Required to run the astropy test suite.
     pytest-astropy>=0.8
     pytest-xdist
+    packaging
+test_all =  # Required for testing, plus packages used by particular tests.
+    pytest-astropy>=0.8
+    pytest-xdist
+    packaging
     objgraph
     ipython>=4.2
     coverage
     skyfield>=1.20
     sgp4>=2.3
-    packaging
 all =
     scipy>=1.1
     certifi
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,20 +63,24 @@ test_all =  # Required for testing, plus packages used by particular tests.
     coverage
     skyfield>=1.20
     sgp4>=2.3
+recommended =
+    scipy>=1.1
+    matplotlib>=3.0,!=3.4.0
+    PyYAML>=3.13
 all =
     scipy>=1.1
+    matplotlib>=3.0,!=3.4.0
+    PyYAML>=3.13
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -48,26 +48,28 @@ asdf_extensions =
...
     pytest-xdist
+    coverage
+test_all =
     objgraph
     ipython
-    coverage
     skyfield>=1.20
     sgp4>=2.3
-all =
+rec =
```
pllim, line 57: "rec" is too cryptic for me.
```suggestion
recommended =
```

mhvk: Yes, quite happy for "recommended" - still OK to use `rec` in `tox.ini`?

pllim: Yeah, I don't think typical user would care about what we name things in `tox.ini`.

pllim:

embray:
This looks straightforward to me.  Any help needed finishing this up?  I agree with @pllim it should be mentioned in the installation docs.

embray:
This looks straightforward to me.  Any help needed finishing this up?  I agree with @pllim it should be mentioned in the installation docs.

mhvk:
@embray - thanks, I was really just waiting to get confirmation this was a good idea beyond @pllim's tentative OK!

mhvk:
@embray - thanks, I was really just waiting to get confirmation this was a good idea beyond @pllim's tentative OK!

mhvk:
OK, I tried to find places where it would be good to mention the changes. Please let me know of any others!

mhvk:
Failure is unrelated - see #11857 (fixed by #11858)

embray:
@pllim Any final thoughts on this one, or should we merge?

pllim:
Pretty sure the failure was fixed in #11858 but since this PR touches on dependencies, I would like to request a rebase, just to make sure. Thanks!

pllim:
I guess repetitions are unavoidable and seems like the benefits outweigh the costs here.

We can worry about testing the new "recommended" deps in CI if the need arises in the future. I think we already have enough jobs as it is.

I'll approve after rebase is done (please ping me). 😸

mhvk:
@pllim - no problem, rebased!

pllim (approved):
Thanks!

PR #10469, opened 2020-06-08
EXP: Run TestManipulation without remote data

pllim:
### Description

This pull request is to address problem reported in https://github.com/astropy/astropy/pull/10311#issuecomment-640073776 with no able to run a test without accessing internet.

This is a follow-up of #10337.

### DIFF
--- a/astropy/coordinates/tests/test_shape_manipulation.py
+++ b/astropy/coordinates/tests/test_shape_manipulation.py
@@ -6,6 +6,7 @@ import pytest
 
 from astropy import units as u
 from astropy.units.quantity_helper.function_helpers import ARRAY_FUNCTION_ENABLED
+from astropy.utils.iers import conf as iers_conf
 from astropy.coordinates import Longitude, Latitude, EarthLocation, SkyCoord
 
 # test on frame with most complicated frame attributes.
@@ -24,7 +25,6 @@ needs_array_function = pytest.mark.xfail(
     reason="Needs __array_function__ support")
 
 
-@pytest.mark.remote_data
 class TestManipulation():
     """Manipulation of Frame shapes.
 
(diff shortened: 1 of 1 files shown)


pllim:
Something funky is going on because I can't reproduce this locally. I did stumble across #10471 though.

astropy/coordinates/tests/test_shape_manipulation.py
```diff
@@ -34,6 +34,10 @@ class TestManipulation():
     """
 
     def setup_class(cls):
+        # This is to be consistent with setting in astropy/conftest.py
+        # that somehow is not getting picked up.
+        iers_conf.auto_download = False
```
pllim, line 39: No sure what happened: Either it is not picking up setting from `astropy/conftest.py` or something else is overwriting the setting. But forcing it (again) here seems to work.

pllim:
I am expecting the Windows job to fail from #10471 but the `html-docs` failure is surprising.
```
Warning, treated as error:
 
autodoc: failed to import method '__call__' from module 'keyword'; the following exception was raised:
 
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sphinx/util/inspect.py", line 323, in safe_getattr
 
    return getattr(obj, name, *defargs)
 
AttributeError: module 'keyword' has no attribute '__call__'
```

mhvk:
Thanks! Though it is very puzzling indeed; the whole beauty was to make it unnecessary to set it in tests individually... I think in my local trial, I also didn't have the problem, but now wonder if perhaps I was not running the whole test suite.  Will check.

pllim:
Ah, interesting, the failure I reported in #10471 is not Windows-specific. See https://travis-ci.org/github/astropy/astropy/jobs/696220533

mhvk:
So, this is really weird: if I just remove `remote_data` and do not apply the fix you have here, I do reproduce the test failure, that internet gets accessed with `tox -e test`. If confirm this if I then do
```
cd .tox/test
source bin/activate
pytest --pyargs astropy
```
However, if I next do
```
pytest --pyargs astropy -P coordinates
```
I do *not* get the failure!

mhvk:
Ah, getting somewhere:
```
pytest --pyargs astropy -P config,coordinates
```
does lead to failure. Maybe config messes up the configuration?!

mhvk:
Fix in #10474! Now your previous work comes to its full use - see how many `remote_data` were removed!!!

pllim:
Thank you very much, @mhvk ! I'll have a look at #10474 after putting out the html-docs fire.

PR #10472, opened 2020-06-08
MNT: Remove PY2 compat for urlparse import

pllim:
### Description

This pull request is to address a PY2 import that we no longer need.


pllim:
p.s. If you want to backport, feel free to change the milestone.

PR #8727, opened 2019-05-18
Make parallel tests insensitive to user configuration

astrofrog:
This is an attempt to fix https://github.com/astropy/astropy/issues/8691 - I discovered that when run in parallel mode, the user configuration is **not** ignored.

The first step here is to add a regression test by writing out to the config file before running the parallel tests on CircleCI. This failed, so I'm now pushing some fixes.

It turns out there are several places where we need to set the config/cache to be temporary directories, not just in the test runner:

* ``astropy/conftest.py`` - we need to set the path to the temporary config/cache directories here for cases when running the tests in parallel mode using e.g.

```
python setup.py test --parallel
```

This is because pytest starts up new subprocesses, which aren't aware of the temporary directories set in the test runner.

* ``conftest.py`` - the temporary paths need to be set here for when pytest is run directly on an astropy package, e.g.

```
pytest astropy/io/votable
```

* ``docs/conftest.py`` - we also need to set them here for when running the tests with ``python setup.py test -t docs/...`` since then the root-level ``conftest.py`` gets ignored (since the tests run from outside the source directory, and ``astropy/conftest.py`` gets ignored since we are only testing things in ``docs``

This will become simpler some day if we switch to only using ``pytest`` directly.

Fixes https://github.com/astropy/astropy/issues/8154

PR #10474, opened 2020-06-09
Remove unnecessary remote_data in (mostly) coordinates.

### LINKED ISSUE #10471: TST: TestManipulation::test_broadcast_to fails with older dependencies
### Description

`TestManipulation::test_broadcast_to` fails on Windows locally when I run `pytest astropy/coordinates/tests/test_shape_manipulation.py`.

### Expected behavior

Test to pass.

### Actual behavior

```
============================= test session starts =============================
platform win32 -- Python 3.7.1, pytest-5.4.1, py-1.7.0, pluggy-0.13.1

Running tests with Astropy version 4.2.dev240+gc3c09942c.
Running tests in astropy/coordinates/tests/test_shape_manipulation.py.

Date: 2020-06-08T17:03:35

Platform: Windows-10-10.0.18362-SP0

Executable: c:\...\miniconda3\envs\py37\python.exe

Full Python Version:
3.7.1 | packaged by conda-forge | (default, Nov 13 2018, 19:01:41) [MSC v.1900 64 bit (AMD64)]

encodings: sys: utf-8, locale: cp1252, filesystem: utf-8
byteorder: little

Package versions:
Numpy: 1.16.3
Scipy: 1.3.1
Matplotlib: not available
h5py: 2.10.0
Pandas: 0.24.2
Cython: 0.29.10
Scikit-image: 0.14.2
asdf: 2.6.0

rootdir: ..., inifile: setup.cfg
plugins: asdf-2.6.0, hypothesis-5.1.5, arraydiff-0.2, astropy-header-0.1.2, cov-2.8.1, doctestplus-0.6.1, filter-subpackage-0.1.1, forked-0.2, openfiles-0.4.0, remotedata-0.3.1, xdist-1.24.1
collected 11 items

astropy\coordinates\tests\test_shape_manipulation.py ..........F         [100%]

================================== FAILURES ===================================
_____________________ TestManipulation.test_broadcast_to ______________________

self = <astropy.coordinates.tests.test_shape_manipulation.TestManipulation object at 0x0000025B0A1C5668>

    def test_broadcast_to(self):
        s1_broadcast = np.broadcast_to(self.s1, (20, 6, 7))
        assert s1_broadcast.shape == (20, 6, 7)
>       assert np.all(s1_broadcast.data.lon == self.s1.data.lon[np.newaxis])
E       AttributeError: 'memoryview' object has no attribute 'lon'

astropy\coordinates\tests\test_shape_manipulation.py:346: AttributeError
```

UPDATE: Looks like `TestManipulation` is completely skipped in the Windows job in CI because they are marked as remote data.

UPDATE: Looks like this is not Windows-specific, as the CI also sees it with Linux job with older dependencies.

mhvk:
In #10311, @pllim made the tests run by default using no download for IERS. This means that a lot of tests can now be run (again) without `remote_data`. It turned out, though, that the tests of `astropy.config` messed up the configuration setting. This PR fixes that, and removes the now no longer needed `remote_data` (I hope I didn't miss any... do we need a test that `remote_data` actually tries to access the internet???)

Note that the first commit is a missed item in #10337, which I include here since without the work here, the fix would not get tested (which was why it passed by in the first place). As such: fixes #10471

EDIT: Close #10469


mhvk:
Hmm, the double test fails too, so there must be another place where the configuration gets changed...

And still the mysterious error from the docs...

pllim:
Docs error is unrelated but the double one definitely is. Turns out the "double run" command is very different from the rest:
```
python -c 'import sys; from astropy import test; test(); sys.exit(test())'
```

I don't know why yet, but I have a feeling that it is not completely equivalent to running `pytest` in the source dir. For example, one documented behavior difference is at #8153 , where Dan even called for getting rid of the test runner entirely.

pllim:
@mhvk , what if we forcefully do a `astropy.utils.iers.conf.auto_download = False` between Run 1 and Run 2 on this line?

https://github.com/astropy/astropy/blob/c17ce8b4593570cbed33a5e0c95251e4d5deaa2c/tox.ini#L90

mhvk:
@pllim - indeed, that may be a hack to get things to pass, but, really, `astropy.test()` should not mess with the configuration! I'll see if I can narrow it down a bit.

mhvk:
OK, I think it was a combination of `set_temp_config` being rather liberal in clearing earlier configuration (I guess assuming it was all on disk, so would just be re-read) and my earlier fix for the configuration tests not being quite good enough. It should now work... (and I rebased on master to pick up the fix for the docs, so ideally all will be green :crossed_fingers:)

pllim:
Hmm... Looks like some of the tests might be relying on the old behavior... The diff makes no sense at a glance.

mhvk:
And then I clearly didn't bother to check the normal testing still worked!

mhvk:
Actually, very weird. I cannot reproduce the travis failures.  What has `io.votable` to do with anything!?!?

mhvk:
So, that was funny: it turned out I couldn't reproduce the travis failures because locally I have a normal, all-default astropy config file. When I replaced that with the temporary one created in `.travis.yml`, I could reproduce the failures. Since it actually seems better to not have to set up a fake config file, I just removed it from `.travis.yml`. But, @astrofrog, since you added the line originally, if you have a chance to have a look whether this all makes sense, that would be very welcome

.travis.yml
```diff
@@ -223,11 +223,6 @@ before_install:
      # Check CC variable
     - echo "CC="$CC
 
-    # Write configuration items to standard location to make sure they are
```
mhvk, line 226: I'm not completely sure why, but with the other changes that ensure the configuration never needs to be reloaded, this writing of a configuration file now causes failures, so I removed it.

astropy/config/paths.py
```diff
@@ -243,16 +243,16 @@ def __enter__(self):
         # Special case for the config case, where we need to reset all the
         # cached config objects
         from .configuration import _cfgobjs
-
-        path = super().__enter__()
+        self._cfgobjs_copy = _cfgobjs.copy()
```
mhvk, line 246: The changes here are to *keep* the configuration instead of deleting it at the end so that it gets reloaded from the configuration file. This is because some config items can have been set programmatically (in particular, `utils.iers.conf.auto_download`)

astropy/config/tests/test_configs.py
```diff
@@ -16,6 +16,19 @@
 from astropy.utils.exceptions import AstropyDeprecationWarning
 
 
+OLD_CONFIG = {}
+
+
+def setup_module():
```
mhvk, line 22: Arguably, all the test in this file should be written such that they do not change the configuration, but this was easier...

astropy/config/tests/test_configs.py
```diff
@@ -47,6 +63,8 @@ def test_func():
         assert paths.get_config_dir(rootname='astropy') == temp_astropy_config
 
     assert not os.path.exists(temp_config_dir)
+    # Check that we have returned to our old configuration.
+    assert configuration._cfgobjs == OLD_CONFIG
```
mhvk, line 67: This verifies that `set_temp_config` does not destroy the old configuration.

mhvk:
I added some comments to make review easier. Beyond the first few files, everything is just removal of `remote_data`.

.travis.yml
pllim: This was to fix a problem with parallel tests. :(

#8727

pllim: What about... Instead of removing this, we add a line to the `astropy.cfg` here to set `auto_download = False`?

mhvk: I wondered about it, but it seemed to me that your solution of a change in `astropy/conftest.cfg` was really more elegant, and works for everybody.

Do you think we still need this at all? It doesn't seem to give failures, and it is hard to see how we can mess up the configuration on travis.

pllim: Oh, I didn't realize I made it unnecessary. I'll have to remember what I did and get back to you.

astropy/config/tests/test_configs.py
```diff
@@ -16,6 +16,19 @@
 from astropy.utils.exceptions import AstropyDeprecationWarning
 
 
+OLD_CONFIG = {}
+
+
+def setup_module():
+    OLD_CONFIG.clear()
+    OLD_CONFIG.update(configuration._cfgobjs)
```
pllim, line 24: The need to call "hidden" stuff might indicate that we should expose the API properly? But that is discussion for another day.

.travis.yml
pllim: I am curious... You said this command made the tests fail in Travis CI. However, the same command was also added in #8727 to CircleCI and that didn't affect your PR. How could that be? 🤔

mhvk: I'm totally confused!  But with this still in place (one but last commit), both Initial tests on travis failed (see https://travis-ci.org/github/astropy/astropy/builds/696611940). I'm actually not quite sure why they failed, but trying to determine why the tests passed locally I realized that the only difference was the config file (and locally tests also fail if I use the now-removed lines as config).

astropy/config/tests/test_configs.py
mhvk: Yes, or one should have a context manager option that ensures the configuration gets returned in its original state. But I'd indeed like to keep that for another PR.

mhvk:
@pllim - thanks for noticing the remote data failures - they are indeed related, since those two specifically test that an up-to-date IERS file is used. They have `remote_data`, but now I should add that we do allow downloads there.

astropy/time/tests/test_basic.py
```diff
@@ -1346,7 +1346,15 @@ def test_TimeFormat_scale():
 
 
 @pytest.mark.remote_data
-def test_scale_conversion():
+def test_scale_conversion(monkeypatch):
+    monkeypatch.setattr('astropy.utils.iers.conf.auto_download', True)
+    with iers.earth_orientation_table.set(iers.IERS_B.open()):
```
mhvk, line 1351: Note: I was very surprised about the single line in the test as written so rewrite to make it similar to the original purpose of this test (which was removed because it seemed tricky to do; indeed, resulting in a discussion that was not actually resolved... https://github.com/astropy/astropy/pull/6606/files#r140902373).

mhvk:
@pllim - you may want to look just at the last commit, which will hopefully fix the remote-data failures (though I fear some numpy-dev stuff may be coming...).

astropy/time/tests/test_basic.py
pllim: Hmm, that new test for exception with bundled table doesn't need to be marked as remote data, right? Can it be in its own function? Or does it still somehow needs internet?

astropy/coordinates/tests/test_intermediate_transformations.py
```diff
@@ -519,11 +519,12 @@ def test_teme_itrf():
 
 
 @pytest.mark.remote_data
-def test_earth_orientation_table():
+def test_earth_orientation_table(monkeypatch):
     """Check that we can set the IERS table used as Earth Reference.
 
     Use the here and now to be sure we get a difference.
     """
+    monkeypatch.setattr('astropy.utils.iers.conf.auto_download', True)
```
pllim, line 527: Usually not a fan of monkeypatching but it does seem less clunky than the `try ... finally ...` implementation I did in #10311, so 👍

mhvk: There is also `with iers.iers.Conf.auto_download.set_temp(True)` but I felt this was a bit more natural. I actually rather liked your `setup_class` and `teardown_class`, but it seemed too much effort for a single function.

mhvk:
Darn, getting that thing that even using `tox` apparently doesn't always solve: an error here on `dev` that I cannot reproduce locally.

astropy/time/tests/test_basic.py
mhvk: Looking at this again, there really is no point in testing the exception, so I removed it again. Instead, I now added a comment so it is clearer why the test exists.

mhvk:
OK, by removing the `catch_warnings` and relying on `pytest` to error, the tests now pass. @pllim - sorry that this picks one piece of your PR!

Anyway, should now be ready!

.travis.yml
pllim: Given that it still exists in CircleCI for "32 bit and parallel" tests and things pass there, probably okay to remove here. If @astrofrog disagrees when he is available again, we can always do a follow-up PR.

astropy/config/paths.py
pllim: Is the "reset all the cached config objects" comment on L243 above still accurate? If not, should remove/modify it.

And this is a subtle API change for `set_temp_config`, right? I wonder if it warrants a change log.

astropy/coordinates/tests/test_intermediate_transformations.py
```diff
@@ -533,26 +519,24 @@ def test_teme_itrf():
...
-        altaz_auto = sc.transform_to(altaz)
-
-    assert len(w) == 0
+    # Note: tests run with warnings turned into errors, so it is
+    # meaningful if this passes.
+    altaz_auto = sc.transform_to(altaz)
 
     with iers.earth_orientation_table.set(iers.IERS_B.open()):
-        with catch_warnings() as w:
+        with pytest.warns(AstropyWarning, match='after IERS data'):
```
pllim, line 538: I must be going crazy. Pretty sure I needed some `.*` in the regex when I worked on it, but tests passed here without. 🤷

astropy/coordinates/tests/test_intermediate_transformations.py
```diff
@@ -533,26 +519,24 @@ def test_teme_itrf():
...
             altaz_b = sc.transform_to(altaz)
-        assert len(w) == 1
```
pllim, line 554: So you don't want to check to see if this is the *only* warning emitted anymore?

mhvk: I think it is fine to just test that we get the right warning. If there are others, pytest would presumably fail.

astropy/config/paths.py
mhvk: Yes, I guess you are right. I now updated the comment and added a changelog entry.

astropy/coordinates/tests/test_intermediate_transformations.py
pllim: Not really. Consider this example:

```python
import warnings

import pytest


def myfunc():
    warnings.warn('Catch this', UserWarning)
    warnings.warn('Nope', UserWarning)  # Even if you change the warning type, same result below


def test1():
    with pytest.warns(UserWarning, match='Catch this'):
        myfunc()  # passes


def test2():
    with pytest.warns(UserWarning, match='Catch this') as w:
        myfunc()
    assert len(w) == 1  # fails
```

pllim: It comes down to whether you care if the functionality emits other warnings you are not checking for or not.

mhvk: I think in this case we check the code enough in other places that I don't worry, but clearly that is not the behaviour I had expected....

mhvk:
Yes, this was a nice collaboration on getting rid of the need for remote data!

PR #10475, opened 2020-06-09
Modifications to transforming frames with data as opposed to SkyCoord

ayshih:
The coordinates transformation framework is currently a bit confusing/inconsistent because users can work with either `SkyCoord` instances or frame instances with data.  Frame instances with data behave differently from `SkyCoord` because they lack the ability to merge frame attributes across a transformation path.  This is particularly confusing when transforming a frame instance with data to a frame *class*, because the behavior is significantly different from the seemingly analogous call when using `SkyCoord`.  See the discussion in #10453.

This PR makes the following changes:

1. Modify `CoordinateAttribute` so that it always uses `SkyCoord` for the internal transformation (with the default setting of `merge_attributes=True`).
2. Revise any documentation examples that instantiate frames directly with data to instead use `SkyCoord`, unless there is some strong reason to avoid using `SkyCoord`.
3. Deprecate being able to pass a class to `BaseCoordinateFrame.transform_to()`, and point users to `SkyCoord`.

To do:
- [x] Add tests for the new behavior of `CoordinateAttribute`
- [x] Find all of the examples of non-use of `SkyCoord`
- [x] Add changelog entries

### DIFF
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -1172,7 +1172,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
 
         Parameters
         ----------
-        new_frame : class or frame object or SkyCoord object
+        new_frame : frame object or SkyCoord object
             The frame to transform this coordinate frame into.
 
         Returns
@@ -1201,6 +1201,12 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
                                       'please comment at https://github.com/astropy/astropy/issues/6280')
 
         if inspect.isclass(new_frame):
+            warnings.warn("Transforming a frame instance to a frame class (as opposed to another "
+                          "frame instance) will not be supported in the future.  Either "
+                          "explicitly instantiate the target frame, or first convert the source "
+                          "frame instance to a `astropy.coordinates.SkyCoord` and use its "
+                          "`transform_to()` method.",
+                          AstropyDeprecationWarning)
             # Use the default frame attributes for this class
             new_frame = new_frame()
 
--- a/astropy/coordinates/builtin_frames/galactocentric.py
+++ b/astropy/coordinates/builtin_frames/galactocentric.py
@@ -217,50 +217,52 @@ doc_footer = """
 
     To transform to the Galactocentric frame with the default
     frame attributes, pass the uninstantiated class name to the
-    ``transform_to()`` method of a coordinate frame or
-    `~astropy.coordinates.SkyCoord` object::
+    ``transform_to()`` method of a `~astropy.coordinates.SkyCoord` object::
 
         >>> import astropy.units as u
         >>> import astropy.coordinates as coord
-        >>> c = coord.ICRS(ra=[158.3122, 24.5] * u.degree,
-        ...                dec=[-17.3, 81.52] * u.degree,
-        ...                distance=[11.5, 24.12] * u.kpc)
+        >>> c = coord.SkyCoord(ra=[158.3122, 24.5] * u.degree,
+        ...                    dec=[-17.3, 81.52] * u.degree,
+        ...                    distance=[11.5, 24.12] * u.kpc,
+        ...                    frame='icrs')
         >>> c.transform_to(coord.Galactocentric) # doctest: +FLOAT_CMP
-        <Galactocentric Coordinate (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
+        <SkyCoord (Galactocentric: galcen_coord=<ICRS Coordinate: (ra, dec) in deg
             (266.4051, -28.936175)>, galcen_distance=8.122 kpc, galcen_v_sun=(12.9, 245.6, 7.78) km / s, z_sun=20.8 pc, roll=0.0 deg): (x, y, z) in kpc
             [( -9.43489286, -9.40062188, 6.51345359),
-            (-21.11044918, 18.76334013, 7.83175149)]>
+             (-21.11044918, 18.76334013, 7.83175149)]>
 
 
     To specify a custom set of parameters, you have to include extra keyword
     arguments when initializing the Galactocentric frame object::
 
         >>> c.transform_to(coord.Galactocentric(galcen_distance=8.1*u.kpc)) # doctest: +FLOAT_CMP
-        <Galactocentric Coordinate (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
+        <SkyCoord (Galactocentric: galcen_coord=<ICRS Coordinate: (ra, dec) in deg
             (266.4051, -28.936175)>, galcen_distance=8.1 kpc, galcen_v_sun=(12.9, 245.6, 7.78) km / s, z_sun=20.8 pc, roll=0.0 deg): (x, y, z) in kpc
             [( -9.41284763, -9.40062188, 6.51346272),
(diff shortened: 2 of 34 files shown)


astropy/coordinates/baseframe.py
```diff
@@ -1201,6 +1201,11 @@ def transform_to(self, new_frame):
                                       'please comment at https://github.com/astropy/astropy/issues/6280')
 
         if inspect.isclass(new_frame):
+            warnings.warn("Transforming a frame instance to a frame class (as opposed to another "
+                          "frame instance) will not be supported in the future.  Either "
+                          "explicitly instantiate the target frame or first convert this frame "
```
mhvk, line 1206: "this frame instance" is perhaps a little unclear. How about "or first convert the source frame instance to a ...SkyCoord and its its transform_to method."

astropy/coordinates/tests/test_sky_coord.py
```diff
@@ -64,8 +64,8 @@ def teardown_function(func):
 
 
 def test_transform_to():
-    for frame in (FK5, FK5(equinox=Time('J1975.0')),
-                  FK4, FK4(equinox=Time('J1975.0')),
+    for frame in (FK5(), FK5(equinox=Time('J1975.0')),
```
mhvk, line 67: This is a test where it is truly obvious this is an improvement!

astropy/coordinates/baseframe.py
ayshih: Done

ayshih:
As I was adding the empty parentheses to explicitly instantiate destination frames, I amusingly discovered that there was a smattering of occurrences of empty parentheses already being used, possibly because the writer wanted to make explicit what the destination frame was.

astropy/coordinates/baseframe.py
```diff
@@ -1201,6 +1201,12 @@ def transform_to(self, new_frame):
                                       'please comment at https://github.com/astropy/astropy/issues/6280')
 
         if inspect.isclass(new_frame):
+            warnings.warn("Transforming a frame instance to a frame class (as opposed to another "
+                          "frame instance) will not be supported in the future.  Either "
+                          "explicitly instantiate the target frame, or first convert the source "
+                          "frame instance to a `astropy.coordinates.SkyCoord` and use its "
+                          "`transform_to()` method.",
+                          AstropyDeprecationWarning)
```
adrn, line 1209: Good! Let's also remember to make an issue to remove this warning and functionality.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -217,50 +217,52 @@ def validate(cls, value):
...
 
         >>> import astropy.units as u
         >>> import astropy.coordinates as coord
-        >>> c = coord.ICRS(ra=[158.3122, 24.5] * u.degree,
-        ...                dec=[-17.3, 81.52] * u.degree,
-        ...                distance=[11.5, 24.12] * u.kpc)
+        >>> c = coord.SkyCoord(ra=[158.3122, 24.5] * u.degree,
+        ...                    dec=[-17.3, 81.52] * u.degree,
+        ...                    distance=[11.5, 24.12] * u.kpc,
+        ...                    frame=coord.ICRS)
```
adrn, line 227: I think we can leave off the `frame=coord.ICRS` and (optionally) add a comment like `# the default frame is ICRS`.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -217,50 +217,52 @@ def validate(cls, value):
...
         >>> c.transform_to(coord.Galactocentric) # doctest: +FLOAT_CMP
-        <Galactocentric Coordinate (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
+        <SkyCoord (Galactocentric: galcen_coord=<ICRS Coordinate: (ra, dec) in deg
```
adrn, line 229: If `CoordinateAttribute` now holds a `SkyCoord`, won't this repr change at `galcen_coord=<ICRS Coordinate:...` as well?

docs/coordinates/frames.rst
```diff
@@ -288,12 +288,12 @@ Transforming between Frames
 
 To transform a frame object with data into another frame, use the
 ``transform_to`` method of an object, and provide it the frame you wish to
-transform to. This frame can either be a frame *class*, in which case
-the default attributes will be used, or a frame object (with or without
-data)::
+transform to.  This frame should be a frame object (with or without data).
```
adrn, line 291: Just to be explicit here for people who may have missed or skipped earlier discussion, maybe
```suggestion
transform to.  This frame should be a frame object (with or without coordinate data).
```

astropy/coordinates/builtin_frames/galactocentric.py
Cadair: Explicit is better than implicit? :wink:

adrn: OK, but then we should [add it everywhere](https://docs.astropy.org/en/latest/coordinates/index.html) and remove the default!?

Cadair: I have always thought it was weird there was a default, but that ship has sailed off to another solar system at this point.

mhvk: I'd leave it out (like we don't use `scale='utc'` in `Time`) or use the string 'icrs'.

docs/coordinates/frames.rst
ayshih: Done

astropy/coordinates/builtin_frames/galactocentric.py
ayshih: Changed it to `frame='icrs'`

ayshih:
Okay, PR updated:

- Changelog entry (API) for `transform_to()` deprecation
- Tests and changelog entry (bugfix) for `CoordinateAttribute`
- Fixed all the examples that I could find where a frame with data was used instead of a `SkyCoord`
- Other minor stuff

astropy/coordinates/builtin_frames/galactocentric.py
ayshih: `CoordinateAttribute` doesn't actually hold a `SkyCoord`.  To preserve the API, it uses a `SkyCoord`-style internal transformation, but then it drops it down to a frame with data.  (It's like the current behavior when a `SkyCoord` is provided as input.)

ayshih:
I just realized that this PR hadn't been merged yet...  I've now rebased it against the latest.

adrn:
Thanks very much @ayshih !

pllim:
This code in Ginga (modified so you can run it outside of Ginga) now raises deprecation warning with astropy dev.

```python
import numpy as np
from astropy import coordinates, units

coordsys = 'fk5'
system = 'icrs'
wcspt = np.array([[0.1234, -34.34], [0.4567, 42.42]])

frame_class = coordinates.frame_transform_graph.lookup_name(coordsys)
ra_deg = wcspt[:, 0]
dec_deg = wcspt[:, 1]
coord = frame_class(ra_deg * units.degree, dec_deg * units.degree)
to_class = coordinates.frame_transform_graph.lookup_name(system)

coord = coord.transform_to(to_class)  # Deprecation warning here
```

Is this the correct solution?

```python
coord = coord.transform_to(to_class())
```

mhvk:
Yes, that is correct! The idea was that down at the frame level one should be specific.

PR #10482, opened 2020-06-11
Fix issue when setting a keyword ending with space

### LINKED ISSUE #10479: Keywords' values are not replaced in fits.Header when keywords end in space
### Description
Normally, when a keyword in `astropy.io.fits.Header` is set twice second value replaces previous value of the same keyword. This does not happen when keywords end in space (`' '`)

### Expected behavior
I would expect that setting a keyword value of a keyword like `'KEY '` to 1 (first time) and to 2 (second time), the second value would replace the first value.

### Actual behavior
When setting a keyword value of a keyword like `'KEY '` to 1 (first time) and to 2 (second time), the second value **does not** replace the first value. Instead the header has two cards with the same keyword with two different values (1 and 2 in this case).

### Steps to Reproduce

```python
In [1]: from astropy.io import fits
   ...: print("--- FIRST iteration ---")
   ...: hdr = fits.Header()
   ...: hdr['KEY1'] = 1
   ...: hdr['KEY2 '] = 2
   ...: print(repr(hdr))
   ...: hdr['KEY1'] = 3
   ...: hdr['KEY2 '] = 4
   ...: print("--- SECOND iteration ---")
   ...: print(repr(hdr))
   ...: print("value of 'KEY2 ' is: {}".format(hdr['KEY2 ']))
--- FIRST iteration ---
KEY1    =                    1
KEY2    =                    2
--- SECOND iteration ---
KEY1    =                    3
KEY2    =                    2
KEY2    =                    4
value of 'KEY2 ' is: 2
```
I would expect the value of `'KEY2 '` to be 4 and that there were just one entry for this keyword in the header.

If space character is supported in keyword names (i.e., if it is simply ignored, like `strip()`), then the behavior should be as expected by me. If space at the end is not supported, then I would expect to see a warning like in the following situation:
```python
In [2]: hdr['a#'] = 9
WARNING: VerifyWarning: Keyword name 'a#' is greater than 8 characters or contains characters not allowed by the FITS standard; a HIERARCH card will be created. [astropy.io.fits.card]

In [3]: hdr['a '] = 9  # No warning!

In [4]:
```

### System Details
```
Darwin-17.7.0-x86_64-i386-64bit
Python 3.7.6 (default, Jan  8 2020, 13:42:34)
Numpy 1.18.0
astropy 4.2.dev167+g8bdd24b04
Scipy 1.3.2
Matplotlib 3.2.0
```

saimn:
Fix #10479.

### DIFF
--- a/astropy/io/fits/tests/test_header.py
+++ b/astropy/io/fits/tests/test_header.py
@@ -2377,8 +2377,9 @@ class TestHeaderFunctions(FitsTestCase):
         """
         hdr = fits.Header()
         hdr['KEY2 '] = 2
-        hdr['KEY2 '] = 4
+        hdr['KEY2  '] = 4
         assert len(hdr) == 1
+        assert hdr['KEY2'] == 4
         assert hdr['KEY2 '] == 4
 
 
--- a/astropy/io/fits/tests/test_header.py
+++ b/astropy/io/fits/tests/test_header.py
@@ -2371,6 +2371,16 @@ class TestHeaderFunctions(FitsTestCase):
         assert pri_hdr == pri_hdr_from_bytes
         assert pri_hdr.tostring() == pri_hdr_from_bytes.tostring()
 
+    def test_set_keyword_with_space(self):
+        """
+        Regression test for https://github.com/astropy/astropy/issues/10479
+        """
+        hdr = fits.Header()
+        hdr['KEY2 '] = 2
+        hdr['KEY2 '] = 4
+        assert len(hdr) == 1
+        assert hdr['KEY2 '] == 4
+
 
 class TestRecordValuedKeywordCards(FitsTestCase):
     """
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -887,6 +887,8 @@ astropy.io.fits
 
 - Fix crash when reading an invalid table file. [#10171]
 
+- Fix duplication issue when setting a keyword ending with space. [#10482]
+
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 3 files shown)


astropy/io/fits/tests/test_header.py
```diff
@@ -2371,6 +2371,16 @@ def test_header_fromstring_bytes(self):
...
 
+    def test_set_keyword_with_space(self):
+        """
+        Regression test for https://github.com/astropy/astropy/issues/10479
+        """
+        hdr = fits.Header()
+        hdr['KEY2 '] = 2
+        hdr['KEY2 '] = 4
+        assert len(hdr) == 1
+        assert hdr['KEY2 '] == 4
```
mcara, line 2382: Just maybe to make things weirder, replace:
```suggestion
        assert hdr['KEY2  '] == 4
```
to have a different number of trailing spaces.

mcara:
This PR seems to have fixed the issue. Thanks!

astropy/io/fits/tests/test_header.py
dhomeier: In that case I'd recommend to keep both, just in case some even weirder bug sneaks in that leaves it at
```
hdr['KEY2 '] = 2
hdr['KEY2  '] = 4
```

saimn: Good ideas, thanks, I updated the test.

mcara: @dhomeier That would be detected by `assert len(hdr) == 1`

mcara:
Happy

nden:
Since this is milestoned as 4.0.2, is it going to go in 4.1?

pllim:
> Since this is milestoned as 4.0.2, is it going to go in 4.1?

Gotta ask @eteq . Definitely in 4.1.1 but not sure about 4.1.

saimn:
Thanks @mcara and @pllim :), merging.

PR #10484, opened 2020-06-11
Add world coordinate names to coordinate map aliases

Cadair:
To be honest I really thought this was already the behaviour, so I made it that way.

As well as using `world_axes_names` for aliases in `CoordinatesMap` this actually fixes some tests so they, you know, test things :eyes:

### DIFF
--- a/astropy/visualization/wcsaxes/coordinates_map.py
+++ b/astropy/visualization/wcsaxes/coordinates_map.py
@@ -96,7 +96,10 @@ class CoordinatesMap:
             # Set up aliases for coordinates
             if isinstance(name, tuple):
                 for nm in name:
-                    self._aliases[nm] = index
+                    # Do not replace an alias already in the map if we have
+                    # more than one alias for this axis.
+                    if nm.lower() not in self._aliases:
+                        self._aliases[nm] = index
             else:
                 self._aliases[name.lower()] = index
 
--- a/astropy/visualization/wcsaxes/coordinates_map.py
+++ b/astropy/visualization/wcsaxes/coordinates_map.py
@@ -96,9 +96,10 @@ class CoordinatesMap:
             # Set up aliases for coordinates
             if isinstance(name, tuple):
                 for nm in name:
+                    nm = nm.lower()
                     # Do not replace an alias already in the map if we have
                     # more than one alias for this axis.
-                    if nm.lower() not in self._aliases:
+                    if nm not in self._aliases:
                         self._aliases[nm] = index
             else:
                 self._aliases[name.lower()] = index
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -465,6 +465,9 @@ astropy.visualization
 - Fixed an issue where ticks were sometimes not drawn at the edges of a spherical
   projection on a WCSAxes. [#10442]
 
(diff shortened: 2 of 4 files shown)


Cadair:
Hmm, the figure test fail is my fault, but unnecessary.

pllim:
Not sure who besides Tom R is qualified to review this...

Cadair:
@astrofrog plz can haz review? :grinning:

astrofrog:
Yep! The image tests are failing though right?

Cadair:
I am reasonably sure it's supposed to, but I will double check.

astropy/visualization/wcsaxes/coordinates_map.py
```diff
@@ -96,7 +96,10 @@ def __init__(self, axes, transform=None, coord_meta=None,
             # Set up aliases for coordinates
             if isinstance(name, tuple):
                 for nm in name:
-                    self._aliases[nm] = index
+                    # Do not replace an alias already in the map if we have
+                    # more than one alias for this axis.
+                    if nm.lower() not in self._aliases:
```
astrofrog, line 101: remove lower? or check that both the lowercase and original are not in aliases

astrofrog (commented):
Looks good but minor question below. Also if image failures are deliberate can you open a PR to add new baseline images?

astropy/visualization/wcsaxes/coordinates_map.py
Cadair: line 102 below is the only place we add things to `self._aliases` and we do it with a lower?

Cadair:
For the record the image change is deliberate and it's to test that when specified the axes are labelled with the world names and not the physical types. This isn't a change introduced in this PR, that is existing behaviour I just noticed it wasn't being tested.

astropy/visualization/wcsaxes/coordinates_map.py
astrofrog: line 102 doesn't show a lower, only line 104 does?

astrofrog:
Ok can you open a PR for the image tests? (to the data repo)

astropy/visualization/wcsaxes/coordinates_map.py
Cadair: that's what I meant :facepalm:

astrofrog: But my point is that on line 102 we also add things to aliases *without* a ``lower``? Shouldn't both lines 102 and 104 use lower?

astrofrog (changes requested):
One remaining comment below and a comment in https://github.com/astropy/astropy-data/pull/97#pullrequestreview-514163141 - once these are resolved this is good to go.

astropy/visualization/wcsaxes/coordinates_map.py
Cadair: I just added the lower to all the things?

Cadair:
Looks like I forgot one of the images: https://github.com/astropy/astropy-data/pull/99

astrofrog (approved):
Looks good, thanks!

PR #10492, opened 2020-06-15
Update astropy-package-template.rst

bmorris3:
### Description

This pull request is to address astropy/package-template#474 by tweaking the wording of the docs to suggest using the `git tag` command rather than changing the version number in `setup.cfg` (which is no longer used).

Fixes astropy/package-template#474.

### DIFF
--- a/docs/development/astropy-package-template.rst
+++ b/docs/development/astropy-package-template.rst
@@ -34,9 +34,10 @@ by ``CHANGES.md`` in the instructions.
    and update the release date, which should currently be set to
    ``unreleased``, to the current date in ``yyyy-mm-dd`` format.
 
-#. Update the version number in ``setup.cfg`` to the version you're about to
-   release, without the ``.dev`` suffix (e.g. ``0.1``). If your package uses
-   setuptools_scm to manage version numbers, you can skip this step.
+#. Update the version number to the version you're about to
+   release (using ``git tag``, for example), without the ``.dev`` suffix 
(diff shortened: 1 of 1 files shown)


bmorris3:
Looking over the rest of the docs a bit closer, we might also need to change some of the later steps which imply the `VERSION` must be changed, which we should consider before this gets merged. I would appreciate advice from an expert like @cadair, for example, on what the ideal workflow should look like.

docs/development/astropy-package-template.rst
```diff
@@ -34,9 +34,10 @@ by ``CHANGES.md`` in the instructions.
    and update the release date, which should currently be set to
    ``unreleased``, to the current date in ``yyyy-mm-dd`` format.
 
-#. Update the version number in ``setup.cfg`` to the version you're about to
-   release, without the ``.dev`` suffix (e.g. ``0.1``). If your package uses
-   setuptools_scm to manage version numbers, you can skip this step.
+#. Update the version number to the version you're about to
```
pllim, line 37: Ops... I think this PR fell through the cracks. Reading this again, I think this whole entry can be deleted. Using `git tag` is the way when you also use `setuptools_scm`, and this entry says you can skip this step if you use `setuptools_scm`.

pllim:
I just went ahead to remove that entry altogether. I don't think this needs a backport.

pllim:
LoL... oh no, stalebot ate up the API limit... 😅  `API rate limit exceeded for installation ID`

pllim (approved):

PR #10497, opened 2020-06-16
MNT: Remove doc build from CircleCI (use RTD PR builder)

pllim:
### Description

This pull request is to remove documentation build job from CircleCI and rely on [RTD PR builder](https://docs.readthedocs.io/en/latest/guides/autobuild-docs-for-pull-requests.html). The feature is already enabled on the admin side, but let's see if it really works. 🤞 

rtd_astropy

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -93,13 +93,15 @@ commands =
     cov-!double: pytest --pyargs astropy {toxinidir}/docs {env:MPLFLAGS} --cov astropy --cov-config={toxinidir}/setup.cfg {posargs}
     double: python -c 'import sys; from astropy import test; test(); sys.exit(test())'
 
+# This lets developers to use tox to build docs and ignores warnings.
+# This is not used in CI; For that, we have RTD PR builder.
 [testenv:build_docs]
 changedir = docs
 description = invoke sphinx-build to build the HTML docs
 extras = docs
 commands =
     pip freeze
-    sphinx-build -W -b html . _build/html
+    sphinx-build -b html . _build/html
 
 [testenv:linkcheck]
 changedir = docs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,63 +109,13 @@ jobs:
       - store_artifacts:
           path: results
 
-  html-docs:
-    docker:
-      - image: circleci/python:3.7
-    steps:
-      - checkout
-      - run:
-          name: Install apt packages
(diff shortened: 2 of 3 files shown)


Cadair:
Something that has come up in the discussions around this with SunPy is RTD doesn't catch sphinx warnings. I don't know of anyway to make it either...

pllim:
Actually, it can. Please see second commit (and thanks for reminding me to do it).

Cadair:
I guess the drawback of that is it's going to stop the actual deploy builds from passing in those cases as well, but as we are building them on all the PRs that's less of an issue.

pllim:
You are correct. I think the pros outweighs the cons here. If failing on warning during deploy is a show-stopper, we can perhaps submit a feature request to RTD so that directive can be ignored for deploy event (although I don't think it is necessary).

tox.ini
```diff
@@ -1,7 +1,6 @@
 [tox]
 envlist =
     py{36,37,38,dev}-test{,-alldeps,-oldestdeps,-devdeps,-numpy116,-numpy117,-numpy118}{,-cov}
-    build_docs
```
Cadair, line 4: We shouldn't remove this.

tox.ini
```diff
@@ -93,14 +92,6 @@ commands =
     cov-!double: pytest --pyargs astropy {toxinidir}/docs {env:MPLFLAGS} --cov astropy --cov-config={toxinidir}/setup.cfg {posargs}
     double: python -c 'import sys; from astropy import test; test(); sys.exit(test())'
 
-[testenv:build_docs]
-changedir = docs
-description = invoke sphinx-build to build the HTML docs
-extras = docs
-commands =
-    pip freeze
-    sphinx-build -W -b html . _build/html
```
Cadair, line 102: We should keep this as well.

I am +1 on removing the `-W` here to fix issues with building docs on mac, but I suggest we actually punt that discussion to another PR / issue.

pllim: Ah, okay. So we keep this so that developers can use tox to build docs but *not* use it in CI, correct?

pllim: Anyways, I reverted the deletion as you requested.

pllim:
Given an approval is given, I am just going to merge this now because new PRs now have duplicate doc builds, which is quite unnecessary. 😬 

Thanks, @Cadair !

PR #10500, opened 2020-06-16
MNT: Update builtin IERS

pllim:
This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.


pllim:
Oooo seem to work. Should we do this update anyway?

mhvk:
Belatedly, :fireworks: 

It definitely looks good. It would seem to make sense to do it regularly - one less thing for the release manager to worry about. On the other hand, there have been quite a few times when the number of lines changed has been very large for reasons I'm not quite sure about; it would be bad if the git history of astropy became dominated by this!

pllim:
It's okay if you want to wait; then just close this without merge. I was just testing the script. 😬

mhvk:
I do think it is great! But if the idea is to run the script shortly before the release, and then it auto-generates relevant PRs, that is even better. So, let me close it now - test worked!!!

pllim:
Automation part will have to wait till @astrofrog is back. At least we know it is doable. 😄

PR #10502, opened 2020-06-17
Fix regression for np.diff(time) and np.sum(timedelta) no longer working

### LINKED ISSUE #10501: TimeDelta does not support np.sum anymore
### Description
In the development version of Astropy the `TimeDelta` object does not work with `np.sum()` anymore. The issue was found in one of our [Gammapy Travis-CI builds](https://travis-ci.org/github/gammapy/gammapy/jobs/699226524#L8080). I presume the issue is somehow related to https://github.com/astropy/astropy/issues/8610, but I could not find a statement whether it is intended behaviour.

### Expected behavior

I would expect that the expression `np.sum(time_delta)` just returns the sum of the time differences

### Actual behavior

The code fails with a `TypeError`:

```
TypeError: no implementation found for 'numpy.sum' on types that implement __array_function__: [<class 'astropy.time.core.TimeDelta'>]
```

### Steps to Reproduce
The following code example reproduces the error:

```python
from astropy.time import Time
import numpy as np

t_1 = Time(['1999-01-01T00:00:00.123456789', '2010-01-01T00:00:00'])
t_2 = Time(['1998-01-01T00:00:00.123456789', '2009-01-01T00:00:00'])
total = np.sum(t_1 - t_2)
```

### System Details

Darwin-18.7.0-x86_64-i386-64bit
Python 3.7.0 | packaged by conda-forge | (default, Nov 12 2018, 12:34:36) 
Numpy 1.18.1
astropy 4.2.dev267+g295c2ab4b
Scipy 1.4.1
Matplotlib 3.1.2

mhvk:
In #10337, an `__array_function__` override was introduced, which allows some numpy functions to behave as expected on `SkyCoord` and `Time`. It was a strict whitelist, however, which breaks numpy functions that sort-of worked already (sort-of, since often an object array was returned).

This PR returns to the old behaviour by passing on arguments to bare numpy function if it is not among the known functions.

fixes #10501 

@taldcroft - I must admit I had no idea that `np.diff(time)` and `np.sum(timedelta)` even worked (if *very* slowly). Obviously, the final goal should be to support those properly, but this PR is just to fix the regression caused by #10337.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -64,7 +64,7 @@ astropy.time
   epoch time). [#10406]
 
 - Numpy functions that broadcast, change shape, or index (like ``np.broadcast_to``,
-  ``np.rot90``, or ``np.roll``) now work on times. [#10337]
+  ``np.rot90``, or ``np.roll``) now work on times. [#10337, #10502]
 
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
--- a/astropy/time/tests/test_functions.py
+++ b/astropy/time/tests/test_functions.py
@@ -1,0 +1,45 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import pytest
+import numpy as np
+
+from astropy.time import Time, TimeDelta
+
+
+class TestFunctionsTime:
+
+    def setup_class(cls):
+        cls.t = Time(50000, np.arange(8).reshape(4, 2), format='mjd',
+                     scale='tai')
+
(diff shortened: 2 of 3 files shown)


mhvk:
Thanks. It probably was a mistake, though it is also not good to rely on a private numpy implementation detail... The problem with `__array_function__` is that it is all or nothing. But seeing that things like `np.sum` are actually used in the wild, it will be good to have implementations for them that do not rely on casting to an object array of `Time` scalars!

PR #10503, opened 2020-06-17
MNT: Started deprecation process for enable_deprecations_as_exceptions

pllim:
### Description

This pull request is to kick off #9718 . While it does not issue any deprecation warning yet for downstream packages, core library now will stop relying on it and the change log has an early message for them that they should think about not using it too.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -523,6 +523,10 @@ astropy.table
 astropy.tests
 ^^^^^^^^^^^^^
 
+- ``enable_deprecations_as_exceptions`` is pending deprecation and will be
+  deprecated in a future release. Downstream packages should switch to using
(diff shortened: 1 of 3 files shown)


pllim:
@saimn indicated that this is still widely used (probably due to `package-template`) and deprecating this might be more trouble than it is worth.

saimn:
We should probably start by making the change in the template, hoping that this will propagate to the maintained downstream packages with the other recent changes, and then deprecate in a more distant future ?

pllim:
This PR simply adds the deprecation with `pending=True`, which means no warning will be emitted, though I am not sure if this is overkill or not. 💭 

But I think it is pretty clear we don't want it so late in the game for 4.3. And there is already an issue open (#9718) so I'll just close this...

Issue #6761, opened 2017-10-17
Deprecate catch_warnings and raises in favor of pytest machinery?


pllim:
For example, some `catch_warnings` can be replaced with `pytest.warns` (see https://docs.pytest.org/en/latest/warnings.html#warns). As a result, do we still need warning related functionality in `astropy/tests/helper.py`?

mhvk:
I've wondered too whether pytest has caught up with our custom machinery; it would certainly be nice to be as compliant as possible in our tests.

pllim:
Started an attempt at #10504

PR #10504, opened 2020-06-17
MNT: Stop using our own catch_warnings and raises

pllim:
### Description

This pull request is to replace our own `catch_warnings` with standard `pytest` stuff. This addresses most of #6761 but not going to deprecate anything.

Also see https://docs.pytest.org/en/stable/warnings.html . `pytest` was at 5.4.1 when this PR was opened.

As for deprecation warning, local testing indicates that it will bubble through and be caught by `filterwarnings = error` in `setup.cfg` even without our own `catch_warnings`.

Similarly, our own `raises` can be replaced with `pytest.raises`.

**TODO**

- [x] Why `astropy/coordinates/tests/test_iau_fullstack.py:180` and `astropy/coordinates/tests/test_regression.py:229` try to access internet on CI? I thought we already set `auto_download=False`. Maybe problem will fix itself after #10474 is merged?
- [x] Fix message being caught in `astropy/io/ascii/tests/test_ecsv.py:479` when `pyyaml` is not installed.
- Replace `catch_warnings` usage in remaining tests:
  - [x] astropy/units/tests/test_format.py
  - [x] astropy/units/tests/test_quantity_ufuncs.py
  - [x] astropy/units/tests/test_equivalencies.py
  - [x] astropy/units/tests/test_quantity.py
  - [x] astropy/utils/iers/tests/test_iers.py
  - [x] astropy/utils/tests/test_decorators.py
  - [x] astropy/utils/tests/test_data.py (this should be its own commit in case it conflicts with #10437)
  - [x] astropy/visualization/wcsaxes/tests/test_misc.py
  - [x] ~astropy/visualization/wcsaxes/tests/test_wcsapi.py~ (false positive)
  - [x] astropy/wcs/tests/test_wcs.py
  - [x] astropy/wcs/tests/test_wcsprm.py
  - [x] astropy/wcs/tests/test_utils.py
  - [x] astropy/wcs/tests/test_profiling.py
  - [x] ~astropy/wcs/wcsapi/tests/test_fitswcs.py~ (false positive)
  - [x] ~astropy/wcs/wcsapi/wrappers/tests/test_sliced_wcs.py~ (false positive)
- [ ] Since this might drag on for a while, might need to `grep` again towards the end to make sure no new `catch_warnings` snuck in upstream in the meanwhile.
  - Tip: Add a `raise RuntimeError` in `raise.__enter__` and `catch_warnings.__enter__` to make sure nothing besides `fitscheck.py` is using them.
- [x] Hmm, what about `astropy.tests.helper.raises`?
- [x] Make sure tests pass, even the remote data job that is allowed to fail.

p.s. Oh dear, I didn't realize I was going to touch so many lines. Initial grep-fu failed me and I underestimated the impact. If this sits too long and conflicts start to creep in, this PR might need to be broken down into smaller pieces that can be merged faster. Or someone can come do a mega-review quick and get this in sooner than later. 😬


astropy/coordinates/tests/test_distance.py
```diff
@@ -233,11 +232,11 @@ def test_parallax():
     with pytest.raises(ValueError):
         Distance(parallax=[10, 1, -1] * u.mas)
 
-    with catch_warnings(AstropyWarning) as w:
+    with pytest.warns(AstropyWarning) as w:
         Distance(parallax=-1 * u.mas, allow_negative=True)
     assert len(w) > 0
```
mhvk, line 237: No need to check the length (or assign to `w`), I would think

astropy/coordinates/tests/test_distance.py
```diff
@@ -233,11 +232,11 @@ def test_parallax():
...
 
-    with catch_warnings(AstropyWarning) as w:
+    with pytest.warns(AstropyWarning) as w:
```
mhvk, line 239: Same here.

astropy/coordinates/tests/test_frames.py
```diff
@@ -821,11 +820,9 @@ def test_represent_as():
     assert isinstance(rep2.differentials['s'], r.CylindricalDifferential)
 
     # single class with positional in_frame_units, verify that warning raised
-    with catch_warnings() as w:
+    with pytest.warns(AstropyWarning, match='argument position') as w:
         icrs.represent_as(r.CylindricalRepresentation, False)
-        assert len(w) == 1
-        assert w[0].category == AstropyWarning
-        assert 'argument position' in str(w[0].message)
+    assert len(w) == 1
```
mhvk, line 825: I guess here there is a more explicit test that there is only one warning, which I would be tempted to remove, but fine to leave too.

astropy/coordinates/tests/test_iau_fullstack.py
```diff
@@ -166,7 +165,7 @@ def test_future_altaz():
     if hasattr(utils, '__warningregistry__'):
         utils.__warningregistry__.clear()
 
-    with catch_warnings() as found_warnings:
+    with pytest.warns(AstropyWarning) as found_warnings:
```
mhvk, line 168: I'd write this with an `ExitStack` context (with entries added as per the `if-then` logic below), but again no problem to just keep it.

EDIT: ie., basically like you do below!

astropy/coordinates/tests/test_distance.py
```diff
@@ -49,7 +48,7 @@ def test_distances():
 
     # Coordinate objects can be assigned a distance object, giving them a full
     # 3D position
-    c = Galactic(l=158.558650*u.degree, b=-43.350066*u.degree,
+    _ = Galactic(l=158.558650*u.degree, b=-43.350066*u.degree,
```
mhvk, line 51: This is a weird one, not checking the result. Strictly, no need to assign anything, although I think I'd prefer a quick `assert c.distance == 12*u.pc`

astropy/coordinates/tests/test_iau_fullstack.py
pllim: Nesting doesn't seem to work as expected. The second context can't see the warnings.

mhvk: O, sorry to hear that (and sorry for wasting your time...)

pllim: No worries! It was a great idea and would have been very elegant if it were to work.

astropy/coordinates/tests/test_distance.py
pllim: Good catch!

astropy/coordinates/tests/test_frames.py
pllim: I think it is valid to ensure that is the only warning being raised, so I kept it.

astropy/coordinates/tests/test_distance.py
pllim: Done.

astropy/coordinates/tests/test_iau_fullstack.py
```diff
@@ -166,26 +165,19 @@ def test_future_altaz():
...
         utils.__warningregistry__.clear()
 
-    with catch_warnings() as found_warnings:
-
-        location = EarthLocation(lat=0*u.deg, lon=0*u.deg)
-        t = Time('J2161')
-
-        SkyCoord(1*u.deg, 2*u.deg).transform_to(AltAz(location=location, obstime=t))
+    location = EarthLocation(lat=0*u.deg, lon=0*u.deg)
+    t = Time('J2161')
```
pllim, line 169: Note: It is good practice to move things that are not expected to emit warning outside of the context manager.

astropy/coordinates/tests/test_iau_fullstack.py
```diff
@@ -166,26 +165,19 @@ def test_future_altaz():
...
-        messages_to_find.append("(some) times are outside of range covered by IERS table.")
-
-    messages_found = [False for _ in messages_to_find]
-    for w in found_warnings:
-        if issubclass(w.category, AstropyWarning):
-            for i, message_to_find in enumerate(messages_to_find):
-                if message_to_find in str(w.message):
-                    messages_found[i] = True
-    assert all(messages_found)
+        messages_found = ["(some) times are outside of range covered by IERS "
```
pllim, line 181: @mhvk , I tried to simplify the logic here but you might want to double check.

astropy/coordinates/tests/test_intermediate_transformations.py
```diff
@@ -543,16 +542,12 @@ def test_earth_orientation_table():
     altaz = AltAz(location=location, obstime=t)
     sc = SkyCoord(1*u.deg, 2*u.deg)
     # Default: uses IERS_Auto, which will give a prediction.
-    with catch_warnings() as w:
-        altaz_auto = sc.transform_to(altaz)
-
-    assert len(w) == 0
+    altaz_auto = sc.transform_to(altaz)
```
pllim, line 545: Note: Now that we turn all warnings into errors, except for the ones we explicitly ignore, check for zero warning is no longer necessary.

astropy/io/ascii/tests/test_c_reader.py
```diff
@@ -1075,12 +1074,15 @@ def test_data_out_of_range(parallel, fast_reader, guess):
         elif TRAVIS:
             pytest.xfail("Multiprocessing can sometimes fail on Travis CI")
 
+    test_for_warnings = fast_reader and not parallel
+
     fields = ['10.1E+199', '3.14e+313', '2048e+306', '0.6E-325', '-2.e345']
     values = np.array([1.01e200, np.inf, np.inf, 0.0, -np.inf])
-    with catch_warnings(AstropyWarning) as w:
+    # NOTE: Warning behavior varies for the parameters being passed in.
+    with pytest.warns(None) as w:
```
pllim, line 1082: This test file is a tricky one. `nullcontext` doesn't work and too strict a `warns` catching doesn't work either; must be due to weird interactions between `pyx` and pure Python. 🤷 

At least this way, I *think* I didn't change the original intent of the tests.

pllim, line 1082: Note to self: Maybe I need to try this too.
```suggestion
    with pytest.warns(AstropyWarning) as w:
```

astropy/io/ascii/tests/test_c_reader.py
```diff
@@ -1207,7 +1208,7 @@ def test_int_out_of_range(parallel, guess):
     # Check with leading zeroes to make sure strtol does not read them as octal
     text = f'P M S\n000{imax:d} -0{-imin:d} 00{huge:s}'
     expected = Table([[imax], [imin], ['00' + huge]], names=('P', 'M', 'S'))
-    with catch_warnings(AstropyWarning) as w:
+    with pytest.warns(None) as w:
```
pllim, line 1211:
```suggestion
    with pytest.warns(AstropyWarning) as w:
```

astropy/io/ascii/tests/test_ecsv.py
```diff
@@ -553,31 +554,29 @@ def test_round_trip_user_defined_unit(table_cls, tmpdir):
...
+        ctx = pytest.warns(u.UnitsWarning, match=r"'bandpass_sol_lum' did not parse .*")
+    else:
+        ctx = nullcontext()
+    # Note: The read might also generate ResourceWarning, in addition to UnitsWarning
+    with ctx:
         t2 = table_cls.read(filename)
     assert isinstance(t2['l'].unit, u.UnrecognizedUnit)
     assert str(t2['l'].unit) == 'bandpass_sol_lum'
     if table_cls is QTable:
-        assert len(w) == 1
```
pllim, line 561: For whatever reason, `catch_warnings` did not see the `ResourceWarning` but `pytest.warns` did. 🤷

docs/development/testguide.rst
```diff
@@ -684,44 +684,24 @@ Testing warnings
...
-  from astropy.tests.helper import catch_warnings
-
-  with catch_warnings(MergeConflictWarning) as warning_lines:
-      # Test code which triggers a MergeConflictWarning
-      out = table.vstack([t1, t2, t4], join_type='outer')
-
-      assert warning_lines[0].category == metadata.MergeConflictWarning
-      assert ("In merged column 'a' the 'units' attribute does not match (cm != m)"
-              in str(warning_lines[0].message))
+situations,
```
pllim, line 688:
```suggestion
```

astropy/coordinates/tests/test_distance.py
dhomeier: No need in fact to return it at all if you don't wanna analyse it.

pllim:
p.s. I thought this was going to be 4 hours of work max. LoL... 🙈

astropy/io/ascii/tests/test_c_reader.py
pllim: Nope.

astropy/io/ascii/tests/test_c_reader.py
pllim: Nope.

astropy/io/misc/tests/test_hdf5.py
```diff
@@ -51,23 +49,19 @@ def test_write_nopath(tmpdir):
     test_file = str(tmpdir.join('test.hdf5'))
     t1 = Table()
     t1.add_column(Column(name='a', data=[1, 2, 3]))
-    with pytest.raises(ValueError) as exc:
+    with pytest.warns(UserWarning, match="table path was not set via the path= argument"):
```
pllim, line 52: This test was never run before because of name clash with the one on L57, hence the dramatic looking change.

pllim: They do look similar though. Should I just delete this one?

astropy/io/votable/tests/converter_test.py
```diff
@@ -48,14 +47,13 @@ def test_char_mask():
 
 def test_oversize_unicode():
     config = {'verify': 'exception'}
-    with catch_warnings(exceptions.W46) as w:
+    with pytest.raises(exceptions.W47):
```
pllim, line 50: @tomdonaldson , I don't understand what is going on here but I had to make thing change to make the test pass changing from `catch_warnings` to `pytest`. Does this make sense to you?

astropy/modeling/tests/test_blackbody.py
```diff
@@ -130,16 +130,12 @@ def test_blackbody_exceptions_and_warnings():
     assert exc.value.args[0] == 'Temperature should be positive: -100.0 K'
 
     # Zero wavelength given for conversion to Hz
-    with catch_warnings(AstropyUserWarning) as w:
+    with pytest.warns(AstropyUserWarning, match='invalid'):
         blackbody_nu(0 * u.AA, 5000)
-    assert len(w) == 1
```
pllim, line 135: There are actually more than 1 warning because of the deprecation warning. Since `blackbody.py` itself is deprecated, I am not worried about how many warnings exactly here.

astropy/table/tests/test_table.py
```diff
@@ -2359,36 +2349,34 @@ def test_replace_update_column_via_setitem_warnings_refcount():
...
-            assert len(w) == 1
-            assert "replaced column 'a'" == str(w[0].message)
+        assert len(w) == 1
 
-            # Make sure the warning points back to the user code line
-            assert w[0].lineno == frameinfo.lineno + 1
-            assert w[0].category is table.TableReplaceWarning
-            assert 'test_table' in w[0].filename
+        # Make sure the warning points back to the user code line
+        assert w[0].lineno == frameinfo.lineno + 2
```
pllim, line 2378: I have no idea why it changed from `+1` to `+2`. Maybe because I moved some orders around? Is this okay or do I have to make it `+1` again?

astropy/utils/tests/test_decorators.py
```diff
@@ -406,22 +383,22 @@ def test_deprecated_argument_multi_deprecation():
...
         assert test(x=1, y=2, z=3) == (1, 2, 3)
-        assert len(w) == 3
+    assert len(w) == 3
 
     # Make sure relax is valid for all arguments
-    with catch_warnings(AstropyUserWarning) as w:
+    with pytest.warns(AstropyUserWarning) as w:
         assert test(x=1, y=2, z=3, b=3) == (1, 3, 3)
-        assert len(w) == 1
+    assert len(w) == 4
```
pllim, line 393: Not sure what `catch_warnings` was doing but now the test emits more warnings, which makes more sense because L388 above got 3 for 3, so here is should have been 4 for 4.

astropy/wcs/tests/test_wcs.py
```diff
@@ -30,21 +25,6 @@
 _WCSLIB_VER = LooseVersion(_wcs.__version__)
 
 
-def _check_v71_dateref_warnings(w, nmax=None):
```
pllim, line 33: Given that `wcslib` is bundled, there is no need to check for older versions? With that same reasoning, I also deleted older versions of "validate.txt" files. Please let me know if I misunderstood how the interactions work.

docs/development/testguide.rst
```diff
@@ -684,44 +684,23 @@ Testing warnings
...
-  from astropy.tests.helper import catch_warnings
-
-  with catch_warnings(MergeConflictWarning) as warning_lines:
-      # Test code which triggers a MergeConflictWarning
-      out = table.vstack([t1, t2, t4], join_type='outer')
-
-      assert warning_lines[0].category == metadata.MergeConflictWarning
-      assert ("In merged column 'a' the 'units' attribute does not match (cm != m)"
-              in str(warning_lines[0].message))
-
```
dhomeier, line 703: Should the example be replaced with an analogous one for `pytest.warns`, perhaps demonstrating the `match` option and how to return the warning? `pytest.raises` has its own example, and it would save one from following an external link for a quick overview.

pllim: > Should the example be replaced with an analogous one

I debated with myself on this one... We can barely keep up with doc updates as it is, adding example that you can find on official site anyway seems like we're just giving ourselves more work for no reason. If there is strong opinion of adding an example back, I'll do it.

astropy/wcs/tests/test_wcs.py
nden: We also support using a system version of wcslib. That's why the checks for older versions.

pllim: Ah, okay. I'll revert those changes. Thanks for the clarifications!

mhvk:
@pllim - not quite clear what the status is, but can have a last look if that's help - probably want to get this in before you get conflicts... Wouldn't you also need an entry to CHANGES.rst and deprecate it?

pllim:
> not quite clear what the status is

I need to undo some `wcs` deletion. Also waiting for another opinion on [adding example back](https://github.com/astropy/astropy/pull/10504#discussion_r446560643). Other than that, nothing on the list currently.

> deprecate it?

Deprecation is a controversial topic because this is probably widely used downstream. I'd rather tackle that as a separate PR.

docs/development/testguide.rst
mhvk: I think on balance I agree with @pllim here. Overall, the recommendation might be to generally follow what is in the test files already. But not in this PR.... Might be worth a separate issue to look over the testguide more generally.

mhvk:
Apropos deprecation: good point about people using it downstream. I was wondering mostly as an easy way for us to know that we are regressing. But it may well be that just not having `catch_warnings` already imported, and having `pytest.warns` throughout, will solve most problems.

pllim:
> I was wondering mostly as an easy way for us to know that we are regressing

Yes, it is a risk we have to take until we're ready to issue that deprecation warning. I removed mention of it from dev docs, so hopefully that would reduce the risk but it is not foolproof. :(

pllim:
> not quite clear what the status is

Update: I undid my damage to `wcs` tests. Should be good for review now.

astropy/io/fits/tests/test_core.py
```diff
@@ -782,10 +763,9 @@ def test_open_bzipped_from_handle(self):
     @pytest.mark.skipif(not HAS_BZ2, reason='Python built without bz2 module')
     def test_detect_bzipped(self):
         """Test detection of a bzip2 file when the extension is not .bz2."""
-        with ignore_warnings():
```
pllim, line 785: Note: I don't know were warnings are ignored blindly and turns out they are not necessary anyway.

astropy/io/fits/tests/test_hdulist.py
```diff
@@ -980,30 +972,26 @@ def test_no_resource_warning_raised_on_non_fits_file(self):
...
+                fits.open(f, ignore_missing_end=True)
 
-            with pytest.raises(OSError):
-                fits.open(filename, ignore_missing_end=False)
+            assert not f.closed
 
-            with pytest.raises(OSError):
-                fits.open(filename, ignore_missing_end=True)
+        with pytest.raises(OSError):
+            fits.open(filename, ignore_missing_end=False)
```
pllim, line 991: Note: The diff is a bit misleading. All I did was un-indent after removing `with catch_warnings(ResourceWarning) as ws` because that catch is now unnecessary.

setup.cfg
```diff
@@ -115,6 +115,8 @@ filterwarnings =
     ignore:::astropy.tests.plugins.display
     ignore:::astropy.tests.disable_internet
     ignore:direct construction of Asdf:pytest.PytestDeprecationWarning
+    ignore:Unknown pytest.mark.mpl_image_compare:pytest.PytestUnknownMarkWarning
```
pllim, line 118: Note: These warnings now appear from `wcsaxes` after replacing `catch_warnings` with `pytest.warns`.

mhvk: Is it worth an issue to fix the tests?

astropy/io/votable/tests/converter_test.py
pllim: Interesting... when I ran this manually (see below), both W46 and W47 are emitted but at different lines. I will modify this to reflect the reality that I see. I think when you pass in `catch_warnings(exceptions.W46)`, it only cares about that warning and ignores everything else silently. So I think what the old code was really testing was "make sure there is only one W46" emitted. I am still not sure why I needed `pytest.raises`, I'll dig around more.
```python
>>> field = tree.Field(None, name='c2', datatype='unicodeChar', config=config)                  
WARNING: W47: ?:?:?: W47: Missing arraysize indicates length 1 [astropy.io.votable.converters]
>>> c.parse("XXX")                                                                              
WARNING: W46: ?:?:?: W46: unicodeChar value is too long for specified length of 1 [astropy.io.votable.converters]
('XXX', False)
```

astropy/io/misc/tests/test_yaml.py
```diff
@@ -18,12 +18,11 @@
 from astropy import units as u
 from astropy.time import Time
 from astropy.table import QTable, SerializedColumn
-from astropy.tests.helper import catch_warnings
 from astropy.coordinates.tests.test_representation import representation_equal
 
 yaml = pytest.importorskip('yaml', minversion='3.12')
 
-from astropy.io.misc.yaml import load, load_all, dump
+from astropy.io.misc.yaml import load, load_all, dump  # noqa
```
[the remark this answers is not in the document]
pllim: "module level import not on top of the file"

astropy/modeling/tests/test_blackbody.py
```diff
@@ -53,7 +53,7 @@ def test_fit(self):
         b_fit = fitter(b, wav, fnu)
 
         assert_quantity_allclose(b_fit.temperature, 2840.7438339457754 * u.K)
-        assert_quantity_allclose(b_fit.bolometric_flux, 6.821837075583734e-08 * u.erg / u.cm**2 / u.s)
+        assert_quantity_allclose(b_fit.bolometric_flux, 6.821837075583734e-08 * u.erg / u.cm**2 / u.s)  # noqa
```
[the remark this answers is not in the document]
pllim: Line too long, but maybe I should just fix it...

astropy/modeling/tests/test_constraints.py
```diff
@@ -199,20 +196,24 @@ def test_bounds_gauss2d_slsqp(self):
                                   x_stddev=4., y_stddev=4., theta=0.5,
                                   bounds=bounds)
         gauss_fit = fitting.SLSQPLSQFitter()
-        with ignore_non_integer_warning():
+        # Warning does not appear in all the CI jobs.
```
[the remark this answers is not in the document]
pllim: I'll leave it to the modeling developer to decide whether it should always warn or not warn at all, but I did add a `# TODO: Rewrite the test for more consistent warning behavior.` comment.

astropy/io/misc/tests/test_yaml.py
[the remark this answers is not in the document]
mhvk: Ah, that makes sense and is pretty unavoidable.

setup.cfg
pllim: I don't think so. This warning is because some tests in `wcsaxes` need the `pytest-mpl` plugin but we cannot install it for all jobs because that would also pull in `matplotlib` and we have jobs that only tests with compulsory dependencies but `matplotlib` is optional. Does that make sense?

astropy/modeling/tests/test_constraints.py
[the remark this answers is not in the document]
mhvk: Super!

astropy/io/fits/tests/test_diff.py
```diff
@@ -787,12 +786,10 @@ def test_file_output_overwrite_vs_clobber(self):
         hb['C'] = 4
         diffobj = HeaderDiff(ha, hb)
         diffobj.report(fileobj=outpath)
-        with catch_warnings(AstropyDeprecationWarning) as warning_lines:
+        with pytest.warns(AstropyDeprecationWarning, match=r'"clobber" was '
```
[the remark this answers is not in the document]
pllim: If you mean removing `clobber` from API, I cannot find an issue for it though it was a bit controversial when the deprecation happen. It was truly deprecated in #6203 (v2.0). @saimn , do you think it is safe to finally remove now? Or is this one of those we just keep around deprecated forever. This is FITS, after all.

astropy/io/votable/tests/converter_test.py
```diff
@@ -48,12 +47,11 @@ def test_char_mask():
 
 def test_oversize_unicode():
     config = {'verify': 'exception'}
-    with catch_warnings(exceptions.W46) as w:
+    with pytest.warns(exceptions.W46) as w:
         field = tree.Field(
             None, name='c2', datatype='unicodeChar',
-            config=config)
+            arraysize='1', config=config)
```
pllim, line 53: @mhvk , so `pytest.raises` was a typo on my part. Nice catch! Also I think the more correct solution is to provide `arraysize` to get rid of the W47 warning that this test is not looking for anyway.

pllim:
@mhvk , I think I applied all the code changes you asked for and hopefully answer all your questions/concerns. Please let me know if I missed anything. Thank you so much for doing the mega-review!

setup.cfg
mhvk: Yes, makes sense!

pllim:
As for `pytest.warns(None)` vs `nullcontext`, these two example below should be equivalent:
```python
with pytest.warns(None) as w:
    some_code()
assert len(w) == 0
```
```python
with nullcontext():
    some_code()
```
Therefore, I think `pytest.warns(None)` should be used when you want to blindly catch any warnings and `nullcontext` is for when you don't want any warnings at all.

astropy/io/votable/tests/converter_test.py
mhvk: Ah, that makes sense. Resolved!

mhvk:
Yes, I think this can (and should!) go in, once tests have passed.

And the `nullcontext` vs `None` makes sense.

pllim:
Given the approval, I'll go ahead and merge then before I need to rebase (again). Codecov failure can be ignored. Thank you so much for the review!

astropy/io/fits/tests/test_diff.py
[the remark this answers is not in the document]
saimn: I would vote to keep it forever ;)

PR #10505, opened 2020-06-18
NFW Profile Addition to Modeling Package

PaulHuwe:
### Description

This pull request is to add the NFW halo profile to the modeling package.

NOTE: this does not include the tests yet, as I am anticipating code review of this function to necessitate changes.

### DIFF
--- a/astropy/modeling/physical_models.py
+++ b/astropy/modeling/physical_models.py
@@ -10,11 +10,12 @@ import numpy as np
 
 from astropy import constants as const
 from astropy import units as u
+from astropy import cosmology
 from astropy.utils.exceptions import AstropyUserWarning
 from .core import Fittable1DModel
 from .parameters import Parameter, InputParameterError
 
-__all__ = ["BlackBody", "Drude1D", "Plummer1D"]
+__all__ = ["BlackBody", "Drude1D", "Plummer1D", "NFW"]
 
 
 class BlackBody(Fittable1DModel):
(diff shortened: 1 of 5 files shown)


eteq:
Thanks for this @PaulHuwe !  I haven't done a detailed review, but as an initial first look , a couple big picture questions:

1. Why is this a 2D model rather than a 1D model?  The NFW profile is a radial *density* profile, not a surface brightness profile. It can be *projected* down to 2D, but that would be a different model than this one...
2. What's the reason for the use of all the `self.<something>` constructs in `evaluate` instead of having it inside the `evaluate` function? In general I think of `evaluate` as being a state-less operation (it's often a static or classmethod in other models)?  I'm not necessarily against it, but I think it adds an unnecessary complication if it's not necessary.

eteq:
Oh, and I'm noticing the commit messages just say "NFW v0.x" or similar. In general in Astropy commit messages should provide some contextual information about what is actually in the commit - so @PaulHuwe, at some point we may need to do an interactive rebase/partial squash to clean that up (but perhaps not so critical until the initial review above is addressed).

PaulHuwe:
@eteq Thanks for the replies. In response:

1. I wasn't given a specific use case, so I thought to go with a 2D projection in order to generate and/or fit mass profile maps. If you dislike this use case and/or have another in mind, please let me know.

2. The reason for the self.X constructs is mainly to handle the overdensity type (e.g. 200c, 500m, virial) and cosmology - two things which aren't Parameter objects. These two items only need to be set once per instance, so they are set in the init and used via self (note: @nden pointed out some class/instance confusion in my variables, which I am presently rectifying in the code). If you think this is inappropriate, pease let me know how you think I should manage them.

3. Thank you for the commit notes - both @nden and @pllim have noted similarly. (Detriment of coming from a different workflow.)

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,412 @@ def input_units(self):
...
+
+class NFW1D(Fittable1DModel):
+    """
+    Navarro–Frenk–White (NFW) profile - model for radial distribution of matter.
+
+    Parameters
+    ----------
+    x_0 : float or :class:`~astropy.units.Quantity`
+        Radial position of the peak (center) of the NFW profile.
+    mass : float or class:`~astropy.units.Quantity`
```
jdavies-st, line 385: I think this needs to be
```
:class:`~astropy.units.Quantity`
```

astropy/modeling/physical_models.py
```diff
@@ -10,11 +10,12 @@
...
 from astropy import constants as const
 from astropy import units as u
+from astropy import cosmology
 from astropy.utils.exceptions import AstropyUserWarning
-from .core import Fittable1DModel
+from .core import Fittable1DModel, Fittable2DModel
 from .parameters import Parameter, InputParameterError
 
-__all__ = ["BlackBody", "Drude1D", "Plummer1D"]
+__all__ = ["BlackBody", "Drude1D", "Plummer1D", "NFW2D", "NFW1D"]
```
jdavies-st, line 18: Maybe switch the order to `"NFW1D", "NFW2D"` since that's the order in the module?

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,412 @@ def input_units(self):
...
+        """
+        # Set default cosmology
+        if cosmo is None:
+            cosmo = cosmology.default_cosmology.get()
+
+        # Set mass overdensity type and factor:
+        #   vir : virial radius
+        #   #c  : radius where density is # that of the critical density
+        #   #m  : radius where density is # that of the mean density
+        if massfactor.lower() == "vir":
```
jdavies-st, line 460: Would it be better to have a more descriptive string here?  `NFW1D(massfactor="virial")` would be clearer.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,412 @@ def input_units(self):
...
+            # Virial Mass
+            delta = None
+            masstype = massfactor.lower()
+        elif massfactor[-1].lower() == 'c' or massfactor[-1].lower() == 'm':
+            # Critical or Mean Overdensity Mass
+            delta = massfactor[0:-1]
```
jdavies-st, line 466: `delta` here is going to be a string.  Probably needs to be converted to a float for use below.

jdavies-st: The other way to do this would be to ask for either a `str` or `tuple`.  I.e.

```python
nfw_model = NFW1D(massfactor="virial")
nfw_model = NFW1D(massfactor=("m", 2.2))
nfw_model = NFW1D(massfactor=("c", 4))
```

That might be more intuitive and less prone to user error?

eteq:
@PaulHuwe - to follow on the thread from above

> I wasn't given a specific use case, so I thought to go with a 2D projection in order to generate and/or fit mass profile maps. If you dislike this use case and/or have another in mind, please let me know.

Ah, gotcha.  Yes, I think in this case the 1d/radial version makes much more sense astrophysically, as that's the typical application.  Some of the other modeling machinery can be combined with that to produce the 2d case anyway.  It looks like you've updated already in that direction, so that's good! The one quibble I'd have is that instead of `x`, I'd say use `r` instead, as that's the tradition in the field (and "x" usually means something different as used in this little corner of astrophysics).

> The reason for the self.X constructs is mainly to handle the overdensity type (e.g. 200c, 500m, virial) and cosmology - two things which aren't Parameter objects. These two items only need to be set once per instance, so they are set in the init and used via self (note: @nden pointed out some class/instance confusion in my variables, which I am presently rectifying in the code). If you think this is inappropriate, pease let me know how you think I should manage them.

Thanks for clarifying!  I think that is reasonable now that you've given the reason. It might be good to add something in the code and/or docstrings saying (I didn't see it, but didn't look *too* carefully, so if you've already done it then all good).

eteq:
Addition/clarification to above comment after some more thinking and out-of-band convo with @PaulHuwe : in the current state there's both a `NFW1D` and `NFW2D` which might seem like it makes sense since there are similar constructs in `modeling`, but for NFW I think this is potentially confusing because in the astrophysics applications, there's an ambiguity whether "2d" means "3d density profile projected to 2d" vs "3d density profile value in an on-sky plane.  The latter is basically what is meant in other part of modeling, but the former is what is most common in the astrophysics context.

So my proposal is to drop `NFW2D` and change `NFW1D` to just `NFW`.  Combined with the `x`->`r` change that will make it clear it's the density profile not the projected profile.

For anyone who actually really does want the other interpretation of "2d", the machinery outlined in https://docs.astropy.org/en/stable/modeling/compound-models.html#model-concatenation should be usable to do the "2d-ifying".

And one other thing I'm realizing - I think in the 1D version, there shouldn't be a `x_0`/`r_0`, because there's no physical meaning (in this case) to a 3D radial profile with an offset. Worse yet, `r_0` is sometimes used in the literature as the variableyou *divide* the `r` coordinate by in the NFW profile.  So I'd say just drop that parameter (the model-concatenation machinery can also provide that anyway if there's a corner-case need for it).


I also think if might be nice to add some "convenience" accessors like converting this parametrization into `r_s` or computing circular velocities (which are just conventional re-scalings of this model), but that can be added in a later PR if preferred.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,412 @@ def input_units(self):
...
+            Concentration of the NFW profile.
+        redshift : float
+            Redshift of the NFW profile.
+        massfactor : str
+            Mass overdensity factor and type for provided profiles:
+                vir : virial radius
+                Nc  : radius where density is N times that of the critical density (e.g. '200c')
+                Nm  : radius where density is N times that of the mean density (e.g. '500m')
+        cosmo : class:`~astropy.cosmology`
+            Background cosmology for density calculation.
```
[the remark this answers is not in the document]
PaulHuwe: Done.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,412 @@ def input_units(self):
...
+                 mass=u.Quantity(mass.default, mass.unit),
+                 concentration=concentration.default, redshift=redshift.default, massfactor="200c",
+                 cosmo=None,  **kwargs):
+        """
+        NFW profile Initialization
+
+        Parameters
+        ----------
+        x : float or :class:`~astropy.units.Quantity`
+            Radial position of density to be calculated for the NFW profile.
```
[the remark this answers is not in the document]
PaulHuwe: Done

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,412 @@ def input_units(self):
...
+        mass : float or class:`~astropy.units.Quantity`
+            Mass of NFW peak within specified overdensity radius.
+        concentration : float
+            Concentration of the NFW profile.
+        redshift : float
+            Redshift of the NFW profile.
+
+        Returns
+        -------
+        density : float
```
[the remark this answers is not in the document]
PaulHuwe: Yup - updated.

PaulHuwe:
Note to reviewers: I am adding some exception testing to comply with the 90+% code test requirement.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,334 @@ def input_units(self):
...
+
+        # Calculate scale radius
+        # Set radial units to kiloparsec by default (unit will be rescaled by units of radius
+        # in evaluate)
+        self.radius_s = (((3.0 * in_mass) / (4.0 * np.pi * self.density_delta)) ** (
+                    1.0 / 3.0)) / concentration
+        self.radius_s = self.radius_s.to(u.kpc)
+
+        # Calculate scale density
+        self.density_s = in_mass / (4.0 * np.pi * self.radius_s ** 3 * self.A_NFW(concentration))
```
nden, line 518: Is there a reason to compute `radius_s` and `density_s` in the init function as well as in the `evaluate` function?
Isn't it sufficient to have the computation in the property?
It will make the code simpler.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,334 @@ def input_units(self):
...
+        if not isinstance(mass, u.Quantity):
+            in_mass = u.Quantity(mass, u.M_sun)
+        else:
+            in_mass = mass
+
+        # Create radial version of input with dimension
+        if hasattr(r, "unit"):
+            in_r = r
+            self.radius_s = self.radius_s.to(r.unit)
+            self.density_s = self.density_s.to(in_mass.unit / r.unit**3)
```
nden, line 559: Why are `radius_s` and `denstiy_s` computed here and further down?

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,334 @@ def input_units(self):
...
+        -----
+
+        Model formula:
+
+        .. math:: A_{NFW} = [\ln(1+y) - \frac{y}{1+y}]
+        """
+        return np.log(1.0 + y) - (y / (1.0 + y))
+
+    @property
+    def rho_s(self):
```
nden, line 602: Is this name common among users of this model? To me `scale_density` or `rho_scale` sounds clearer. I'll differ to @eteq on this one.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,334 @@ def input_units(self):
...
+        """
+        Scale density of the NFW profile.
+        """
+        return self.density_s
+
+    @property
+    def r_s(self):
```
nden, line 609: Same comment for the name here - to a non-user like me using the full name sounds better.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,334 @@ def input_units(self):
...
+    References
+    ----------
+    .. [1] https://arxiv.org/pdf/astro-ph/9508025
+    .. [2] https://en.wikipedia.org/wiki/Navarro%E2%80%93Frenk%E2%80%93White_profile
+    .. [3] https://en.wikipedia.org/wiki/Virial_mass
+    """
+
+    # Model Parameters
+
+    # NFW Profile mass and mass type
```
nden, line 423: If you move the comments below the attribute definitions they will appear in the documentation.

astropy/modeling/physical_models.py
PaulHuwe: Two reasons:
1. The virial radius (however a user defines it) is needed for circular velocities, which a user may be interested in acquiring without evaluating densities at arbitrary radii. (i.e. if I remove the radius_s calculation from init, then a user would have to evaluate the profile before solving for circular velocities, when that technically isn't needed).
2. Since init has all the information to calculate the scale radius & density, calculating both within init enables a user to utilize all the NFW accessors without having to evaluate first. It is a convenience to the user.

astropy/modeling/physical_models.py
PaulHuwe: This block is labeled with:
# Update NFW Scale radius
...
# Update NFW Scale density

To highlight that these values are being updated. The reason this update code exists is in the case a user updates the parameters in their evaluate call from what they gave in the init (i.e. like the fitter does). Accordingly, these parameters would need to also be updated.

astropy/modeling/physical_models.py
PaulHuwe: This is the name commonly used in the literature, but I am happy to change it if @eteq so desires.

astropy/modeling/physical_models.py
PaulHuwe: This is the name commonly used in the literature, but I am happy to change it if @eteq so desires.

astropy/modeling/physical_models.py
PaulHuwe: Discussed and resolved offline

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,334 @@ def input_units(self):
...
+    mass : float or :class:`~astropy.units.Quantity`
+        Mass of NFW peak within specified overdensity radius.
+    concentration : float
+        Concentration of the NFW profile.
+    redshift : float
+        Redshift of the NFW profile.
+    massfactor : tuple or str
+        Mass overdensity factor and type for provided profiles:
+            Tuple version:
+                ("virial") : virial radius
```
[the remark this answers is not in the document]
PaulHuwe: Fixed and committed.

astropy/modeling/physical_models.py
PaulHuwe: Private function added as requested.

astropy/modeling/physical_models.py
PaulHuwe: Private function added as requested.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,353 @@ def input_units(self):
...
+            self.density_delta = delta * cosmo.critical_density(redshift) * cosmo.Om(redshift)
+        else:
+            raise ValueError("Invalid masstype '" + str(masstype) +
+                             "'. Should be one of 'virial','c', or 'm'")
+        return self.density_delta
+
+    @staticmethod
+    def A_NFW(y):
+        r"""
+        Commonly used expression in several calculations.
```
eteq, line 571:
```suggestion
        Dimensionless volume integral of the NFW profile, used as an intermediate step in some calculations for this model.
```

astropy/modeling/physical_models.py
eteq: I think for this one I agree with @nden and I like `rho_scale`, because this is not quite as universally used as `r_s`.  But I think it's probably good to add some text like ``often written in the literature as :math:`\rho_s` `` or something like that?

astropy/modeling/physical_models.py
eteq: In this case I think this is good as it stands.  In my exeperience anyone working with the NFW profile has seen `r_s` written that way in math before, and know that that's the scale radius. And especially given that `radius_s` is an alternate I think we're fine here as it stands.

astropy/modeling/physical_models.py
```diff
@@ -371,3 +372,353 @@ def input_units(self):
...
+                ("mean", N)  : radius where density is N times that of the mean density
+
+            String version:
+                "virial" : virial radius
+
+                "Nc"  : radius where density is N times that of the critical density (e.g. "200c")
+
+                "Nm"  : radius where density is N times that of the mean density (e.g. "500m")
+    cosmo : :class:`~astropy.cosmology.Cosmology`
+        Background cosmology for density calculation.
```
eteq, line 405:
```suggestion
        Background cosmology for density calculation. If None, the default cosmology will be used.
```

nden:
Thanks for the review @eteq . 
Both failing tests seem unrelated, perhaps the next commit will clear up the errors.

pllim:
Might need a rebase to clear failures.

PaulHuwe:
Thanks for the reviews. I will put those changes is. We will see about the failures, but I am pessimistic as I have yet to have all the tests pass to this point for some reason.

PaulHuwe:
Even with a rebase (which I can do again).

pllim:
At least for the failures that I see about `distutils` and obscure RTD warnings, pretty sure they are fixed on `master`. 😄

astropy/modeling/physical_models.py
PaulHuwe: Change made.

astropy/modeling/physical_models.py
PaulHuwe: I really like this comment. Change made.

astropy/modeling/physical_models.py
PaulHuwe: Implemented as per Eric's comment.

astropy/modeling/physical_models.py
PaulHuwe: Unchanged as per @eteq's recommendation.

astropy/modeling/physical_models.py
PaulHuwe: Changed.

PaulHuwe:
After rebasing, I see some small formatting issues in docstrings which I will fix (though there is one in a wcs file that my code doesn't touch).

PR #10509, opened 2020-06-19
Update development_workflow.rst

embray:
Fix syntax: this link was written in markdown format instead of an rst format linke

[ci skip]


pllim:
I re-milestoned so we backport to LTS. Thanks for catching the mistake!

pllim:
For future reference, for doc changes, please use `[skip travis]`, as we would still want CircleCI to render the doc and double check that before merging. FYI.

PR #10510, opened 2020-06-20
Replace use of deprecated numpy aliases for python int, float, bool.

mhvk:
I'm actually not sure why our `numpy-dev` builds using `NIGHTLY` are not picking this up -- is anaconda that far behind? -- but in my own packages, tests against `astropy-dev` and `numpy-dev` are now failing because of astropy. So, might as well fix it...

Note that this can be backported safely: `np.int` always was just an alias for `int`, and similarly for `float`, `bool`, `str` and `unicode`.

EDIT: failures now happening also with `NIGHTLY` - and I think I recall it is actually weekly, so that makes sense.

### DIFF
--- a/astropy/io/ascii/core.py
+++ b/astropy/io/ascii/core.py
@@ -902,9 +902,9 @@ def convert_numpy(numpy_type):
     ----------
     numpy_type : numpy data-type
         The numpy type required of an array returned by ``converter``. Must be a
-        valid `numpy type <https://docs.scipy.org/doc/numpy/user/basics.types.html>`_,
-        e.g. int, numpy.uint, numpy.int8, numpy.int64, numpy.float,
-        numpy.float64, numpy.str.
+        valid `numpy type <https://docs.scipy.org/doc/numpy/user/basics.types.html>`_
+        (e.g., numpy.uint, numpy.int8, numpy.int64, numpy.float64) or a python
+        type covered by a numpy type (e.g., int, float, str, bool).
 
     Returns
     -------
--- a/astropy/io/ascii/core.py
+++ b/astropy/io/ascii/core.py
@@ -846,7 +846,7 @@ class BaseData:
         """Replace string values in col.str_vals and set masks"""
         if self.fill_values:
             for col in (col for col in cols if col.fill_values):
-                col.mask = numpy.zeros(len(col.str_vals), dtype=numpy.bool)
+                col.mask = numpy.zeros(len(col.str_vals), dtype=bool)
                 for i, str_val in ((i, x) for i, x in enumerate(col.str_vals)
                                    if x in col.fill_values):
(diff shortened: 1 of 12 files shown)


mhvk:
Failure is just a timeout.

astropy/io/ascii/core.py
```diff
@@ -903,7 +903,7 @@ def convert_numpy(numpy_type):
     numpy_type : numpy data-type
         The numpy type required of an array returned by ``converter``. Must be a
         valid `numpy type <https://docs.scipy.org/doc/numpy/user/basics.types.html>`_,
-        e.g. numpy.int, numpy.uint, numpy.int8, numpy.int64, numpy.float,
+        e.g. int, numpy.uint, numpy.int8, numpy.int64, numpy.float,
         numpy.float64, numpy.str.
```
pllim, line 907: What about `str`? Not mentioned in title but mentioned in the issue body.
```suggestion
        numpy.float64, str.
```

astropy/table/tests/test_row.py
```diff
@@ -340,7 +340,7 @@ def test_uint_indexing():
     """
     t = table.Table([[1., 2., 3.]], names='a')
     assert t['a'][1] == 2.
-    assert t['a'][np.int(1)] == 2.
+    assert t['a'][np.int_(1)] == 2.
```
pllim, line 343: Why not `int` here like above?
```suggestion
    assert t['a'][int(1)] == 2.
```

astropy/io/ascii/core.py
```diff
@@ -903,7 +903,7 @@ def convert_numpy(numpy_type):
     numpy_type : numpy data-type
         The numpy type required of an array returned by ``converter``. Must be a
         valid `numpy type <https://docs.scipy.org/doc/numpy/user/basics.types.html>`_,
-        e.g. numpy.int, numpy.uint, numpy.int8, numpy.int64, numpy.float,
+        e.g. int, numpy.uint, numpy.int8, numpy.int64, numpy.float,
```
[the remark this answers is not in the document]
mhvk: Good point, should have looked at the docstring!  I rephrased it a bit further.

astropy/io/ascii/core.py
mhvk: And indeed `str` as well.

astropy/table/tests/test_row.py
mhvk: Here, I think the goal was really to use a guaranteed numpy int, which is what `np.int_(1)` gives; as written, the test just confirmed that a regular python int works...

mhvk:
OK, rebased and pushed a version without the ci skip.

mhvk:
Obviously should merge only if `-dev` passes!!

pllim:
Yay, all green, even the dev job. Thanks!

PR #10511, opened 2020-06-22
TST: Fix stats doctest failures due to upstream changes

pllim:
### Description

This pull request is to address `stats` doctest suddenly failing, possibly due to new ~numpy 1.19.0 release (not even on conda yet but already on PyPI)~ upstream changes. Could be numpy, scipy, pytest... not sure. I cannot reproduce the new results locally but I also don't have time to clone the env.

p.s. Not sure how these tests passed all these years without `FLOAT_CMP`.

Jobs in initial stage looking for
```
    array([[1.38777878e-17, 7.92174125e-02, 2.15973276e-01, 8.33333042e-01],
           [1.66666958e-01, 4.20782762e-01, 6.17360116e-01, 1.00000000e+00]])
```
while the job with older dependencies wants
```
    array([[0.        , 0.07921741, 0.21597328, 0.83333304],
           [0.16666696, 0.42078276, 0.61736012, 1.        ]])
```
even though the two sets of data pass `np.isclose()` check. 🤷 

UPDATE: Looks like the failure is a manifestation of astropy/pytest-doctestplus#92 .

🤞

### DIFF
--- a/astropy/stats/funcs.py
+++ b/astropy/stats/funcs.py
@@ -216,39 +216,40 @@ def binom_conf_interval(k, n, confidence_level=0.68269, interval='wilson'):
     --------
     Integer inputs return an array with shape (2,):
 
-    >>> binom_conf_interval(4, 5, interval='wilson')
+    >>> binom_conf_interval(4, 5, interval='wilson')  # doctest: +FLOAT_CMP
     array([0.57921724, 0.92078259])
 
     Arrays of arbitrary dimension are supported. The Wilson and Jeffreys
     intervals give similar results, even for small k, N:
 
-    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='wilson')
+    >>> result = binom_conf_interval([0, 1, 2, 5], 5, interval='wilson')
+    >>> print(np.array_repr(result, precision=8, suppress_small=True))  # doctest: +FLOAT_CMP
     array([[0.        , 0.07921741, 0.21597328, 0.83333304],
            [0.16666696, 0.42078276, 0.61736012, 1.        ]])
 
-    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='jeffreys')
+    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='jeffreys')  # doctest: +FLOAT_CMP
     array([[0.        , 0.0842525 , 0.21789949, 0.82788246],
            [0.17211754, 0.42218001, 0.61753691, 1.        ]])
 
-    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='flat')
+    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='flat')  # doctest: +FLOAT_CMP
     array([[0.        , 0.12139799, 0.24309021, 0.73577037],
            [0.26422963, 0.45401727, 0.61535699, 1.        ]])
 
     In contrast, the Wald interval gives poor results for small k, N.
     For k = 0 or k = N, the interval always has zero length.
 
-    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='wald')
+    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='wald')  # doctest: +FLOAT_CMP
     array([[0.        , 0.02111437, 0.18091075, 1.        ],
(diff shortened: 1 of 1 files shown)


pllim:
p.p.s. I milestoned to LTS, assuming the tests would fail there too. If that is not the case, feel free to re-milestone.

pllim:
(Whoops... I guess we'll have to live with duplicate doc jobs in both CircleCI and RTD until #10497 is merged.

astropy/stats/funcs.py
```diff
@@ -216,39 +216,40 @@ def binom_conf_interval(k, n, confidence_level=0.68269, interval='wilson'):
...
-    >>> binom_conf_interval(4, 5, interval='wilson')
+    >>> binom_conf_interval(4, 5, interval='wilson')  # doctest: +FLOAT_CMP
     array([0.57921724, 0.92078259])
 
     Arrays of arbitrary dimension are supported. The Wilson and Jeffreys
     intervals give similar results, even for small k, N:
 
-    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='wilson')
+    >>> result = binom_conf_interval([0, 1, 2, 5], 5, interval='wilson')
+    >>> print(np.array_repr(result, precision=8, suppress_small=True))  # doctest: +FLOAT_CMP
```
pllim, line 226: I don't know how to make this pretty...

mhvk: Sigh. Really this is a bit of a bug in `FLOAT_CMP`, though in fairness I wouldn't know how it can know when small and zero are the same...  But I think we shouldn't go this route -- the disadvantage to the reader of the docstring is larger than the advantage of testing that this works... How about just skipping the 0 in the example (and removing the first entry of each row in the result)?

pllim: Like this?
```python
    >>> binom_conf_interval([0, 1, 2, 5], 5, interval='wilson')  # doctest: +FLOAT_CMP +ELLIPSIS
    array([[..., 0.07921741, 0.21597328, 0.83333304],
           [..., 0.42078276, 0.61736012, 1.        ]])
```

I did try to replace `0` with `...` but that didn't work, but I also didn't try to remove the first entry in the second row.

ayshih:
As you note, this is due to the same issue as astropy/pytest-doctestplus#92.  The workaround I settled on is to wrap the doctests in `testsetup`/`testcleanup` directives, e.g.:
```rst
.. testsetup::
  # The next test is run with fixed-precision printing to ensure no whitespace appears when tested
  >>> import numpy as np
  >>> old_floatmode = np.get_printoptions()['floatmode']
  >>> np.set_printoptions(floatmode='fixed')

>>> binom_conf_interval([0, 1, 2, 5], 5, interval='wilson')
array([[0.00000000, 0.07921741, 0.21597328, 0.83333304],
       [0.16666696, 0.42078276, 0.61736012, 1.00000000]])

# and so forth

.. testcleanup::
  >>> np.set_printoptions(floatmode=old_floatmode)
```

Thus, the print formatting will be hidden in the rendered docs.

astropy/stats/funcs.py
mhvk: I think it would be nicer to remove the 0 in the input and remove both now dotted entries.  Really, `0, 1, 2, 5` was just taken as an example, so one might as well reduce it to `[1, 2, 5]`.

mhvk: Basically, I think that the docs should remain aimed at someone *reading* them, not at testing. So, also completely happy to just add `doctest: +SKIP` and be done with it!!

mhvk:
@ayshih - that is a good trick! That said, this is inside a docstring, so the test setup/teardown will be visible to anyone looking at it with, e.g., `binom_conf_interval?` in ipython, which I think is not good, as it just distracts from the docstring. So, my sense is we either adjust the example so that it doesn't fail or skip it.

ayshih:
Agreed: If it's not too contorted to adjust the example to dodge the whitespace issue, that's the cleanest way to go.

(I wouldn't outright skip the doctest; you might miss a real issue down the line.)

astropy/stats/funcs.py
pllim: Good idea! Probably the best workaround between setting print options and skipping it outright.

pllim:
Tests passed.

PR #10513, opened 2020-06-24
Shorten D2IM and CPDIS keyword comments

mcara:
This PR shortens comment strings for some `D2IM` and `CPDIS` FITS keywords to reduce the number of warning when working with WCSes containing lookup table distortions.

Here is an example of how some comments are currently truncated in FITS headers:
```
D2IM1   = 'AXIS.1: 1' / Axis number of the jth independent variable in a d2im fu
D2IM1   = 'AXIS.2: 2' / Axis number of the jth independent variable in a d2im fu
```

As a result, when running `stwcs.updatewcs.updatewcs()` function on such WCSes produces warnings like these:
```
WARNING: VerifyWarning: Card is too long, comment will be truncated. [astropy.io.fits.card]
```

This PR solves this issue by removing the word `independent` from the comment for "axis number". This PR also modifies the comment for `CPDIS` to contain the distortion function name (`'CPDIS'`) in a similar way to `D2IM` distortion.

CC: @nden


astropy/wcs/wcs.py
```diff
@@ -902,10 +902,10 @@ def write_d2i(num, det2im):
             hdulist[0].header[f'{d_kw}{num:d}.EXTVER'] = (
                 num, 'Version number of WCSDVARR extension')
             hdulist[0].header[f'{d_kw}{num:d}.NAXES'] = (
-                len(det2im.data.shape), 'Number of independent variables in d2im function')
+                len(det2im.data.shape), 'Number of independent variables in D2IM function')
             for i in range(det2im.data.ndim):
                 hdulist[0].header['{}{:d}.AXIS.{:d}'.format(d_kw, num, i + 1)] = (
-                    i + 1, 'Axis number of the jth independent variable in a d2im function')
```
pllim, line 908: I like the reduction of warning, but is "independent" really redundant here?

Also, will changing this trips DMS data verification step when we deliver pipeline products? (This is STScI specific and we can discuss offline if you wish.)

pllim:
I'd rather we not backport to LTS.

mcara:
These are all good questions. With regard to DMS - I have no idea. I assume they could verify that nothing malicious  happened 😄 

With regard to "jth independent variable". Here I assume it is like "jth coordinate (along jth axis)" - probably it can be safely assumed "independent" - see https://docs.astropy.org/en/stable/api/astropy.wcs.DistortionLookupTable.html `(x, y)`.

I wonder if `'jth'` should be replaced with `'1st'` or `'2nd'` in the comments but that is obvious from the keyword name and value. Or even better, remove it completely? Like this:
```
Axis number for (independent) variable in a D2IM function
```

mcara:
@pllim In addition to shortening the comments, this PR also modifies `CPDIS` comments from the generic "distortion function" to "`CPDIS` function" to match the `D2IM` comment and to be more specific what function we are talking about.

pllim:
As for the specific verbiage, I'll leave that to the experts. I think I have wrecked enough havoc here already. Thanks for your patience! 😉

mcara:
@stsci-hack What is your opinion on triggering verification warnings?

stsci-hack:
This change would trigger verification warnings, however, such warnings are triggered fairly often for other changes to the headers and, if expected, can be handled and approved easily.  So, I have no concerns about the impact of this change and support merging this PR.

nden:
@mcara When you get a chance please add a log entry under 4.2.

astropy/wcs/wcs.py
```diff
@@ -894,16 +894,19 @@ def _write_det2im(self, hdulist):
...
+
             hdulist[0].header[f'{dist}{num:d}'] = (
                 'LOOKUP', 'Detector to image correction type')
             hdulist[0].header[f'{d_kw}{num:d}.EXTVER'] = (
                 num, 'Version number of WCSDVARR extension')
             hdulist[0].header[f'{d_kw}{num:d}.NAXES'] = (
-                len(det2im.data.shape), 'Number of independent variables in d2im function')
+                len(det2im.data.shape), 'Number of independent variables in D2IM function')
+
+            jth = {1: '1st', 2: '2nd', 3: '3rd'}.get(num, f'{num}th')
```
nden, line 905: If I'm not mistaken this line needs to go in the loop and use `I=1`, not `num`, otherwise the result is like
```
DP1     = 'AXIS.1: 1' / Axis number of the 1st variable in a CPDIS function     
DP1     = 'AXIS.2: 2' / Axis number of the 1st variable in a CPDIS function   
DP2     = 'AXIS.1: 1' / Axis number of the 2nd variable in a CPDIS function     
DP2     = 'AXIS.2: 2' / Axis number of the 2nd variable in a CPDIS function     


```

astropy/wcs/wcs.py
```diff
@@ -894,16 +894,19 @@ def _write_det2im(self, hdulist):
...
+
             for i in range(det2im.data.ndim):
                 hdulist[0].header['{}{:d}.AXIS.{:d}'.format(d_kw, num, i + 1)] = (
-                    i + 1, 'Axis number of the jth independent variable in a d2im function')
+                    i + 1, 'Axis number of the {jth} variable in a D2IM function')
```
nden, line 909:
```suggestion
                    i + 1, f'Axis number of the {jth} variable in a D2IM function')
```

astropy/wcs/wcs.py
```diff
@@ -1023,12 +1026,14 @@ def write_dist(num, cpdis):
             hdulist[0].header[f'{d_kw}{num:d}.EXTVER'] = (
                 num, 'Version number of WCSDVARR extension')
             hdulist[0].header[f'{d_kw}{num:d}.NAXES'] = (
-                len(cpdis.data.shape), 'Number of independent variables in distortion function')
+                len(cpdis.data.shape), f'Number of independent variables in {dist} function')
+
+            jth = {1: '1st', 2: '2nd', 3: '3rd'}.get(num, f'{num}th')
```
nden, line 1031: This line also needs to go in the loop and use `i+1`.

nden:
After the changes are addressed I suggest writing a test using the file below.

```
f=fits.open('astropy/wcs/tests/data/dist_lookup.fits.gz')
w = wcs.WCS(f[1].header, f)
fw=w.to_fits()
```
Then test the FITS cards are as expected.

astropy/wcs/wcs.py
mcara: Oh, yes. Thanks!

PR #10515, opened 2020-06-24
Removing hardcoded filepaths in tests.

### LINKED ISSUE #10344: Ideally should avoid hardcoded file path in tests

Iamsoto:
### Description

This pull request follows #10344 and is attempting to remove hardcoded filepaths in tests in situations where `get_pkg_data_filename` or `get_pkg_data_fileobj` can be better used instead.

One problem I'm having right now is with` FitsTestCase ` in  **/io/fits/tests/__init__.py**
It carries an attribute **data_dir** that uses a os.path() for the data path. I have removed this attribute in this PR but it's referenced elsewhere. Once on line 733 in **/fits/tests/test_hdulist.py**, here I believe I've written the correct equivalent without using _data_dir. 
And again in 
`def test_path(self, capsys)` in **/fits/tests/test_fitsdiff.py** , there I haven't figured out a correct equivalent. 
Do we think it's beneficial to remove this data_dir attribute?  

Overall, this is the work I've done so far, so feel free to tell me if there's something you want added/removed/changed. 

Fixes #10344

### DIFF
--- a/astropy/io/ascii/tests/test_connect.py
+++ b/astropy/io/ascii/tests/test_connect.py
@@ -6,10 +6,10 @@ import pytest
 from astropy.table import Table, Column
 
 from astropy.table.table_helpers import simple_table
+from astropy.utils.data import get_pkg_data_filename
 
 import numpy as np
 
-ROOT = os.path.abspath(os.path.dirname(__file__))
 
 files = ['data/cds.dat', 'data/ipac.dat', 'data/daophot.dat', 'data/latex1.tex',
          'data/simple_csv.csv']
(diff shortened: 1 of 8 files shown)


pllim:

Iamsoto:
Ok

I forgot this PR existed...

Iamsoto:
Ultimately decided to leave the hardcoded filepath in **/io/fits/tests/init.py**  `test_fitsdiff.test_path()` depends on it and I'm not confident in re-writing that test.
Think I got the rest though.

pllim (approved):
Thank you! RTD and remote data failures are unrelated.

PR #10517, opened 2020-06-26
Add copy=False option to Quantity.to

### LINKED ISSUE #10514: Add copy=False option to Quantity.to
### Description
Currently `Quantity.to` always makes a copy of the data, even if the target unit is the same. It would be useful to call `q.to(u.m, copy=False)` which would only copy if necessary.

### Additional context
While the newish `<<` operator helps to reduce copies, it can lead to mistakes because it does two unrelated things:
1. If the left side is a quantity, it performance unit conversion (like `Quantity.to`, but only copying if necessary);
2. If the left side is some other array-like, it performs Quantity construction (like `*`, but only copying if necessary).

If a function is supposed to take a Quantity and uses `<<` to ensure that it is working in a particular unit (maybe so that future arithmetic operations won't need to do conversions, or so that unit errors are caught early), the user could accidentally pass a plain ndarray (which could be in any units) and instead of failing, it will be assumed that it's in whatever units are being converted to.

Additionally, `<<` is not convenient for working with equivalencies.

bmerry:
### Description

This allows the cost of making a copy to be avoided when there is no
change in units.

Closes #10514.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -74,6 +74,8 @@ astropy.uncertainty
 
 astropy.units
 ^^^^^^^^^^^^^
+- ``Quantity.to`` has gained a ``copy`` option to allow copies to be avoided
+  when the units do not change. [#10517]
 
 astropy.utils
 ^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


bmerry:
I don't follow what codecov is trying to say - it seems to be indicating a decrease in coverage, but the files it refers to are totally unrelated to the change.

mhvk:
@bmerry - yes, there is something not quite right with the setup of codecov - I've started to simply ignore it...

mhvk:
OK, in it goes! Thanks, @bmerry!

PR #10530, opened 2020-07-06
DOC: Unpin numpydoc, add nitpick exception rules

### LINKED ISSUE #10529: MNT: Undo numpydoc pin when we can
This is a follow-up of #10524 . When a new version of something (numpydoc, Sphinx, automodapi, etc) comes out upstream that fixes the `:py:class reference target not found:` warnings, we should remove this pinning:

https://github.com/astropy/astropy/blob/3a8732fd2e99653fca8e766fc72245ca98b240cd/setup.cfg#L76

If even then, some fix to our own docstrings is still needed, that fix should be applied as well.

pllim:
Follow up of #10524 . Fix #10529. Thanks to https://github.com/numpy/numpydoc/issues/275#issuecomment-654418024 .

### DIFF
--- a/astropy/wcs/docstrings.py
+++ b/astropy/wcs/docstrings.py
@@ -824,9 +824,9 @@ coordinate : coordinate pair
 """
 
 get_cdelt = """
-get_cdelt() -> ``double array[naxis]``
+get_cdelt() -> numpy.ndarray
 
-Coordinate increments (``CDELTia``) for each coord axis.
+Coordinate increments (``CDELTia``) for each coord axis as ``double array[naxis]``.
 
 Returns the ``CDELT`` offsets in read-only form.  Unlike the
 `~astropy.wcs.Wcsprm.cdelt` property, this works even when the header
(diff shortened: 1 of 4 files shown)


pllim:
So close....

astropy/wcs/docstrings.py
```diff
@@ -824,9 +824,9 @@ def RA_DEC_ORDER(indent=0):
 """
 
 get_cdelt = """
-get_cdelt() -> ``double array[naxis]``
+get_cdelt() -> numpy.ndarray
```
pllim, line 827: Note: Apparently linking was still attempted despite the double-backticks. This is some C-extension thingy, so I am not going to question it... 🤷

pllim:
RTD passed without numpydoc pinning! I'll let the full CI run too, just in case.

docs/conf.py
```diff
@@ -200,7 +200,16 @@
 # resolve.
 
 nitpicky = True
-nitpick_ignore = []
+# This list is from https://github.com/numpy/numpydoc/issues/275
+nitpick_ignore = [
```
mhvk, line 204: I think we should add them to nitpick_exceptions; I found it needs to be without the quotes:
```
# sphinx doing strange things with dict methods
py:class None.  Remove all items from D.
py:class a set-like object providing a view on D's items
py:class a set-like object providing a view on D's keys
py:class v, remove specified key and return the corresponding value.
py:class None.  Update D from dict/iterable E and F.
py:class an object providing a view on D's values
py:class a shallow copy of D
```

astrofrog:
Thanks! Is there a reason the nitpick exceptions can’t live in the dedicated file?

pllim:
> Is there a reason the nitpick exceptions can’t live in the dedicated file?

No. I wasn't even aware the is a dedicated file. Personally, I like all the settings in one place but I'll move it as requested.

docs/conf.py
pllim: Done. Thanks!

astrofrog:
I am backporting this to v4.1.x to fix some docs failures there

PR #10531, opened 2020-07-06
MNT: Remove test helper dependency from fitscheck

### LINKED ISSUE #10528: fitscheck should not be using astropy.tests.helper.catch_warnings
### Description

`fitscheck` uses `astropy.tests.helper.catch_warnings`:

https://github.com/astropy/astropy/blob/49708febbdaa1d1bf53f96981bf1f1d71295d33d/astropy/io/fits/scripts/fitscheck.py#L134

It should not be using a function from "test helper" module because that makes `pytest` a compulsory dependency whenever this import happens:

https://github.com/astropy/astropy/blob/49708febbdaa1d1bf53f96981bf1f1d71295d33d/astropy/io/fits/scripts/fitscheck.py#L47

### Expected behavior

Any non-test function should not be using stuff in `astropy.tests`.

### Blocker

An attempt to replace it with `warnings.catch_warnings` failed in #10504 , which @saimn explained beautifully at https://github.com/astropy/sphinx-astropy/issues/21#issuecomment-653692416 : 

*Basically Astropy's `catch_warnings` is overwriting the warnings defined in `setup.cfg` for pytest, by calling `treat_deprecations_as_exceptions` in its `__exit__` method. So the global `error` clause in pytest's config was ignored once Astropy's `catch_warnings` was called... When using `warnings.catch_warnings` this does not happen, so the global `error` clause is there, and the checksum verification in `io.fits` is issuing warning when the checksum is wrong, which turns into errors, which break the code's logic...*

pllim:
Fix #10528 . This should remove the dependency of `pytest` while building docs as mentioned in astropy/sphinx-astropy#21 , hopefully.

Basically, this takes the essence of `astropy.tests.helper.catch_warnings` and move that to `fitscheck` so it works as before without needing test helper module. Too hacky?

### DIFF
--- a/astropy/io/fits/scripts/fitscheck.py
+++ b/astropy/io/fits/scripts/fitscheck.py
@@ -39,14 +39,12 @@ Example uses of fitscheck:
 
 """
 
-
 import argparse
 import logging
 import sys
+import warnings
 
-from astropy.tests.helper import catch_warnings
 from astropy.io import fits
-
 
 log = logging.getLogger('fitscheck')
 
@@ -124,6 +122,20 @@ def setup_logging():
     log.addHandler(handler)
 
 
+class _catch_warnings(warnings.catch_warnings):
+    def __init__(self):
+        super().__init__(record=True)
+
+    def __enter__(self):
+        warning_list = super().__enter__()
+        warnings.resetwarnings()
+        warnings.simplefilter('always')
+        return warning_list
+
+    def __exit__(self, type, value, traceback):
+        warnings.resetwarnings()
+
+
 def verify_checksums(filename):
     """
     Prints a message if any HDU in `filename` has a bad checksum or datasum.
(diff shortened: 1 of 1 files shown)


saimn:
This is a bit more tricky than that ... Calling `warnings.resetwarnings()` in the `__exit__` means that anything happening after this will not use the global warning filters, which is a side effect that we want to remove.
But the problem is not with `verify_checksums`, where we could use directly `warnings.catch_warnings`. The problem that now having `error` as a global filter break the updating case, where a file needs first to be verified (raising warnings) before being fixed.
So I was thinking at resetting temporarily the warnings filter in the `update` method. I also have a wip branch but if you want to try it that's fine too ;).

pllim:
@saimn , you have a much better understanding of this than I do, so I'll wait for your PR then. Thank you!

PR #10542, opened 2020-07-08
still use `out` even if out.dtype is not float64

r-xue:
### Description

`model.render(out=)` still creates a new array when out.dtype is not float64 (see below).
In such a situation, models will not be rendered into the existing array.

```python
In [1]: from astropy.modeling import models as apmodels 
   ...: import numpy as np 
   ...: g=apmodels.Gaussian2D(x_mean=20,y_mean=20)                                              

In [2]: out=np.zeros((40,40)) 
   ...: g.render(out) 
   ...: print(out.sum())                                                                        
6.2831853741872

In [3]: out=np.zeros((40,40),dtype=np.float32) 
   ...: g.render(out) 
   ...: print(out.sum())                                                                        
0.0
```

This pull request is supposed to address the issue.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -214,6 +214,9 @@ astropy.io.votable
 astropy.modeling
 ^^^^^^^^^^^^^^^^
 
+- Fixed an issue of ``Model.render`` when the input ``out`` datatype is not 
+  float64. [#10542]
+
 astropy.nddata
 ^^^^^^^^^^^^^^
 
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -1494,7 +1494,7 @@ class Model(metaclass=_ModelMeta):
                 out = np.zeros(coords[0].shape)
 
         if out is not None:
-            out = np.asanyarray(out, dtype=float)
+            out = np.asanyarray(out)
             if out.ndim != ndim:
                 raise ValueError('the array and model must have the same '
                                  'number of dimensions.')
@@ -3256,7 +3256,7 @@ class CompoundModel(Model):
                 out = np.zeros(coords[0].shape)
 
         if out is not None:
-            out = np.asanyarray(out, dtype=float)
+            out = np.asanyarray(out)
             if out.ndim != ndim:
                 raise ValueError('the array and model must have the same '
                                  'number of dimensions.')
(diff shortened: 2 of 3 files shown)


pllim:
Thank you! I'll let the reviewer(s) set the milestone.

* When milestone is set, a change log under "bug fix" needs to be added.
* Perhaps also add a test in the test suite to avoid regression.
* Would be nice if the commits can be squashed.

perrygreenfield:
This seems reasonable, though I think there needs to be a check that the specified output array is a float array (what happens if someone provides a int8 array?)

Also, a test should be added to show that out is in fact being used in this case (and perhaps one that shows that use of the wrong type of output array raises an error).

r-xue:
Added the test and filled an entry in the changelog.

Another unexpected (not very relevant here) behavior I came across in `model.render`: 

I was slightly confused by the definition of `bounding_box`:
For direct model evaluation, `bounding_box` is in the model coordinate system.

```python
In [8]: g=models.Gaussian1D(amplitude=10, mean=0, stddev=.3) 
   ...: print(g.bounding_box)                                                                                                                                                                       
(-1.65, 1.65)

In [9]: x=np.linspace(-5,5,10) 
   ...: y=g(x, with_bounding_box=True) 
   ...: print(y)                                                                                                                                                                                    
[       nan        nan        nan        nan 1.80021743 1.80021743
        nan        nan        nan        nan]
```

However, in `model.render(coords=)`, it's used as pixel indices.

```python
In [10]: y=g.render(coords=x) 
    ...: print(y)                                                                                                                                                                                   
[4.80089224e-60 3.24282237e-36 2.41615094e-18 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00]
```

I guess the function is not supposed to use like this. It will become too complicated to implement `bounding_box` as a box in the model coordinate. Then slicing & add will require an inverse coordinate transform of the one specified by ```coords```. 
But perhaps this "inconsistency" can be mentioned somewhere in the documentation.

pllim:
@perrygreenfield , you sure you don't want it in 4.1.1? (The change log needs to move there too if milestone changes.)

nden:
@pllim Is the Windows error here real? It looks like it's intermittent. If so, this can be merged.

pllim:
@nden , I have never seen that error before. I restarted the job. If it persist, this PR might need a rebase, just in case. But we'll see. 🤞

nden:
@pllim Thanks! Merging now.

PR #10543, opened 2020-07-08
Change Table.columns keys(), values() to return generator

### LINKED ISSUE #10525: Should TableColumns keys and values be generators now?
I stumbled upon this:

https://github.com/astropy/astropy/blob/1f75aa164539949723971d5b767f9517a3d77344/astropy/table/table.py#L285-L290

The comment said they were defined for PY2/PY3 compatibility. We have dropped PY2 since v3.0. Ideally, we should also be able to remove these if they were indeed only for compatibility. However, they revert to PY2 behavior, which makes removal impossible in the PY3 world. Should `keys` and `values` really by generators now?

taldcroft:
### Description

This pull request is to make `Table.columns` (which is a dict subclass) return a generator for `keys()` and `values()` methods, instead of the current Py2-compatible `list`.

Fixes #10525

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -53,10 +53,6 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
-- Change ``Table.columns.keys()`` and ``Table.columns.values()`` to both return
-  generators instead of a list. This matches the behavior for Python ``dict``
-  objects. [#10543]
-
 - Allow initializing a ``Table`` using a list of ``names`` in conjunction with
   a ``dtype`` from a numpy structured array. The list of ``names`` overrides the
   names specified in the ``dtype``. [#10419]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -53,6 +53,10 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
+- Change ``Table.columns.keys()`` and ``Table.columns.values()`` to both return
+  generators instead of a list. This matches the behavior for Python ``dict``
+  objects. [#10543]
+
 - Allow initializing a ``Table`` using a list of ``names`` in conjunction with
   a ``dtype`` from a numpy structured array. The list of ``names`` overrides the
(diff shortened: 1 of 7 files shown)


CHANGES.rst
```diff
@@ -53,6 +53,10 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
+- Change ``Table.columns.keys()`` and ``Table.columns.values()`` to both return
```
pllim, line 56: Is the duplicate entry intentional?

pllim:
Looks like you missed a doctest example:
```
095   # Make a new table with the same number of rows and add columns to original table
096   >>> t2 = Table(np.arange(25).reshape(5, 5), names=('e', 'f', 'g', 'h', 'i'))
097   >>> t.add_columns(t2.columns.values())
UNEXPECTED EXCEPTION: TypeError("'odict_values' object does not support indexing")
  File "/opt/python/3.7.1/lib/python3.7/doctest.py", line 1329, in __run
    compileflags, 1), test.globs)
  File "<doctest modify_table.rst[29]>", line 1, in <module>
  File "/home/travis/build/astropy/astropy/.tox/py37-test/lib/python3.7/site-packages/astropy/table/table.py", line 2086, in add_columns
    self.add_column(cols[ii], index=indexes[ii], name=names[ii],
TypeError: 'odict_values' object does not support indexing
/home/travis/build/astropy/astropy/docs/table/modify_table.rst:97: UnexpectedException
```

CHANGES.rst
taldcroft: Yes.  It's definitely an API change.  I put it in the new features just to give it more visibility, but I can see the point for taking it out of new features. Either way is fine by me.

taldcroft:
I fixed the doctest problem.

pllim:
Technically, there is no new feature introduced, just API change, but I'll defer to @mhvk .

mhvk:
Agreed with @pllim - just API change. Thanks!

Issue #10487, opened 2020-06-12
4.1rc1: test_wtbarr_print fails during test


olebole:
### Description

When I compile and and test 4.1rc1 for Debian (i.e. with an external wcslib; version 7.2), I get a failure during the build time test.

### Expected behavior

All tests pass

### Actual behavior

The following failure appears:

```
______________________________ test_wtbarr_print _______________________________

tab_wcs_2di = WCS Keywords

Number of WCS axes: 2
CTYPE : 'RA---TAB'  'DEC--TAB'  
CRVAL : 1.0  1.0  
CRPIX : 1.0  1.0  
PC1_1 PC1_2  : 1.0  0.0  
PC2_1 PC2_2  : 0.0  1.0  
CDELT : 1.0  1.0  
NAXIS : 150  200
capfd = <_pytest.capture.CaptureFixture object at 0x7f86e08ef190>

    def test_wtbarr_print(tab_wcs_2di, capfd):
        tab_wcs_2di.wcs.wtb[0].print_contents()
        captured = capfd.readouterr()
        s = str(tab_wcs_2di.wcs.wtb[0])
        lines = s.split('\n')
>       assert captured.out == s
E       AssertionError: assert '\n\n\n\n\n\n...: 0x49e43b8\n' == '     i: 1\n  ...: 0x49e43b8\n'
E         - 
E         - 
E         - 
E         - 
E         - 
E         - 
E         - ...
E         
E         ...Full output truncated (39 lines hidden), use '-vv' to show

astropy/wcs/tests/test_wtbarr.py:51: AssertionError
= 1 failed, 14395 passed, 440 skipped, 53 xfailed, 12 xpassed in 307.52 seconds =
E: pybuild pybuild:352: test: plugin distutils failed with: exit code=1: cd '/build/astropy-4.1~rc1/.pybuild/cpython3_3.8/build'; python3.8 -m pytest 
dh_auto_test: error: pybuild --test --test-pytest -i python{version} -p 3.8 returned exit code 13
```

### Steps to Reproduce

Build and test astropy on Debian sid, with  `ASTROPY_USE_SYSTEM_ALL=1`.

### System Details

```
Python 3.8.3 (default, May 14 2020, 11:03:12) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform; print(platform.platform())
Linux-5.4.0-4-amd64-x86_64-with-glibc2.29
>>> import sys; print("Python", sys.version)
[...]

PR #10544, opened 2020-07-08
Fix failing wtbarr *test* on Debian - not root cause

mcara:
Kind of "fixes" https://github.com/astropy/astropy/issues/10487 in the sense it makes test pass but it does not fix the root cause of the extra lines appearing in the "captured" output.

CC: @pllim

### DIFF
--- a/astropy/wcs/tests/test_wtbarr.py
+++ b/astropy/wcs/tests/test_wtbarr.py
@@ -46,9 +46,10 @@ def test_wtbarr_ndim(tab_wcs_2di):
 def test_wtbarr_print(tab_wcs_2di, capfd):
     tab_wcs_2di.wcs.wtb[0].print_contents()
     captured = capfd.readouterr()
+    c = captured.out.lstrip('\n')  # see #10487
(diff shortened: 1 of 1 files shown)


pllim:
cc @olebole and @bnavigator

bnavigator:
I would not call it a fix and leave #10487 open but that is just my opinion. Package maintainers can always just skip the test as a workaround.

pllim:
`wcslib` behavior is kinda out of scope if it is not bundled. Unfortunately, it is also not easy to punt this upstream. As long as the test pass, it is fixed in my "astropy core" book. Of course, `wcs` maintainers can always override my decision.

mcara:
> I would not call it a fix and leave #10487 open but that is just my opinion. Package maintainers can always just skip the test as a workaround.

I agree. I meant it as a fix _to the test_ not to the issue. I did not mean that this fixes the issue of extra lines in #10487

pllim:
#10487 is about the test failing. If it is really about inconsistent wcslib behavior in bundled vs external, the title should be changed and perhaps the body content needs to be updated.

PR #10545, opened 2020-07-08
Remove newline characters from header file 3d_cd.hdr

### LINKED ISSUE #10487: 4.1rc1: test_wtbarr_print fails during test

mcara:
My hope is that this might fix #10487 

EDIT: Close #10544 

CC: @pllim @bnavigator

### DIFF
--- a/astropy/wcs/tests/data/3d_cd.hdr
+++ b/astropy/wcs/tests/data/3d_cd.hdr
@@ -1,16 +1,1 @@
-CD1_2   =            -3.72E-05                                                  
-CD1_3   =                    0                                                  
-CD1_1   =            -4.12E-05                                                  
-CUNIT3  = 'nm      '                                                            
-CUNIT2  = 'deg     '                                                            
-CTYPE1  = 'RA---TAN'                                                            
-NAXIS   =                    3                                                  
-CTYPE3  = 'AWAV    '                                                            
-CD2_1   =            -3.72E-05                                                  
-CTYPE2  = 'DEC--TAN'                                                            
-CD2_3   =                    0                                                  
-CD2_2   =             4.12E-05                                                  
-CUNIT1  = 'deg     '                                                            
-CD3_1   =                    0                                                  
-CD3_2   =                    0                                                  
-CD3_3   =                  0.2                                                  
+CD1_2   =            -3.72E-05                                                  CD1_3   =                    0                                                  CD1_1   =            -4.12E-05                                                  CUNIT3  = 'nm      '                                                            CUNIT2  = 'deg     '                                                            CTYPE1  = 'RA---TAN'                                                            NAXIS   =                    3                                                  CTYPE3  = 'AWAV    '                                                            CD2_1   =            -3.72E-05                                                  CTYPE2  = 'DEC--TAN'                                                            CD2_3   =                    0                                                  CD2_2   =             4.12E-05                                                  CUNIT1  = 'deg     '                                                            CD3_1   =                    0                                                  CD3_2   =                    0                                                  CD3_3   =                  0.2                                                  
--- a/astropy/wcs/tests/test_wcs.py
+++ b/astropy/wcs/tests/test_wcs.py
@@ -791,11 +791,10 @@ def test_printwcs(capsys):
         captured = capsys.readouterr()
         assert 'WCS Keywords' in captured.out
     h = get_pkg_data_contents('data/3d_cd.hdr', encoding='binary')
-    with pytest.warns(AstropyUserWarning):
-        w = wcs.WCS(h)
-        w.printwcs()
-        captured = capsys.readouterr()
-        assert 'WCS Keywords' in captured.out
+    w = wcs.WCS(h)
+    w.printwcs()
+    captured = capsys.readouterr()
+    assert 'WCS Keywords' in captured.out
 
 
 def test_invalid_spherical():


pllim:
So, does this supersede #10544 ?

mcara:
If it works.

bnavigator:
Okay, I tried it out. Yes, the new file lets `Wcsprm()` not print out the newlines. The tests pass.

So, has the "original" 3d_cd.hdr an invalid format with the newlines or is this allowed by the specification?

mcara:
I think this should be merged regardless of #10487

mcara:
@bnavigator Many thanks!

I think so. Headers do not have `'\n'` at the end of the line. Lines are supposed to be 80 character long and one can compute the number of lines as `len(header)/80`.

When I tried to create a header from this `3d_cd.hdr` file, I get:

```
In [110]: hdr = fits.Header.fromfile('/.../astropy/wcs/tests/data/3d_cd.hdr', endcard=False, padding=False)
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:

CD1_3   =                    0                                                  [astropy.io.fits.card]
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:

CD1_1   =            -4.12E-05                                                 [astropy.io.fits.card]
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:

CUNIT3  = 'nm      '                                                          [astropy.io.fits.card]
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:

CUNIT2  = 'deg     '                                                         [astropy.io.fits.card]
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:

CTYPE1  = 'RA---TAN'                                                        [astropy.io.fits.card]
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:

NAXIS   =                    3                                             [astropy.io.fits.card]
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:

CTYPE3  = 'AWAV    '                                                      [astropy.io.fits.card]
```

mcara:
> So, has the "original" 3d_cd.hdr an invalid format with the newlines or is this allowed by the specification?

FITS standard does not allow `'\n'` at the end of the lines. However, in principle, there is nothing preventing `WCVSLIB` or `astropy` being able to deal with this kind of string if they wanted to support non-standard strings. My guess is that `WCSLIB` follows standard because it uses `CFITSIO`.

pllim:
This file wasn't touched for 9 years. It was moved from `pywcs` to `astropy` by Mike D. 🤷‍♀️

pllim:
Also, you can remove the warning catch now. 😄 
```
 Failed: DID NOT WARN. No warnings of type (<class 'astropy.utils.exceptions.AstropyUserWarning'>,)
```

bnavigator:
And it doesn't create a problem with the bundled wcslib. Only with system provided one. Although the source for wcslib 7.3 is 100% identical between the openSUSE package and the cextern/wcslib after #10433 (minus trim_wcslib.sh) directory. Maybe the compiler flags?

mcara:
So, if `'\n'` was left in the header as it was here, it was practically the 81st character and it was shifting every (next) line by a character. This creates a mess in the header. For example, with the header created above in https://github.com/astropy/astropy/pull/10545#issuecomment-655723185 if you tried to access keywords defined in `3d_cd.hdr`, you would get:
```
In [121]: list(hdr.keys())
Out[121]:
['CD1_2',
 'CD1_3',
 'CD1_1',
 'CUNIT',
 'CUNI',
 'CTY',
 'NA',
 'C',
 '',
 '',
 '',
 '',
 '',
 '',
 '',
 '',
 '']

In [122]: print(hdr['CD1_2'])
     ...: print(hdr['CD1_3'])
     ...: print(hdr['CD1_1'])
     ...: print(hdr['CUNIT1'])
-3.72e-05
 =                    0
  =            -4.12E-05
<ipython-input-122-e9219e73fae0> in <module>
      2 print(hdr['CD1_3'])
      3 print(hdr['CD1_1'])
----> 4 print(hdr['CUNIT1'])
KeyError: "Keyword 'CUNIT1' not found."
```

bnavigator:
What about sub-segfault.hdr?
https://github.com/astropy/astropy/blob/9afd551f1246da2eaf9a63aa0500f96aa298434a/astropy/wcs/tests/data/sub-segfault.hdr#L1-L28

mcara:
@bnavigator That one works fine because header is created using `fits.Header.fromtextfile` as opposite to `fits.Header.fromfile`

pllim:
I tried digging a bit with help from @jhunkeler but didn't any new info. I think the file has been "wrong" since the beginning.
```
plnott> svn log 3d_cd.hdr
------------------------------------------------------------------------
r2176 | stsci_embray | 2011-07-01 18:57:30 -0400 (Fri, 01 Jul 2011) | 1 line

Merging the setup_refactoring branch into trunk.  This should only affect pysynphot, pywcs, stsci_python, and coords at this point
------------------------------------------------------------------------
r2135 | stsci_embray | 2011-06-17 15:36:39 -0400 (Fri, 17 Jun 2011) | 1 line

Merging up to r2127 from trunk.  This brings the branch up to date with the current trunk.
------------------------------------------------------------------------
r2091 | stsci_mdroe | 2011-05-27 10:09:56 -0400 (Fri, 27 May 2011) | 1 line

Adding missing file
------------------------------------------------------------------------

```

nden:
I agree this is the solution. The error was hidden by the `with` statement. Removing it shows the file does not create a valid header. Thanks all for figuring it out!

PR #10546, opened 2020-07-09
Fix a forgotten change in requirement checking in __init__.py

mhvk:
Without it, I got a message that with pyerfa not installed, it wanted
version 1.16 -- which won't exist for a while yet.

Fixes small mistake introduced in #10329 - just goes to show that anything that is not tested may break...

p.s. Note that this fix only affects the error message.

### DIFF
--- a/astropy/__init__.py
+++ b/astropy/__init__.py
@@ -76,7 +76,7 @@ else:
 
 def _check_requirement(name, minimum_version):
     """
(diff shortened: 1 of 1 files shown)


mhvk:
OK, adjusted the docstring (should've caught that as well; part of the trouble with giant PRs...). Unfortunately, I thought I would just amend the commit, which means all the tests run again...

PR #10549, opened 2020-07-09
Stop using astropy's catch_warnings in fitscheck

### LINKED ISSUE #10528: fitscheck should not be using astropy.tests.helper.catch_warnings

saimn:
Fix #10528.

### DIFF
--- a/astropy/io/fits/scripts/fitscheck.py
+++ b/astropy/io/fits/scripts/fitscheck.py
@@ -106,6 +106,8 @@ def handle_options(args):
 
     if OPTIONS.checksum_kind == 'none':
         OPTIONS.checksum_kind = False
+    elif OPTIONS.checksum_kind == 'standard':
+        OPTIONS.checksum_kind = True
     elif OPTIONS.checksum_kind == 'remove':
         OPTIONS.write_file = True
         OPTIONS.force = True
--- a/astropy/io/fits/scripts/fitscheck.py
+++ b/astropy/io/fits/scripts/fitscheck.py
@@ -114,6 +114,8 @@ def handle_options(args):
 
 
 def setup_logging():
+    log.handlers.clear()
+
     if OPTIONS.verbose:
         log.setLevel(logging.INFO)
     else:
--- a/astropy/io/fits/scripts/fitscheck.py
+++ b/astropy/io/fits/scripts/fitscheck.py
@@ -40,13 +40,12 @@ Example uses of fitscheck:
 """
 
 
-import argparse
-import logging
 import sys
+import logging
+import argparse
+import warnings
 
-from astropy.tests.helper import catch_warnings
 from astropy.io import fits
-
 
 log = logging.getLogger('fitscheck')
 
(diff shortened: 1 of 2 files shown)


astropy/io/fits/scripts/fitscheck.py
```diff
@@ -106,6 +105,8 @@ def handle_options(args):
 
     if OPTIONS.checksum_kind == 'none':
         OPTIONS.checksum_kind = False
+    elif OPTIONS.checksum_kind == 'standard':
```
pllim, line 108: This doesn't look related.

saimn:
> My FITS-fu is not high enough to know how this will change actual fitscheck behavior outside of CI but I'll just trust the CI. grimacing

It shouldn't change anything, just making sure that warnings are not turned into exceptions.

> My only nitpick is that the import order was alphabetical in fitscheck.py and there is no reason to change it? wink

That's the way isort is configured on the repo ;)

astropy/io/fits/scripts/fitscheck.py
saimn: Not really, just making sure that the option value makes sense, 'standard' evaluates as True but it has no special meaning and `True` should be used here.

pllim:
There's a remote failure but doesn't seem related.
```
        with iers.earth_orientation_table.set(iers.IERS_B.open()):
>           with pytest.warns(AstropyWarning, match='after IERS data'):
E           NameError: name 'AstropyWarning' is not defined
../../.tox/py37-test-devdeps/lib/python3.7/site-packages/astropy/coordinates/tests/test_intermediate_transformations.py:536: NameError
```

pllim:
I approved it but I'll leave it for someone else to merge in case more pairs of eyes are needed.

PR #10550, opened 2020-07-09
TST: Fix missing import in test_intermediate_transformations.py

pllim:
### Description

This pull request is to address remote data test failure due to missing import.


pllim:
Hmm, the import was there in #10474 but then somehow removed in #10329 , which was merged later.

mhvk:
p.s. I vaguely recall removing the import after clearing up an `ErfaWarning` - and actually checking that it really wasn't there. Perhaps it was a problem in the rebase?  In any case, I should have checked the remote data run!

PR #10556, opened 2020-07-11
Make angle parser and lexer thread-local

### LINKED ISSUE #7168: Threadsafe angle parsing
The current angle parsing code is not thread safe with string inputs. This can be reproduced with:
```python
import astropy.coordinates, threading
angles = ['00:00:00']*10000
def parse_test(i=0):
   try:
      print(i, len(astropy.coordinates.Angle(angles, unit='hour')))
   except:
      print('{0} failed.'.format(i))
for i in range(10):
    threading.Thread(target=parse_test, args=(i,)).start()
```
Many of the threads should fail due to parsing errors.

The issue appears to be due to the reliance on the external ply.yacc for parsing. From the yacc header:
```python
# The current implementation is only somewhat object-oriented. The
# LR parser itself is defined in terms of an object (which allows multiple
# parsers to co-exist).  However, most of the variables used during table
# construction are defined in terms of global variables.  Users shouldn't
# notice unless they are trying to define multiple parsers at the same
# time using threads (in which case they should have their head examined).
```
yacc is also used in several places in astropy.units but I have not been able to find a reproducible failure case--I have encountered a rare error with Time parsing where a bad initialization appears to cause a failure in all time parses until Python is restarted.

Changing `_AngleParser.__init__` and `_AngleParser.parse` to use an object-level `threading.RLock` mitigates the failure:
```python
def __init__(self):
    import threading
    self.__lock = threading.RLock()
    if '_parser' not in _AngleParser.__dict__:
        _AngleParser._parser, _AngleParser._lexer = self._make_parser()

def parse(self, angle, unit, debug=False):
    try:
        with self.__lock:
            found_angle, found_unit = self._parser.parse(
                   angle, lexer=self._lexer, debug=debug)
    except ValueError as e:
        if str(e):
            raise ValueError("{0} in angle {1!r}".format(
                str(e), angle))
        else:
            raise ValueError(
                "Syntax error parsing angle {0!r}".format(angle))

    if unit is None and found_unit is None:
        raise u.UnitsError("No unit specified")

    return found_angle, found_unit
```
The description in the yacc header suggests this is not a complete solution.

bnikolic:
Should fix #7168

### Description

Fixes #<Issue Number>

### DIFF
--- a/astropy/coordinates/angle_utilities.py
+++ b/astropy/coordinates/angle_utilities.py
@@ -18,6 +18,7 @@ of data to another.
 """
 
 import os
+import threading
 from warnings import warn
 
 import numpy as np
@@ -56,6 +57,10 @@ class _AngleParser:
     This class should not be used directly.  Use `parse_angle`
     instead.
     """
+    # For safe multi-threaded operation all class (but not instance)
+    # members that carry state should be thread-local. They are stored
+    # in the following class member
+    _thread_local = threading.local()
 
     def __init__(self):
         # TODO: in principle, the parser should be invalidated if we change unit
@@ -65,8 +70,9 @@ class _AngleParser:
         # generate the parser for each release (as done for unit formats).
         # For some discussion of this problem, see
         # https://github.com/astropy/astropy/issues/5350#issuecomment-248770151
-        if '_parser' not in _AngleParser.__dict__:
-            _AngleParser._parser, _AngleParser._lexer = self._make_parser()
+        if '_parser' not in _AngleParser._thread_local.__dict__:
+            (_AngleParser._thread_local._parser,
+             _AngleParser._thread_local._lexer) = self._make_parser()
 
     @classmethod
     def _get_simple_unit_names(cls):
@@ -334,8 +340,8 @@ class _AngleParser:
 
     def parse(self, angle, unit, debug=False):
         try:
-            found_angle, found_unit = self._parser.parse(
-                angle, lexer=self._lexer, debug=debug)
+            found_angle, found_unit = self._thread_local._parser.parse(
+                angle, lexer=self._thread_local._lexer, debug=debug)
         except ValueError as e:
             if str(e):
                 raise ValueError("{} in angle {!r}".format(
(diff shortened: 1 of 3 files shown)


astropy/coordinates/angle_utilities.py
```diff
@@ -56,6 +57,10 @@ class _AngleParser:
     This class should not be used directly.  Use `parse_angle`
     instead.
     """
+    # For safe multi-threaded operation all class (but not instance)
+    # members that carry state should be thread-local. They are stored
+    # in the following class member
+    _ThLoc=threading.local()
```
mhvk, line 63: I'd just call this `_thread_local`.

mhvk:
@eteq - requesting you to review as while my own sense is that this is a very nice improvement, I feel a bit out of my league and you were part of the discussion in #7168 and seemed to have thought a bit more about threads than I have.

bnikolic:
I've pushed changes as requested.

astropy/coordinates/angle_utilities.py
```diff
@@ -56,6 +57,10 @@ class _AngleParser:
     This class should not be used directly.  Use `parse_angle`
     instead.
     """
+    # For safe multi-threaded operation all class (but not instance)
+    # members that carry state should be thread-local. They are stored
+    # in the following class member
+    _thread_local=threading.local()
```
mhvk, line 63: Super nitpicky, but spaces around ` = `

astropy/coordinates/tests/test_angles.py
```diff
@@ -1009,3 +1009,13 @@ def test_longitude_nan():
 def test_latitude_nan():
     # Check that passing a NaN to Latitude doesn't raise a warning
     Latitude([0, np.nan, 1] * u.deg)
+
+
+def test_angle_multithreading():
+    import  threading
```
mhvk, line 1015: Could you add a comment with a link back to the issue here? For future selves, it would be particularly useful to know that the loop over multiple threads would fail if the angle parser was not thread-save.

mhvk: Also, just move the `import threading` to the top (and one space, not two)

bnikolic:
OK done (hopefully I've done the squashing etc right)

PR #10557, opened 2020-07-11
Convert 'ctype_name' to uppercase before appending to 'types'

kakirastern:
### Description

This pull request is to address the problem of some error(s) in wcs auto-linking that could arise if the input `CTYPEn` is not in uppercase. During some work on the `glue-solar` project for GSoC 2020, it was found that the `world_axis_physical_types` of a `wcs` object would not pick up a keyword that is not in all uppercase, which in our case was the `CTYPE3 = 'Time    '` for an IRIS SJI data cube. This would not be picked up by `wcs.world_axis_physical_types` as a valid entry as all and would output a NoneType instead. But after we have edited the corresponding `FITS` header to read `CTYPE3 = 'TIME    '` in lieu of the original, we would get the `wcs.world_axis_physical_types` as expected. 

So in case of situations such as these, we (me at the suggestion of @Cadair) are proposing pre-processing at the `astropy`-level to turn the CTYPE into uppercase before further processing downstream. Any discussion regarding this matter is welcome.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -353,6 +353,9 @@ astropy.visualization
 astropy.wcs
 ^^^^^^^^^^^
 
+- Add .upper() to ctype or ctype names to wcsapi/fitwcs.py to mitigate bugs from
+  unintended lower/upper case issues [#10557]
+
 Other Changes and Additions
 ---------------------------
 
--- a/astropy/wcs/wcsapi/fitswcs.py
+++ b/astropy/wcs/wcsapi/fitswcs.py
@@ -250,7 +250,7 @@ class FITSWCSAPIMixin(BaseLowLevelWCS, HighLevelWCSMixin):
         types = []
         # TODO: need to support e.g. TT(TAI)
         for ctype in self.wcs.ctype:
-            if ctype.startswith(('UT(', 'TT(')):
+            if ctype.upper().startswith(('UT(', 'TT(')):
                 types.append('time')
             else:
                 ctype_name = ctype.split('-')[0]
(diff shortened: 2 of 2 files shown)


pllim:
Thanks! If this proceeds, please add change log and unit test.

kakirastern:
@pllim Sure, I will do so as soon as it is confirmed to be a bug.

Cadair:
:+1: This makes sense to me to do these comparisons type-independently, even though the standard says uppercase.

nden:
Strictly speaking this is a data issue but it's not unreasonable to fix it in the code. If there are other similar issues in this file we could fix them now. For example,[ this](https://github.com/astropy/astropy/blob/12b0295df37abe761ba07e3ca6a6fbc061ab78f7/astropy/wcs/wcsapi/fitswcs.py#L253) may be a candidate. @kakirastern Do you have time to look over all uses of ctype in the file and see if there are cases which can be problematic?

kakirastern:
@nden Sure, I will look over all uses of `CTYPE` in the file and watch out for any problematic cases, then report back soon.

kakirastern:
I did a systematic search of the file and found one other spot at https://github.com/astropy/astropy/blob/12b0295df37abe761ba07e3ca6a6fbc061ab78f7/astropy/wcs/wcsapi/fitswcs.py#L398
that could potentially be problematic due to lowercase/uppercase issues.

kakirastern:
So we have two more snippets besides the original one that might need patching, including the one involving the following line:

https://github.com/astropy/astropy/blob/12b0295df37abe761ba07e3ca6a6fbc061ab78f7/astropy/wcs/wcsapi/fitswcs.py#L253

nden:
@kakirastern If you can fix those two cases and add a change log then this is good to go.

kakirastern:
@nden Yup, sure. Will add to this PR to fix those "bugs" as well, and get back within the next few days.

kakirastern:
@nden I have patched the three spots which might be "buggy."

kakirastern:
My pleasure @nden!

Issue #10241, opened 2020-04-30
Provide an easy way to generate zero/random coordinates


astrofrog:
As discussed in https://github.com/astropy/astropy/pull/10185#discussion_r416878148, it would be nice to avoid repetitive code such as

```python
observer = ICRS(ra=0 * u.degree, dec=0 * u.degree,
                pm_ra_cosdec=0 * u.mas/u.yr, pm_dec=0 * u.mas/u.yr,
                distance=0 * u.pc, radial_velocity=0 * u.km/u.s)
```

in tests (or even implementations). Having a way to do ``ICRS.zeros`` or ``np.zeros(ICRS)`` would be convenient, and similarly for random values.

taldcroft:
This feels more like something that should be a test helpers function.  It isn't obvious how this would benefit normal users. The implementation in #10559 is very limited and I don't think it should be on the `ICRS` class.

nstarman:
I recall reading discussion of greater mutability being added to the `BaseCoordinateFrames`. If this is ever implemented than pre-allocating memory before a loop would be quite nice. I certainly use `np.zeros` often to avoid `list.append` over a long array.

taldcroft:
@nstarman - there is https://docs.astropy.org/en/stable/coordinates/inplace.html that can help with pre-allocation for performance-critical code.

nstarman:
That's great. Thanks.

mhvk:
Happened to note #10559 again which brought me here: I think this is indeed best done via a helper function, not a method, being fairly rare.

Thought I'd note that it could also be done by overriding a numpy function rather than adding a new one (via `__array_function__`; see #8610 for the general issue). In particular, one could override `np.zeros_like`. This would effectively do the same as:
```
In [28]: sc = SkyCoord([10, 20], [30, 40], unit='deg')

In [29]: sc._apply(np.zeros_like, shape=())
Out[29]: 
<SkyCoord (ICRS): (ra, dec) in deg
    (0., 0.)>
```
I think this may be closer to what @astrofrog had in mind, since generally one would want a `SkyCoord` similar to another one.

Though note that doing exactly as suggested above could lead to surprises: *any* array-valued attributes on the `SkyCoord` would be zeroed! So one would take some care if one actually defined an override in `SkyCoord.__array_function__`.

adrn:
[...]

PR #10559, opened 2020-07-12
Added zeros class method in ICRS [#10241]

DoctorCronut:
This pull request is to address #10241. I added a class method to the ICRS doc frame
which will eliminate the need to write repetitive code if the 6 arguments used are all
zeros. I have not started on implementing the ICRS random class method as I don't
know the range of values expected as arguments in ICRS. 

Contributes to #10241

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,10 @@ astropy.convolution
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
 
+- Allow user to call ``ICRS.zeros()`` to construct ``ICRS`` class with arguments
+  ``ra``, ``dec``, ``pm_ra_cosdec``, ``pm_dec``, ``distance``, and
+  ``radial_velocity`` equal to 0. [#10559]
+
 - Numpy functions that broadcast, change shape, or index (like ``np.broadcast_to``,
   ``np.rot90``, or ``np.roll``) now work on coordinates, frames, and
   representations. [#10337]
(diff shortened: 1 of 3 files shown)


pllim:
Hello and thanks! Needs the following:

* Add docstring.
* Add change log.
* Add unit test.
* Fix CI.

nstarman:
> I have not started on implementing the ICRS random class method as I don't
> know the range of values expected as arguments in ICRS.

The  ICRS frame expects
Spherical: `ra` [0, 360), `dec` [-90, 90], `distance` [0, oo).
Cartesian: `x, y, z` (-oo, oo)
Cylindrical : `rho` [0, oo), `z` (-oo, oo), `phi` (-oo, oo)
    Note that this last one might be a bug and should probably be implemented as [0, 360)

StuartLittlefair (changes requested):
Hi @DoctorCronut 

Since this is milestoned for 4.2 I had a look at the code. What you've added is certainly simple and meets the immediate need raised in the issue.

However, I think it would be a better approach to fix the underlying problem. ICRS is only one frame class of many and I think the better approach would be to implement a ```zeros``` and perhaps a ```ones``` method for all frames. This would involve changing the base classes such as BaseRADecFrame and friends.

Since 4.2 feature freeze is coming up, and this is not an urgent fix, I am going to re-milestone as 4.3

eteq:
(Sorry, this PR should have expired due to lack of activity, but the bot is glitchy right now.  So BEEP BOOP Erik-bot closes this PR)

PR #10563, opened 2020-07-13
DOC: Fix linkcheck ref target not found

pllim:
For some reason, this only appears in the link check (cron) job. 🤷 

```
Warning, treated as error:
astropy/timeseries/__init__.py:docstring of astropy.timeseries.BoxLeastSquaresResults.popitem::py:class reference target not found: (k, v), remove and return some (key, value) pair as a
ERROR: InvocationError for command .../sphinx-build -W -b linkcheck . _build/html (exited with code 2)
```

Example log: https://travis-ci.org/github/astropy/astropy/jobs/707690795

### DIFF
--- a/docs/nitpick-exceptions
+++ b/docs/nitpick-exceptions
@@ -81,7 +81,6 @@ py:obj distutils.version.LooseVersion
 py:obj pkg_resources.parse_version
 py:class pandas.DataFrame
 
-
 # Pending on python docs links issue #11975
 py:class list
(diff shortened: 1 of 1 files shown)


pllim:
Thanks for the review! I'll keep an eye on the next cron run. 🤞

pllim:
Should have forced that job to run on CI first. This PR didn't fix the problem. My bad.

Issue #10568, opened 2020-07-14
BUG: `constants.pc` definition does not use small angle approximation


dhomeier:
### Description
As reported on the [mailing list](https://mail.python.org/pipermail/astropy/2020-July/004772.html) the definition of the parsec in https://github.com/astropy/astropy/blob/28349153fcc25a9ca01bb5e563c86177c50aed1d/astropy/constants/iau2015.py#L29
uses the exact distance for a 1"-parallax object `c.au / np.tan(u.arcsec.to('radian'))` rather than the small-angle approximation `c.au / u.arcsec.to('radian')`. However the [IAU 2015 resolution B2](https://www.iau.org/static/resolutions/IAU2015_English.pdf) states (footnote 4) that 
> The parsec is defined as exactly (648 000/π) au (e.g. Cox 2000, Binney & Tremaine 2008).

([small-angle approximation](https://en.wikipedia.org/wiki/Parsec#cite_note-11)).
### Expected behavior
```
from astropy import constants as apc
from astropy import units as apu
print(apc.au.value * 180 * 3600 / np.pi)
3.085677581491367e+16 
```
### Actual behavior
```
print(apc.pc.value)
print(apc.au.value / np.tan(np.pi / (180 * 3600)))
3.0856775814671916e+16
3.0856775814671916e+16
```

Fixing the value in `iau2015.py` is trivial; perhaps to be discussed if the change should also be copied into `iau2012.py` (which implicitly certainly assumed the same definition, but did not state this).
Also `equivalencies=u.parallax()` will probably need to be adapted as well.

pllim:
cc @stargaser

stargaser:
Thank you @dhomeier for putting in this issue and replying to the email. You beat me to it!

dhomeier:
Do you want to start a PR @stargaser ? As indicated I have not really looked into what to do about the equivalency; it might be straightforward as well...

pllim:
Hopefully the tests will show us that the results will not change in a way that `FLOAT_CMP` or `allclose` cares, otherwise we might be in trouble. 😅 🤞

stargaser:
I probably won't get to a PR for a few hours or maybe a couple of days...if someone else wants to start one before I get to it.

dhomeier:
> Hopefully the tests will show us that the results will not change in a way that `FLOAT_CMP` or `allclose` cares

[...]

PR #10569, opened 2020-07-14
Change parsec definition to small-angle (linear) approximation following IAU 2015 B2

dhomeier:
Addresses #10568 – this replaces the `np.tan` in the definition of `constants.iau2015.pc` with the linear angle, making 1 pc exactly `(648 000/np.pi) constants.au` as noted in the IAU resolution.

This commit only changes `iau2015.py`, it remains open if `iau2012.py` should adopt the same definition. If not, an additional test in `test_prior_version.py` might check for the differences between the 2.
So far the `u.arcminute.to(u.au, 1, u.parallax())` check in `test_parallax` detects the difference with tightened `rtol`.

Lastly for discussion, the `parallax_converter` in `equivalencies` already uses the linear form; one might argue that this should in fact use the trigonometric function for really large parallaxes(as in the `u.arcminute` test), but since the equivalency is defined commutative in angle and distance, there is no direct way to implement this – and maybe would go beyond the concept of an equivalency.
EDIT: I have worked out how to use `np.tan` and `np.arctan`, respectively, in `parallax_converter` when used as the `forward` or `backward` converter function, so this _could_ be implemented, but would of course change results. Might be better to implement this through an optional argument to `u.parallax`, if at all.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1009,6 +1009,8 @@ astropy.uncertainty
 astropy.units
 ^^^^^^^^^^^^^
 
+- Corrected definition of parsec to 648 000 / pi AU following IAU 2015 B2 [#10569]
+
 astropy.utils
 ^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -857,6 +857,8 @@ astropy.config
 astropy.constants
 ^^^^^^^^^^^^^^^^^
 
+- Corrected definition of parsec to 648 000 / pi AU following IAU 2015 B2 [#10569]
+
 astropy.convolution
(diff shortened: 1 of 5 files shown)


mhvk:
p.s. I'll leave this open for now so that @stargaser can have a look too.

dhomeier:
Agreed, I had lost track a bit what is defined in `constants` and what in `units` myself...
I have added a comparison to the 2012 version to the tests in `constants` now.

dhomeier:
I've given up trying to understand what codecov is complaining about. Not only is it flagging a file that has changed neither here nor in master during the entire process, it points to some code that has not been touched in 9 years!

PR #10570, opened 2020-07-15
Make finding home directory more general

dstansby:
In sunpy we are seeing CI failures due to astropy not being able to find the home directory on macos: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=8869&view=logs&jobId=6e7bd578-a810-548b-8fbf-8fa6b1a2d4f6&j=8b26c057-390d-57e7-9a94-ed463b57befe&t=7cf2f17d-5054-541f-491b-2e6078562aef

This seems to be because the 'HOME' environment variable isn't set. This PR makes finding home more lenient by trying to use `os.path.expanduser('~')` as a backup if the 'HOME' environment variable isn't set. This is in line with what IPython now do, which according to the comments this function is based on
https://github.com/ipython/ipython/blob/f8c9ea7db42d9830f16318a4ceca0ac1c3688697/IPython/utils/path.py#L190

### DIFF
--- a/astropy/config/paths.py
+++ b/astropy/config/paths.py
@@ -33,8 +33,11 @@ def _find_home():
         if 'HOME' in os.environ:
             homedir = os.environ['HOME']
         else:
-            raise OSError('Could not find unix home directory to search for '
-                          'astropy config dir')
+            try:
+                return os.path.expanduser('~')
+            except Exception as e:
+                raise OSError('Could not find unix home directory to search for '
+                              'astropy config dir')
     elif os.name == 'nt':  # This is for all modern Windows (NT or after)
         if 'MSYSTEM' in os.environ and os.environ.get('HOME'):
             # Likely using an msys shell; use whatever it is using for its
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -915,6 +915,9 @@ Bug fixes
 astropy.config
 ^^^^^^^^^^^^^^
 
+- Added an extra fallback to ``os.expanduser('~')`` when trying to find the
+  user home directory. [#10570]
+
 astropy.constants
 ^^^^^^^^^^^^^^^^^
 


astropy/config/paths.py
```diff
@@ -33,8 +33,11 @@ def _find_home():
         if 'HOME' in os.environ:
             homedir = os.environ['HOME']
         else:
-            raise OSError('Could not find unix home directory to search for '
-                          'astropy config dir')
+            try:
```
taldcroft, line 36: Can this entire function be effectively replaced by `os.expanduser('~')`?  This claims to be applicable for all platforms so it seems best to delegate to this core library function.  I guess it isn't clear if all the custom logic below for Windows is reflected in that function.

taldcroft: Looking at the code for `expanduser` it takes a very different approach and uses the `pwd` module (password database) to get the home directory if `HOME` is not defined.  That seems at first glance reliable, maybe?

pllim:
Not sure why CI not kicking off properly. Maybe a change log under "API change" in 4.2? Is there a way to add a test?

astropy/config/paths.py
dstansby: I think it could - I took the current approach to maintain any previous behaviour, but in ipython they just use `os.expanduser`. If there's openess to changing it I'm happy to modify in that direction?

taldcroft: I recall that @embray had good Windows knowledge back in the day.  Git blame confirms that some of the code came from @embray.  Thoughts?

embray: I'm not exactly sure what you're asking me, but IIRC using [`os.expanduser`](https://docs.python.org/3.6/library/os.path.html#os.path.expanduser) on Windows is reasonable.

dstansby: 👍 if there is consensus that `os.expanduser('~')` is a reasonable approach to take then I will change the PR and see if CI is happy.

taldcroft: @embray - my question is whether we can replace the *entire* `_get_home` code with a single line `return os.expanduser('~')`.  Are all those special cases on Windows still necessary or are they a relic of how things were 10 years ago?

dstansby: Is there any chance this change can be merged (to fix the issues we are having in sunpy, whilst definitely maintaining back-compat), and discussion about whether to replace the whole function with `os.path.expanduser('~')` moved to a new issue?

taldcroft: Yes, good plan, sorry for the delay. Can you open that issue? For merging this needs a bugfix changelog entry indicating fix of problems in MacOS with no HOME dir var.  I can't see any reason not to backport to LTS, so put it in the 4.0.2 section.

taldcroft:
@pllim - do you agree this could be called a bug fix and go to LTS 4.0.2?

taldcroft:
For testing, maybe a test could temporarily unset HOME to exercise the new functionality?

pllim:
Milestone moved. Needs a change log. Thanks! 😄

dstansby:
👍 changelog added

PR #10571, opened 2020-07-15
MNT: Replace some distutils imports with setuptools

pllim:
### Update

This follows up on #10576 to replace some `distutils` imports with `setuptools`. This should now be uncontroversial but it does not completely remove the warning that #10576 is ignoring.

Also see #10578 for the future.

### Original Description

~This pull request is to address~ this error that started showing up during test collection in CI.

```
==================================== ERRORS ====================================
_ ERROR collecting .tox/py37-test/lib/python3.7/site-packages/astropy/tests/command.py _
../../.tox/py37-test/lib/python3.7/site-packages/astropy/tests/command.py:16: in <module>
    from setuptools import Command
../../.tox/py37-test/lib/python3.7/site-packages/setuptools/__init__.py:7: in <module>
    import setuptools.distutils_patch  # noqa: F401
../../.tox/py37-test/lib/python3.7/site-packages/setuptools/distutils_patch.py:59: in <module>
    warn_distutils_present()
../../.tox/py37-test/lib/python3.7/site-packages/setuptools/distutils_patch.py:26: in warn_distutils_present
    "Distutils was imported before Setuptools. This usage is discouraged "
E   UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
```

I think this is due to new `setuptools` release 49.2.0 but I am not 100% sure.


pllim:
Wut....?

astropy/__init__.py
```diff
@@ -6,6 +6,10 @@
 managing them.
 """
 
+# setuptools needs to be imported before distutils import happens
+# in various setup_package.py to avoid pytest collection error.
+import setuptools  # noqa
```
pllim, line 11: Hmm... this makes setuptools a runtime dependency, does it? 💭

pllim: Is this acceptable or do I need to hunt down all the distutils import and then import setuptools before each of them?

pllim:
Hmm... there is still a warning in the logs:
```
py37-test run-test: commands[1] | pytest --pyargs astropy /home/travis/build/astropy/astropy/docs
[5775] .../pytest --pyargs astropy /home/travis/build/astropy/astropy/docs
.../setuptools/distutils_patch.py:26: UserWarning: Distutils was imported before Setuptools.
    This usage is discouraged and may exhibit undesirable behaviors or errors.
    Please use Setuptools' objects directly or at least import Setuptools first.
  "Distutils was imported before Setuptools. This usage is discouraged "
```
ModuleNotFoundError: No module named 'wcsapi_test'
...
FAILED ../../.tox/py37-test/lib/python3.7/site-packages/astropy/wcs/tests/extension/test_extension.py::test_wcsapi_extension
Not sure how to fix this or how it is even related. 🤷 
```


```

saimn:
Yes, seems related to recent changes in setuptools:
```
v49.2.0
    #2230: Now warn the user when setuptools is imported after distutils modules have been loaded 
(exempting PyPy for 3.6), directing the users of packages to import setuptools first.
```
But the real motivation for this is a bit older:
```
v48.0.0
    #2143: Setuptools adopts distutils from the Python 3.9 standard library and no longer depends on distutils
 in the standard library. When importing setuptools or setuptools.distutils_patch, Setuptools will expose its bundled 
version as a top-level distutils package (and unload any previously-imported top-level distutils package), retaining 
the expectation that distutils' objects are actually Setuptools objects. To avoid getting any legacy behavior from the
 standard library, projects are advised to always "import setuptools" prior to importing anything from distutils. 
This behavior happens by default when using pip install or pep517.build. Workflows that rely on setup.py (anything) 
will need to first ensure setuptools is imported. One way to achieve this behavior without modifying code is to invoke
 Python thus: python -c "import setuptools; exec(open('setup.py').read())" (anything).
```
https://github.com/pypa/setuptools/blob/master/CHANGES.rst

All of this makes sense when executing setup.py and with the recent changes in the packaging ecosystem, but I guess that also means that we should really avoid using setuptools in the package itself.

astrofrog:
Instead of importing at the top level, which would add a runtime dependency on setuptools, I think we should fix any of the distutils imports in ``setup_package.py`` files to just be setuptools imports instead, so:

```
from distutils.core import Extension                                                                                                               
from disutils.dep_util import newer_group   
```

would become:

```
from setuptools import Extension                                                                                                               
from setuptools.dep_util import newer_group   
```

I think there is one instance not in a setup_package.py file which is:

https://github.com/astropy/astropy/blob/7a81d70882d9797859954134489a5c802a067606/astropy/wcs/tests/extension/setup.py

and I think for that it's also safe to use setuptools (if needed, the test using it - if any - can be made to be conditional on setuptools being installed).

Basically we should just not have any direct distutils imports.

pllim:
Not sure how to get rid of usage of `from distutils.version import LooseVersion` and I am not really sure what exactly is triggering the error, as it keeps getting stuck at `from setuptools import Command`.

pllim:
tl;dr -- Just fixing `distutils` in `setup_package.py` does not seem to fix the problem. Is it time we get rid of the test runner?

mhvk:
Numpy ended up pinning the version: https://github.com/numpy/numpy/pull/16870. From the discussion at https://github.com/pypa/setuptools/issues/2261, it would seem best to do as @astrofrog noted, try to use `setuptools` throughout. I'm not against replacing `LooseVersion` with a simpler version - we seem to need it only for really simple version strings (like numpy's).

p.s. I still like the idea of the package providing a way to test itself

pllim:
Thanks for the links, @mhvk ! Given https://github.com/pypa/setuptools/issues/2261#issuecomment-658214508, maybe we want to wait a bit before going forward with this PR.

In the meantime, I think @jdavies-st 's suggestion to whitelist is a very good one -- see #10576 . (Sorry for the inconvenience caused, James!)

pllim:
Update: I reduced the scope of this PR. Should now be uncontroversial. There are still some leftover `distutils` imports that we can worry about later.

```
./astropy/io/misc/asdf/tags/transform/tests/test_transform.py:from distutils.version import LooseVersion
./astropy/io/votable/util.py:from distutils import version
./astropy/tests/helper.py:from distutils.version import LooseVersion
./astropy/tests/command.py:from distutils import log
./astropy/utils/introspection.py:from distutils.version import LooseVersion
./astropy/visualization/scripts/fits2bitmap.py:from distutils.version import LooseVersion
./astropy/visualization/tests/test_norm.py:from distutils.version import LooseVersion
./astropy/visualization/wcsaxes/tests/test_misc.py:from distutils.version import LooseVersion
./astropy/wcs/tests/test_wcs.py:from distutils.version import LooseVersion
./astropy/wcs/tests/test_wcsprm.py:from distutils.version import LooseVersion
./astropy/wcs/wcsapi/tests/test_fitswcs.py:from distutils.version import LooseVersion
./docs/nitpick-exceptions:py:obj distutils.version.LooseVersion
```

pllim:
Oh weird. What I thought was uncontroversial made `test_wcsapi_extension` fail: https://travis-ci.org/github/astropy/astropy/jobs/709276303

docs/development/building.rst
```diff
@@ -15,7 +15,7 @@ extensions to be defined inside each sub-package.
 The way extension-helpers works is that it looks for ``setup_package.py`` files
 anywhere in the package, and then looks for a function called ``get_extensions``
 inside each of these files. This function should return a list of
-:class:`distutils.core.Extension` objects, and these are combined into an
+``setuptools.Extension`` objects, and these are combined into an
```
mhvk, line 18: Would be good to fix this to point to the right documentation. Would adding an `intersphinx_mapping['setuptools']` to `docs/conf.py` help?

astropy/tests/command.py
```diff
@@ -10,10 +10,10 @@
 import subprocess
 import sys
 import tempfile
-from distutils import log
 from contextlib import contextmanager
 
 from setuptools import Command
+from distutils import log
```
mhvk, line 16: Since it is basically not used except to give some `info` messages, I think it can be either replaced with `import logging as log` or the commands replaced with `print`. In favour of the earlier option is:
https://github.com/python/cpython/blob/164b04c47e61bd35d55e61bc74f9fd646eba81bb/Lib/distutils/log.py#L3-L4

jdavies-st:
> Not sure how to get rid of usage of `from distutils.version import LooseVersion`

Once can probably use `packaging.version.parse`, but that will be a new dependency on `packaging`, though `packaging` is vendorized in `setuptools` also.  But it will be good to do that because 1) distutils LooseVersion is undocumented, and 2) it does not implement that latests PEPs that deal with version handling (PEP 440).

mwcraig:
>  But it will be good to do that because 1) distutils LooseVersion is undocumented, and 2) it does not implement that latests PEPs that deal with version handling (PEP 440).

Yes, this is...interesting: https://github.com/astropy/astropy/blob/master/astropy/utils/introspection.py#L144

My inclination is to use the vendored version of `packaging` from `setuptools` to avoid the extra dependency. 

Shall I open a separate PR to do that?

pllim:
> Shall I open a separate PR to do that?

Yes, please. 😉

astropy/tests/command.py
pllim: I'll leave this as future exercise but I'll look into intersphinx. Thanks for the review!

mwcraig:
ok, I'm going to go ahead and merge this since it has been approved. Thanks @pllim!

docs/development/building.rst
pllim: Ops, this is already merged. See #10579 😸

pllim:
While `LooseVersion` might have a way forward, test still fails when this is replaced with `setuptools`. No idea what's going on here:

https://github.com/astropy/astropy/blob/25ac198cfe0040d34f139ae6d7417c3a3e12768c/astropy/wcs/tests/extension/setup.py#L14

astropy/tests/command.py
pllim: See #10580

mhvk:
Note: one advantage of `LooseVersion` is that it is from the standard library - we currently do not require `setuptools` as a run-time dependency and I think just version checks are not quite worth adding it. (People who use astropy via a distribution or conda won't have setuptools by default.) Main question is whether we ever need more than `__version__.split('.')` (or even straight string comparison...)

pllim:
FWIW, maybe the reply at https://github.com/pypa/setuptools/issues/2261#issuecomment-660343237 is helpful.

mwcraig:
> Note: one advantage of LooseVersion is that it is from the standard library - we currently do not require setuptools as a run-time dependency and I think just version checks are not quite worth adding it. 

It is odd that something like this is not part of the standard library. I think many of the places it is used it is in files that are part of the test suite; adding `setuptools` or `packaging` itself as a dependency seems less burdensome there. 

In any event, I'll open a PR later today for further (and more concrete) discussion.

eteq:
@pllim - this turns out to be pretty complicated of a backport to v4.0.x . Before trying to do it, I want to confirm: Does it really need to be in 4.0.x ? I'm a bit worried of unintended consequences in things like logging that might be surprising in an LTS...

pllim:
Yeah, okay, no need to backport then, as long as this is in 4.0.x, should be fine until Python org really pulls the plug but I doubt they would do it so soon.

https://github.com/astropy/astropy/blob/d65dabbcc3152ad0d4839e696d85e0e4c70d5966/setup.cfg#L112

pllim:
p.s. I'll leave it for you to re-milestone. Sorry for the trouble, @eteq !

eteq:
> Sorry for the trouble, @eteq !

No trouble at all!

PR #10574, opened 2020-07-16
Remove flawed transform serialization test

eslavich:
### Description

@pllim pointed out that this test is using the `tmpdir` fixture but `tmpdir` is not included in the test method signature.  I thought the `pytest.raises` was covering up the bug, but it turns out the test doesn't even get to the that line -- the error is being raised by the call to `fix_inputs`.

It doesn't appear possible to create a model that would fail ASDF schema validation, so I think we should remove this test.  I'll need @perrygreenfield or @nden to confirm.


pllim:

PR #10576, opened 2020-07-17
MNT: Temporarily ignore setuptools warning about distutils

pllim:
Hopefully #10571 or its replacement will undo this later. This is a temporarily bandaid to allow CI to be green again until we find a proper solution.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -109,6 +109,7 @@ xfail_strict = true
 qt_no_exception_capture = 1
 filterwarnings =
     error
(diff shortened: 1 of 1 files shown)


pllim:
RTD warnings are unrelated but I'll try to patch them here too.

pllim:
CI green again. Remote failure unrelated.

pllim:
Actually we might want this in LTS, so I changed the milestone. Feel free to change it back if we don't.

mhvk:
Great that tests pass again, thanks both!

PR #10579, opened 2020-07-17
DOC: Add setuptools intersphinx

pllim:
### Description

This pull request is to add intersphinx to `setuptools`.

Direct follow up of #10571 .

p.s. Looks like [setuptools use TXT files, not RST](https://github.com/pypa/setuptools/tree/master/docs), so not sure if this will actually work.


pllim:
I don't think intersphinx is possible, @mhvk . It was a good idea though.
```
docs/development/building.rst:15: WARNING: py:obj reference target not found: setuptools.Extension
docs/development/ccython.rst:7: WARNING: py:obj reference target not found: setuptools.Extension
docs/development/ccython.rst:26: WARNING: py:obj reference target not found: setuptools.Extension
```

Searching through https://setuptools.readthedocs.io does not indicate otherwise. Oh well, at least we tried.

mhvk:
OK, thanks for trying!

Aside: the `setuptools` docs seem a bit behind the times more generally...

PR #10582, opened 2020-07-19
fix fitting test

nden:
This fixes a problem with setting up the values for the test discovered in #10552.

#10552 also revealed a more general issue  with validators in models (not addressed here) -  parameter validators are not run on fitted parameters because the values are assigned using `model.parameters=fitted parameters` and do not use the parameter setters. Example is the fitting test of `SmoothlyBrokenPowerLaw` in test_models_quantities where the data has such values that the amplitude is 0. Although an amplitude of 0. is not a valid value, it is not caught by the validator.

@cshanahan1


pllim:
If this is really a bug in the test, you might wanna backport. Up to you.

nden:
Included in #10552

PR #10583, opened 2020-07-19
Don't ignore blank-valued attributes with the c-based parser.

### LINKED ISSUE #10572: SVO votable fails to parse, but looks valid?
This VOTABLE:
http://svo2.cab.inta-csic.es/theory/fps/fps.php?FORMAT=metadata

will not load because of the following error:
```
>>> import requests, io
>>> response = requests.get('http://svo2.cab.inta-csic.es/theory/fps/fps.php?FORMAT=metadata')
>>> votable = io.BytesIO(response.content)
>>> from astropy.io.votable import  parse
>>> parse(votable, invalid='mask', verify='ignore')
  File "<ipython-input-32-f5eb581b75ff>", line 1, in <module>
    parse(votable, invalid='mask', verify='ignore')
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/utils/decorators.py", line 521, in wrapper
    return function(*args, **kwargs)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/table.py", line 167, in parse
    config=config, pos=(1, 1)).parse(iterator, config)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/tree.py", line 3568, in parse
    iterator, tag, data, config, pos)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/tree.py", line 3478, in _add_resource
    resource.parse(self, iterator, config)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/tree.py", line 3280, in parse
    iterator, tag, data, config, pos)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/tree.py", line 3240, in _add_param
    param.parse(iterator, config)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/tree.py", line 1472, in parse
    self.values.parse(iterator, config)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/tree.py", line 1034, in parse
    vo_raise(E09, 'OPTION', config, pos)
  File "/home/adam/anaconda3/lib/python3.7/site-packages/astropy/io/votable/exceptions.py", line 101, in vo_raise
    raise exception_class(args, config, pos)
E09: None:92:8: E09: 'OPTION' must have a value attribute
```
The line it's failing on is  `<OPTION name="All" value=""/>`, which has a value specified, it's just empty.

Is this an invalid table, or is this a real bug in the reader?

tomdonaldson:
### Description

This pull request is to address the failure to parse VOTables that have `OPTION` elements with attribute `value=""`.  More generally, all blank-valued attributes will no longer be ignored.

The previous behavior differed between the "fast" C-based parser and the slower Python-based parser.  The fast parser was ignoring blank-valued attributes, treating them the same as if the attribute were not present.  The slow parser was passing the empty string values of those attributes through to the `tree.py` code that handles the VOTable elements.  The slow parser was correct, so this PR updates the fast parser to do the same thing.

In most cases the difference didn't matter, but in cases like issue #10572, treating a blank value as unspecified caused the parser to think that `OPTION` had no `value`.

Fixes #10572

### DIFF
--- a/astropy/io/votable/tests/data/regression.xml
+++ b/astropy/io/votable/tests/data/regression.xml
@@ -18,7 +18,12 @@ The VOTable format is an XML standard for the interchange of data represented as
   This is a resource description
 </DESCRIPTION>
 <PARAM ID="awesome" datatype="float" name="INPUT" value="0.000000,0.000000" arraysize="*" unit="deg"></PARAM>
-<PARAM ID="empty_value" name="empty_value" arraysize="*" datatype="char" value="" unit="foo"/>
+<PARAM ID="empty_value" arraysize="*" datatype="char" name="empty_value" unit="foo" value="">
+  <VALUES>
+    <OPTION name="empty_value" value=""/>
+    <OPTION name="full_value" value="90prime"/>
+  </VALUES>
+</PARAM>
 <LINK href="http://www.foo.com/" gref="DECPRECATED">
   <DESCRIPTION>Really, this link is totally bogus.</DESCRIPTION>
 </LINK>
--- a/astropy/io/votable/tests/data/regression.xml
+++ b/astropy/io/votable/tests/data/regression.xml
@@ -21,7 +21,7 @@ The VOTable format is an XML standard for the interchange of data represented as
 <PARAM ID="empty_value" arraysize="*" datatype="char" name="empty_value" unit="foo" value="">
   <VALUES>
     <OPTION name="empty_value" value=""/>
-    <OPTION name="full_value" value="90prime"/>
+    <OPTION value="90prime"/>
   </VALUES>
 </PARAM>
 <LINK href="http://www.foo.com/" gref="DECPRECATED">
--- a/astropy/utils/xml/src/iterparse.c
+++ b/astropy/utils/xml/src/iterparse.c
@@ -22,6 +22,7 @@
  *     library.
  ******************************************************************************/
 
+#include <stdio.h>
 #include <Python.h>
 #include "structmember.h"
 
@@ -384,25 +385,24 @@ startElement(IterParser *self, const XML_Char *name, const XML_Char **atts)
                 goto fail;
             }
             do {
-                if (*(*(att_ptr + 1)) != 0) {
-                    key = PyUnicode_FromString(*att_ptr);
-                    if (key == NULL) {
-                        goto fail;
-                    }
-                    val = PyUnicode_FromString(*(att_ptr + 1));
-                    if (val == NULL) {
-                        Py_DECREF(key);
-                        goto fail;
-                    }
-                    if (PyDict_SetItem(pyatts, key, val)) {
-                        Py_DECREF(key);
-                        Py_DECREF(val);
-                        goto fail;
-                    }
+                key = PyUnicode_FromString(*att_ptr);
+                if (key == NULL) {
+                    goto fail;
+                }
+                val = PyUnicode_FromString(*(att_ptr + 1));
+                if (val == NULL) {
+                    Py_DECREF(key);
+                    goto fail;
+                }
+                if (PyDict_SetItem(pyatts, key, val)) {
                     Py_DECREF(key);
                     Py_DECREF(val);
-                    key = val = NULL;
+                    goto fail;
                 }
+                Py_DECREF(key);
+                Py_DECREF(val);
+                key = val = NULL;
+
                 att_ptr += 2;
             } while (*att_ptr);
         } else {
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -306,6 +306,9 @@ astropy.io.misc
 astropy.io.votable
 ^^^^^^^^^^^^^^^^^^
 
+- No longer ignore attributes whose values were specified as empty
+  strings. [#10583]
+
 astropy.modeling
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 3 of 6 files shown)


pllim:
Thanks, @tomdonaldson ! cc @carlosrodrigoblanco and @keflavich

pllim:
I am hesitant to backport this to LTS due to the API change. Is 4.1.x safer or do we really need this in LTS? Please advise. Once we decide on the milestone, a change log should be added.

Also needs a rebase against `master` to pick up fixes for the CI. Thanks!

tomdonaldson:
> I am hesitant to backport this to LTS due to the API change. Is 4.1.x safer or do we really need this in LTS? Please advise. Once we decide on the milestone, a change log should be added.
> 
> Also needs a rebase against `master` to pick up fixes for the CI. Thanks!

I agree.  I'll add it to 4.1.1 and rebase to hopefully clear up the CI.

pllim:
Sorry, when I said API change, I really meant behavioral change. Now that I re-read what I wrote, I am not sure if I agree with my Past Self ™️ . If the old behavior was really a bug, maybe this should go into LTS after all. But if the old behavior was somewhat acceptable due to ambiguous standards (ref: https://github.com/astropy/astropy/issues/10572#issuecomment-659781897), then maybe not. What do you think (again), @tomdonaldson ?

astropy/utils/xml/src/iterparse.c
```diff
@@ -384,25 +385,24 @@ startElement(IterParser *self, const XML_Char *name, const XML_Char **atts)
...
-                    if (PyDict_SetItem(pyatts, key, val)) {
-                        Py_DECREF(key);
-                        Py_DECREF(val);
-                        goto fail;
-                    }
+                key = PyUnicode_FromString(*att_ptr);
+                if (key == NULL) {
+                    goto fail;
+                }
+                val = PyUnicode_FromString(*(att_ptr + 1));
```
pllim, line 392: So it won't crash here without the `if (*(*(att_ptr + 1)) != 0)` check if there is no `(att_ptr + 1)`? Or is `(att_ptr + 1)` always guaranteed?

astropy/io/votable/tests/data/regression.xml
```diff
@@ -18,7 +18,12 @@ The VOTable format is an XML standard for the interchange of data represented as
   This is a resource description
 </DESCRIPTION>
 <PARAM ID="awesome" datatype="float" name="INPUT" value="0.000000,0.000000" arraysize="*" unit="deg"></PARAM>
-<PARAM ID="empty_value" name="empty_value" arraysize="*" datatype="char" value="" unit="foo"/>
+<PARAM ID="empty_value" arraysize="*" datatype="char" name="empty_value" unit="foo" value="">
+  <VALUES>
+    <OPTION name="empty_value" value=""/>
+    <OPTION name="full_value" value="90prime"/>
```
pllim, line 24: If I remember correctly, the actual data reported in #10572 does not have `name` defined for all the options. So maybe remove `name` from the second option in the regression test data?

```suggestion
    <OPTION value="90prime"/>
```

astropy/utils/xml/src/iterparse.c
tomdonaldson: I don't think this adds any new danger.  Since the `if` statement I removed also referenced that pointer,  it would have crashed there before if  `att_ptr + 1` were not a valid pointer:
`if (*(*(att_ptr + 1)) != 0)`

We could add a new check here, but I think there are multiple places in this loop that assume that the attribute structure delivered by expat was a fairly solid contract , having `att_ptr` be an array of an even number of valid strings (char *'s) followed by a `NULL` to end the loop.  So if `att_ptr` is not `NULL`, then `att_ptr` would also be not `NULL`.  So while we could check for `NULL`, we couldn't check for random garbage which might be just as likely if the expat attribute structure was corrupt.

astropy/io/votable/tests/data/regression.xml
tomdonaldson: A fair point.  I'll update this and the truth values.

tomdonaldson:
> Sorry, when I said API change, I really meant behavioral change. Now that I re-read what I wrote, I am not sure if I agree with my Past Self ™️ . If the old behavior was really a bug, maybe this should go into LTS after all. But if the old behavior was somewhat acceptable due to ambiguous standards (ref: [#10572 (comment)](https://github.com/astropy/astropy/issues/10572#issuecomment-659781897)), then maybe not. What do you think (again), @tomdonaldson ?

I agree this is a bug fix, and it would be good to get it into LTS, but not critical.  My reluctance to get it into LTS was about not knowing what's left in the release process.  I.e., since there is a behavior change (albeit subtle), is there enough testing left in the release cycle to make us confident enough we didn't introduce bad side-effects?

pllim:
I briefly thought about maybe LTS but a bit later like 4.0.3 but then there is no place for such a change log yet, so let's just stick with 4.1.1. If people really need it in LTS, we can revisit as a separate issue.

Good point about `if` crashing. I think it is fine to remove then. I'll approve after your one change to test data when tests pass after. Thanks!

tomdonaldson:
Sounds good.  Thanks for the review!

embray:
This seems to be broken in Python 3.8+.  The parser no longer receives the empty 'value' from the `<OPTION>` tag with the empty value.

embray:
Confirmed that the problem is not reproducible with `_debug_python_based_parser = True`.  I'm not sure what changed between Python 3.7 and 3.8 that would affect this though.

embray:
My mistake.  I think I switched to an older branch that didn't have this fix, and didn't rebuild the extension modules when switching back.  Big waste of time!

tomdonaldson:
Whew!  I'm sorry about the wild goose chase, but I'm definitely glad Python 3.8 didn't change in a way that affected this.

Issue #10585, opened 2020-07-19
method `.si` has unexpected output. 


lunarspectrum:
### Description

The list of SI units in `astropy.units.si` contains many units that are not within the SI. Further, using `.si` on a quantity to convert to SI units results in incorrect behavior.
### Expected behavior

Applying .si to a quantity would output the quantity in SI units. Those units should either be Base SI units, or it could be argued on the list of SI derived units with special names. C.f. [https://physics.nist.gov/cuu/Units/units.html](https://physics.nist.gov/cuu/Units/units.html) (preferably this would be with an option).
### Actual behavior

The quantity was output in other units. Also, steradians are reduced to radians^2, which, while possibly being necessary for certain conversions, is not typically done. 

### Steps to Reproduce

`(u.sr).si` outputs radians^2
`(u.kg*u.s**-2* u.sr**-1 * u.nm**-1).si` outputs bar/rad^2. Bar is not part of the SI. 
`(u.kg*u.s**-2* u.sr**-1 * u.m**-1).si` outputs Pa/rad^2. Pascal is an SI derived unit, and so possibly ok, but still unexpected. 

### System Details

Python 3.8.2 | packaged by conda-forge | (default, Apr 24 2020, 07:34:03) [MSC v.1916 64 bit (AMD64)]
Numpy 1.18.4
astropy 4.0.1.post1
Scipy 1.3.2
Matplotlib 3.2.1

pllim:
Thanks for bringing this to our attention. `bar` was added back in #1468, apparently for CDS support. @mhvk , I see that you commented in that PR, do you happen to still remember anything about it?

mhvk:
@lunarspectrum - agreed that `bar` for sure should not be there -- see #10586. While checking the NIST list, I realized we have quite a few further units defined in `units.si` that arguably do not really belong (e.g., Curie, Angstrom). But at least for Curie it is also not obvious to me that they should be in `astrophys` -- perhaps we need some other `derived` units module.

For `sr`, I think we need some adjustment to the "scoring system" in [`to_system`](https://github.com/astropy/astropy/blob/25ac198cfe0040d34f139ae6d7417c3a3e12768c/astropy/units/core.py#L1269).

As for just getting SI base units, that is possible already with `.decompose`. E.g.,
```
(u.erg/(u.cm**2*u.s*u.AA)).decompose(u.si.bases)
# Unit("1e+07 kg / (m s3)")
```
[...]

PR #10586, opened 2020-07-20
Move bar out of SI, since it is not recommended to use it.

mhvk:
Partially addresses #10585 

I mile-stoned it for 4.2 since there are more units in `units.si` that really do not belong, but which also do not make sense in `astrophys` - so, this could become a larger overhaul. Also, it is not really a bug in that the result is not strictly "wrong", just not recommended.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -119,8 +119,6 @@ astropy.units
 - Added the ``spat`` unit of solid angle that represents the full sphere.
   [#10726]
 
-- The ``bar`` unit is no longer wrongly considered an SI unit. [#10586]
-
 astropy.utils
 ^^^^^^^^^^^^^
 
@@ -221,6 +219,10 @@ astropy.uncertainty
 
 astropy.units
 ^^^^^^^^^^^^^
+
+- The ``bar`` unit is no longer wrongly considered an SI unit, meaning that
+  SI decompositions like ``(u.kg*u.s**-2* u.sr**-1 * u.nm**-1).si`` will
+  no longer include it. [#10586]
 
 astropy.utils
 ^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -112,11 +112,14 @@ astropy.uncertainty
 
 astropy.units
 ^^^^^^^^^^^^^
+
 - ``Quantity.to`` has gained a ``copy`` option to allow copies to be avoided
   when the units do not change. [#10517]
 
 - Added the ``spat`` unit of solid angle that represents the full sphere.
   [#10726]
+
+- The ``bar`` unit is no longer wrongly considered an SI unit. [#10586]
 
 astropy.utils
 ^^^^^^^^^^^^^
--- a/astropy/units/astrophys.py
+++ b/astropy/units/astrophys.py
@@ -100,6 +100,12 @@ def_unit(['Ry', 'rydberg'],
          "constant",
          format={'latex': r'R_{\infty}', 'unicode': 'R∞'})
 
+##########################################################################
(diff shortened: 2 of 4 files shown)


pllim:
Looks like failure is real; e.g.
```
  File ".../astropy/units/photometric.py", line 30, in <module>
    def_unit(['ST', 'STflux'], 10.**(21.1/-2.5) * cgs.erg * cgs.cm**-2 / si.s / si.AA,
AttributeError: module 'astropy.units.si' has no attribute 'AA'
```

Yeah, this can break a lot of things downstream, so I agree with the 4.2 milestone.

> this could become a larger overhaul

Do you wish to overhaul stuff in this same PR, or as a series of PRs? If the former, maybe can utilize the "draft PR" feature until you think it is ready.

> it is not really a bug

Hmm... Do we need a "bug but not really" label? 😹

mhvk:
@pllim - ah, oops, I had a revert locally that didn't get included in the PR.

I don't think I will have time for the larger overhaul soon, so would rather just do this one obvious problem now. I'm happy to call it "refactoring" ...

pllim:
Hmm, why isn't this PR affected by #10588 ...

pllim:
@lunarspectrum , are you satisfied with handling of `bar` in this PR? Do you want to take this PR for a spin? Thanks!

lunarspectrum:
> @lunarspectrum , are you satisfied with handling of `bar` in this PR? Do you want to take this PR for a spin? Thanks!

@pllim , I like the idea of doing a quick change followed by a larger overhaul later. 

Are you asking for me to just run through the process removing bar only? Or are you interested in a larger attempt now?

There are a bunch of units in there that are astronomy-specific that are definitely not SI but it may break a lot of code if removed.

mhvk:
@lunarspectrum - I think @pllim was just asking whether you had wanted to check that this PR worked as advertised (never a bad idea, though I'd hope the test would prove that!).

As long as we just move units out of `.si` to another module, I doubt we'll cause too many problem, but it does need some thought...

adrn:
:+1: to this, but not sure it makes sense to move to `astrophys.py`. Looking elsewhere in that module, it seems that maybe we should have a `misc.py` for things like bar, barn, cycle, byte, to name a few more general units I noticed on a quick look.

mhvk:
@adrn - I agree that we should have a `misc`, but as you note, it is not just `bar` which would go there, so can I weasel out of this and call it a separate issue? See #10769...

adrn:
@mhvk Yes, fine by me! Unfortunately, this requires a rebase though to fix the changelog 😬

mhvk:
OK, rebased

CHANGES.rst
```diff
@@ -112,12 +112,15 @@ astropy.uncertainty
...
 astropy.units
 ^^^^^^^^^^^^^
+
 - ``Quantity.to`` has gained a ``copy`` option to allow copies to be avoided
   when the units do not change. [#10517]
 
 - Added the ``spat`` unit of solid angle that represents the full sphere.
   [#10726]
 
+- The ``bar`` unit is no longer wrongly considered an SI unit. [#10586]
```
adrn, line 122: Does this belong under API Changes instead?

mhvk: Hmm, yes, you're right - will put it under bug-fix (with a `[ci skip]`, so current commit pass is what will count). I think API change is not quite right since it doesn't change how one accesses the unit.

mhvk: Actually, let me put it in API changes after all - it is not something we really want to back-port, and it does change some decomposition results.

adrn (approved):

PR #10589, opened 2020-07-20
TST: Fix linkcheck in cron job

pllim:
### Description

This pull request is to fix what #10563 was supposed to fix but didn't really.

- [x] Find something that actually works. See https://github.com/astropy/astropy/pull/10589#issuecomment-661374535
- [x] Drop temporary commits.

### DIFF
--- a/astropy/utils/decorators.py
+++ b/astropy/utils/decorators.py
@@ -708,6 +708,8 @@ class classproperty(property):
         return fget
 
 
+# Adapted from the recipe at
+# http://code.activestate.com/recipes/363602-lazy-property-evaluation
 class lazyproperty(property):
     """
     Works similarly to property(), but computes the value only once.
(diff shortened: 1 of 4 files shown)


pllim:
So... I tried a few different things:

* Removing `-W` from `linkcheck` command -- failed
* Changing the `nitpick-exceptions` text to match "remove and return some (key, value) pair" -- failed
* Tried to escape the parenthesis in `nitpick-exceptions` with `\` -- failed
* Removing Sphinx pinning for `linkcheck` job -- failed and things got even worse

The only thing that worked is completely removing `nitpicky = True` from `docs/conf.py`. Probably a little drastic just to get rid of one warning in a `linkcheck` job, but it is also annoying that we'll never have a green `linkcheck` job again.

I am out of ideas...

pllim:
I forgot we unpinned numpydoc! Hmm... Let's see if pinning it back just for this job works.

pllim:
Hmm, getting rate limited. I guess linkcheck works now. We can worry about "broken" links after the next cron run.

pllim:
Thanks for the review, @mhvk !

Issue #10578, opened 2020-07-17
MNT: Un-ignore Distutils was imported before Setuptools


pllim:
### Description

When it is clear how to port the rest of the stuff in https://github.com/astropy/astropy/pull/10571#issuecomment-660262832, we can undo this from #10576: 

https://github.com/astropy/astropy/blob/d10d0f2e27e23ef3e94631ca83126554a3083c75/setup.cfg#L112

### Additional context

#10571,  numpy/numpy#16870, pypa/setuptools#2261

PR #10590, opened 2020-07-20
Remove LooseVersion

mwcraig:
### Description

This addresses part of #10578 by removing `distutils.version.LooseVersion`.  

- In the tests, `packaging` is added as a dependency and `packaging.version.Version` is used instead of `LooseVersion`. It does not seem problematic to me to add this as a test dependency.
- In a couple of places ([here](https://github.com/astropy/astropy/pull/10590/files#diff-cb06e7e54ea6c0cc5013e1e85f3cda5bR102) and [here](https://github.com/astropy/astropy/pull/10590/files#diff-5a4193c12a96fa7bfa06b29a41e95120R287)) the version test was simple to replace.
- [x] One place I'm not sure how to handle: https://github.com/astropy/astropy/blob/master/astropy/utils/introspection.py#L153 . Writing a bit of code to do the check ourselves would be straightforward and avoid `LooseVersion`; alternatively, we could just move the import of `LooseVersion` inside of the function, which should remove the setuptools warnings.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,6 +57,7 @@ test =
     coverage
     skyfield>=1.20
     sgp4>=2.3
+    packaging
 all =
     scipy>=0.18
     dask[array]
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,6 @@ test =
     coverage
     skyfield>=1.20
     sgp4>=2.3
-    packaging
 all =
     scipy>=0.18
     dask[array]
--- a/astropy/io/misc/asdf/tags/transform/tests/test_transform.py
+++ b/astropy/io/misc/asdf/tags/transform/tests/test_transform.py
@@ -2,8 +2,8 @@
 # -*- coding: utf-8 -*-
 
 import warnings
-from distutils.version import LooseVersion
 
+from packaging.version import Version
 import pytest
 import numpy as np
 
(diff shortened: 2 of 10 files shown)


mwcraig:
@mhvk and @pllim -- here is the pull request I promised to take a stab at removing `LooseVersion`.

setup.cfg
```diff
@@ -57,6 +57,7 @@ test =
     coverage
     skyfield>=1.20
     sgp4>=2.3
+    packaging
```
pllim, line 60: Shouldn't packaging be a runtime requirement now, like `numpy` and `pyerfa`?

mwcraig: I didn't think so because it is only imported in the tests. Or at least that was the goal -- using `packaging` in the tests but not in the main package to avoid adding a new dependency.

mhvk: Aha, I hadn't realized this would be just for testing - in that case we actually do not need to add a requirement at all, since `pytest` itself already requires packaging! (It must do its version tests in, e.g., `pytest.importorskip` with it...).

mwcraig:
> For introspection, I'm fine with going the easy route and simply move the import inside, with a comment on why. Alternatively, also fine with just seeing whether a simple use of the output of the regexp that is already there isn't good enough...

I think the regex works for getting the version number -- I don't want to write the code to do the comparison.

For now I've just moved the distutils import inside the function in `introspection`, but since `pytest` is already a hard requirement for astropy, and `pytest` requires `packaging`, couldn't we switch this to using `packagin.version.Version` too? 

I had not realized that `pytest` requires `packaging`....

pllim:
> but since pytest is already a hard requirement for astropy

It shouldn't be though. Requiring it at runtime is a bug.

mwcraig:
> It shouldn't be though. Requiring it at runtime is a bug.

In that case the current fix should do the trick.

PR #10592, opened 2020-07-21
fix earth location info

### LINKED ISSUE #10591: IndexError when accessing EarthLocation.info
### Description

EarthLocation info display is broken on current master

### Expected behavior

display the site info

### Steps to Reproduce

```python
from astropy.coordinates import EarthLocation

keck = EarthLocation.of_site('Keck Observatory')
keck.info
```
Produces the following exception
```
  File "/home/hannes/work/repos/astropy/astropy/utils/data_info.py", line 488, in __repr__
    self.__call__(out=out)
  File "/home/hannes/work/repos/astropy/astropy/utils/data_info.py", line 472, in __call__
    info['length'] = len(dat)
  File "/home/hannes/work/repos/astropy/astropy/coordinates/earth.py", line 799, in __len__
    raise IndexError('0-d EarthLocation arrays cannot be indexed')
IndexError: 0-d EarthLocation arrays cannot be indexed
```

### System Details

```
Linux-4.15.0-111-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0]
Numpy 1.20.0.dev0+2ccfd79
astropy 4.2.dev434+gd1e34fb0c
Scipy 1.5.1
Matplotlib 3.2.1.post2723+g89fa0e43b
```

astromancer:
### Description

Fixes #10591 

```python
from astropy.coordinates import EarthLocation

keck = EarthLocation.of_site('Keck Observatory')
keck.info
```

Now produces the site info
```
name = W. M. Keck Observatory
dtype = void192
unit = m
class = EarthLocation
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -902,6 +902,9 @@ astropy.coordinates
 - Make the parsing of angles thread-safe so that ``Angle`` can be used in
   Python multithreading. [#10556]
 
+- Fixed reporting of ``EarthLocation.info`` which previously raised an exception.
+  [#10592]
+
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -348,6 +348,7 @@ class EarthLocation(u.Quantity):
         dtype = void192
         unit = m
         class = EarthLocation
+        n_bad = 0
         >>> keck.info.meta  # doctest: +REMOTE_DATA
         {'source': 'IRAF Observatory Database', 'timezone': 'US/Aleutian'}
 
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -343,6 +343,11 @@ class EarthLocation(u.Quantity):
         >>> keck = EarthLocation.of_site('Keck Observatory')  # doctest: +REMOTE_DATA
         >>> keck.geodetic  # doctest: +REMOTE_DATA +FLOAT_CMP
         GeodeticLocation(lon=<Longitude -155.47833333 deg>, lat=<Latitude 19.82833333 deg>, height=<Quantity 4160. m>)
+        >>> keck.info  # doctest: +REMOTE_DATA
+        name = W. M. Keck Observatory
+        dtype = void192
+        unit = m
+        class = EarthLocation
         >>> keck.info.meta  # doctest: +REMOTE_DATA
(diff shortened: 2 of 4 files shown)


mhvk:
Thanks, that's a nice simple fix! But could you add a test? (in `test_earth.py`). Also an entry in the changelog (under bug fixes in 4.0.2).

mhvk:
@astromancer - This looks good but the failures are real: the test does need to be marked `@pytest.mark.remote_data(source="astropy")` since it accesses the internet, and you'd need to fix the pep8 errors.

While you do that, could you rebase and squash the commits?

Also, one suggestion (feel free to ignore): since this makes the `.info` quite nice, how about giving the output in the docstring of `of_site` (it already shows `.info.meta`, which is good too, so perhaps just before that)?

astromancer:
OK. Re decorating the test: I thought that using the 'greenwich' example would not need internet access since 
```
    If a site cannot be found in the registry (i.e., an internet
    connection is not available), it will fall back on a built-in list,
    In the future, this bundled list might include a version-controlled
    list of canonical observatories extracted from the online version,
    but it currently only contains the Greenwich Royal Observatory as an
    example case.
```
from `EarthLocation.of_site` docs.

mhvk:
@astromancer - yes, that makes sense, but it may be that it still tries to connect. cc @pllim who tried to make sure these things didn't happen (but I vaguely recall `of_site` was a particular problem).

pllim:
I forgot what I did now but I think what you need is in #10304 ?

mhvk:
@astromancer - if it is easy to get something to work following the above hint, do so, but if not, I think it is absolutely fine to just mark the test as `remote_data`.

mhvk:
The code now looks all OK, but unfortunately the doctest fails:
```
346         >>> keck.info  # doctest: +REMOTE_DATA
    @@ -3,2 +3,4 @@
     unit = m
     class = EarthLocation
    +n_bad = 0
    +<BLANKLINE>
```

See https://travis-ci.org/github/astropy/astropy/jobs/712153720 (the other failure is unrelated)

astromancer:
@mhvk This looks all good now, do i still need to rebase?

mhvk:
@astromancer - generally I like single commits, but I think we've been through enough here, and 2 is fine too! So, I'll merge. Thanks for sticking with it, adding all the extras to the super-small and good fix you started with!

PR #10599, opened 2020-07-24
Fix ResourceWarning with fits.writeto and pathlib object

### LINKED ISSUE #10594: HDUList does not close file if filename is pathlib.Path
If I write a FITS file and provide the filename as a `pathlib.Path` object, the file is not closed

### How to reproduce

```python
from astropy.io import fits
import pathlib
hdulist = fits.HDUList([fits.PrimaryHDU()])
print("Use pathlib.Path")
hdulist.writeto(pathlib.Path("e.fits"), overwrite=True)
print("Use string")
hdulist.writeto("e.fits", overwrite=True)
```

Running this:

```
python -Wd writeto_path.py
Use pathlib.Path
/home/zonca/zonca/p/software/healpy/healpyvenv/lib/python3.7/site-packages/astropy/utils/decorators.py:535: ResourceWarning: unclosed file <_io.FileIO name='e.fits' mode='wb' closefd=True>
  return function(*args, **kwargs)
Use string
```

### Possible issue

I think the issue is at https://github.com/astropy/astropy/blob/a6a350570aec6600eb907fe9903585dd18dc0874/astropy/io/fits/hdu/hdulist.py#L927
It checks if the input file is `str`, it would be nice to check also for `pathlib.Path`.

saimn:
Fix #10594

Note: the `ResourceWarning` is hidden by the filter in `setup.cfg`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -928,6 +928,9 @@ astropy.io.fits
 
 - Fix duplication issue when setting a keyword ending with space. [#10482]
 
+- Fix ResourceWarning with ``fits.writeto`` and ``pathlib.Path`` object.
+  [#10599]
+
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
--- a/astropy/io/fits/hdu/hdulist.py
+++ b/astropy/io/fits/hdu/hdulist.py
@@ -38,7 +38,7 @@ def fitsopen(name, mode='readonly', memmap=None, save_backup=False,
 
     Parameters
     ----------
-    name : file path, file object, file-like object or pathlib.Path object
+    name : str, file-like or `pathlib.Path`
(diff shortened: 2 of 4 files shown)


saimn:
I guess there are too many `ResourceWarning` warnings to turn them into errors.
Also there is an unrelated issue on the numpy-dev build: #10600

pllim:
Right, there were too many `ResourceWarning` occurrences in our CI. If I remember correctly, not just in FITS, but also `io.ascii` and maybe even `votable`. In short, it is a mess.

Maybe we can have `setup.cfg` to only ignore `ResourceWarning` from certain modules and then tackle it part by part... 💭 

Thanks, everyone! Given this is approved and failures are unrelated, I'll go ahead and merge.

PR #10601, opened 2020-07-24
TST: Fix cannot convert NaN to int in nddata test

pllim:
### Description

This pull request is to address test failure when a test tries to cast NaN into integer.

Fixes #10600

**Note to reviewer:** Check the remote data job that is allowed to fail.

PR #10602, opened 2020-07-24
Fix extract_array with integer input array and NaN fill value

### LINKED ISSUE #10600: Numpy dev failure with test_extract_array_return_pos
https://travis-ci.org/github/astropy/astropy/jobs/711330525
```
        # Extracting on the edges is presumably a rare case, so treat special here
        if (extracted_array.shape != shape) and (mode == 'partial'):
            extracted_array = np.zeros(shape, dtype=array_large.dtype)
>           extracted_array[:] = fill_value
E           ValueError: cannot convert float NaN to integer
../../.tox/py37-test-devdeps/lib/python3.7/site-packages/astropy/nddata/utils.py:246: ValueError
```

larrybradley:
`numpy 1.20dev` now raises `ValueError: cannot convert float NaN to integer` when attempting to assign to `np.nan` in an integer array.  In this PR, I've updated the `extract_array` function to change the `dtype` of the output array to float if the fill value is used and its value is `np.nan`.  That preserves the current behavior with `numpy <= 1.19`.

https://github.com/astropy/astropy/pull/10601 simply fixes an `astropy` test.  This PR fixes the root issue (I noticed this in `photutils`).

Fix #10600.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1074,6 +1074,9 @@ Other Changes and Additions
 - Updated the bundled CFITSIO library to 3.480. See
   ``cextern/cfitsio/docs/changes.txt`` for additional information. [#10256]
 
+- ``extract_array`` raises a ``ValueError`` if the data type of the
+  input array is inconsistent with the ``fill_value``. [#10602]
+
 
 4.0.1 (2020-03-27)
 ==================
--- a/astropy/nddata/tests/test_utils.py
+++ b/astropy/nddata/tests/test_utils.py
@@ -8,7 +8,7 @@ from astropy.tests.helper import assert_quantity_allclose
 from astropy.nddata import (extract_array, add_array, subpixel_indices,
                             overlap_slices, NoOverlapError,
                             PartialOverlapError, Cutout2D)
-
+from astropy.utils import minversion
 from astropy.wcs import WCS, Sip
 from astropy.wcs.utils import proj_plane_pixel_area
 from astropy.coordinates import SkyCoord
@@ -274,7 +274,7 @@ def test_extract_array_return_pos():
     The result will differ by mode. All test here are done in 1d because it's
     easier to construct correct test cases.
     '''
-    large_test_array = np.arange(5)
+    large_test_array = np.arange(5).astype(float)
     for i in np.arange(-1, 6):
         extracted, new_pos = extract_array(large_test_array, 3, i,
                                            mode='partial',
@@ -290,6 +290,14 @@ def test_extract_array_return_pos():
         extracted, new_pos = extract_array(large_test_array, (3,), (i,),
                                            mode='trim', return_position=True)
         assert new_pos == (expected, )
+
+
+def test_extract_array_nan_fillvalue():
+    if minversion(np, '1.20'):
+        with pytest.raises(ValueError) as err:
+            extract_array(np.ones((10, 10), dtype=int), (5, 5), (1, 1),
+                          fill_value=np.nan)
+        assert 'fill_value cannot be set to np.nan if' in err.value.args[0]
 
 
 def test_add_array_odd_shape():
--- a/astropy/nddata/tests/test_utils.py
+++ b/astropy/nddata/tests/test_utils.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+from packaging.version import Version
 import pytest
 import numpy as np
 from numpy.testing import assert_allclose, assert_array_equal
@@ -8,7 +9,6 @@ from astropy.tests.helper import assert_quantity_allclose
 from astropy.nddata import (extract_array, add_array, subpixel_indices,
                             overlap_slices, NoOverlapError,
                             PartialOverlapError, Cutout2D)
-from astropy.utils import minversion
 from astropy.wcs import WCS, Sip
 from astropy.wcs.utils import proj_plane_pixel_area
 from astropy.coordinates import SkyCoord
@@ -293,7 +293,7 @@ def test_extract_array_return_pos():
 
 
 def test_extract_array_nan_fillvalue():
-    if minversion(np, '1.20'):
+    if Version(np.__version__) >= Version('1.20'):
         msg = 'fill_value cannot be set to np.nan if the input array has'
         with pytest.raises(ValueError, match=msg):
             extract_array(np.ones((10, 10), dtype=int), (5, 5), (1, 1),
--- a/astropy/nddata/tests/test_utils.py
+++ b/astropy/nddata/tests/test_utils.py
@@ -274,7 +274,7 @@ def test_extract_array_return_pos():
     The result will differ by mode. All test here are done in 1d because it's
     easier to construct correct test cases.
     '''
-    large_test_array = np.arange(5).astype(float)
+    large_test_array = np.arange(5, dtype=float)
     for i in np.arange(-1, 6):
         extracted, new_pos = extract_array(large_test_array, 3, i,
                                            mode='partial',
@@ -294,10 +294,10 @@ def test_extract_array_return_pos():
 
 def test_extract_array_nan_fillvalue():
     if minversion(np, '1.20'):
-        with pytest.raises(ValueError) as err:
+        msg = 'fill_value cannot be set to np.nan if the input array has'
+        with pytest.raises(ValueError, match=msg):
             extract_array(np.ones((10, 10), dtype=int), (5, 5), (1, 1),
                           fill_value=np.nan)
-        assert 'fill_value cannot be set to np.nan if' in err.value.args[0]
 
 
 def test_add_array_odd_shape():
--- a/astropy/nddata/utils.py
+++ b/astropy/nddata/utils.py
@@ -244,8 +244,8 @@ def extract_array(array_large, shape, position, mode='partial',
 
     # Extracting on the edges is presumably a rare case, so treat special here
     if (extracted_array.shape != shape) and (mode == 'partial'):
+        extracted_array = np.zeros(shape, dtype=array_large.dtype)
         try:
-            extracted_array = np.zeros(shape, dtype=array_large.dtype)
             extracted_array[:] = fill_value
         except ValueError:
             raise ValueError('fill_value cannot be set to np.nan if the '
--- a/astropy/nddata/utils.py
+++ b/astropy/nddata/utils.py
@@ -248,10 +248,12 @@ def extract_array(array_large, shape, position, mode='partial',
         try:
             extracted_array[:] = fill_value
         except ValueError:
-            raise ValueError('fill_value cannot be set to np.nan if the '
-                             'input array has integer type. Please '
-                             'change either the input array dtype or the '
-                             'fill_value.') from None
+            raise ValueError('fill_value is inconsistent with the data type '
+                             'of the input array (e.g., fill_value cannot be '
+                             'set to np.nan if the input array has integer '
+                             'type). Please change either the input array '
+                             'dtype or the fill_value.') from None
+
         extracted_array[small_slices] = array_large[large_slices]
         if return_position:
             new_position = [i + s.start for i, s in zip(new_position,
--- a/astropy/nddata/utils.py
+++ b/astropy/nddata/utils.py
@@ -194,8 +194,10 @@ def extract_array(array_large, shape, position, mode='partial',
     fill_value : number, optional
         If ``mode='partial'``, the value to fill pixels in the extracted
         small array that do not overlap with the input ``array_large``.
-        ``fill_value`` must have the same ``dtype`` as the
-        ``array_large`` array.
+        ``fill_value`` will be changed to have the same ``dtype`` as the
+        ``array_large`` array, with one exception. If ``array_large``
+        has integer type and ``fill_value`` is ``np.nan``, then a
+        `ValueError` will be raised.
     return_position : bool, optional
         If `True`, return the coordinates of ``position`` in the
         coordinate system of the returned array.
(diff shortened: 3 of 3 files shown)


astropy/nddata/utils.py
```diff
@@ -242,7 +242,11 @@ def extract_array(array_large, shape, position, mode='partial',
 
     # Extracting on the edges is presumably a rare case, so treat special here
     if (extracted_array.shape != shape) and (mode == 'partial'):
-        extracted_array = np.zeros(shape, dtype=array_large.dtype)
+        if np.isnan(fill_value):
```
pllim, line 245: Do you think there needs to be info or warning when array type is changed like this?

larrybradley: Perhaps, but we weren't doing it before.

larrybradley: We could also add a note in the docstring.

pllim: Yeah, maybe in the docstring then.

saimn: The docstring says "``fill_value`` must have the same ``dtype`` as the ``array_large`` array." which could be improved.
But indeed if the output array dtype is changed because of the fill value type, I think a warning would be appropriate.
Another option would be to have another default fill value for integers, I think 0 would be fine (by default).

pllim: Not `-999`? 😏

larrybradley: OK, I'll add the warning.  I think that's better than having different default `fill_values` based on the input `dtype`.

saimn: It would not work for int8 or unsigned int ;). 
And for a function that is extracting a piece of an array, I think padding with 0 by default is reasonable. At least it's less surprising that getting a float array if you give it some integer type, imo.

larrybradley: Using `0` as a default sentinel value for "no data" can be surprising too.  The user can always specify `fill_value=0` if that is what they really want.

Having different defaults would also be surprising to get a different result based only on the input dype.  I can see users getting confused by this:

```python
>>> extract_array(np.ones((5, 5)), (3, 3), (0, 0))
array([[nan, nan, nan],
       [nan,  1.,  1.],
       [nan,  1.,  1.]])

>>> extract_array(np.ones((5, 5), dtype=int), (3, 3), (0, 0))
array([[0,  0,  0],
       [0,  1,  1],
       [0,  1,  1]])
```

larrybradley: I know -- masked arrays! jk :)

larrybradley: I added the warning and updated the docstring.

astropy/nddata/utils.py
```diff
@@ -242,7 +245,15 @@ def extract_array(array_large, shape, position, mode='partial',
...
+            dtype = float
+            warnings.warn('Output array dtype was changed to float to '
+                          'fill missing values with np.nan. Please '
+                          'change the fill_value keyword if this is not '
+                          'desirable.')
```
pllim, line 253: And you need to catch it in the test too. 😉 
```suggestion
                          'desirable.', AstropyUserWarning)
```

larrybradley: Ugh, there are now 10 failing tests, include some in the narrative docs.

pllim: For the docs, you can use https://github.com/astropy/pytest-doctestplus#ignoring-warnings

larrybradley:
@mhvk I think you're right.  I'll change this to raise an informative error and update the tests.  I expect this to break some people's code (using our test suite and docs as an example), so I'll add a note in the API section of the changelog.

pllim:
Maybe remilestone to... 4.2?

larrybradley:
Yeah, I was wondering about that.  It's a breaking change (but using numpy 0.20 without us doing anything would also break things -- with a less friendly error message).  Perhaps we want to fix this sooner than later?

pllim:
Hrm... if this goes in LTS, might want to add in numpy version check, so people using `numpy<1.20` will not notice any difference.

larrybradley:
How do I do that *without* using `distutils` and not introducing `packaging/setuptools` as a runtime dependency?  🤔

pllim:
> How do I do that

The `distutils` headache only happens at build time, I think. @mwcraig moved it to local import for `astropy.utils.introspection.minversion` in #10590. I am hoping that you can use `minversion` here without having that warning pop up again. Only one way to find out...

pllim:
No, wait, wait! @mwcraig made `packaging` a test dependency, so now you can use `from packaging.version import Version` directly. See his PR above for examples!

larrybradley:
But here I would need to use `Version` in actual code (not tests), thus requiring `packaging` as a runtime dependency.

pllim:
Oh yeah... Try `minversion` then. It is still used in other places and CI passed.

mhvk:
Indeed, I'd define a new `NUMPY_LT_1_20` in `utils/compat/numpycompat.py` and use that.

Though: is this really necessary? An alternative would be to error in newer astropy but in LTS stick with a (more informative) deprecation warning -- after all, numpy's behaviour is not yet changing.

larrybradley:
I decided to simply catch the new `numpy` exception and re-raise it with a more informative message.  I only need to check the numpy version in a test.

astropy/nddata/tests/test_utils.py
```diff
@@ -292,6 +292,14 @@ def test_extract_array_return_pos():
         assert new_pos == (expected, )
 
 
+def test_extract_array_nan_fillvalue():
+    if minversion(np, '1.20'):
+        with pytest.raises(ValueError) as err:
```
mhvk, line 297: could be `with pytest.raises(ValueError, match='fill_value cannot be set to np.nan if'):`

astropy/nddata/tests/test_utils.py
```diff
@@ -274,7 +274,7 @@ def test_extract_array_return_pos():
     The result will differ by mode. All test here are done in 1d because it's
     easier to construct correct test cases.
     '''
-    large_test_array = np.arange(5)
+    large_test_array = np.arange(5).astype(float)
```
mhvk, line 277: total nitpick, but I'd write `5.` or add `dtype=float`.

astropy/nddata/utils.py
```diff
@@ -242,8 +244,14 @@ def extract_array(array_large, shape, position, mode='partial',
 
     # Extracting on the edges is presumably a rare case, so treat special here
     if (extracted_array.shape != shape) and (mode == 'partial'):
-        extracted_array = np.zeros(shape, dtype=array_large.dtype)
-        extracted_array[:] = fill_value
+        try:
+            extracted_array = np.zeros(shape, dtype=array_large.dtype)
```
mhvk, line 248: Maybe move the definition out of the `try/except`.

astropy/nddata/tests/test_utils.py
larrybradley: Changed, thanks!

astropy/nddata/utils.py
larrybradley: Done

astropy/nddata/tests/test_utils.py
larrybradley: Fixed

pllim:
I guess the RTD warning isn't related?
```

/home/docs/checkouts/readthedocs.org/user_builds/astropy/checkouts/10602/docs/visualization/rgb.rst:57: WARNING: Exception occurred in plotting rgb-2
 from /home/docs/checkouts/readthedocs.org/user_builds/astropy/checkouts/10602/docs/visualization/rgb.rst:
  File "/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/10602/lib/python3.8/site-packages/matplotlib/sphinxext/plot_directive.py", line 472, in run_code
    exec(code, ns)
  File "<string>", line 8, in <module>
  File "/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/10602/lib/python3.8/site-packages/astropy/utils/data.py", line 622, in get_pkg_data_filename
    return download_file(conf.dataurl + data_name, cache=True,
  File "/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/10602/lib/python3.8/site-packages/astropy/utils/data.py", line 1165, in download_file
    f_name, hexdigest = _download_file_from_source(
  File "/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/10602/lib/python3.8/site-packages/astropy/utils/data.py", line 1000, in _download_file_from_source
    with urlopener.open(req, timeout=timeout) as remote:
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/urllib/request.py", line 525, in open
    response = self._open(req, data)
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/urllib/request.py", line 542, in _open
    result = self._call_chain(self.handle_open, protocol, protocol +
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/urllib/request.py", line 1346, in http_open
    return self.do_open(http.client.HTTPConnection, req)
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/urllib/request.py", line 1321, in do_open
    r = h.getresponse()
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/http/client.py", line 1322, in getresponse
    response.begin()
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/http/client.py", line 303, in begin
    version, status, reason = self._read_status()
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/http/client.py", line 264, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/home/docs/.pyenv/versions/3.8.0/lib/python3.8/socket.py", line 669, in readinto
    return self._sock.recv_into(b)
ConnectionResetError: [Errno 104] Connection reset by peer
/home/docs/checkouts/readthedocs.org/user_builds/astropy/checkouts/10602/docs/visualization/rgb.rst:83: WARNING: Exception occurred in plotting rgb-3
 from /home/docs/checkouts/readthedocs.org/user_builds/astropy/checkouts/10602/docs/visualization/rgb.rst:
  File "/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/10602/lib/python3.8/site-packages/matplotlib/sphinxext/plot_directive.py", line 472, in run_code
    exec(code, ns)
NameError: name 'i' is not defined
```

astropy/nddata/tests/test_utils.py
```diff
@@ -292,6 +292,14 @@ def test_extract_array_return_pos():
         assert new_pos == (expected, )
 
 
+def test_extract_array_nan_fillvalue():
+    if minversion(np, '1.20'):
```
pllim, line 296: Sure you don't want to use `packaging` here like what Matt C did to the other tests for the sake of consistency?

astropy/nddata/utils.py
```diff
@@ -243,7 +245,13 @@ def extract_array(array_large, shape, position, mode='partial',
     # Extracting on the edges is presumably a rare case, so treat special here
     if (extracted_array.shape != shape) and (mode == 'partial'):
         extracted_array = np.zeros(shape, dtype=array_large.dtype)
-        extracted_array[:] = fill_value
+        try:
+            extracted_array[:] = fill_value
+        except ValueError:
```
pllim, line 250: I think assuming all `ValueError` is *that* error might be dangerous. Should at least check for the error message before re-raising? Can we guarantee that *only* this kind of error will produce `ValueError` here?

pllim:
And given the decision in https://github.com/astropy/astropy/pull/10602#issuecomment-664588035 , should the milestone be moved back to 4.0.x?

larrybradley:
@pllim Can an API change go in a bugfix release?

astropy/nddata/tests/test_utils.py
larrybradley: OK, I switched to use `packaging`.

astropy/nddata/utils.py
larrybradley: I updated the error message to be more generic (e.g. `fill_value='a string'` also raises `ValueError`).  The new error message should cover any `ValueError` raised here because of this assignment.

larrybradley:
I changed the milestone back to 4.0.2 and moved the changelog entry.

CHANGES.rst
```diff
@@ -1066,6 +1066,9 @@ Other Changes and Additions
 - Updated the bundled CFITSIO library to 3.480. See
   ``cextern/cfitsio/docs/changes.txt`` for additional information. [#10256]
 
+- ``extract_array`` raises a ``ValueError`` if the data type of the
```
pllim, line 1069: Note for future self: This was discussed offline. In the end, it is neither API change or bug fix, as it clarifies the error message but does not change any behavior.

astropy/nddata/utils.py
mhvk: How about keeping the exception chained?  I've also see usage like:
```
except ValueError as exc:
    exc.args += ('Did you perhaps try to insert a float into an integer array',)
    raise exc
```
(But really, we're perhaps going too far with this!!)

pllim:
Yikes, I think failures caused by new `pandas` release. Unrelated but it also blocks full CI, including the numpy-dev job I was gonna check.

astropy/nddata/utils.py
pllim: Hmm `+=` isn't a bad idea. We can see the original error for debugging purpose in the future...

pllim:
CI should be fixed now. Please rebase. Thanks for your patience! 🤞

pllim:
@larrybradley , please rebase. Thanks!

larrybradley:
I'm updating this PR to append to the error message.

larrybradley:
Rebased.

PR #10603, opened 2020-07-26
Use the handy sys.modules.keys() to iterate over sys.modules

jerobado:
### Description

Since `sys.modules` is already a dictionary, we can use its `sys.modules`**.keys()** to iterate over each key (module) without the need to use `list()`. This also improves the for-loop's readability.

Let me know your thoughts.

Fixes #<Issue Number> N/A

### DIFF
--- a/astropy/__init__.py
+++ b/astropy/__init__.py
@@ -231,10 +231,8 @@ def _initialize_astropy():
         # exception logger before we raise the import error:
         _teardown_log()
 
-        # Roll back any astropy sub-modules that have been imported thus
-        # far
-
-        for key in list(sys.modules):
+        # Roll back any astropy sub-modules that have been imported thus far
+        for key in sys.modules.keys():
             if key.startswith('astropy.'):
                 del sys.modules[key]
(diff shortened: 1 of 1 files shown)


astropy/__init__.py
```diff
@@ -231,10 +231,8 @@ def _rollback_import(message):
         # exception logger before we raise the import error:
         _teardown_log()
 
-        # Roll back any astropy sub-modules that have been imported thus
-        # far
-
-        for key in list(sys.modules):
+        # Roll back any astropy sub-modules that have been imported thus far
+        for key in sys.modules.keys():
```
mhvk, line 235: The reason the list was here is that we're deleting modules while iterating over them; hence, we need to be sure that `.keys()` doesn't change during the loop.  See https://stackoverflow.com/questions/5384914/how-to-delete-items-from-a-dictionary-while-iterating-over-it

jerobado: I think I missed looking in that angle. Anyway, thanks for the review @mhvk , appreciate it 👍.

PR #10610, opened 2020-07-29
Fix SigmaClip output data values for masked=True, copy=True, axis is not None

### LINKED ISSUE #10605: Inconsistent and "lossy" return values from `SigmaClip`
### Description
In June 2018, `SigmaClip.__call__` was split into `_sigmaclip_noaxis` and `_sigmaclip_withaxis` branches. The former branch optimizes the simplest cases by removing deviant pixels during processing, whilst the latter [now tracks deviant values by substituting `NaN`s in the data array instead of using a `MaskedArray`](https://github.com/astropy/astropy/commit/96f20287634c40faef7a43e4ee90140010e05652). This has led to a couple of probably-unintended changes in API when using the `axis` parameter:

* With `masked=True` (the default & previous behaviour), the `MaskedArray` that gets returned now has `NaN` values in its main data array. These special values are unnecessary because the corresponding elements are already masked. Moreover, changing them to `NaN` discards useful information about what the masked values were. This is undesirable eg. when plotting the data with deviant values annotated.

* When providing an integer array as input, it gets upcast to `float64`, which also seems unexpected and shouldn't happen implicitly if avoidable.

These are **inconsistent with the behaviour for `axis=None` and for `copy=False`** -- options which shouldn't affect what the return value looks like.

Specifying `masked=False` also produces `NaN`s -- but in that case the behaviour is a documented new form of usage, where `NaN` values are used in place of a mask (and thus the return type has to be some kind of float).

### Expected behavior
I expect to get back a `MaskedArray` that's the same as the input `data` but with a mask attached. I don't expect the output `dtype` or the presence of `NaN` values to depend on the `axis` or `copy` parameters.

### Actual behavior
With `masked=True` and `axis=0`, the `MaskedArray` returned has `NaN` at masked pixels and is upcast to `float64`.

### Steps to Reproduce

Note that `axis=None` and `axis=0` mean exactly the same thing here (clip over the only axis), but they trigger different branches in the code.
```python
import numpy as np
from astropy.stats import sigma_clip

data_in = np.array([-2, 5, -5, -6, 20, 14, 1])

print('In               ', data_in)

data_out = sigma_clip(data_in, sigma=1.5, maxiters=3)

print('\nNoaxis_data      ', data_out.data)
print('Noaxis_mask      ', data_out.mask)

data_out = sigma_clip(data_in, sigma=1.5, maxiters=3, axis=0)

print('\nWithaxis_data    ', data_out.data)
print('Withaxis_mask    ', data_out.mask)

data_out = sigma_clip(data_in, sigma=1.5, maxiters=3, axis=0, copy=False)

print('\nWithaxis_no_copy ', data_out.data)

data_out = sigma_clip(data_in, sigma=1.5, maxiters=3, axis=0, masked=False)

print('\nWithaxis_no_mask ', data_out)
```
```
In                [-2  5 -5 -6 20 14  1]

Noaxis_data       [-2  5 -5 -6 20 14  1]
Noaxis_mask       [False  True False False  True  True False]

Withaxis_data     [-2. nan -5. -6. nan nan  1.]
Withaxis_mask     [False  True False False  True  True False]

Withaxis_no_copy  [-2  5 -5 -6 20 14  1]

Withaxis_no_mask  [-2. nan -5. -6. nan nan  1.]
```

### System Details
Linux-3.10.0-1127.13.1.el7.x86_64-x86_64-with-centos-7.8.2003-Core
[GCC 7.3.0]
Numpy 1.17.2
astropy 4.1.dev1967+g02fb00a
Scipy 1.3.1
Matplotlib 3.1.1

### Solution

This can be fixed in 1 line by adding `filtered_data = np.ma.MaskedArray(data, filtered_data.mask, copy=True)` after [L381](https://github.com/astropy/astropy/blob/055c0ad984869538a0d6fa97391ff5486ec922fd/astropy/stats/sigma_clipping.py#L381) of `_sigmaclip_withaxis`. I'm already working on a PR for this same class and can just include the fix there, following any discussion here. Thanks.

jehturner:
### Description

This fixes #10605 by re-copying the input data into the final `MaskedArray` at the end of `_sigmaclip_withaxis`. It adds a regression test to confirm that the output data (as opposed to mask) are the same as the input for `masked=True` with different `axis` and `copy` arguments and also that the input array gets copied or referenced as appropriate.

I've made one change WRT what I proposed in #10605, which is to add `copy=False` at L381 to avoid an additional copy operation. Does that look good? Thanks.

### DIFF
--- a/astropy/stats/sigma_clipping.py
+++ b/astropy/stats/sigma_clipping.py
@@ -378,7 +378,9 @@ class SigmaClip:
         if masked:
             # create an output masked array
             if copy:
-                filtered_data = np.ma.masked_invalid(filtered_data)
+                filtered_data = np.ma.masked_invalid(filtered_data, copy=False)
+                filtered_data = np.ma.MaskedArray(data, filtered_data.mask,
+                                                  copy=True)
             else:
                 # ignore RuntimeWarnings for comparisons with NaN data values
                 with np.errstate(invalid='ignore'):
--- a/astropy/stats/sigma_clipping.py
+++ b/astropy/stats/sigma_clipping.py
@@ -378,8 +378,7 @@ class SigmaClip:
         if masked:
             # create an output masked array
             if copy:
-                filtered_data = np.ma.masked_invalid(filtered_data, copy=False)
-                filtered_data = np.ma.MaskedArray(data, filtered_data.mask,
+                filtered_data = np.ma.MaskedArray(data, ~np.isfinite(filtered_data),
                                                   copy=True)
             else:
                 # ignore RuntimeWarnings for comparisons with NaN data values
--- a/astropy/stats/tests/test_sigma_clipping.py
+++ b/astropy/stats/tests/test_sigma_clipping.py
@@ -302,3 +302,44 @@ def test_sigma_clippped_stats_all_masked():
     mask = arr < 20
     result = sigma_clipped_stats(arr, mask=mask)
     assert result == (np.ma.masked, np.ma.masked, np.ma.masked)
+
+
+def test_sigma_clip_masked_data_values():
+    """
+    Test that the data values & type returned by sigma_clip are the same as
+    its input when using masked=True (rather than being upcast to float64 &
+    containing NaNs as in issue #10605) and also that the input data get
+    copied or referenced as appropriate.
+    """
+
+    data = np.array([-2, 5, -5, -6, 20, 14, 1])
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=None, masked=True,
+                        copy=True)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
+    assert not np.shares_memory(result.data, data)
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=None, masked=True,
+                        copy=False)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
+    assert np.shares_memory(result.data, data)
+    # (The fact that the arrays share memory probably also means they're the
+    # same, but doesn't strictly prove it, eg. one could be reversed.)
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=0, masked=True,
+                        copy=True)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
+    assert not np.shares_memory(result.data, data)
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=0, masked=True,
+                        copy=False)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
+    assert np.shares_memory(result.data, data)
--- a/astropy/stats/tests/test_sigma_clipping.py
+++ b/astropy/stats/tests/test_sigma_clipping.py
@@ -318,14 +318,14 @@ def test_sigma_clip_masked_data_values():
                         copy=True)
 
     assert result.dtype == data.dtype
-    assert np.all(result.data == data)
+    assert_equal(result.data, data)
     assert not np.shares_memory(result.data, data)
 
     result = sigma_clip(data, sigma=1.5, maxiters=3, axis=None, masked=True,
                         copy=False)
 
     assert result.dtype == data.dtype
-    assert np.all(result.data == data)
+    assert_equal(result.data, data)
     assert np.shares_memory(result.data, data)
     # (The fact that the arrays share memory probably also means they're the
     # same, but doesn't strictly prove it, eg. one could be reversed.)
@@ -334,12 +334,12 @@ def test_sigma_clip_masked_data_values():
                         copy=True)
 
     assert result.dtype == data.dtype
-    assert np.all(result.data == data)
+    assert_equal(result.data, data)
     assert not np.shares_memory(result.data, data)
 
     result = sigma_clip(data, sigma=1.5, maxiters=3, axis=0, masked=True,
                         copy=False)
 
     assert result.dtype == data.dtype
-    assert np.all(result.data == data)
+    assert_equal(result.data, data)
     assert np.shares_memory(result.data, data)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -978,7 +978,7 @@ astropy.stats
 - Fixed an API regression where ``SigmaClip.__call__`` would convert masked
   elements to ``nan`` and upcast the dtype to ``float64`` in its output
   ``MaskedArray`` when using the ``axis`` parameter along with the defaults
-  ``masked=True`` and ``copy=True``. [?]
+  ``masked=True`` and ``copy=True``. [#10610]
 
 - Fixed an issue where fully masked ``MaskedArray`` input to
   ``sigma_clipped_stats`` gave incorrect results. [#10099]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -975,10 +975,15 @@ astropy.samp
 astropy.stats
 ^^^^^^^^^^^^^
 
+- Fixed an API regression where ``SigmaClip.__call__`` would convert masked
+  elements to ``nan`` and upcast the dtype to ``float64`` in its output
+  ``MaskedArray`` when using the ``axis`` parameter along with the defaults
+  ``masked=True`` and ``copy=True``. [?]
+
 - Fixed an issue where fully masked ``MaskedArray`` input to
   ``sigma_clipped_stats`` gave incorrect results. [#10099]
 
-- Fixed an issued where ``sigma_clip`` and ``SigmaClip.__call__``
+- Fixed an issue where ``sigma_clip`` and ``SigmaClip.__call__``
   would return a masked array instead of a ``ndarray`` when
   ``masked=False`` and the input was a full-masked ``MaskedArray``.
   [#10099]


jehturner:
Do you happen to know anything about all those tox conftest errors, @pllim? Presumably they have nothing to do with this PR.

saimn:
@jehturner - Yes, CI failures are unrelated, caused by the recent pytest 6.0 release.

pllim:
Yeah, we need to pin pytest<6 in ci-helpers, merge that real fix in doctestplus, release doctestplus, and then unpin it in ci-helpers. 😢  I can't do the merging nor release, so this will remain broken until someone who can does it. Maybe I should open a proper issue even though there is nothing much we can do here in the core.

jehturner:
Thanks, so should we re-start the CI now, @pllim?

astropy/stats/sigma_clipping.py
```diff
@@ -378,7 +378,9 @@ def _sigmaclip_withaxis(self, data, axis=None, masked=True,
         if masked:
             # create an output masked array
             if copy:
-                filtered_data = np.ma.masked_invalid(filtered_data)
+                filtered_data = np.ma.masked_invalid(filtered_data, copy=False)
+                filtered_data = np.ma.MaskedArray(data, filtered_data.mask,
+                                                  copy=True)
```
larrybradley, line 383:
```suggestion
                filtered_data = np.ma.MaskedArray(data, ~np.isfinite(filtered_data),
                                                  copy=True)
```

This is ~10x faster (tested with a 1k x 1k array) than using `np.ma.masked_invalid`.

astropy/stats/tests/test_sigma_clipping.py
```diff
@@ -302,3 +302,44 @@ def test_sigma_clippped_stats_all_masked():
...
+    copied or referenced as appropriate.
+    """
+
+    data = np.array([-2, 5, -5, -6, 20, 14, 1])
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=None, masked=True,
+                        copy=True)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
```
larrybradley, line 321:
```suggestion
    assert_equal(result.data, data)
```

Please use `np.testing.assert_equal` instead (https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html).  `arr1 == arr2` elementwise comparison is deprecated and it also doesn't catch issues with NaNs (`np.nan == np.nan` returns False) or mismatched shapes, etc.

astropy/stats/tests/test_sigma_clipping.py
```diff
@@ -302,3 +302,44 @@ def test_sigma_clippped_stats_all_masked():
...
+    assert not np.shares_memory(result.data, data)
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=None, masked=True,
+                        copy=False)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
```
larrybradley, line 328:
```suggestion
    assert_equal(result.data, data)
```

astropy/stats/tests/test_sigma_clipping.py
```diff
@@ -302,3 +302,44 @@ def test_sigma_clippped_stats_all_masked():
...
+    assert np.shares_memory(result.data, data)
+    # (The fact that the arrays share memory probably also means they're the
+    # same, but doesn't strictly prove it, eg. one could be reversed.)
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=0, masked=True,
+                        copy=True)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
```
larrybradley, line 337:
```suggestion
    assert_equal(result.data, data)
```

astropy/stats/tests/test_sigma_clipping.py
```diff
@@ -302,3 +302,44 @@ def test_sigma_clippped_stats_all_masked():
...
+    assert not np.shares_memory(result.data, data)
+
+    result = sigma_clip(data, sigma=1.5, maxiters=3, axis=0, masked=True,
+                        copy=False)
+
+    assert result.dtype == data.dtype
+    assert np.all(result.data == data)
```
larrybradley, line 344:
```suggestion
    assert_equal(result.data, data)
```

pllim:
Re: CI -- Not quite yet. Need to figure out how to fix pandas doctest until pandas can figure out why they have a regression.

astropy/stats/sigma_clipping.py
jehturner: OK. But surely this is going to be true for the other instances of `masked_invalid` as well? The 10x difference must be in the existing calls, rather than the line I added, so perhaps we should change all of them? As far as I can see, the only differences are that `masked_invalid` uses `where` and that it ORs the result with any existing mask (if I understand the docs). The latter thing might need doing separately at L346 (but that presumably can't account for most of the 10x). The other 2 instances could probably just be replaced in the same way as above?

astropy/stats/tests/test_sigma_clipping.py
jehturner: OK. Yes, I had done something closer to that (`assert array_equal()`) in my other PR. This was something I thought about changing then forgot; obviously I don't write enough tests, hehe.

jehturner:
I had better squash your `assert_equal` commits, @larrybradley, since they all make the same change and I inadvertently accepted one with a default commit message. Not sure if you want me to squash anything else, but the other commits are more distinct (maybe the change log)?

larrybradley:
I think just the 3 `assert_equal` should be squashed.  Thanks.

astropy/stats/sigma_clipping.py
saimn: `masked_invalid` is using `np.isfinite` so I'm surprised by the difference. And one thing that `masked_invalid` does additionally is to combine the masks if the input is a masked array. I think we need to keep that, as here `data` may be a masked array ?

jehturner: Me too. Maybe constructing the indices in `masked_where` is fiddly (and I think unnecessary)? I think we need to keep the mask combination as you say at L346, but after that the data get converted to an `ndarray` for intermediate processing.

jehturner: I should really just time it when I get a minute (but I have to finish other things this week).

saimn: Yes actually the masked values in the input data have been set to NaN, so there is no need to combine with the input mask.

pllim:
Re: CI -- hopefully fixed now, please rebased! 🤞

astropy/stats/sigma_clipping.py
larrybradley: ` filtered_data = np.ma.masked_invalid(filtered_data, copy=False)` above was being used *only* to get a new mask (that is then used in the next line) -- effectively it was being used only to do `~np.isfinite(filtered_data)` (`filtered_data` is a plain ndarray), so I suggested that we just use `np.isfinite` directly to get the new mask.

My timing test was a simple comparison of only `np.ma.masked_invalid` vs. `np.isfinite` on a 1k x 1k array.  The former took 2.1 ms and the latter 311 us (~7x slower).  The difference is likely due to the creation/initialization of a `MaskedArray` object.

larrybradley:
@pllim I think a rebase is required to have CI pass :).

Actually, in my review I missed that the `~np.isfinite` change wasn't added.  I think that is better, but I won't hold up this PR.

jehturner:
I still need to change the `isfinite` thing though. Does that tag merge the PR automatically? I was going to change all but one of the other occurrences of `masked_invalid` (let me just do that), so does @larrybradley want to check that first?

pllim:
Don't worry, our bot is not at the point to auto-merge yet, just auto-close. 😅 

But I just want to clarify in case some other maintainers come along, see the approval, and thought it is good to merge.

larrybradley:
I'll re-review before merging.

astropy/stats/sigma_clipping.py
jehturner: Oh! I think I understand what you're getting at now. I thought you were saying that `MaskedArray` with `np.isfinite` was 10x as fast as `masked_invalid` plus `MaskedArray`, which would have implied that `masked_invalid` will be slow everywhere. But you're just saying that, assuming there's always a `MaskedArray`, adding a `masked_invalid` takes 10x as long as adding an `isfinite` to it (so the 2 lines probably take ~twice as long as your suggested one). OK, I'll just accept your suggestion then...

pllim:
What now?! 🤯 
```
[1musing tox.ini: /home/travis/build/astropy/astropy/tox.ini (pid 7355)
[0mcould not satisfy requires PackageNotFoundError('tox-pypi-filter')
...
Looking in indexes: http://localhost:59275
Processing ./.tox/.tmp/package/1/astropy-4.2.dev459+g0b77237e2.tar.gz
  Installing build dependencies: started
  Installing build dependencies: finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /home/travis/build/astropy/astropy/.tox/py38-test-alldeps/bin/python /home/travis/build/astropy/astropy/.tox/py38-test-alldeps/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-pqxypr8h/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i http://localhost:59275 -- setuptools setuptools_scm wheel cython==0.29.14 jinja2==2.10.3 oldest-supported-numpy extension-helpers
       cwd: None
  Complete output (8 lines):
  Looking in indexes: http://localhost:59275
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPConnection object at 0x7f00409aa0d0>: Failed to establish a new connection: [Errno 111] Connection refused')': /setuptools/
  WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPConnection object at 0x7f00409cd7f0>: Failed to establish a new connection: [Errno 111] Connection refused')': /setuptools/
  WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPConnection object at 0x7f00409cd2e0>: Failed to establish a new connection: [Errno 111] Connection refused')': /setuptools/
  WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPConnection object at 0x7f00409cd100>: Failed to establish a new connection: [Errno 111] Connection refused')': /setuptools/
  WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPConnection object at 0x7f00409cd820>: Failed to establish a new connection: [Errno 111] Connection refused')': /setuptools/
  ERROR: Could not find a version that satisfies the requirement setuptools (from versions: none)
  ERROR: No matching distribution found for setuptools
```

larrybradley:
I think that is saying that `setuptools` wasn't found....transient network failure?

pllim:
Better be... there is only so much CI fixing I want to do in a week.

jehturner:
I've just rebased this anyway (which took a few minutes as I messed up slightly by doing things locally & on github at the same time), so it should have re-started.

jehturner:
I notice that Larry's suggestion wraps to 84 characters, FWIW, but I don't want to keep @pllim watching the CI all day :grin:. Otherwise, it should be finished now.

pllim:
Larry, why 84 and not 88? 😉

jehturner:
That's just the length of the line with characters added.

larrybradley:
Roads?!  Where we're going we don't need roads!  :delorian:  🤣

larrybradley:
But seriously -- I use the 79 character limit, but github's suggestion interface doesn't show the column numbers so I was flying blind. We can fix later.

jehturner:
Good thing to watch out for...

pllim:
Tests passed! The 3 remote data failures are unrelated.

larrybradley:
Thanks, @jehturner!

PR #10609, opened 2020-07-28
TST: doctest compat with pandas 1.1.0

pllim:
### Description

This pull request is to address doctest failures with `pandas` 1.1.0. Given we don't test for older `pandas`, should be okay? 🤞 

Fixes #10608

p.s. Milestoned to LTS so the CI there would pass too, but feel free to re-milestone if that is not applicable.

Issue #10608, opened 2020-07-28
TST: Table doctest failures with pandas 1.1.0


pllim:
`pandas` 1.1.0 was released about 6 hours ago. Now we see this failure "Python 3.8 with all optional dependencies" job in Travis CI initial stage.
```
============================= test session starts ==============================
platform linux -- Python 3.8.1, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
cachedir: .tox/py38-test-alldeps/.pytest_cache
Running tests with Astropy version 4.2.dev455+g4d853f887.
Running tests in astropy docs.
Date: 2020-07-28T19:37:45
Platform: Linux-5.3.0-1029-gcp-x86_64-with-glibc2.2.5
Executable: /home/travis/build/astropy/astropy/.tox/py38-test-alldeps/bin/python
Full Python Version: 
3.8.1 (default, Dec 20 2019, 10:06:11) 
[GCC 7.4.0]
encodings: sys: utf-8, locale: UTF-8, filesystem: utf-8
byteorder: little
float info: dig: 15, mant_dig: 15
Package versions: 
Numpy: 1.19.1
Scipy: 1.5.2
Matplotlib: 3.3.0
h5py: 2.10.0
Pandas: 1.1.0
PyERFA: 1.7.0
Using Astropy options: remote_data: none.
rootdir: /home/travis/build/astropy/astropy, inifile: setup.cfg
plugins: filter-subpackage-0.1.1, hypothesis-5.23.3, xdist-1.34.0, openfiles-0.5.0, remotedata-0.3.2, forked-1.3.0, arraydiff-0.3, cov-2.10.0, asdf-2.7.0, doctestplus-0.7.0, astropy-header-0.1.2
collected 15405 items                                                

...

=================================== FAILURES ===================================
_______________ [doctest] astropy.table.table.Table.from_pandas ________________
3487           >>> import numpy as np
3488           >>> import pandas as pd
3489           >>> from astropy.table import QTable
3490 
3491           >>> time = pd.Series(['1998-01-01', '2002-01-01'], dtype='datetime64[ns]')
3492           >>> dt = pd.Series(np.array([1, 300], dtype='timedelta64[s]'))
3493           >>> df = pd.DataFrame({'time': time})
3494           >>> df['dt'] = dt
3495           >>> df['x'] = [3., 4.]
3496           >>> with pd.option_context('display.max_columns', 20):
Differences (unified diff with -expected +actual):
[...]

PR #10614, opened 2020-07-30
TST: Temporarily pin to pandas<1.1 for table doctest

pllim:
### Description

This pull request is to pin to `pandas<1.1` for `table.py` doctest to prevent failure in other PRs until pandas-dev/pandas#35463 is resolved one way or another upstream.

Hack for #10608

Close #10609


pllim:
`mpldev` failure is unrelated.

Issue #10612, opened 2020-07-29
TST: Unpin pytest when pytest-doctestplus 0.8 is released


pllim:
### Description

CI is breaking bad. (Get the TV show reference? 😉 😉 )

### Expected behavior

CI to pass when there is nothing wrong with the relevant code (in PR or `master`).

### Actual behavior

Breaking bad.

### Steps to fix

- [x] Pin to `pytest<6` in astropy/ci-helpers#453 .
- [x] Fix doctest errors caused by `pandas` (#10608), see proposed patch at #10609 .
- [x] Fix astropy/pytest-doctestplus#118 (fixed by astropy/pytest-doctestplus#120, thanks to @saimn ).
- [x] Release a new version of `pytest-doctestplus`.
- [x] Undo the pin; i.e., revert astropy/ci-helpers#453, see astropy/ci-helpers#455

pllim:
Nothing else we can do here. astropy/ci-helpers#455 should resolve this.

PR #10620, opened 2020-07-31
DO NOT MERGE: Test unpin of pytest<6

pllim:
### Description

This pull request is a sanity check for astropy/ci-helpers#455 .

Also see #10612


pllim:
Ignore the Windows failure. I used Unix commands to change directory. The Linux jobs passing are good enough of an indication that astropy/ci-helpers#455 works.

pllim:
This has proven all it needs to prove. See CI in the first commit. Nothing else to be done here.

PR #10621, opened 2020-08-03
Fix mpldev build in v4.1.x branch

astrofrog:
This backports just a few lines from #10504 which should resolve the failures in the mpldev job. Since this is not a full backport, we should keep that PR milestoned as v4.2.


astrofrog:
Looks like this did the trick for fixing mpldev and can be merged

pllim:
Given the CI failures and lack of milestone, I'll let @eteq handle this one since he is backporting stuff anyway.

eteq:
LGTM. Hopefully we will not later need to backport other parts of  #10504  since that'll get pretty confusing, but I think that's unlikely so it seems like this is the best guess.

PR #10624, opened 2020-08-04
Galactocentric frame defaults registry

nstarman:
### Description
`galactocentric_frame_defaults` allows for a user-defined set of parameters to be manually set. This PR extends that capability to a registry system such that parameter sets may be named and get/set just like the built-in option (pre-v4.0, v4.0).

```python
import astropy.units as u
from astropy.coordinates import galactocentric_frame_defaults

galactocentric_frame_defaults.register(
    name="test",
    value={"galcen_coord": ICRS(ra=266.0 * u.degree, dec=-29 * u.degree),
                "galcen_distance": 8.1 * u.kpc,
                "galcen_v_sun": r.CartesianDifferential([12.9, 245.6, 7.8] * u.km / u.s),
                "z_sun": 21.0 * u.pc,
                 "roll": 0 * u.deg},
    references=None
)

with galactocentric_frame_defaults.set("test"):
    params = galactocentric_frame_defaults.get()
```

Additional metadata, such as the value standard deviations,  is supported by the ``_state`` dictionary attribute, whose items are also exposed to the user by key access on the main class.

```python
galactocentric_frame_defaults._registry["test"]["stdv"] = {
    "galcen_coord": {"ra": 1 * u.deg, "dec": 1 * u.deg},
    "galcen_distance": 0.1 * u.kpc,
    "galcen_v_sun": {"x": 2 * u.km / u.s, "y": 1 * u.km / u.s, "z": 1* u.km / u.s},
    "z_sun": 1 * u.pc,
    "roll": 0 * u.deg,
}

galactocentric_frame_defaults.set("test")

test_stdv = galactocentric_frame_defaults["stdv"]

```

If metadata is added to one parameter set, for consistency it should be added to all. Though the built-in options are read-only for "value" and "references", they permit additional fields, such as the standard deviation.

```python
galactocentric_frame_defaults._registry["v4.0"]["stdv"] = {
    "galcen_coord": {"ra": 1 * u.deg, "dec": 1 * u.deg},
    "galcen_distance": 0.1 * u.kpc,
    "galcen_v_sun": {"x": 2 * u.km / u.s, "y": 1 * u.km / u.s, "z": 1* u.km / u.s},
    "z_sun": 1 * u.pc,
    "roll": 0 * u.deg,
}

galactocentric_frame_defaults.set("v4.0")

stdv = galactocentric_frame_defaults["stdv"]

```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,7 @@ astropy.coordinates
   Depending on needed precision and the obstime array in question, speed ups
   reach factors of 10x to >100x. [#10647]
 
+- ``galactocentric_frame_defaults`` can now also be used as a registry, with user-defined parameter values and metadata. [#10624]
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
--- a/astropy/__init__.py
+++ b/astropy/__init__.py
@@ -170,25 +170,7 @@ class base_constants_version(ScienceState):
         if 'astropy.constants' in sys.modules:
             raise RuntimeError('astropy.constants is already imported')
 
-        class _Context:
-            def __init__(self, parent, value):
-                self._value = value
-                self._parent = parent
-
-            def __enter__(self):
-                pass
-
-            def __exit__(self, type, value, tb):
-                self._parent._value = self._value
-
-            def __repr__(self):
-                return ('<ScienceState {}: {!r}>'
-                        .format(self._parent.__name__, self._parent._value))
-
-        ctx = _Context(cls, cls._value)
-        value = cls.validate(value)
-        cls._value = value
-        return ctx
+        return super().set(cls, value)
 
 
 class physical_constants(base_constants_version):
--- a/astropy/coordinates/builtin_frames/galactocentric.py
+++ b/astropy/coordinates/builtin_frames/galactocentric.py
@@ -1,11 +1,15 @@
 # -*- coding: utf-8 -*-
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import copy
+from collections.abc import MappingView
+from types import MappingProxyType
+
 import numpy as np
 
 from astropy import units as u
 from astropy.utils.state import ScienceState
-from astropy.utils.decorators import format_doc
+from astropy.utils.decorators import format_doc, classproperty, deprecated
 from astropy.coordinates.angles import Angle
 from astropy.coordinates.matrix_utilities import rotation_matrix, matrix_product, matrix_transpose
 from astropy.coordinates import representation as r
@@ -30,6 +34,25 @@ __all__ = ['Galactocentric']
 _ROLL0 = Angle(58.5986320306*u.degree)
 
 
+class _StateProxy(MappingView):
+    """
+    `~collections.abc.MappingView` with a read-only ``getitem`` through
+    `~types.MappingProxyType`.
+
+    """
+
+    def __init__(self, mapping):
+        super().__init__(mapping)
+        self._mappingproxy = MappingProxyType(self._mapping)  # read-only
+
+    def __getitem__(self, key):
+        """Read-only ``getitem``."""
+        return self._mappingproxy[key]
+
+    def __deepcopy__(self, memo):
+        return copy.deepcopy(self._mapping, memo=memo)
+
+
 class galactocentric_frame_defaults(ScienceState):
     """This class controls the global setting of default values for the frame
     attributes in the `~astropy.coordinates.Galactocentric` frame, which may be
@@ -41,12 +64,29 @@ class galactocentric_frame_defaults(ScienceState):
     ``Galactocentric`` or ``Galactocentric()`` with no explicit arguments.
 
     This class controls the parameter settings by specifying a string name,
-    which can be one of:
+    with the following pre-specified options:
 
     - 'pre-v4.0': The current default value, which sets the default frame
       attribute values to their original (pre-astropy-v4.0) values.
     - 'v4.0': The attribute values as updated in Astropy version 4.0.
     - 'latest': An alias of the most recent parameter set (currently: 'v4.0')
+
+    Alternatively, user-defined parameter settings may be registered, with
+    :meth:`~astropy.coordinates.galactocentric_frame_defaults.register`,
+    and used identically as pre-specified parameter sets. At minimum,
+    registrations must have unique names and a dictionary of parameters
+    with keys "galcen_coord", "galcen_distance", "galcen_v_sun", "z_sun",
+    "roll". See examples below.
+
+    This class also tracks the references for all parameter values in the
+    attribute ``references``, as well as any further information the registry.
+    The pre-specified options can be extended to include similar
+    state information as user-defined parameter settings -- for example, to add
+    parameter uncertainties.
+
+    The preferred method for getting a parameter set and metadata, by name, is
+    :meth:`~galactocentric_frame_defaults.get_from_registry` since
+    it ensures the immutability of the registry.
 
     See :ref:`astropy-coordinates-galactocentric-defaults` for more information.
 
@@ -82,73 +122,157 @@ class galactocentric_frame_defaults(ScienceState):
         <Galactocentric Frame (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
             (266.4051, -28.936175)>, galcen_distance=8.0 kpc, galcen_v_sun=(11.1, 232.24, 7.25) km / s, z_sun=27.0 pc, roll=0.0 deg)>
 
+    Additional parameter sets may be registered, for instance to use the
+    Dehnen & Binney (1998) measurements of the solar motion. We can also
+    add metadata, such as the 1-sigma errors. In this example we will modify
+    the required key "parameters", change the recommended key "references" to
+    match "parameters", and add the extra key "error" (any key can be added)::
+
+        >>> state = galactocentric_frame_defaults.get_from_registry("v4.0")
+        >>> state["parameters"]["galcen_v_sun"] = (10.00, 225.25, 7.17) * (u.km / u.s)
+        >>> state["references"]["galcen_v_sun"] = "https://ui.adsabs.harvard.edu/full/1998MNRAS.298..387D"
+        >>> state["error"] = {"galcen_v_sun": (0.36, 0.62, 0.38) * (u.km / u.s)}
+        >>> galactocentric_frame_defaults.register(name="DB1998", **state)
+
+    Just as in the previous examples, the new parameter set can be retrieved with::
+
+        >>> state = galactocentric_frame_defaults.get_from_registry("DB1998")
+        >>> print(state["error"]["galcen_v_sun"])  # doctest: +FLOAT_CMP
+        [0.36 0.62 0.38] km / s
+
     """
 
     _latest_value = 'v4.0'
-    _references = None
     _value = None
+    _references = None
+    _state = dict()  # all other data
+
+    # Note: _StateProxy() produces read-only view of enclosed mapping.
+    _registry = {
+        "v4.0": {
+            "parameters": _StateProxy(
+                {
+                    "galcen_coord": ICRS(
+                        ra=266.4051 * u.degree, dec=-28.936175 * u.degree
+                    ),
+                    "galcen_distance": 8.122 * u.kpc,
+                    "galcen_v_sun": r.CartesianDifferential(
+                        [12.9, 245.6, 7.78] * (u.km / u.s)
+                    ),
+                    "z_sun": 20.8 * u.pc,
+                    "roll": 0 * u.deg,
+                }
+            ),
+            "references": _StateProxy(
+                {
+                    "galcen_coord": "https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R",
+                    "galcen_distance": "https://ui.adsabs.harvard.edu/abs/2018A%26A...615L..15G",
+                    "galcen_v_sun": [
+                        "https://ui.adsabs.harvard.edu/abs/2018RNAAS...2..210D",
+                        "https://ui.adsabs.harvard.edu/abs/2018A%26A...615L..15G",
+                        "https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R",
+                    ],
+                    "z_sun": "https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1417B",
+                    "roll": None,
+                }
+            ),
+        },
+        "pre-v4.0": {
+            "parameters": _StateProxy(
+                {
+                    "galcen_coord": ICRS(
+                        ra=266.4051 * u.degree, dec=-28.936175 * u.degree
+                    ),
+                    "galcen_distance": 8.3 * u.kpc,
+                    "galcen_v_sun": r.CartesianDifferential(
+                        [11.1, 220 + 12.24, 7.25] * (u.km / u.s)
+                    ),
+                    "z_sun": 27.0 * u.pc,
+                    "roll": 0 * u.deg,
+                }
+            ),
+            "references": _StateProxy(
+                {
+                    "galcen_coord": "https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R",
+                    "galcen_distance": "https://ui.adsabs.harvard.edu/#abs/2009ApJ...692.1075G",
+                    "galcen_v_sun": [
+                        "https://ui.adsabs.harvard.edu/#abs/2010MNRAS.403.1829S",
+                        "https://ui.adsabs.harvard.edu/#abs/2015ApJS..216...29B",
+                    ],
+                    "z_sun": "https://ui.adsabs.harvard.edu/#abs/2001ApJ...553..184C",
+                    "roll": None,
+                }
+            ),
+        },
+    }
+
+    @classproperty  # read-only
+    def parameters(cls):
+        return cls._value
+
+    @classproperty  # read-only
+    def references(cls):
+        return cls._references
 
     @classmethod
-    def get_solar_params_from_string(cls, arg):
-        """Return Galactocentric solar parameters given string names for the
-        parameter sets.
+    def get_from_registry(cls, name: str):
         """
+        Return Galactocentric solar parameters and metadata given string names
+        for the parameter sets. This method ensures the returned state is a
+        mutable copy, so any changes made do not affect the registry state.
 
-        # Resolve the meaning of 'latest': The latest parameter set is from v4.0
+        Returns
+        -------
+        state : dict
+            Copy of the registry for the string name.
+            Should contain, at minimum:
+
+            - "parameters": dict
+                Galactocentric solar parameters
+            - "references" : Dict[str, Union[str, Sequence[str]]]
+                References for "parameters".
+                Fields are str or sequence of str.
+
+        Raises
+        ------
+        KeyError
+            If invalid string input to registry
+            to retrieve solar parameters for Galactocentric frame.
+
+        """
+        # Resolve the meaning of 'latest': latest parameter set is from v4.0
         # - update this as newer parameter choices are added
-        if arg == 'latest':
-            arg = cls._latest_value
+        if name == 'latest':
+            name = cls._latest_value
 
-        params = dict()
-        references = dict()
+        # Get the state from the registry.
+        # Copy to ensure registry is immutable to modifications of "_value".
+        # Raises KeyError if `name` is invalid string input to registry
+        # to retrieve solar parameters for Galactocentric frame.
+        state = copy.deepcopy(cls._registry[name])  # ensure mutable
 
-        # Currently, all versions use the same sky position for Sgr A*:
-        params['galcen_coord'] = ICRS(ra=266.4051*u.degree,
-                                      dec=-28.936175*u.degree)
-        references['galcen_coord'] = \
-            'https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R'
+        return state
 
-        # The roll angle is the same for both frames:
-        params['roll'] = 0 * u.deg
+    @deprecated("v4.2", alternative="`get_from_registry`")
+    @classmethod
+    def get_solar_params_from_string(cls, arg):
+        """
+        Return Galactocentric solar parameters given string names
+        for the parameter sets.
 
-        if arg == 'pre-v4.0':
-            params['galcen_distance'] = 8.3 * u.kpc
-            references['galcen_distance'] = \
-                'https://ui.adsabs.harvard.edu/#abs/2009ApJ...692.1075G'
+        Returns
+        -------
+        parameters : dict
+            Copy of Galactocentric solar parameters from registry
 
-            params['galcen_v_sun'] = r.CartesianDifferential([11.1,
-                                                              220+12.24,
-                                                              7.25]*u.km/u.s)
-            references['galcen_v_sun'] = \
-                ['https://ui.adsabs.harvard.edu/#abs/2010MNRAS.403.1829S',
-                 'https://ui.adsabs.harvard.edu/#abs/2015ApJS..216...29B']
+        Raises
+        ------
+        KeyError
+            If invalid string input to registry
+            to retrieve solar parameters for Galactocentric frame.
 
-            params['z_sun'] = 27.0 * u.pc
-            references['z_sun'] = \
-                'https://ui.adsabs.harvard.edu/#abs/2001ApJ...553..184C'
-
-        elif arg == 'v4.0':
-            params['galcen_distance'] = 8.122 * u.kpc
-            references['galcen_distance'] = \
-                'https://ui.adsabs.harvard.edu/abs/2018A%26A...615L..15G'
-
-            params['galcen_v_sun'] = r.CartesianDifferential([12.9,
-                                                              245.6,
-                                                              7.78]*u.km/u.s)
-            references['galcen_v_sun'] = \
-                ['https://ui.adsabs.harvard.edu/abs/2018RNAAS...2..210D',
-                 'https://ui.adsabs.harvard.edu/abs/2018A%26A...615L..15G',
-                 'https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R']
-
-            params['z_sun'] = 20.8 * u.pc
-            references['z_sun'] = \
-                'https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1417B'
-
-        else:
-            raise ValueError(f'Invalid string input to retrieve solar '
-                             f'parameters for Galactocentric frame: "{arg}"')
-
-        return params, references
+        """
+        return cls.get_from_registry(arg)["parameters"]
 
     @classmethod
     def validate(cls, value):
@@ -156,25 +280,61 @@ class galactocentric_frame_defaults(ScienceState):
             value = cls._latest_value
 
         if isinstance(value, str):
-            params, refs = cls.get_solar_params_from_string(value)
-            cls._references = refs
-            return params
+            state = cls.get_from_registry(value)
+            cls._references = state["references"]
+            cls._state = state
+            parameters = state["parameters"]
 
         elif isinstance(value, dict):
-            return value
+            parameters = value
 
         elif isinstance(value, Galactocentric):
             # turn the frame instance into a dict of frame attributes
-            attrs = dict()
+            parameters = dict()
             for k in value.frame_attributes:
-                attrs[k] = getattr(value, k)
-            cls._references = value.frame_attribute_references()
-            return attrs
+                parameters[k] = getattr(value, k)
+            cls._references = value.frame_attribute_references.copy()
+            cls._state = dict(parameters=parameters,
+                              references=cls._references)
 
         else:
             raise ValueError("Invalid input to retrieve solar parameters for "
                              "Galactocentric frame: input must be a string, "
                              "dict, or Galactocentric instance")
+
+        return parameters
+
+    @classmethod
+    def register(cls, name: str, parameters: dict, references=None,
+                 **meta: dict):
+        """Register a set of parameters.
+
+        Parameters
+        ----------
+        name : str
+            The registration name for the parameter and metadata set.
+        parameters : dict
+            The solar parameters for Galactocentric frame.
+        references : dict or None, optional
+            References for contents of `parameters`.
+            None becomes empty dict.
+        **meta: dict, optional
+            Any other properties to register.
+
+        """
+        # check on contents of `parameters`
+        must_have = {"galcen_coord", "galcen_distance", "galcen_v_sun",
+                     "z_sun", "roll"}
+        missing = must_have.difference(parameters)
+        if missing:
+            raise ValueError(f"Missing parameters: {missing}")
+
+        references = references or {}  # None -> {}
+
+        state = dict(parameters=parameters, references=references)
+        state.update(meta)  # meta never has keys "parameters" or "references"
+
+        cls._registry[name] = state
 
 
 doc_components = """
@@ -324,7 +484,7 @@ class Galactocentric(BaseCoordinateFrame):
         # for the solar parameters defined above
         default_params = galactocentric_frame_defaults.get()
         self.frame_attribute_references = \
-            galactocentric_frame_defaults._references.copy()
+            galactocentric_frame_defaults.references.copy()
 
         for k in default_params:
             if k in kwargs:
--- a/astropy/coordinates/tests/test_frames.py
+++ b/astropy/coordinates/tests/test_frames.py
@@ -1419,6 +1419,26 @@ def test_galactocentric_defaults(reset_galactocentric_defaults):
         else:
             assert getattr(galcen_40, k) == getattr(galcen_latest, k)
 
+    # test validate Galactocentric
+    with galactocentric_frame_defaults.set('latest'):
+        params = galactocentric_frame_defaults.validate(galcen_latest)
+        references = galcen_latest.frame_attribute_references
+        state = dict(parameters=params, references=references)
+
+        assert galactocentric_frame_defaults.parameters == params
+        assert galactocentric_frame_defaults.references == references
+        assert galactocentric_frame_defaults._state == state
+
+    # Test not one of accepted parameter types
+    with pytest.raises(ValueError):
+        galactocentric_frame_defaults.validate(ValueError)
+
+    # test parameters property
+    assert (
+        galactocentric_frame_defaults.parameters
+        == galactocentric_frame_defaults.parameters
+    )
+
 
 def test_galactocentric_references(reset_galactocentric_defaults):
     # references in the "scientific paper"-sense
--- a/astropy/utils/tests/test_state.py
+++ b/astropy/utils/tests/test_state.py
@@ -1,0 +1,25 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+from astropy.utils.state import ScienceState
+
+
+def test_ScienceState_and_Context():
+    """
+    Tests a ScienceState and spawned contexts.
+    """
+
+    class MyState(ScienceState):
+        _value = "A"
+        _state = dict(foo="bar")
+
+    state = {"foo": "bar"}
+
+    # test created ScienceState
+    assert MyState.get() == "A"
+    assert MyState.validate("B") == "B"
+    assert MyState._state == state
+
+    # test setting
+    with MyState.set("B"):
+        assert MyState.get() == "B"
+    assert MyState.get() == "A"  # test returning
--- a/docs/coordinates/galactocentric.rst
+++ b/docs/coordinates/galactocentric.rst
@@ -171,7 +171,13 @@ up-to-date list of valid names, see the docstring of
 like ``'pre-v4.0'``, which sets the default parameter values to their original
 definition (i.e. pre-astropy-v4.0) values, and ``'v4.0'``, which sets the
 default parameter values to a more modern set of measurements as updated in
-Astropy version 4.0.
+Astropy version 4.0. Also, custom sets of measurements can be registered to
+`~astropy.coordinates.galactocentric_frame_defaults` and used like the
+built-in options.
+
+`~astropy.coordinates.galactocentric_frame_defaults` also tracks the
+references (i.e. scientific papers that define the parameter values) for all
+parameter values, as well as any further specified metadata information.
 
 As with other `~astropy.utils.state.ScienceState` subclasses, the
 `~astropy.coordinates.galactocentric_frame_defaults` class can be used to
@@ -222,6 +228,21 @@ attributes that are explicitly specified::
     ...     print(Galactocentric(galcen_distance=8.0*u.kpc)) # doctest: +FLOAT_CMP
     <Galactocentric Frame (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
         (266.4051, -28.936175)>, galcen_distance=8.0 kpc, galcen_v_sun=(11.1, 232.24, 7.25) km / s, z_sun=27.0 pc, roll=0.0 deg)>
+
+
+Additional parameter sets may be registered, for instance to use the Dehnen & Binney (1998) measurements of the solar motion. We can also add metadata, such as the 1-sigma errors::
+
+    >>> state = galactocentric_frame_defaults.get_from_registry("v4.0")
+    >>> state["parameters"]["galcen_v_sun"] = (10.00, 225.25, 7.17) * (u.km / u.s)
+    >>> state["references"]["galcen_v_sun"] = "http://www.adsabs.harvard.edu/full/1998MNRAS.298..387D"
+    >>> state["error"] = {"galcen_v_sun": (0.36, 0.62, 0.38) * (u.km / u.s)}
+    >>> galactocentric_frame_defaults.register(name="DB1998", **state)
+
+Just as in the previous examples, the new parameter set can be get / set::
+
+    >>> state = galactocentric_frame_defaults.get_from_registry("DB1998")
+    >>> print(state["error"]["galcen_v_sun"])  # doctest: +FLOAT_CMP
+    [0.36 0.62 0.38] km / s
 
 ..
   EXAMPLE END
--- a/docs/credits.rst
+++ b/docs/credits.rst
@@ -267,6 +267,7 @@ Core Package Contributors
 * Nabil Freij
 * Nadia Dencheva
 * Nathanial Hendler
+* Nathaniel Starkman
 * Neal McBurnett
 * Neil Crighton
 * Neil Parley


nstarman:
If this PR aligns with Astropy's goals for ScienceState, I think there's room to push further. In particular, the ``_value`` attribute could be replaced by the ``_state`` attribute, which allows for more flexibility. Backward compatibility and attribute-style access to ``_value`` can be maintained by the ScienceState metaclass as `@property`, or perhaps, something like the BaseCoordinateFrame method of injecting attributes. Also, the ``_Context`` returned by ``set`` could store the old state and redirect all `getitem` to ``self._parent``. This would allow for the ``__enter__`` statement to return the context for use in `with` statements.

```
MyState._value = 1
with MyState.set(value=2) as context:
    assert context._value == MyState._value == 2
assert MyState._value == 1
with pytest.raises(Exception):
    context._value
```

This currently does not work as the context would have ``_value``=1 in `with`. 
This PR already implements most of the underlying mechanics to make this possible, including clearing the context `__dict__` on `__exit__`.

adrn:
Cool, thanks for this @nstarman! I'll review this sometime over the next week or so.

astropy/coordinates/tests/test_frames.py
```diff
@@ -1473,3 +1487,25 @@ def test_galactocentric_references(reset_galactocentric_defaults):
...
+    galactocentric_frame_defaults.register(
+        name="test",
+        value={
+            "galcen_coord": ICRS(ra=266.0 * u.degree, dec=-29 * u.degree),
+            "galcen_distance": 8.1 * u.kpc,
+            "galcen_v_sun": r.CartesianDifferential([12.9, 245.6, 7.8] * u.km / u.s),
+            "z_sun": 21.0 * u.pc,
+            "roll": 0 * u.deg},
+        references=None
+    )
```
mhvk, line 1506: The test should probably include that the frame is in fact registered. Also, do put everything in a `try/finally` so that in the end one can be sure the state is back to what it was originally (even if the test failed).

astropy/utils/state.py
```diff
@@ -7,7 +7,79 @@
...
+        self._parent = parent
+
+    def __enter__(self):
+        """Enter ``with`` statement."""
+        pass
+
+    def __exit__(self, type, value, tb):
+        """Exit ``with`` statement, resetting value and clearing context."""
+        self._parent.set(self._value)  # reset
+        self.__dict__.clear()  # delete on exit
```
mhvk, line 70: Is this to remove any possible references? Though wouldn't generally the context itself go out of scope so be cleaned up?

astropy/utils/state.py
```diff
@@ -7,7 +7,79 @@
...
+        return namespace
+
+    def __getitem__(cls, key):
+        """
+        Items in ``_state`` can be accessed as items of the class.
+        """
+        return cls._state[key]
+
+
+class _Context(metaclass=StateMixinMeta):
```
mhvk, line 43: I'm confused why this class needs the `StateMixinMeta`.

astropy/__init__.py
```diff
@@ -170,25 +170,7 @@ def set(cls, value):
...
-
-            def __repr__(self):
-                return ('<ScienceState {}: {!r}>'
-                        .format(self._parent.__name__, self._parent._value))
-
-        ctx = _Context(cls, cls._value)
-        value = cls.validate(value)
-        cls._value = value
-        return ctx
+        return super().set(cls, value)
```
mhvk, line 173: Nice catch, this was never needed!

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -88,76 +98,95 @@ class galactocentric_frame_defaults(ScienceState):
     _references = None
     _value = None
 
+    _registry = {
+        "v4.0": {
+            "value": MappingProxyType(
```
mhvk, line 103: I had to look up that this "produces a read-only view" - nice! Maybe add a one-line comment above the registry?

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -88,76 +98,95 @@ class galactocentric_frame_defaults(ScienceState):
...
+                        "https://ui.adsabs.harvard.edu/abs/2018RNAAS...2..210D",
+                        "https://ui.adsabs.harvard.edu/abs/2018A%26A...615L..15G",
+                        "https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R",
+                    ],
+                    "z_sun": "https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1417B",
+                    "roll": None,
+                }
+            ),
+        },
+        "pre-v4.0": {
```
mhvk, line 130: Might it make sense for the outer one to be a `MappingProxyType` too?

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -88,76 +98,95 @@ class galactocentric_frame_defaults(ScienceState):
...
-                ['https://ui.adsabs.harvard.edu/abs/2018RNAAS...2..210D',
-                 'https://ui.adsabs.harvard.edu/abs/2018A%26A...615L..15G',
-                 'https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R']
-
-            params['z_sun'] = 20.8 * u.pc
-            references['z_sun'] = \
-                'https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1417B'
-
+        if arg in cls._registry:
+            state = cls._registry[arg].copy()
```
mhvk, line 170: This also took me bit: the copy is to ensure one cannot just replace, say, "value" and then also change the system state. Correct? Again worth a comment, I think, for our future selves...

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -88,76 +98,95 @@ class galactocentric_frame_defaults(ScienceState):
...
+            params = dict(state.get("value"))  # ensure mutable
```
mhvk, line 171: The name change makes me wonder: would it make more sense to call the entry "parameters"?

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -88,76 +98,95 @@ class galactocentric_frame_defaults(ScienceState):
...
+            references = dict(state.get("references", dict()))
         else:
-            raise ValueError(f'Invalid string input to retrieve solar '
-                             f'parameters for Galactocentric frame: "{arg}"')
+            raise ValueError(
```
mhvk, line 174: Hmm, does feel like it should be a `KeyError`... In fact, as we're changing API here anyway by returning an extra `state`, I'd be tempted to remove the whole clause and just let the indexing fail if there's no match.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -41,13 +43,21 @@ class galactocentric_frame_defaults(ScienceState):
...
 
     - 'pre-v4.0': The current default value, which sets the default frame
       attribute values to their original (pre-astropy-v4.0) values.
     - 'v4.0': The attribute values as updated in Astropy version 4.0.
     - 'latest': An alias of the most recent parameter set (currently: 'v4.0')
 
+    Alternatively, user-defined parameter settings may be registered and set.
+
+    This class also tracks the references for all parameter values in the
+    attribute ``_references``, as well as any further information in
```
mhvk, line 56: Seeing this makes me realize this was always a bit weird: we should *never* expose private symbols as public API! I suggest we change this to plain attributes. Or, to keep backward compatibility, store the references and state in the private attributes, but expose them as read-only `@classproperty`.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -88,76 +98,95 @@ class galactocentric_frame_defaults(ScienceState):
...
     @classmethod
     def validate(cls, value):
         if value is None:
             value = cls._latest_value
 
         if isinstance(value, str):
-            params, refs = cls.get_solar_params_from_string(value)
+            params, refs, state = cls.get_solar_params_from_string(value)
             cls._references = refs
+            cls._state = state
```
mhvk, line 189: I wonder if we shouldn't simply store the parameters as well.

astropy/utils/state.py
```diff
@@ -7,7 +7,79 @@
 __all__ = ['ScienceState']
 
 
-class ScienceState:
+class StateMixinMeta(type):
```
mhvk, line 10: I'm very uncertain about this new meta class (see general comments). If we do it at all, my tendency would be to expose not a `__getitem__` but rather a `__getattr__`.

astropy/utils/state.py
nstarman: Yes. As it currently stands, the `_Context.__enter__` does not return anything, so it would just go out of scope and be cleaned up. However, if the ``__enter__`` method returns the context for use in `with` statements, this is needed.

```
MyState._value = 1
with MyState.set(value=2) as context:
    pass
```

Although, as noted above `_Context` would need to be slightly reconfigured to redirect `_value` etc to `self._parent` because they store the old state.

astropy/utils/state.py
nstarman: I guess it doesn't currently. I was hoping to do another PR after this to extend the ``_Context`` returned by ``set`` to store the old state and redirect all `getitem` to ``self._parent``. This would allow the ``__enter__`` method to return the context for use in `with` statements. Having `_Context` use the same metaclass is more in line with that PR.

```
MyState._value = 1
with MyState.set(value=2) as context:
    assert context._value == MyState._value == 2
assert MyState._value == 1
with pytest.raises(Exception):
    context._value
```

This currently does not work as the context would have ``_value``=1 in `with`.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: done.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I refrained from doing that so that the pre-registered options are editable by enterprising users who want to maintain consistency between their added metadata and the pre-registered options. For instance if a new option "test" with extra field "stdv" is added, something can be added to "v4.0" or "pre-v4.0" for consistency.

> If metadata is added to one parameter set, for consistency it should be added to all. Though the built-in options are read-only for "value" and "references", they permit additional fields, such as the standard deviation.

> ```python
> galactocentric_frame_defaults._registry["v4.0"]["stdv"] = None
> ```

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Yes. I had a few mishaps before I diagnosed that. I put in a comment.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: It might. I just called it "value" because it gets assigned to the "_value" attribute. Also, looking forward, if the attribute becomes an `@property` calling `_state["value"]` then the key "value" might have lower cognitive complexity.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I agree. I didn't change things like this so that this PR had the same I/O was before.
I'll change it now.

astropy/utils/state.py
mhvk: I do like the idea of being able to do `with ctx as ...:` (as follow-up...)

astropy/coordinates/builtin_frames/galactocentric.py
mhvk: OK, makes sense.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I agree. The `_state` attribute could coordinate all the information

```python
class ScienceState...
    ...
    @classproperty
    def value(cls):
         return cls._state["value"]
    
    @classproperty
    def _value(cls):  # for backward compatibility
         return cls._state["value"]

    @classproperty
    def references(cls):
         return cls._state["references"]
    
    @classproperty
    def _references(cls):  # for backward compatibility
         return cls._state["references"]
```

astropy/coordinates/tests/test_frames.py
nstarman: done.

astropy/utils/state.py
nstarman: I agree about the  __getattr__ over __getitem__. I have resolved the issue in another branch. My general comment (below) will address this and the meta class.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Scratch the above. Unfortunately, class properties cannot be used on attributes when there is no instance of the class. This is why a metaclass is necessary as only a metaclass (or a modification to how `__dict__` functions) can create properties that work on non-instantiated classes.

as example
```python
class A:
    _value = 2

    @property
    @classmethod
    def value(cls):
        return cls.value
```

`A.value` is as a property object, not `2` until an instance of A is created.

```python
class AMeta(type):
    @property
    @classmethod
    def value(cls):
        return cls.value

class A(metaclass=AMeta):
    _value = 2
```

mhvk: Are you sure? We do have https://docs.astropy.org/en/latest/api/astropy.utils.decorators.classproperty.html#astropy.utils.decorators.classproperty for this purpose (which is what I was referring to)

nstarman:
**The question is which to do first?**

> My main suggestion would be to not add the StateMixinMeta metaclass, at least not in this PR. The ScienceState class is used in quite a few other places, and I think we should first make sure it provides greatest benefit to all (and that it doesn't make expanding them more difficult). I also think the benefit here is small at best - I don't really see why the ScienceState classes should have dict-like behaviour.

Based upon the other comments, particularly the implementation details to make a `__getattr__` work over a `__getitem__`, I agree. I think I should remove the metaclass and most modifications to ScienceState to get the registry working. A subsequent PR can address the more complex issue of ScienceState.

>  My overall sense would be just to keep things minimal, but I could see instead adding a few @classproperty to help get more information on the current state (or perhaps expand what .get() can return).

I do have a fix for all this, but I think should be the subject of a different PR since it is a more substantial change to ScienceState. The changes to `galactocentric_frame_defaults` would be minimal. **The question is which to do first?** pushing a minimal change here that adds the registry will need to be changed a little again if the following PR is accepted.

As a preview for a ScienceState PR. It is a drop-in replacement for all the current ScienceState classes. It should allow the `solar_system_ephemeris` ScienceState to automatically implement the correct kernel, for which the class currently has special fixes. the state has a central value (`value`) but any other related state values are accessible as attributes. The `Context` class can be used in with statements and correctly references the current set state, etc.

[galactocentric.py.txt](https://github.com/astropy/astropy/files/5043152/galactocentric.py.txt)
[state.py.txt](https://github.com/astropy/astropy/files/5043156/state.py.txt)

(Unfortunately these had to be attached as .txt)

mhvk:
@nstarman - looks like we agree - let's just split this into a first part that adds the registration and a follow-up PR that improves `ScienceState`. I do not think it is a problem at all to make a further small change to the `galactocentric_frame_defaults` class in that second PR. And I like the idea of simplifying the solar system one - the implementation Erik, Stuart and I made always felt a bit like a hack!

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I stand corrected. That would work. They are read-only, which is probably an advantage given that ``set`` is desired means to change the state.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I think this should be in the other PR.

astropy/utils/state.py
nstarman: - [x] For followup PR.

nstarman:
With the exception of moving `_Context` out of `ScienceState.set` this PR is now just the galactocentric registry.

CHANGES.rst
```diff
@@ -19,6 +19,7 @@ astropy.coordinates
 - Numpy functions that broadcast, change shape, or index (like ``np.broadcast_to``,
   ``np.rot90``, or ``np.roll``) now work on coordinates, frames, and
   representations. [#10337]
+- ``galactocentric_frame_defaults`` registry for parameter values and metadata.
```
mhvk, line 22: This needs a PR number added to it to get rid of the bot failure. I'd also expand the sentence a bit; maybe simply:
```
``galactocentric_frame_defaults`` can now also be used as a registry, with user-defined parameter values and metadata. [#10624]
```
I thought there normally also was an empty line between entries, which used to be required for sphinx to typeset correctly, but perhaps that doesn't matter any more.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -41,13 +43,21 @@ class galactocentric_frame_defaults(ScienceState):
     ``Galactocentric`` or ``Galactocentric()`` with no explicit arguments.
 
     This class controls the parameter settings by specifying a string name,
-    which can be one of:
+    which has the pre-specified options:
```
mhvk, line 46: maybe "the following pre-specified options"

astropy/coordinates/builtin_frames/galactocentric.py
mhvk, line 56: With your new `references` class property, we can change it here. I think for `_state` we should either also make it public (i.e., have a `state` class property) or not mention it; I'd lean towards making it public as I quite like the idea.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -85,79 +95,124 @@ class galactocentric_frame_defaults(ScienceState):
...
-                'https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1417B'
+        # Get the state from the registry. Need to copy to ensure
+        # registry is immutable to modifications of "_value".
+        state = cls._registry[arg].copy()
+        params = dict(state.pop("value"))  # ensure mutable
 
-        else:
-            raise ValueError(f'Invalid string input to retrieve solar '
-                             f'parameters for Galactocentric frame: "{arg}"')
+        # get references form the state
```
mhvk, line 201: Maybe add the on-the-line comments, which are now split over 2 lines, here too. Also, do capitalize first word (i.e., treat as regular text, following PEP8).

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -168,14 +223,40 @@ def validate(cls, value):
...
                              "Galactocentric frame: input must be a string, "
                              "dict, or Galactocentric instance")
 
+    @classmethod
+    def register(
+        cls,
+        name: str,
+        value: dict,
+        references=None,
+        **state: dict,
```
mhvk, line 241: Since this is not what we actually return as `state`, would it make sense to call this `meta`? And then, of course, below
```
state = dict(value=value, references=references)
state.update(meta)
```

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -85,79 +95,124 @@ class galactocentric_frame_defaults(ScienceState):
...
 
-        else:
-            raise ValueError(f'Invalid string input to retrieve solar '
-                             f'parameters for Galactocentric frame: "{arg}"')
+        # get references form the state
+        references = dict(state.pop("references", None) or {})  # get, if has
+        references = references or None  # empty references -> None
 
-        return params, references
+        return params, references, state  # TODO return references?
```
mhvk, line 205: As we are changing API here anyway, would it make sense to return just `state` itself (well, the mutable copy)?

astropy/coordinates/tests/test_frames.py
```diff
@@ -1437,6 +1437,25 @@ def test_galactocentric_defaults(reset_galactocentric_defaults):
         else:
             assert getattr(galcen_40, k) == getattr(galcen_latest, k)
 
+    # test validate Galactocentric
+    with galactocentric_frame_defaults.set('latest'):
+        params = galactocentric_frame_defaults.validate(galcen_latest)
+        references = galcen_latest.frame_attribute_references
+
+        assert galactocentric_frame_defaults._value == params
```
mhvk, line 1445: Total detail, but I'd test the public attributes.

astropy/coordinates/tests/test_frames.py
```diff
@@ -1437,6 +1437,25 @@ def test_galactocentric_defaults(reset_galactocentric_defaults):
...
+        assert galactocentric_frame_defaults._references == references
+        assert galactocentric_frame_defaults._state == dict()
+
+    # Test not one of accepted parameter types
+    with pytest.raises(ValueError):
+        galactocentric_frame_defaults.validate(ValueError)
+
+    # test value property
```
mhvk, line 1453: With that, I think one can remove this - private properties should be implementation details, ideally.

astropy/coordinates/tests/test_frames.py
mhvk: Ideally, the `try` should be before the registering is done (guard against bug in registration function; yes, it is a bit over the top!)

astropy/coordinates/tests/test_frames.py
```diff
@@ -1473,3 +1492,44 @@ def test_galactocentric_references(reset_galactocentric_defaults):
...
+        references=None
+    )
+
+    # test registration
+    # using try/finally s.t. state is reset even if tests fail
+    try:
+        assert "test" in galactocentric_frame_defaults._registry
+        with galactocentric_frame_defaults.set("test"):
+            assert set(galactocentric_frame_defaults._state.keys()) == set()
+    except AssertionError as e:  # oops, something unexpected
```
mhvk, line 1522: No need to do any `except` - the `AssertionError` will still be captured by pytest.

astropy/coordinates/tests/test_frames.py
```diff
@@ -1473,3 +1492,44 @@ def test_galactocentric_references(reset_galactocentric_defaults):
...
+        tests_passed = True
+    finally:  # remove from registry
+        galactocentric_frame_defaults._registry.pop("test", None)
+
+    if not tests_passed:  # throw exception if above tests failed
+        raise AssertionError(message)
+
+    # test when something isn't in the registry
+    with pytest.raises(KeyError):
+        galactocentric_frame_defaults.get_solar_params_from_string("raises")
```
mhvk, line 1535: Might as well test here that 'test' no longer in the registry!

astropy/utils/tests/test_state.py
```diff
@@ -0,0 +1,31 @@
...
+    class MyState(ScienceState):
+        _value = "A"
+        _state = dict(foo="bar")
+
+    state = {"foo": "bar"}
+
+    # test created ScienceState
+    assert MyState.get() == "A"
+    assert MyState.validate("B") == "B"
+    assert MyState._state == state
```
mhvk, line 21: This would change to the public variable if we go that route.

astropy/utils/state.py
```diff
@@ -7,6 +7,44 @@
 __all__ = ['ScienceState']
 
 
+class _Context:
```
mhvk, line 10: If taking it out, we probably should give it a bit nicer name, like `_ScienceStateContext`.

astropy/utils/state.py
```diff
@@ -7,6 +7,44 @@
...
+    """
+    Science state context. Returned by :func:`~ScienceState.set`.
+
+    ::
```
mhvk, line 14: I don't find the example very helpful... Maybe just some text along the lines of
```
This is initialized and returned inside :meth:`~ScienceState.set`, so that one can
set the state in the context of a `with` statement.
```

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: My two hesitations with making `_state` public are:

1. that the subsequent ScienceState PR will make each field of `_state` accessible as an attribute.
As a side note, this will actually obviate the need for the "value" and "references" class properties, though I think it would be more explicit to keep them around, regardless of strict necessity).

2. `_state` is mutable. This could be resolved using `MappingProxyType`, but I'll need to check whether there are downstream consequences. For instance, I'm looking at `Galactocentric` which appears to expect mutability of "references".

astropy/utils/state.py
nstarman: I agree. It could just be `ScienceStateContext` so that people overriding the `set` method of ScienceStates can subclass it more transparently.

astropy/coordinates/tests/test_frames.py
nstarman: I was mostly doing it for code coverage. I don't think it was ever hit in a test.

CHANGES.rst
nstarman: - [ ]  Added number
- [ ]  used suggested sentence
- [ ]  added line

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: - [ ] implemented

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: - [ ] done

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: - [ ] Implemented

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I agree.
- [ ] Implemented

astropy/coordinates/tests/test_frames.py
nstarman: - [ ] Done

astropy/utils/state.py
nstarman: This will be in the next PR.

astropy/coordinates/tests/test_frames.py
nstarman: Good to know.

astropy/coordinates/builtin_frames/galactocentric.py
mhvk: Good arguments both. Since you went with my other suggestion of returning only `state` in the explicit `get_solar_params_from_string` method, we also do not really need it any more. So, maybe just refer to that as a way to get the state?

nstarman: Given the API change to returning the whole state, what about deprecating the method name `get_solar_params_from_string` in favor of `get_from_registry`?

```
@deprecated("v4.2", message="renamed `get_from_registry`")
@classmethod
def get_solar_params_from_string(cls, arg):
    """..."""
    return cls.get_from_registry(arg)
```

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -85,80 +95,134 @@ class galactocentric_frame_defaults(ScienceState):
     """
 
     _latest_value = 'v4.0'
-    _references = None
     _value = None
+    _references = None
+    _state = dict()  # all other data
+
+    @classproperty  # read-only
+    def value(cls):
```
adrn, line 103: Internally, this info has to be saved as `._value` to work with `ScienceState`, but I wonder if we want to expose this as `parameters_name` or something more descriptive? No strong feelings though..

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -85,80 +95,134 @@ class galactocentric_frame_defaults(ScienceState):
...
-                 'https://ui.adsabs.harvard.edu/abs/2004ApJ...616..872R']
-
-            params['z_sun'] = 20.8 * u.pc
-            references['z_sun'] = \
-                'https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1417B'
+        # Get the state from the registry.
+        # Copy to ensure registry is immutable to modifications of "_value".
+        # Raises KeyError if `arg` is invalid string input to registry
+        # to retrieve solar parameters for Galactocentric frame.
+        state = cls._registry[arg].copy()
```
adrn, line 204: I believe `dict.copy()` is a shallow copy, so maybe we want `copy.deepcopy(cls._registry[arg])` to be safe? For current parameter classes, I guess it doesn't matter though..

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -85,80 +95,134 @@ class galactocentric_frame_defaults(ScienceState):
...
 
-        else:
-            raise ValueError(f'Invalid string input to retrieve solar '
-                             f'parameters for Galactocentric frame: "{arg}"')
+        value = dict(state.get("value"))  # ensure mutable
+        state["value"] = value
+
+        # Get references form the state
+        references = dict(state.get("references", None) or {})  # get, if has
```
adrn, line 210: Given the logic below, couldn't this be:
```suggestion
        references = dict(state.get("references", {}))  # get, if has
```

docs/coordinates/galactocentric.rst
```diff
@@ -171,7 +171,9 @@ up-to-date list of valid names, see the docstring of
 like ``'pre-v4.0'``, which sets the default parameter values to their original
 definition (i.e. pre-astropy-v4.0) values, and ``'v4.0'``, which sets the
 default parameter values to a more modern set of measurements as updated in
-Astropy version 4.0.
+Astropy version 4.0. Also, custom sets of measurements can be registered to
+`~astropy.coordinates.galactocentric_frame_defaults` and used like the
+built-in options.
```
adrn, line 177: I think it would be useful to add an example to the end of the examples below demonstrating how to create a custom parameter set. I'm assuming these changes came because you are using a parameter set that is not one of the default options, so maybe you could use that set of values as your example case? :)

astropy/coordinates/builtin_frames/galactocentric.py
adrn: That seems sensible to me!

adrn:
Mainly minor comments from me, but otherwise this is looking great!

astropy/coordinates/builtin_frames/galactocentric.py
mhvk: Yes, makes sense, though in that case I would let `get_solar_params_from_string` return what it originally did, so that working code doesn't break but can be adjusted.

astropy/coordinates/builtin_frames/galactocentric.py
mhvk: I think I made the same comment but @nstarman had a good answer - which I've forgotten... Was it something about possibly having `uncertainties` in the future?  I still wonder the same. If we change, I'd do `parameters` (or perhaps `attributes` - since in the end these become attributes to the class? I think I like parameters better, given analogy with a model).

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: good point. I'll change it.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: if ``state["references"] = None`` then ``dict(state["references"])`` fails. This doesn't happen for any of the registered options, but might for a user-specified one.

docs/coordinates/galactocentric.rst
nstarman: You assume correctly, particularly the need to keep track of some 1-sigma errors. I'll write something up.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I agree that `parameters` is a better name than `value`. With the next PR, any key in `_state` will automatically be accessible as attributes. I suppose it's not too bad to have parameters -> `_value`. With both of you in agreement, I'll change it.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: I agree a deep copy is necessary. Unfortunately deep copy does not work on a MappingProxyType, but I have found a fix for this, which is useful here, but maybe also elsewhere in Astropy

```
from types import MappingProxyType
import copyreg, copy, pickle

def pickle_mappingproxytype(mpt):
    return dict, (tuple(mpt.items()), )

copyreg.pickle(MappingProxyType, pickle_mappingproxytype)
```

As example
```
d = {"value": MappingProxyType({"A": 1}), }

c = copy.deepcopy(d)
c["value"]["A"] = 2

print(c)
print(d)
```
> {'value': {'A': 2}}
> {'value': mappingproxy({'A': 1})}

mhvk: How about not making a copy at all, and leaving it up to the user to be careful? It would seem fairly standard behaviour (and the `MappingProxy` avoids the most obvious errors).

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Done, `get_solar_params_from_string` calls `get_from_registry` and returns the `parameters` from `_state`

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: In the current `galactocentric_frame_defaults`, when the `_value` is changed this doesn't change the hardcoded registry. I was trying to keep that expectation. The problem with making a shallow copy or no copy is the mutability of the registry. A different fix I looked at was to MappingProxy the whole state when set from the registry. Unfortunately this only makes the top-level read-only, so nested dictionaries, like "parameters" and "references" are mutable. This can approximately be fixed by wrapping both of these in MappingProxy, but for any custom metadata this is not a solution. I think that `deepcopy` is the only easily implemented means of maintaining separation between the state and the registry.

Thankfully, the `copyreg` solution uses only built-in libraries. The only downside is that registering a serialization scheme for MappingProxy here might work in all of Astropy. This isn't actually a problem as 1) MappingProxy is used nowhere else in Astropy and 2) it might even be a desirable outcome to be able to deepcopy a MappingProxy and get a dictionary, not an error. 

This code snippet should be somewhere in `astropy.utils`, not here, but I don't see an obvious place. Thoughts?

docs/coordinates/galactocentric.rst
nstarman: done.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -41,13 +58,25 @@ class galactocentric_frame_defaults(ScienceState):
...
     This class controls the parameter settings by specifying a string name,
-    which can be one of:
+    with the following pre-specified options:
 
     - 'pre-v4.0': The current default value, which sets the default frame
       attribute values to their original (pre-astropy-v4.0) values.
     - 'v4.0': The attribute values as updated in Astropy version 4.0.
     - 'latest': An alias of the most recent parameter set (currently: 'v4.0')
 
+    Alternatively, user-defined parameter settings may be registered and set.
```
adrn, line 68: Perhaps add a link here to documentation showing how to do this, or link to example below?

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,83 +111,176 @@ class galactocentric_frame_defaults(ScienceState):
...
+        """Return Galactocentric solar parameters and metadata given string names
+        for the parameter sets. This method ensures the returned state is mutable,
+        but the registry state remains unchanged. This is the preferred means
+        of accessing the registry.
+
+        Returns
+        -------
+        state : dict
+            Copy of the registry for the string name.
+            At minimum has fields "parameters" and "references", which hold the
```
adrn, line 215:
```suggestion
            At minimum, this must include the fields "parameters" and "references", which hold the
```

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,83 +111,176 @@ class galactocentric_frame_defaults(ScienceState):
...
+            Galactocentric solar parameters and associated references,
+            respectively.
```
adrn, line 217:
```suggestion
            respectively (but all other keys are stored as well).
```

docs/coordinates/galactocentric.rst
adrn: It looks like you added an example to the docstring, but it would also be useful to have something in the narrative docs page. Would you be willing to add a few sentences here, or at minimum link to the docstring example?

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Good idea. Putting a description and link to example below.

docs/coordinates/galactocentric.rst
nstarman: I've added examples and description

mhvk:
@nstarman - just to let you know this hasn't been forgotten! I hope to get to it soonish, if @adrn doesn't get to it before.

docs/coordinates/galactocentric.rst
```diff
@@ -171,7 +171,11 @@ up-to-date list of valid names, see the docstring of
...
+
+`~astropy.coordinates.galactocentric_frame_defaults` also tracks the references for all parameter values, as well as any further specified metadata information.
```
adrn, line 178:
```suggestion
`~astropy.coordinates.galactocentric_frame_defaults` also tracks the references (i.e. scientific papers that define the parameter values) for all parameter values, as well as any further specified metadata information.
```
Could you also wrap this line to be consistent with the other text in this file?

docs/coordinates/galactocentric.rst
```diff
@@ -223,6 +227,21 @@ attributes that are explicitly specified::
     <Galactocentric Frame (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
         (266.4051, -28.936175)>, galcen_distance=8.0 kpc, galcen_v_sun=(11.1, 232.24, 7.25) km / s, z_sun=27.0 pc, roll=0.0 deg)>
 
+
+Additional parameter sets may be registered, for instance to use the Dehnen & Binney (1998) measurements of the solar motion. We can also add metadata, such as the 1-sigma errors::
```
adrn, line 231: Wrap text?

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,100 +102,229 @@ class galactocentric_frame_defaults(ScienceState):
...
+    Additional parameter sets may be registered, for instance to use the
+    Dehnen & Binney (1998) measurements of the solar motion. We can also
+    add metadata, such as the 1-sigma errors::
+
+        >>> state = galactocentric_frame_defaults.get_from_registry("v4.0")
+        >>> state["parameters"]["galcen_v_sun"] = (10.00, 225.25, 7.17) * (u.km / u.s)
+        >>> state["references"]["galcen_v_sun"] = "http://www.adsabs.harvard.edu/full/1998MNRAS.298..387D"
+        >>> state["error"] = {"galcen_v_sun": (0.36, 0.62, 0.38) * (u.km / u.s)}
+        >>> galactocentric_frame_defaults.register(name="DB1998", **state)
+
```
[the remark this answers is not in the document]
nstarman: Done. I modified the preceding paragraph to read

```
    Additional parameter sets may be registered, for instance to use the
    Dehnen & Binney (1998) measurements of the solar motion. We can also
    add metadata, such as the 1-sigma errors. In this example we will modify
    the required key "parameters", change the recommended key "references" to
    match "parameters", and add the extra key "error" (any key can be added)::
```

docs/coordinates/galactocentric.rst
nstarman: Done. Wrapped at 78 characters.

nstarman:
Thanks for the comments @adrn . I think I have addressed / committed them all.
Is there an astropy-wide RTD failure happening? I'm getting an error that does not seem related to my edits.

`ModuleNotFoundError: No module named 'setuptools._distutils'`

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -41,13 +44,30 @@ class galactocentric_frame_defaults(ScienceState):
...
       attribute values to their original (pre-astropy-v4.0) values.
     - 'v4.0': The attribute values as updated in Astropy version 4.0.
     - 'latest': An alias of the most recent parameter set (currently: 'v4.0')
 
+    Alternatively, user-defined parameter settings may be registered, with
+    :meth:`~astropy.coordinates.galactocentric_frame_defaults.register`,
+    and used identically as pre-specified parameter sets. At minimum,
+    registrations must have unique names and a dictionary of parameters
+    with keys "galcen_coord", "galcen_distance", "galcen_v_sun", "z_sun",
+    "roll". See :ref:`this example <galactocentric-frame-defaults-registry-ex>`.
```
mhvk, line 59: I think a cross-link in a docstring is not very helpful - these are mostly going to be seen inside `ipython` or so.  I'd just delete this, but if anything, just put "See examples below".

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,100 +102,231 @@ class galactocentric_frame_defaults(ScienceState):
...
 
     @classmethod
-    def get_solar_params_from_string(cls, arg):
-        """Return Galactocentric solar parameters given string names for the
-        parameter sets.
+    def get_from_registry(cls, name: str):
         """
+        Return Galactocentric solar parameters and metadata given string names
+        for the parameter sets. This method ensures the returned state is
+        mutable, but the registry state remains unchanged. This is the
```
mhvk, line 204: Maybe "is a mutable copy, so that any changes made do not affect the registry state"

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,100 +102,231 @@ class galactocentric_frame_defaults(ScienceState):
...
+        preferred means of accessing the registry.
```
mhvk, line 205: I'd delete "This is the preferred..." - you have it on top.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,100 +102,231 @@ class galactocentric_frame_defaults(ScienceState):
...
+            name = cls._latest_value
 
-        else:
-            raise ValueError(f'Invalid string input to retrieve solar '
-                             f'parameters for Galactocentric frame: "{arg}"')
+        # Get the state from the registry.
+        # Copy to ensure registry is immutable to modifications of "_value".
+        # Raises KeyError if `name` is invalid string input to registry
+        # to retrieve solar parameters for Galactocentric frame.
+        state = copy.deepcopy(cls._registry[name])  # ensure mutable
```
mhvk, line 237: I'm not quite sure why we do not just return `state` here. Up to the user to deal with the absence of references, I'd think. And for `parameters`, it is unclear to me that setting it to `None` if they do not exist is helpful...

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,100 +102,231 @@ class galactocentric_frame_defaults(ScienceState):
...
+            The solar parameters for Galactocentric frame.
+        references : dict or None, optional
+            References for contents of `parameters`.
+            None becomes empty dict.
+        **meta: dict, optional
+            Any other properties to register.
+
+        """
+        # check on contents of `parameters`
+        must_have = ("galcen_coord", "galcen_distance", "galcen_v_sun",
```
mhvk, line 319: I'm a big fan of sets - how about
```
missing = {"galcen_coord", "galcen_distance", "galcen_v_sun",
           "z_sun", "roll"}.difference(parameters)
```

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,100 +102,231 @@ class galactocentric_frame_defaults(ScienceState):
...
+                     "z_sun", "roll")
+        missing = [k for k in must_have if k not in parameters]
+        if missing:
+            raise ValueError(f"Missing parameters: {missing}")
+
+        state = dict(parameters=parameters, references=references)
```
mhvk, line 325: If we change `references=None` to `{}` here, can we not get rid of the extra checks on `references` since now we know the `state` is OK?

astropy/utils/state.py
```diff
@@ -3,10 +3,63 @@
...
+    Tuple[Tuple[Tuple[Any, Any]]]
+        A 1-element tuple containing the tuple of the itemsview of
+        the MappingProxyType mappable.
+
+    """
+    return dict, (tuple(mpt.items()), )
+
+
+# register MappingProxyType copy method
+copyreg.pickle(MappingProxyType, _reduce_mappingproxy)
```
mhvk, line 29: Is this truly needed? It doesn't seem a good idea for astropy to be modifying how `MappingProxyType` is pickled - which I think would affect all packages that are being run. If we really need it, I would subclass it (I guess inside `galactocentric` - right now, this is too much action at a distance!!)

astropy/utils/tests/test_state.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # test setting
```
mhvk, line 23: Can we write this as `with MyState.set("B")` - and then test that `.get() == 'B'` but that it returns to 'A' after the context? Our tests should check the public API, not the implementation!

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Very true. implementing.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: ✔︎

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: ✔︎

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: ✔︎
It was, admittedly, overly cautious. I'll let the user live or die by their own registration 🙃.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: ✔︎. must more elegant.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Done. and extraneous checks are now gone.

astropy/utils/state.py
nstarman: Unfortunately, `MappingProxyType` cannot be subclassed. I have implemented a fix by subclassing `MappingView` with a `MappingProxy` object. Now the `deepcopy` works without any AaaD.

I chose to do it within `state` since my next PR, which will make all `ScienceStates` have `_state` for improved metadata support. Is that alright?

astropy/utils/tests/test_state.py
nstarman: Agreed. Done.

astropy/utils/state.py
mhvk: Ideally, just do it in `galactocentric` - sorry to be making a hassle, but I think it is really best to separate steps...

nstarman: Fair enough.

nstarman:
Tried to make it squash to 1, but there was a conflict on force push, so the minimum appears to be 2.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -30,6 +35,47 @@
...
+    def __getitem__(self, key):
+        """Read-only ``getitem``."""
+        return self._mappingproxy[key]
+
+    def copy(self):
+        """A mutable dictionary deepcopy."""
+        return copy.deepcopy(self)
+
+
+def _reduce_stateproxy(sp: _StateProxy):
```
mhvk, line 58: I think the advantage of the class above is that it can now define `__getstate__` instead of registering this override. Or am I missing something? I am actually slightly confused why this needed - why not adjust the `.copy` method itself?

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -30,6 +35,47 @@
 _ROLL0 = Angle(58.5986320306*u.degree)
 
 
+class _StateProxy(MappingView):
+    """
+    `~collections.abc.MappingView` with a read-only ``getitem`` through
+    `~types.MappingProxyType`.
```
mhvk, line 41: Nice!

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -41,13 +87,30 @@ class galactocentric_frame_defaults(ScienceState):
...
 
+    Alternatively, user-defined parameter settings may be registered, with
+    :meth:`~astropy.coordinates.galactocentric_frame_defaults.register`,
+    and used identically as pre-specified parameter sets. At minimum,
+    registrations must have unique names and a dictionary of parameters
+    with keys "galcen_coord", "galcen_distance", "galcen_v_sun", "z_sun",
+    "roll". See examples below.
+
+    This class also tracks the references for all parameter values in the
+    attribute ``references``, as well as any further information in
```
mhvk, line 105: Don't know why I see this only now, but shouldn't we remove the "as well as ... `_state`" sentence? I think you rightly note below that the preferred way is to use `get_from_registry` -- which seems better than exposing a private attribute!

astropy/coordinates/tests/test_frames.py
```diff
@@ -1499,6 +1519,39 @@ def test_galactocentric_references(reset_galactocentric_defaults):
                 assert k in galcen_custom.frame_attribute_references
 
 
+def test_galactocentric_registry():
+    """Test registerration in galactocentric_frame_defaults"""
+    from astropy.coordinates import (ICRS,
```
mhvk, line 1524: Import might as well be on top.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Unfortunately, a `__getstate__` method doesn't work in this case. I could create a custom ``__deepcopy__`` method, but the `copyreg` method also allows the object to be pickled.

nstarman: https://docs.python.org/3/library/copy.html
ultimate and penultimate paragraphs

astropy/coordinates/tests/test_frames.py
nstarman: I agree. I didn't have it up top because all the other tests `test_frames` import inside the function definition. If I move it up top, I'll should also delete the in-function imports elsewhere.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: changing it to "as well as any further information the registry."

astropy/coordinates/tests/test_frames.py
mhvk: let's leave it as is for now (a separate clean-up PR would be most welcome, though!)

astropy/coordinates/builtin_frames/galactocentric.py
mhvk: But with this code, doesn't it mean that if you unpickle the pickled instance, you get a `dict` rather than a `_StateProxy`?

I'm actually not quite sure whether it is possible to pickle and unpickle the class itself, but it would not seem good that if I pickled the registry for use in another process, all the registries would become `dict`... Especially if the goal is to ensure to avoid a place being made to propagate back.

Overall, I do wonder if we're not over-engineering this. Might it not make more sense to just create a new dict (i.e., do the deep copy) inside the `get_registry` method? (I vaguely fear this is returning to a previous design -- if so, sorry!).

mhvk:
Out of the code thread, maybe good to explicitly discuss (cc @adrn): the idea of the `MappingProxy` here is to guard against people naively getting the state and just changing it, which would change the state internals, and might well lead to very inconsistent result.

I can see the benefit, but am wondering whether it is worth the complexity and performance, given the extra copies any time the state is queried. I also worry about giving a false sense of security, if not for this code, then for the rest of astropy!

p.s. Don't want to make a big deal of this, though! Just thinking the extra code now introduced to make the `MappingProxy` work is quite large, making it harder for a new person seeing the code to understand what is going on - this very much in contrast to the rest, which I think makes the state quite a bit clearer!

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: If I switch from `copyreg` to implementing `__deepcopy__` in `_StateProxy`, then the deep copy will happen inside the `get_from_registry` method and it will have identical behavior to `MappinProxy` (ie read-only), but also implement `getitem`. It will not be able to be pickled, so should not have any unexpected behavior.

nstarman:
> Anyway, it allows me to note that the way to get one commit (and avoid a merge commit in your branch, which we try to avoid [if mostly for backporting which is not relevant here]) is to do a rebase on master before `git push -f`.

Thanks for the tip.

nstarman:
> Out of the code thread, maybe good to explicitly discuss (cc @adrn): the idea of the `MappingProxy` here is to guard against people naively getting the state and just changing it, which would change the state internals, and might well lead to very inconsistent result.
> 
> I can see the benefit, but am wondering whether it is worth the complexity and performance, given the extra copies any time the state is queried. I also worry about giving a false sense of security, if not for this code, then for the rest of astropy!
> 
> p.s. Don't want to make a big deal of this, though! Just thinking the extra code now introduced to make the `MappingProxy` work is quite large, making it harder for a new person seeing the code to understand what is going on - this very much in contrast to the rest, which I think makes the state quite a bit clearer!

I think the new implementation is a lot cleaner. Instead of direct `MappingProxyType`, the read-only copies are maintained by a `MappingView` subclass — `StateProxy` -- that adds a `getitem` and custom `__deepcopy__`. The standard `MappingView` deepcopy returns a `MappingView`, `StateProxy` returns a dict.

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -1,11 +1,16 @@
 # -*- coding: utf-8 -*-
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import copy
+import copyreg
```
mhvk, line 5: Oops, no longer needed!

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -82,100 +123,222 @@ class galactocentric_frame_defaults(ScienceState):
         <Galactocentric Frame (galcen_coord=<ICRS Coordinate: (ra, dec) in deg
             (266.4051, -28.936175)>, galcen_distance=8.0 kpc, galcen_v_sun=(11.1, 232.24, 7.25) km / s, z_sun=27.0 pc, roll=0.0 deg)>
 
+    .. _galactocentric-frame-defaults-registry-ex:
```
mhvk, line 126: No longer used...

astropy/coordinates/builtin_frames/galactocentric.py
```diff
@@ -30,6 +35,25 @@
...
+
+    def __init__(self, mapping):
+        super().__init__(mapping)
+        self._mappingproxy = MappingProxyType(self._mapping)  # read-only
+
+    def __getitem__(self, key):
+        """Read-only ``getitem``."""
+        return self._mappingproxy[key]
+
+    def __deepcopy__(self, memo):
```
mhvk, line 53: The `__deepcopy__` solution is a good one.

astropy/coordinates/builtin_frames/galactocentric.py
nstarman: Indeed.

nstarman:
@adrn @mhvk 
Thanks for all the help getting this ready. I had all this custom code that was just accumulating and I finally decided to do something about it. This gave me a taste of the PR life and now I've got 4 more and a couple issues going.

mhvk:
Darn, was just looking at this to merge it and now it has a conflict... Could you rebase/sqash one more time, @nstarman?  Thanks!

nstarman:
>  Could you rebase/sqash one more time

@mhvk, done. As far as I can tell, the failures are unrelated.

adrn (approved):

mhvk:
Thanks, @nstarman!

adrn:
🎉 A huge thanks to you @nstarman for this!

(CI failures were indeed unrelated.)

nstarman:
Thanks for all the comments and suggestions! This was a great learning experience.

PR #7033, opened 2017-12-24
Feature request: human readable units  

ananyashreyjain:
The function human_help(..) can convert units to human readable form.It picks the right Si prefix so that the value of the given quantity has appropriate units . It can also show the list of values of the given quantity in units with same physical type.
Fix:#6913

PR #4373, opened 2015-12-06
make human_file_size accept astropy quantities

keflavich:

PR #10627, opened 2020-08-06
ENH: Free space getter and checker to handle Quantity

pllim:
### Description

This pull request is to allow `get_free_space_in_dir` and `check_free_space_in_dir` to handle `Quantity`. I tried to use it once and then had to look up what unit the number is supposed to be in. Wouldn't have to do that if it returns `Quantity`.

p.s. Por que no los dos?

### DIFF
--- a/astropy/utils/data.py
+++ b/astropy/utils/data.py
@@ -910,25 +910,26 @@ def _find_hash_fn(hexdigest, pkgname='astropy'):
     return None
 
 
-def get_free_space_in_dir(path, return_quantity=False):
+def get_free_space_in_dir(path, unit=False):
     """
-    Given a path to a directory, returns the amount of free space (in
-    bytes) on that filesystem.
+    Given a path to a directory, returns the amount of free space
+    on that filesystem.
 
     Parameters
     ----------
     path : str
         The path to a directory.
 
-    return_quantity : bool
-        Return ``bytes`` as a Quantity.
-        Default is `False` for backward-compatibility.
+    unit : bool or `~astropy.units.Unit`
+        Return the amount of free space as Quantity in the given unit,
+        if provided. Default is `False` for backward-compatibility.
 
     Returns
     -------
     free_space : int or `~astropy.units.Quantity`
-        The amount of free space (in bytes) on the partition that the directory
-        is on.
+        The amount of free space on the partition that the directory is on.
+        If ``unit=False``, it is returned as plain integer (in bytes).
+
     """
     if not os.path.isdir(path):
         raise OSError(
@@ -937,9 +938,12 @@ def get_free_space_in_dir(path, return_quantity=False):
         # Actually you can on Linux but I want to avoid code that fails
         # on Windows only.
     free_space = shutil.disk_usage(path).free
-    if return_quantity:
+    if unit:
         from astropy import units as u
-        free_space = free_space * u.byte
+        # TODO: Automatically determine best prefix to use.
+        if unit is True:
+            unit = u.byte
+        free_space = u.Quantity(free_space, u.byte).to(unit)
     return free_space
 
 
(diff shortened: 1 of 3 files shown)


pllim:
Hmm. CircleCI didn't kick off. 🤷

astropy/utils/data.py
```diff
@@ -914,20 +914,24 @@ def _find_hash_fn(hexdigest, pkgname='astropy'):
...
     Given a path to a directory, returns the amount of free space (in
     bytes) on that filesystem.
 
     Parameters
     ----------
     path : str
-        The path to a directory
+        The path to a directory.
+
+    return_quantity : bool
```
mhvk, line 927: I wondered a bit about this: How about instead asking for the unit someone wants it in? With `None` meaning no unit ( and perhaps in future `True` for the "best" unit).

astropy/utils/data.py
```diff
@@ -936,32 +940,37 @@ def get_free_space_in_dir(path):
             "not files.")
         # Actually you can on Linux but I want to avoid code that fails
         # on Windows only.
-    return shutil.disk_usage(path).free
+    free_space = shutil.disk_usage(path).free
+    if return_quantity:
```
mhvk, line 944: If we go with the above, then this would become:
```
if unit is not None:
    free_space = u.Quantity(free_space, u.byte).to(unit)
```

astropy/utils/data.py
```diff
@@ -936,32 +940,37 @@ def get_free_space_in_dir(path):
...
     Raises
     -------
-    OSError : There is not enough room on the filesystem
+    OSError
+        There is not enough room on the filesystem.
     """
-    from astropy.utils.console import human_file_size
-
-    space = get_free_space_in_dir(path)
+    space = get_free_space_in_dir(path, return_quantity=hasattr(size, 'unit'))
```
mhvk, line 968: And here it would be `unit=getattr(size, 'unit', None)`.

astropy/utils/data.py
pllim: Very nice idea, but I am not sure how to implement that cleanly. There is this:

https://github.com/astropy/astropy/blob/b5d381f8d576a47ad4c873c015b0c913caec878c/astropy/utils/console.py#L440

But it seems to take great length to avoid using `units`. Do you remember why?

mhvk: I'd leave the "True" for another time - I'd probably rewrite human-readable (there's a closed PR somewhere...)

astropy/utils/data.py
```diff
@@ -910,54 +910,67 @@ def _find_hash_fn(hexdigest, pkgname='astropy'):
     return None
 
 
-def get_free_space_in_dir(path):
+def get_free_space_in_dir(path, unit=False):
```
pllim, line 913: @mhvk , I think I applied all your suggestions, except setting default to `False` instead of `None`. I think `False` is more consistent (staying in the realm of `bool`) when we start to support `True`.

astropy/utils/data.py
pllim: Are you thinking about #4373 ?

mhvk:
All OK, but maybe good to let it sit till tomorrow in case someone else wants to have a look (and good to let the tests finish, of course!)

astropy/utils/data.py
mhvk: No, I was thinking of #7033 - which generalizes it for all units (well, non-composed ones - which to me is the tricky bit: what does one do with, e.g., velocity).

pllim:
Tests passed. Remote data failure is unrelated (it's actually a XPASSED).

mhvk:
Approved this ages ago but clearly we all seem to have forgotten about it. Merging!

pllim:
Thanks! (I have no recollection writing this... 😆 )

PR #10637, opened 2020-08-10
Fix PyInstaller cron job/tox env

astrofrog:
This pins Matplotlib in the pyinstaller job until PyInstaller is compatible with Matplotlib 3.3

<s> ⚠️ contains temporary commit to ensure CI passes, don't merge until this is removed! ⚠️ </s> - passed!


pllim:
Thanks!

PR #10641, opened 2020-08-11
DOC: Fix linkcheck error maybe

pllim:
### Description

This pull request is an attempt to address failing linkcheck for the link to `logging`. Might as well use intersphinx. Gotta check RTD PR build first to see if it works. If not, maybe just get rid of the linking altogether.


pllim:
Update: Intersphinx worked.

PR #10642, opened 2020-08-12
Add a stopping criterion to FittingWithOutlierRemoval

jehturner:
### Description

Currently `FittingWithOutlierRemoval` keeps iterating until `niter` is reached, even when the set of masked points has stopped changing between iterations, rather than [detecting convergence like `stats.SigmaClip`](https://github.com/astropy/astropy/blob/03a99e9ef071541834cb204fcc3429f41bb47867/astropy/stats/sigma_clipping.py#L374) and stopping. This is a basic thing that I should have fixed last time I worked on it...

This mini PR stops iteration when the number of masked points is no longer changing (which, with the current cumulative method of masking, implies that the points are the same ones).

When the equivalent change was made in `SigmaClip`, [the parameter there was renamed from `iters` to `maxiters`](https://github.com/astropy/astropy/commit/662cd2fa41b6657f1b385720b5c6793f044d4107#diff-9a14850376ba3cb948038a9ac71a2bf6). At Gemini we found that a bit disruptive for just a small improvement in clarity, to be honest, so I'm not necessarily advocating changing it here, but am pointing it out for completeness.

With the current API, there isn't really any new test to be added here -- all results should be identical to the previous ones and `__call__` returns a fitted model and mask, which don't provide any way for a test to check what the _actual_ number of iterations was. However, this change does cause the existing `TestWeightedFittingWithOutlierRemoval` to perform 2 iterations instead of the previous 3 and that test still passes unmodified, validating the new behaviour.

We could also add the actual number of iterations as an attribute of `self`, then it could be obtained from the fitter instance immediately after calling it, if that's useful for whatever reason (not that it has any bearing on the results). That would allow a new test to confirm that the actual number of iterations is less than `niter` but still produces the same results, FWIW. What do you think? What would we call the attribute in that case (`niter_actual`?). Thanks.

### DIFF
--- a/astropy/modeling/fitting.py
+++ b/astropy/modeling/fitting.py
@@ -559,8 +559,9 @@ class LinearLSQFitter(metaclass=_FitterMeta):
 class FittingWithOutlierRemoval:
     """
     This class combines an outlier removal technique with a fitting procedure.
-    Basically, given a number of iterations ``niter``, outliers are removed
-    and fitting is performed for each iteration.
+    Basically, given a maximum number of iterations ``niter``, outliers are
+    removed and fitting is performed for each iteration, until no new outliers
+    are found or ``niter`` is reached.
 
     Parameters
     ----------
@@ -577,7 +578,7 @@ class FittingWithOutlierRemoval:
         each model separately; otherwise, the same filtering must be performed
         in a loop over models, which is almost an order of magnitude slower.
     niter : int, optional
-        Number of iterations.
+        Maximum number of iterations.
     outlier_kwargs : dict, optional
         Keyword arguments for outlier_func.
     """
@@ -683,9 +684,9 @@ class FittingWithOutlierRemoval:
         if filtered_data.mask is np.ma.nomask:
             filtered_data.mask = False
         filtered_weights = weights
+        last_n_masked = filtered_data.mask.sum()
 
         # Perform the iterative fitting:
-        # TO DO: add a stopping criterion when results aren't changing?
         for n in range(self.niter):
 
             # (Re-)evaluate the last model:
@@ -762,6 +763,13 @@ class FittingWithOutlierRemoval:
                 fitted_model = self.fitter(fitted_model, *coords,
                                            filtered_data,
                                            weights=filtered_weights, **kwargs)
+
+            # Stop iteration if the masked points are no longer changing (with
+            # cumulative rejection we only need to compare how many there are):
+            this_n_masked = filtered_data.mask.sum()  # (minimal overhead)
+            if this_n_masked == last_n_masked:
+                break
+            last_n_masked = this_n_masked
 
         return fitted_model, filtered_data.mask
 
--- a/astropy/modeling/fitting.py
+++ b/astropy/modeling/fitting.py
@@ -585,10 +585,10 @@ class FittingWithOutlierRemoval:
     Attributes
     ----------
     fit_info : dict
-        The `fit_info` (if any) from the last iteration of the wrapped `fitter`
-        during the most recent fit. An entry is also added with the keyword
-        `niter` that records the actual number of fitting iterations performed
-        (as opposed to the user-specified maximum).
+        The ``fit_info`` (if any) from the last iteration of the wrapped
+        ``fitter`` during the most recent fit. An entry is also added with the
+        keyword ``niter`` that records the actual number of fitting iterations
+        performed (as opposed to the user-specified maximum).
     """
 
     def __init__(self, fitter, outlier_func, niter=3, **outlier_kwargs):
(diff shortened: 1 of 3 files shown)


jehturner:
Oh, and I think we'd have a problem if we rename `iters` to `maxiters`, because then when `FittingWithOutlierRejection` is used with `outlier_func=sigma_clip`, their respective `maxiters` parameters would collide in `__init__`, resulting in that parameter not getting passed to `sigma_clip` when it should be set to 1 to avoid doing the default 5 sigma clipping iterations for each iteration of `FittingWithOutlierRemoval`. This is difficult to avoid in general, since we can't know what that parameter will be called for any given `outlier_func` (but usually it's `sigma_clip`).

astropy/modeling/fitting.py
```diff
@@ -763,6 +764,13 @@ def __call__(self, model, x, y, z=None, weights=None, **kwargs):
                                            filtered_data,
                                            weights=filtered_weights, **kwargs)
 
+            # Stop iteration if the masked points are no longer changing (with
+            # cumulative rejection we only need to compare how many there are):
+            this_n_masked = filtered_data.mask.sum()  # (minimal overhead)
+            if this_n_masked == last_n_masked:
+                break
+            last_n_masked = this_n_masked
```
jehturner, line 772: I considered skipping this whole block when the loop has reached `niter` anyway, mostly to avoid the unnecessary `sum()`, but the difference in performance seems to be in the noise for a small (100 element) array, so I think it's not worth complicating the code slightly for this.

jehturner:
To 4.0.2? I suppose unnecessary iteration with no benefit could be considered a bug, but it is also the documented behaviour (the description string just says "number of iterations" not "maximum number"). Either way this should be a harmless speed up  that doesn't change the behaviour. I don't really have a strong opinion (I've submitted a vaguely-related PR #10613 for 4.2 that I think we'll want to use in conjunction with it). I suppose it would be nice to have in the widely-used LTS version.

pllim:
I milestoned to 4.2 for now because it feels like an enhancement but if this is a bug fix, please remilestone.

Also, is a new test necessary?

nden:
Can anyone restart the documentation build?

pllim:
No, there is no restart button. Try close and reopen? If that doesn't work, a new push event would retrigger it.

jehturner:
See my 4th paragraph above re. the test. There isn't really anything to test unless we add a new attribute to the fitter that records how many iterations were actually used. Maybe that's worth doing though (especially if we're going for 4.2 rather than 4.0) -- some of the other fitters similarly store information calculated during the fit.

Eg. for `LevMarLSQFitter`, we have _"The `fit_info` dictionary contains the values returned by `scipy.optimize.leastsq` for the most recent fit, including the values from the `infodict` dictionary it returns_". This actually raises 2 questions. First, I think `FittingWithOutlierRemoval` is **discarding** this information from the fitters it wraps (even the linear fitter, now that I look at it). We should probably make it copy `fit_info` from the underlying fitter to `self`, shouldn't we? Presumably we'd just save that `dict` from the final iteration and not from all the iterations? Second, if we want to return the actual number of iterations performed, should we likewise keep that information in `fit_info` rather than a new attribute (in which case the entry could just be called `niter` like the parameter)?

Actually, I suppose the user _can_ still get the final `fit_info` because they instantiate the fitter before passing it to `FittingWithOutlierRemoval`. But relying on that would preclude the convenient usage ```fitter = FittingWithOutlierRemoval(LevMarLSQFitter(), sigma_clip```, so it might be more convenient to copy `LevMarLSQFitter.fit_info` into `FittingWithOutlierRemoval.fit_info` and add `niter` there at the same time??

nden:
Moving `fit_info` to `FittingWithOutlierRemoval` sounds reasonable.

saimn:
Having a way to get the number of iterations seems very useful to check if the algorithm converged, same for `fit_info`.

jehturner:
How about this?

Arguably for model sets we should track the iterations separately for each model and remove it from the set to be passed to the fitter once it converges, otherwise you could end up re-fitting 1000 models just because there's a bad pixel left in 1 of them. But that would be considerably more complicated and perhaps no faster if "slicing" a model set ends up being too complicated. In the meantime, this version is already an improvement on always iterating until `niter`. Moreover, I would think that similar models will all converge at similar rates, so this might not be too far from optimal in practice.

There's something odd about the existing tests. They all leave the value of `maxiters` for `sigma_clip` at its default of `5`, which means that every iteration of `FittingWithOutlierRemoval`will do 5 iterations of `sigma_clip` before re-fitting. Although that can evidently work, I think a more standard and probably sensible (not to mention faster) algorithm is to re-fit the model(s) in between each clipping iteration, rather than clipping to convergence around some fit that's not yet finalized. So I think when using `FittingWithOutlierRemoval`, one should normally always pass `maxiters=1` in the arguments to `sigma_clip`, which is not happening here.

jehturner:
Thanks!

PR #10644, opened 2020-08-12
MNT: Replace RandomState with default Generator

### LINKED ISSUE #11519: DOC: remove references to RandomState
### Description

I propose here to remove from the documentation any reference to `RandomState` and wrong (not best practice) usage of random number generators.

We went through this in SciPy. More on the PR https://github.com/scipy/scipy/pull/13778

Let me know if you would accept this contribution and I would propose to do it 😃 

### Additional context

Since NumPy 1.17 (and as far as I understand, the supported minimal version here is 1.17), the [canonical way](https://numpy.org/doc/stable/reference/random/index.html) to create random numbers is:

```python
import numpy as np
from numpy.random import default_rng
rng = default_rng()
```
And If a seed is to be used. User should not use a common value such as 0 or 123456. This is to prevent bias (everyone is using the same values) and ensure sufficient entropy for the generator. More [details](https://numpy.org/doc/stable/reference/random/bit_generators/index.html#seeding-and-entropy). The good practice (from NumPy's doc) is to generate a seed like this:

```python
from numpy.random import SeedSequence
print(SeedSequence().entropy)
```

Then you can use the value like
```python
rng = default_rng(my_seed_generated_and_not_common_value)
```

pllim:
### Description

This pull request is to address the fact that [RandomState is now considered legacy](https://numpy.org/doc/stable/reference/random/legacy.html).

However, during local testing, it appears that replacing it with default `Generator` isn't a simple search-and-replace. That is, the results change enough to cause test failures.

Should we proceed with this or trying something else like our own [NumpyRNGContext](https://docs.astropy.org/en/latest/api/astropy.utils.misc.NumpyRNGContext.html#astropy.utils.misc.NumpyRNGContext)?

@mhvk , do you have any opinion on this?

Fix #11519

**Choices**

1. Update the results. **But** `default_rng` is not available in Numpy 1.16 or older. ~Do we copy it into `compat` or something, or maybe wait to do this till we can drop support for Numpy 1.16?~ Not a problem now as we have dropped support for Numpy 1.16.
2. ~Use something else.~
3. ~Keep using `RandomState`.~

**TODO**

- [x] Wait for #10664 to merge. ~If its milestone changes, the milestone here needs to be updated to match.~
- [x] Rebase to pick up the PR above.
- [x] Update test results so CI pass. Remember to check the dev/remote job that is allowed to fail.

**After merge**

- [ ] Open issue to stop using random seed for doctest, see https://github.com/astropy/astropy/pull/10644#pullrequestreview-634247783
- [ ] Open issue to follow up on hardcoding test cases in `astropy/timeseries/periodograms/bls/tests/test_bls.py` and `astropy/modeling/tests/test_models.py` and assign to @astrofrog , see https://github.com/astropy/astropy/pull/10644#pullrequestreview-523134061


mhvk:
p.s. The failures are of course rather annoying! For tests, we really shouldn't be using random data (unless consciously to flush out bugs, like with `hypothesis`), while for documentation I can see the point, but for those we may be best off just not checking the final result.

pllim:
> we really shouldn't be using random data

Perhaps not, but fixing that is out of scope, as I have no idea what those devs had in mind when they designed the tests. I'll see about updating the results based on local run.

pllim:
I need input from subpackage maintainers about the failures and unexpected result changes before I can proceed with this.

astropy/stats/tests/test_histogram.py
```diff
@@ -63,23 +60,23 @@ def test_freedman_bin_width(N=10000, rseed=0):
...
 @pytest.mark.skipif('not HAS_SCIPY')
 def test_knuth_bin_width(N=10000, rseed=0):
-    rng = np.random.RandomState(rseed)
-    X = rng.randn(N)
+    rng = np.random.default_rng(rseed)
+    X = rng.standard_normal(N)
 
     dx, bins = knuth_bin_width(X, return_bins=True)
-    assert_allclose(len(bins), 59)
+    assert_allclose(len(bins), 58)
```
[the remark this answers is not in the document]
larrybradley: Yes, I think this is fine.  Different bins for different data.

astropy/stats/tests/test_histogram.py
```diff
@@ -129,45 +126,43 @@ def test_histogram_range_with_bins_list(N=1000, rseed=0):
...
-                           -0.17288406, 0.42214237, 1.01716881, 1.61219525,
-                           2.20722169, 2.80224813])
+    assert_allclose(counts, [2, 11, 16, 18, 22, 14, 13, 4])
+    assert_allclose(bins, [-2.32503077, -1.74087192, -1.15671306, -0.5725542,  0.01160465,
+                           0.59576351, 1.17992237, 1.76408122, 2.34824008], rtol=2e-7)
 
     counts, bins = histogram(X, bins='blocks')
-    assert_allclose(counts, [10, 61, 29])
-    assert_allclose(bins, [-2.55298982, -1.24381059, 0.46422235, 2.26975462])
+    assert_allclose(counts, [3, 97])
```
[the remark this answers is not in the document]
larrybradley: Ok for different data.

larrybradley (approved):

nden (approved):

pllim:
RTD failure is unrelated (`h5py` related). Remote data job passed.

astrofrog (approved):
Looks good to me but I think in a follow up we should really try and get rid of the hard coded arrays. I have some ideas for finding a seed so if someone opens the issue it can be assigned to me.

pllim:
@astrofrog , if you can find the seed soon, maybe this PR can wait. If not, feel free to merge. Thanks.

saimn:
This seems ready but needs a rebase.

astropy/stats/biweight.py
```diff
@@ -99,10 +99,10 @@ def biweight_location(data, c=6.0, M=None, axis=None, *, ignore_nan=False):
 
     >>> import numpy as np
     >>> from astropy.stats import biweight_location
-    >>> rand = np.random.RandomState(12345)
-    >>> biloc = biweight_location(rand.randn(1000))
+    >>> rand = np.random.default_rng(12345)
```
tupui, line 102:
```suggestion
    >>> rand = np.random.default_rng()
```

astropy/stats/biweight.py
```diff
@@ -99,10 +99,10 @@ def biweight_location(data, c=6.0, M=None, axis=None, *, ignore_nan=False):
...
+    >>> biloc = biweight_location(rand.standard_normal(1000))
     >>> print(biloc)    # doctest: +FLOAT_CMP
-    -0.0175741540445
+    0.01535330525461019
```
tupui, line 105:
```suggestion
    0.01535330525461019 # random
```

tupui (commented):
I am a bit late sorry. We did this in SciPy and the only difference is that we removed all seeding from examples. This is to enforce best practices. Using small seeds is bad for the entropy of the generator and also common values like 0, 42, 12345 end up in every code which leads to bias in results. Seeing should only be used for testing.

Here I am proposing to remove the seed and just allow variation in the results to pass the doctests (in SciPy I can add `# random` or `# may vary`, I am not sure if this is standard or not. Maybe this is your `# doctest: +FLOAT_CMP`?).

I can do a PR on this branch if you want some help. Also, I would suggest to add somewhere in the doc a paragraph about seeding. See what we did for example https://scipy.github.io/devdocs/tutorial/stats.html#random-number-generation.

mhvk:
@tupui - that sounds like a very good idea, but we do not have the `may vary` and `random` markers yet (might actually be good to replace our `FLOAT_CMP` with that), and I think we should definitely do that in a follow-up PR, not here.

tupui:
> @tupui - that sounds like a very good idea, but we do not have the `may vary` and `random` markers yet (might actually be good to replace our `FLOAT_CMP` with that), and I think we should definitely do that in a follow-up PR, not here.

Then do you suggest to use `FLOAT_CMP` here an in another PR to have `may vary` and `random`? If yes, I could propose the change for this PR and work on the following PR. Let me know 😃

pllim:
Re: No seed for doctest -- `FLOAT_CMP` only works for small variations. I have no idea what removing the seed would do to the doctest. If you insist on doctest not being deterministic, might as well do a `# doctest: +SKIP`?

mhvk:
@tupui - I'm suggesting that for this PR we leave the explicit seed in place and use `FLOAT_CMP` - we can then consider removing the explicit seeding in a separate PR (which will require not using `FLOAT_CMP`).

adrn:
@tupui - I agree that it is a good idea to figure out how to remove seeding from user-facing examples, but I do agree with @mhvk here that this PR should be restricted to replacing `RandomState`. Let's make a new issue to follow up on the broader point?

astropy/modeling/tests/test_models.py
```diff
@@ -452,9 +451,31 @@ def test_deriv_1D(self, model_class, test_parameters):
         model_no_deriv = create_model(model_class, test_parameters,
                                       use_constraints=False)
 
+        # NOTE: PR 10644 replaced deprecated usage of RandomState but could not
+        #       find a new seed that did not cause test failure, resorted to hardcoding.
```
adrn, line 455: Hm - seems like this should maybe become a TODO, with a new issue for @nden?

adrn (commented):
Took a quick look at this - seems fine to make these changes after rebasing. But in a new issue we should discuss modifying examples to remove seeds. 

One other thing that's not clear to me is whether the [default bit generator](https://numpy.org/doc/stable/reference/random/generator.html) could ever change in numpy? If it does, all the tests could start failing. So if that is a possibility, I wonder if we want to have a test helper that can produce a random number generator given a seed, but with the PCG64 bit generator hard coded? May not be an issue, but others might know...

tupui:
> @tupui - I agree that it is a good idea to figure out how to remove seeding from user-facing examples, but I do agree with @mhvk here that this PR should be restricted to replacing `RandomState`. Let's make a new issue to follow up on the broader point?

Sure makes sense. I can propose a following PR if that's ok then.

> One other thing that's not clear to me is whether the default bit generator could ever change in numpy? If it does, all the tests could start failing. So if that is a possibility, I wonder if we want to have a test helper that can produce a random number generator given a seed, but with the PCG64 bit generator hard coded? May not be an issue, but others might know...

We had a similar discussion on SciPy. While the underlying bit generator might change, it's a bit unlikely as people are now using it. Still, from NumPy's documentation they explicitly warn about not wanting to ensure reproducible results over the versions. If you must support a lot of versions `RandomState` is still there and is ensured to provide the same results. But it might not stay forever in NumPy. On SciPy, we still use `RandomState` in tests because we need to support NumPy 1.16. Maybe this will change once we switch to 1.17.

astropy/modeling/tests/test_models.py
pllim: Yeah, haven't gotten any replies here, so follow-up issue is the way to go.

astropy/timeseries/periodograms/bls/tests/test_bls.py
```diff
@@ -35,40 +35,366 @@ def assert_allclose_blsresults(blsresult, other, **kwargs):
         assert_quantity_allclose(v, other[k], **kwargs)
 
 
+# NOTE: PR 10644 replaced deprecated usage of RandomState but could not
+# find a new seed that did not cause test failure, resorted to hardcoding.
 @pytest.fixture
 def data():
-    rand = np.random.RandomState(123)
-    t = rand.uniform(0, 10, 500)
+    t = np.array([
```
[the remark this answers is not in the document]
pllim: Probably should open a follow-up issue about this too... cc @adrn

pllim:
I rebased. The one failure is unrelated. Given there are multiple approvals, I added some follow-up action items and will merge when CI finishes. Thanks for the reviews and discussions, everyone!

p.s. FWIW, I don't think people should be using random numbers in tests but addressing that is out of scope of this PR.

mhvk:
Yay! And, yes, I agree that one should not use random numbers in tests in the first place.

tupui:
Plus, in the end if the seed is fixed, it's not really a "random" number generator anymore 😉

pllim:
I can't read minds, but I think the "random" numbers got into the tests in one of two ways:

* Someone copy-pasted science code into a unit test.
* Attempt to mimic "real" data but want to do it as a one-liner.

I don't think randomness was the *intention* but rather means to an end.

pllim:
I think RTD is stuck due to concurrency limits. #yolo

pllim:
To everyone who is interested in the outstanding issues, please see #11566 and #11567 . Thanks again!

Issue #10156, opened 2020-04-16
Allow return annotations other than Quantity


LukasK13:
### Description

The decorator `quantity_input()` tries to convert the annotated return value to a Quantity [here](https://github.com/astropy/astropy/blob/3d1200bcb6bd400db5156000608e7698447df029/astropy/units/decorators.py#L236). When a return annotation other than Quantity is used, this conversion fails of course resulting in the error below.

### Expected behavior

The decorator `quantity_input()` should ignore annotated return values which are not of type Quantity. This has also been disicussed and fixed [here](https://github.com/astropy/astropy/pull/5606#discussion_r92433023) but the fix has been discarded.

### Actual behavior

The following error message is shown:
```
Traceback (most recent call last):
  File "<input>", line 6, in <module>
  File "<<redacted>>/venv/lib/python3.7/site-packages/astropy/units/decorators.py", line 236, in wrapper
    return return_.to(wrapped_signature.return_annotation)
AttributeError: 'str' object has no attribute 'to'
```

### Steps to Reproduce

1. [Run the code below in a python console]

```python
import astropy.units as u

@u.quantity_input(temp=[u.Kelvin, u.Celsius])
def foo(temp: u.Quantity) -> str:
    return "bar"
print(foo(300 * u.K)())
```

### System Details

```
Darwin-19.4.0-x86_64-i386-64bit
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21) 
[Clang 6.0 (clang-600.0.57)]
Numpy 1.18.2
astropy 4.0.1.post1
Scipy 1.4.1
```

Cadair:
:+1: To supporting this.

The usecase under this, of coming up with a way of supporting type annotations and unit validation at the same time is interesting, and not one the current implementation of `quantity_input` was built for. This is the first example of a thing that makes me want to keep they keyword argument syntax to `quantity_input` around for reasons other than backwards compatibility. (Up to this point I would have liked to deprecate it in favour of function annotations).

I do wonder if there is something we could provide which is a unit aware version of a Quantity type annotation, i.e. `'Quantity[u.Kelvin]'` or similar. My current experience with the type hinting support in Python in limited so this probably needs someone who knows more about it in the kind of contexts it would be useful.

nstarman:
[...]

PR #10646, opened 2020-08-13
Add isclose() method to Time and TimeDelta classes

taldcroft:
### Description

This pull request is to add an `isclose()` method to the `Time` and `TimeDelta` classes to make it easier to compare times to within a specified tolerance. The default tolerance is basically 1-bit in the mantissa of a float64 (`np.finfo(float).eps` = 2.2e-16 days or about 20 picosec), but the `atol` argument can be provided to be less strict.

This partially addresses #8742.


pllim:
Hmm, I guess `rtol` doesn't make sense in this context. Does this needs a "what's new" entry?

astropy/time/core.py
```diff
@@ -1180,6 +1180,34 @@ def __setitem__(self, item, value):
...
+                            f'{other.__class__.__name__} instead')
+
+        if atol is None:
+            atol = np.finfo(float).eps * u.day
+
+        if not isinstance(atol, (u.Quantity, TimeDelta)):
+            raise TypeError("'atol' argument must be a Quantity or TimeDelta instance, got "
+                            f'{atol.__class__.__name__} instead')
+
+        # Note: np.abs(self - other) results in an object array of TimeDelta
```
mhvk, line 1207: There should be no need to do that - `TimeDelta` should be smart enough! (And I think it is - if not, that's a bug...)

astropy/time/core.py
```diff
@@ -1180,6 +1180,34 @@ def __setitem__(self, item, value):
...
+        Parameters
+        ----------
+        other : `~astropy.time.Time`
+            Time object for comparison.
+        atol : `~astropy.units.Quantity` or `~astropy.time.TimeDelta`
+            Tolerance for equality with units of time (e.g. ``u.s`` or ``u.day``).
+            Default is one bit in the 128-bit JD time representation, equivalent
+            to about 20 picosecs.
+        """
+        if not isinstance(other, Time):
```
mhvk, line 1196: I'm wondering whether the checks are necessary, since ideally the comparison just breaks. Indeed,
```
In [20]: TimeDelta(1*u.day) < Time.now()                                                              
TypeError: '<' not supported between instances of 'TimeDelta' and 'Time'
```

But very strangely,
```
In [21]: TimeDelta(1*u.day, scale='tai') < Time('J2000')                                                                            
Out[21]: True
```
It looks like it has something to do with the specific scale picked - I'll investigate.

Ignoring that, would it be an idea to remove the checks? It might allow `Time` and `TimeDelta` can share the implementation.

astropy/time/core.py
```diff
@@ -2460,6 +2488,32 @@ def _make_value_equivalent(self, item, value):
...
+
+        Parameters
+        ----------
+        other : `~astropy.units.Quantity` or `~astropy.time.TimeDelta`
+            Quantity or TimeDelta object for comparison.
+        atol : `~astropy.units.Quantity` or `~astropy.time.TimeDelta`
+            Tolerance for equality with units of time (e.g. ``u.s`` or ``u.day``).
+            Default is one bit in the 128-bit JD time representation, equivalent
+            to about 20 picosecs.
+        """
```
mhvk, line 2503: Possibly contrary to my argument above, `TimeDelta` perhaps should have a different implementation since `rtol` does make sense for `TimeDelta`.

astropy/time/core.py
taldcroft: See #10653

mhvk: As noted there, plain `abs(...)` should work.

astropy/time/core.py
mhvk: I guess it is the usual "look before you leap" vs "easier to ask for forgiveness than permission" - https://docs.python.org/3/glossary.html#term-eafp

Generally, python seems to prefer the latter, and I think it is not illogical here either. E.g., suppose someone writes a different `MyTime` class (say, one that is integer + fraction in seconds), which they try to ensure can interact with `Time`, e.g., by defining subtraction with astropy's `Time` (which they can make work, because python will give their `__rsub__` a chance after our `__sub__` fails). If this also returns something that can have its absolute value taken and be compared with a quantity or `TimeDelta` (as they might well, since those are methods we've had for a long time), then the actual `isclose()` code will "just work". But in the current implementation, the `isinstance` check makes it impossible to ever work. (This is similar to numpy having `np.asarray` sprinkled throughout their code, making it not work with `Quantity`, and indeed `np.abs` sort-of working rather than just plain failing).

p.s. In principle, in EAFP, one does try/except and gives perhaps more informative error messages, but I've slowly moved to a state where I think one should just count on the user looking at the docstring when an error is raised. And in this case, one would generally get a quite informative `TypeError` from the subtraction anyway.

taldcroft: I see your point and it is valid. OTOH, it happens frequently (based on real astropy issues I've seen) that users get confused because of an obscure error message raised in an unrelated bit of code when they pass a bad input. So there is a balance between addressing a problem that it almost sure to occur (bad inputs) vs. a new `MyTime` class that is less likely.

It's also worth noting that a major theme of recent Python development is type annotation. I've actually started paying some attention to this since there must be a reason the core devs are spending so much time on it. From this perspective we would annotate the input with *something*. Agreed that I'm not sure how this really works with duck-typing, which is certainly an incredibly useful concept.

I guess the middle ground is putting the main bit of the function in a try / except TypeError and re-raising with a better exception.

mhvk: Agreed that the `MyTime` class is not very likely, and I should add that this is not a big deal! But if one does it without anything, the type of exception one gets is already quite clear:
```
In [11]: class TTime(Time): 
    ...:     def isclose(self, other, atol=1*u.ns): 
    ...:         return abs(self - other)<atol 
    ...:               
    ...:                                                                                               

In [12]: t = TTime.now()                                                                               

In [13]: dt = TimeDelta(100, format='jd')                                                              

In [14]: t.isclose('J2000')                                                                            
<ipython-input-14-15f66b730fe3> in <module>
----> 1 t.isclose('J2000')

<ipython-input-11-b8b243443d6c> in isclose(self, other, atol)
      1 class TTime(Time):
      2     def isclose(self, other, atol=1*u.ns):
----> 3         return abs(self - other)<atol
      4 
      5 

TypeError: unsupported operand type(s) for -: 'TTime' and 'str'

In [15]: t.isclose(dt)                                                                                 
<ipython-input-15-f21ef40edf4f> in <module>
----> 1 t.isclose(dt)

<ipython-input-11-b8b243443d6c> in isclose(self, other, atol)
      1 class TTime(Time):
      2     def isclose(self, other, atol=1*u.ns):
----> 3         return abs(self - other)<atol
      4 
      5 

TypeError: bad operand type for abs(): 'TTime'
```

mhvk: p.s. The annotations are indeed intriguing. I think currently they are only to help writing correct code, not execution, but I think it is a good way to make the intent clearer. On annotations, our very own `quantity_input` decorator is an exception in that it actually uses annotations at runtime. Actually, that is seeing some nice discussion over at #10156 with a draft new implementation at #10655 - the latter might even get to your long-standing suggestion of having a way to ask for Quantity input that gives the function plain arrays in the right units! Would be good to get your input!

astropy/time/core.py
```diff
@@ -1180,6 +1180,36 @@ def __setitem__(self, item, value):
...
+
+          abs(self - other) <= atol
+
+        Parameters
+        ----------
+        other : `~astropy.time.Time`
+            Time object for comparison.
+        atol : `~astropy.units.Quantity` or `~astropy.time.TimeDelta`
+            Absoute tolerance for equality with units of time (e.g. ``u.s`` or
+            ``u.day``). Default is one bit in the 128-bit JD time representation,
```
mhvk, line 1197: I wonder a bit about the default: when we were doing the hypothesis testing, I found that for UTC I had to allow 2 bits (since taking the difference with a UTC time means one has to do a scale change): 
https://github.com/astropy/astropy/blob/f56d8abf5dd39671c695e1e295293ef1490e2485/astropy/time/tests/test_precision.py#L502-L512

Maybe we should do that here too? (Also, I guess in principle, one can just put the default in in the initializer, rather than make it `None`.)

taldcroft: I made the change.

About putting the atol in the signature, I thought about this and am on the edge, but somehow seeing a somewhat non-trivial expression in the function signature feels wrong: 
```
def isclose(self, other, atol=2 * np.finfo(float).eps * u.day):
```
I think the repr of `<Quantity 4.4408921e-16 d>` won't really be that helpful in the method docstring.

mhvk: I'd definitely leave it in `ns` if we put it in the signature... But, yes, clearly a user would have to read the docstring anyway, so really it doesn't matter - let's stick with what you have.

MaxNoe:
Is isclose a ufunc? If yes, does it make sense to create a WIP branch adding all the ufuncs we need that different people can contribute?


I now see at least three issues / PRs that started implememting `Time.__array_ufunc__`, maybe we should bundle the effort, work in a separste branch until all ufincs we want are implemented and then make a PR from that branch.

I usually don't like longer lived dev branches, but in this case it might be appropriate.

mhvk:
:+1: to a separate branch to work on `ufunc` support! You may want to take a look at `quantity_helpers` to see how it was done for units (since eventually we'll want to support functions like `isclose` too...).

```

In [3]: type(np.isclose)
Out[3]: function

In [4]: type(np.isnat)
Out[4]: numpy.ufunc
```

taldcroft:
> since eventually we'll want to support functions like isclose too...

I still don't see that `np.isclose` should be supported. 

For `Time`, the `np.isclose` arg `rtol` literally makes no sense, and `equal_nan` at best requires a stretch to interpret "NaN" as missing values. Furthermore the default for `atol=1e-8` is not appropriate since it is a float and we require a `Quantity` or `TimeDelta`.

MaxNoe:
> For Time, the np.isclose arg rtol literally makes no sense

For calendar dates i would agree. For timescales base on linear time offsets from an epoch, I think rtol would be well-defined. (Unix, TT, TAI, ...)

taldcroft:
Sure, for unix time there is technically a way to define `rtol`. But in practice this would become quite dodgy since those relative values don't have units that are programmatically specified. In other words the `TimeUnix` class does not know that the unix representation is in `u.s`.

Perhaps more importantly, I cannot see any practical use for `rtol` even with unix time. Do you have a use case?

MaxNoe:
@taldcroft no, no really. Maybe unit testing of arithmetic with Time / TimeDelta?

taldcroft:
@mhvk - I think I have addressed all the points, including a compromise on duck-typing the input.  The code that relies on `other` is now in a try/except with a reasonable error message and some care taken to make it easier for the user to see what went wrong.

taldcroft:
> Maybe unit testing of arithmetic with Time / TimeDelta?

That testing is already solid. I think that using `rtol` would only degrade the testing by introducing additional multiplies done at 64-bit resolution (vs. 128 everywhere else).

MaxNoe:
> That testing is already solid. I think that using rtol would only degrade the testing by introducing additional multiplies done at 64-bit resolution (vs. 128 everywhere else).

I did not mean unit testing inside astropy but of packages using astropy, but you are probably right that there is no real use case here.

mhvk:
@MaxNoe - I'm a bit more with you on this, that `np.isclose(t1, t2)` might as well do something sensible (i.e., with `rtol=0` assumed). But where it makes most sense is for `TimeDelta`: `np.isclose(dt1, dt2, atol=...*u.ns, rtol=1e-9)` is reasonable and would have the benefit that it doesn't matter if `dt1/2` are quantities and `TimeDelta`.

Anyway, the method here is good and discoverable and definitely seems the best way to go for `Time` itself (@taldcroft - I'll review later today).

astropy/time/core.py
```diff
@@ -914,6 +914,45 @@ def __setitem__(self, item, value):
...
+            ``u.day``). Default is two bits in the 128-bit JD time representation,
+            equivalent to about 40 picosecs.
+        """
+        if atol is None:
+            # Note: use 2 bits instead of 1 bit based on experience in precision
+            # tests, since taking the difference with a UTC time means one has
+            # to do a scale change.
+            atol = 2 * np.finfo(float).eps * u.day
+
+        if not isinstance(atol, (u.Quantity, TimeDelta)):
```
mhvk, line 940: Is this still necessary given your (very nice) error message below?

astropy/time/core.py
```diff
@@ -2457,6 +2496,40 @@ def _make_value_equivalent(self, item, value):
...
+        atol : `~astropy.units.Quantity` or `~astropy.time.TimeDelta`
+            Absolute tolerance for equality with units of time (e.g. ``u.s`` or
+            ``u.day``). Default is one bit in the 128-bit JD time representation,
+            equivalent to about 20 picosecs.
+        rtol : float
+            Relative tolerance for equality
+        """
+        try:
+            other_day = other.to_value(u.day)
+        except Exception as err:
```
mhvk, line 2520: I'd follow `Time.isclose()` and do `atol_day = atol.to_value(u.day)` inside here (and adjust the error message), with the `atol is None` statement moved in front.

mhvk (approved):
Looks very nice. Two nitpicks, but OK to ignore, so approving.

taldcroft:
@mhvk - thanks for the review. I thought about your comments for a bit and in the end I think for `atol` it is reasonable to be strict about the type. I think it would be unusual for a program using astropy to find it inconvenient to specify `atol` as a quantity.

In general I still lean toward guaranteeing correctness with tests that run on the advertised API data types, rather than hoping for correctness based only on not generating an exception. Understood this is a big fuzzy line that we'll continually be straddling.

PR #10647, opened 2020-08-13
Improve speed of coordinate transformations for arrays of obstime

MaxNoe:
This is a followup to #6068, which was outdated.

Here, I squashed the commits in that branch to make rebasing easier and then rebased versus the current master.

Please see the discussion there.

I will start doing benchmarks with this code again and doing some cleaning up, e.g. using units for the precision and removing python 2 support in the new files.

EDIT: Close #6068


MaxNoe:
After fixing some minor issues with dtypes and quantities, the tests should be passing now.

Starting to do some benchmarks.

@bwinkel do you still have the scripts you used to create the plots from #6068?

pllim:
Thanks! cc @bwinkel

MaxNoe:
Tests for the actual interpolation are missing and it does currently not work due to a shape issue, looking into it.

MaxNoe:
```python

from astropy.coordinates.astrom_manager import astrom_interpolation_resolution
from astropy.coordinates import SkyCoord, EarthLocation, AltAz
from astropy.time import Time
import astropy.units as u
import numpy as np

from time import perf_counter


t = Time('2020-01-01T20:00:00') + np.random.uniform(0, 24, 100000) * u.hour
location = EarthLocation.of_site('Roque de los Muchachos')
crab = SkyCoord.from_name("Crab Nebula")

altaz = AltAz(obstime=t, location=location)


t0 = perf_counter()
no_interp = crab.transform_to(altaz)
print(f'No Interpolation took {perf_counter() - t0:.4f} s')


seps = {}

for resolution in [10 * u.s, 5 * u.min, 30 * u.min]:
    t0 = perf_counter()
    astrom_interpolation_resolution.set(resolution)
    interp = crab.transform_to(altaz)
    print(f'Interpolation with {resolution.value: 5} {str(resolution.unit):3s} resolution took {perf_counter() - t0:.4f} s')

    seps[resolution.to_value(u.s)] = no_interp.separation(interp)


fig, axs = plt.subplots(len(seps), 1, constrained_layout=True)
for ax, (key, sep) in zip(axs, seps.items()):
    ax.hist(sep.to_value(u.microarcsecond), bins=100)
    ax.set_title(f'{key} s resolution')

    ax.text(
        0.5, 0.95,
        (
            f'mean: {sep.to_value(u.microarcsecond).mean():.4f} µas'
            f'\np95:  {np.percentile(sep.to_value(u.microarcsecond), 95):.4f} µas'
            f'\nmax:  {sep.to_value(u.microarcsecond).max():.4f} µas'
        ),
        va='top',
        ha='center',
        transform=ax.transAxes
    )
    ax.set_xlabel('Angular Separation / µas')

fig.savefig('interpolation_error.png', dpi=300)
```


interpolation_error

mhvk:
@MaxNoe - as you will probably have seen from the original issue, I was quite wary of these speed-related work-arounds. That said, clearly you and others need it and looking at it again I think the approach taken, with a generic `get_astrom` function, is something that can be made to address my concerns, if we change how the `ScienceState` operates. In particular, I feel quite strongly that a goal in any solution should be to separate the regular and interpolating settings -- basically, I would like the interpolation case to be one that I *know* will never affect regular use of astropy, and separate enough that it can be maintained separately.

Before going there, it may be good to just make sure that what we implement is the right level of generality, so that users perhaps can change also in other ways how the transformations operate. I have not read back the whole original thread, but is it obviously why one would not override parts of the transformation graph with interpolating routines (and have a `ScienceState` which essentially does that)? Or is that too high level?  Somewhat similarly, I can see more general use for being able to override `erfa`; would that be sufficient? (I.e., the interpolating case would generally just call the standard routines, but provide overrides for the relevant slow routines to check whether times should be interpolated).

I think it may well be that the `get_astrom` level is in fact the right one!  Right now, though, the actual `ScienceState` is very limited, in that it just gives an interpolating timescale, one that also very much interleaves the current code with interpolation stuff, thus making the regular case more difficult to follow.  Could we instead make it a higher-level `ScienceState`, where the state is something like a module or class which provides all the astrom functions as attributes or methods (I think that would be better than a long if-else-else sequence, but that is somewhat of an implementation detail); in a way, it would then becomes more similar to the `ScienceState` used for [IERS data](https://docs.astropy.org/en/stable/api/astropy.utils.iers.earth_orientation_table.html#astropy.utils.iers.earth_orientation_table), where the IERS tables have different constraints for the precision of the data provided). In this scheme, by default, what is returned just contains the regular code, but it can be replaced by interpolating code (which of course can use the regular module when needed; a class-based scheme would seem easiest for this purpose).

The implementation might look something like
```
class GetAstrom:
    def apci(self, frame):
        ...

    def apcs(self, frame):
        ...

    def __call__(self, frame, tcode):
        return getattr(self, tcode)(frame)


def GetAstromInterpolate(GetAstrom):
    def __init__(self, interpolation_resolution):
        self.interpolation_resolution = interpolation_resolution

    def apci(self, frame):
        <get interpolated frame>
        result = super().apci(interpolated_frame)
        <interpolate back>
        return final_result
```

So, internally, whenever needed, code would do something like
```
coords.get_astrom_methods.set(coords.GetAstromInterpolate(10*u.s)):
```

Or perhaps with string equivalents, like for the [ephemerides](https://docs.astropy.org/en/stable/api/astropy.coordinates.solar_system_ephemeris.html#astropy.coordinates.solar_system_ephemeris), 
```
coords.get_astrom_methods.set('interpolate', 10*u.s):
```

bwinkel:
@mhvk I think your proposal makes a lot of sense! It seems a bit more work than the current solution, but it would indeed be a much more generic solution and leave the default behavior intact. It's been a while, since I looked into this part of the Astropy code, but I'll try to find some time next week to work on this. Perhaps you or @MaxNoe would be available for some pair programming session?

bwinkel:
> 
> 
> After fixing some minor issues with dtypes and quantities, the tests should be passing now.
> 
> Starting to do some benchmarks.
> 
> @bwinkel do you still have the scripts you used to create the plots from #6068?

One of the plots was posted in the original PR, [with the spatial dependency](https://github.com/astropy/astropy/pull/6068#issuecomment-301816446). The other plots are inferior to your histograms, I think, but I could dig that out if necessary.

MaxNoe:
> In particular, I feel quite strongly that a goal in any solution should be to separate the regular and interpolating settings -- basically, I would like the interpolation case to be one that I _know_ will never affect regular use of astropy, and separate enough that it can be maintained separately.

I agree, right now the code is very much entangled and this should be cleanly separated.


 
> would not override parts of the transformation graph with interpolating routines 

I started crude approximate transformations (implemented after formulas in the "Explanatory Supplement to the Astronomical Almanac" with a context manager to switch out transformations completely here:

https://github.com/MaxNoe/approximate_coords

But I think the interpolation approach of the astrom is a nice way of not having to replace a full transformation.
Since it is something *all* transformations, irregardless of how they work, might benefit from. 
So having this decoupled from the actual graph is a nice property I think.


> 
> The implementation might look something like
> 

I like this proposal a lot better then the imperative structure of the current function, I'll give it a shot.

> Perhaps you or @MaxNoe would be available for some pair programming session?

Sure!

MaxNoe:
First I am going to write some basic unit tests and the refactor into the class based approach.

MaxNoe:
@bwinkel @mhvk I implemented the class based approach and I like it a lot better than what was there before, please have a look.

bwinkel:
Note, that in the original PR, @StuartLittlefair mentioned a possible issue with [Moon coordinates calculation speed](https://github.com/astropy/astropy/pull/6068#issuecomment-305791341). I just leave this here as a reminder to make sure it's not forgotten about.

astropy/coordinates/astrom_provider.py
```diff
@@ -0,0 +1,220 @@
...
+        '''
+        Calculate support points for the interpolation.
+        This includes the minimum and maximum timestamp in obstime
+        to improve precision when the time resolution is wider than
+        the time range or only a single value is requested.
+        '''
+        mjd_scaled = np.ravel(obstime.mjd / self. mjd_resolution)
+
+        # ndmin=1 needed in case only a single coordinate is transformed
+        mjd_lower = np.array(np.floor(mjd_scaled), ndmin=1, copy=False)
```
bwinkel, line 130: I could imagine, that a simple cast to int64 would be faster. Also, mjd_upper is redundant, isn't it? Could just append `mjd_lower[-1] + 1` to the array, if I'm not mistaken. (Would working with integers speed up the `np.unique` function? I guess not.)

astropy/coordinates/astrom_provider.py
```diff
@@ -0,0 +1,220 @@
...
+        self.mjd_resolution = time_resolution.to_value(u.day)
+
+    def _get_support_points(self, obstime):
+        '''
+        Calculate support points for the interpolation.
+        This includes the minimum and maximum timestamp in obstime
+        to improve precision when the time resolution is wider than
+        the time range or only a single value is requested.
+        '''
+        mjd_scaled = np.ravel(obstime.mjd / self. mjd_resolution)
```
bwinkel, line 127: With float64, the minimal `mjd_resolution` will be of the order of microseconds before numerical accuracy issues kick in (please confirm). I wonder, if one should restrict the range of mjd resolutions the user is allowed to set?

astropy/coordinates/astrom_provider.py
bwinkel: > Also, mjd_upper is redundant, isn't it? Could just append `mjd_lower[-1] + 1` to the array, if I'm not mistaken.
Strike that, it would not work for inhomogeneous data. However, `mjd_upper = mjd_lower + 1` should be ok.

bwinkel:
@MaxNoe This looks really nice. Thanks so much for contributing it!

astropy/coordinates/astrom_provider.py
MaxNoe: You mean raising an error instead of just a warning?

bwinkel: The thing is, when float64 is not sufficiently accurate for the `mjd/res` operation (float64 is already problematic for mjd anyway, if high time resolution is needed), the result could simply be wrong. This is perhaps reason to raise an error, but I don't have strong feelings, if you prefer a warning. Nevertheless, this scenario should raise a different warning/error than the one, which is already included (about the performance implications).

astropy/coordinates/astrom_provider.py
MaxNoe: I'll check performance of these variants. But this anyhow does not really contribute to the performance costs here.

astropy/coordinates/astrom_provider.py
MaxNoe: Yes, should be:

```
>>> import numpy as np
>>> import astropy.units as u
>>> (Time.now().mjd * np.finfo(np.float64).eps * u.day).to(u.us)
>>> <Quantity 1.13335838 us>
```

So warning for < 1ms and exception for < 10 μs?

MaxNoe:
> I'm not sure we need the base class

I was not sure either, I'll drop it. It is not like everyone will start implementing their own astrom providers.

> Of course, another question is the name! Would erfa_astrom and ErfaAstrom[Interpolator] be nicer?

Sounds good, will rename.

>  this gives a nice symmetry to having those methods on the interpolating case

Ok

astropy/coordinates/astrom_provider.py
mhvk: I'd stick with warning, but expand the docstring and refer to that in the warning.

mhvk: p.s. Basically, if we start going the warning route, we should also warn against overly long intervals - I feel this is specialized enough that users should really read the documentation. 

A propos documentation, this obviously deserves a nice entry in the performance section!

astropy/coordinates/astrom_provider.py
MaxNoe:
```

In [1]: import numpy as np                                                                                                                    

In [2]: mjd = np.random.uniform(50000, 60000, 10000)                                                                                          

In [3]: %timeit np.ceil(mjd)                                                                                                                  
2.92 µs ± 47 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: %timeit np.int64(mjd)                                                                                                                 
5.47 µs ± 6.99 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [5]: %timeit np.floor(mjd)                                                                                                                 
2.83 µs ± 12.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: lower = np.floor(mjd)                                                                                                                 

In [7]: %timeit upper = lower + 1                                                                                                             
3.42 µs ± 13.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [8]: %timeit upper = np.ceil(mjd)                                                                                                          
3.01 µs ± 14.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Performance seems best for what it is now, calling floor and ceil

astropy/coordinates/astrom_provider.py
MaxNoe: Ok, I'll raise another warning for resolution < 10µs and start working on an example for the performance section.

MaxNoe: Concerning overly long, that really depends on the needed accuracy. People will anyways have to make sure by doing comparisons and checks, that the interpolation version delivers the precision they need. 

In my test, even intervals as long as 6 hours still produced < mas resolution.

MaxNoe:
@StuartLittlefair 

In the previous PR, you commented:

> Another thing to check is that this doesn't have unintended consequences - e.g with an older state of this PR I noticed that get_moon had slowed down significantly, perhaps due to slowdowns in the transformation from GeocentricTrueEcliptic to ICRS in calc_moon. I would benchmark a series of typical coordinate transformations, and ensure they don't slowdown significantly.

For clarification: You noticed a slowdown for conversions even with the default settings? This can now basically only happen because of the overhead of calling `ScienceState.get` and nested functions as opposed to more imperative code. 

Or did `get_moon` get slower when the interpolation approach was used.
If this is the case, I would not worry to much, as the decision to use the interpolation should anyways only be made after performance and precision checks for the specific use case at hand. We are not changing astropy's default behaviour.

MaxNoe:
I ran a similar benchmark for `get_moon`:

```python
from astropy.coordinates.erfa_astrom import erfa_astrom, ErfaAstromInterpolator
from astropy.coordinates import get_moon
from astropy.time import Time
import astropy.units as u
import numpy as np

from time import perf_counter


t = Time('2020-01-01T20:00:00') + np.random.uniform(0, 12, 10000) * u.day


t0 = perf_counter()
no_interp = get_moon(t)
reference = perf_counter() - t0
print(f'No Interpolation took {reference:.4f} s')


seps = {}

for resolution in [5 * u.min, 30 * u.min, 6 * u.hour, 24 * u.hour]:
    with erfa_astrom.set(ErfaAstromInterpolator(resolution)):
        t0 = perf_counter()
        interp = get_moon(t)
        duration = perf_counter() - t0

    print(
        f'Interpolation with {resolution.value: 5} {str(resolution.unit):3s}'
        f' resolution took {duration:.4f} s'
        f' ({reference / duration:5.1f}x faster) '
    )
    seps[resolution.to_value(u.s)] = no_interp.separation(interp)

fig, axs = plt.subplots(len(seps), 1, constrained_layout=True, figsize=(6, len(seps) * 2))
for ax, (key, sep) in zip(axs, seps.items()):
    ax.hist(sep.to_value(u.microarcsecond), bins=100)
    ax.set_title(f'{key} s resolution')

    ax.text(
        0.5, 0.95,
        (
            f'mean: {sep.to_value(u.microarcsecond).mean():.5f} µas'
            f'\np95:   {np.percentile(sep.to_value(u.microarcsecond), 95):.5f} µas'
            f'\nmax:   {sep.to_value(u.microarcsecond).max():.5f} µas'
        ),
        va='top',
        ha='center',
        transform=ax.transAxes
    )
    ax.set_xlabel('Angular Separation / µas')

fig.savefig('moon_interpolation_error.pdf', dpi=300)

```


```

Results are not as impressive as with the AltAz <-> ICRS transforms, but still, a noticeable speedup for very minor (< 0.05 µas) precision penalty.
```
No Interpolation took 4.3266 s
Interpolation with   5.0 min resolution took 3.0303 s (  1.4x faster) 
Interpolation with  30.0 min resolution took 2.4782 s (  1.7x faster) 
Interpolation with   6.0 h   resolution took 2.3711 s (  1.8x faster) 
Interpolation with  24.0 h   resolution took 2.3645 s (  1.8x faster) 
```

![moon-1](https://user-images.githubusercontent.com/5488440/90386990-a2832f80-e085-11ea-9b35-772c9fb38f1a.png)

**Comment by StuartLittlefair** [comment]
Thanks @maxnoe. This looks great!

**Comment by MaxNoe** [comment]
And I also compared to the current master, get_moon did not change in performance for the default case.

**Comment by MaxNoe** [comment]
@StuartLittlefair I also have now a version where I also interpolate the call to `erfa.nut06` in `get_moon`.

This makes it 2.5x faster at still µas resolution for the 5 minute interpolation case:

```
from astropy.coordinates.erfa_astrom import erfa_astrom, ErfaAstromInterpolator
from astropy.coordinates import get_moon
from astropy.time import Time
import astropy.units as u
import numpy as np

from time import perf_counter


np.random.seed(42)
t = Time('2020-01-01T20:00:00') + np.random.uniform(0, 12, 10000) * u.hour


t0 = perf_counter()
no_interp = get_moon(t)
reference = perf_counter() - t0
print(f'No Interpolation took {reference:.4f} s')


seps = {}

for resolution in [1 * u.s, 10 * u.s, 5 * u.min, 30 * u.min, 6 * u.hour]:
    with erfa_astrom.set(ErfaAstromInterpolator(resolution)):
        t0 = perf_counter()
        interp = get_moon(t)
        duration = perf_counter() - t0

    print(
        f'Interpolation with {resolution.value: 5} {str(resolution.unit):3s}'
        f' resolution took {duration:.4f} s'
        f' ({reference / duration:5.1f}x faster) '
    )
    seps[resolution.to_value(u.s)] = no_interp.separation(interp)

fig, axs = plt.subplots(len(seps), 1, constrained_layout=True, figsize=(6, len(seps) * 2))
for ax, (key, sep) in zip(axs, seps.items()):
    ax.hist(sep.to_value(u.microarcsecond), bins=100)
    ax.set_title(f'{key} s resolution')

    ax.text(
        0.5, 0.95,
        (
            f'mean: {sep.to_value(u.microarcsecond).mean():.5f} µas'
            f'\np95:   {np.percentile(sep.to_value(u.microarcsecond), 95):.5f} µas'
            f'\nmax:   {sep.to_value(u.microarcsecond).max():.5f} µas'
        ),
        va='top',
        ha='center',
        transform=ax.transAxes
    )
    ax.set_xlabel('Angular Separation / µas')

fig.savefig('moon_interpolation_error.pdf', dpi=300)
```

```
No Interpolation took 4.0575 s
Interpolation with   1.0 s   resolution took 5.6140 s (  0.7x faster) 
Interpolation with  10.0 s   resolution took 2.6680 s (  1.5x faster) 
Interpolation with   5.0 min resolution took 1.6539 s (  2.5x faster) 
Interpolation with  30.0 min resolution took 1.6201 s (  2.5x faster) 
Interpolation with   6.0 h   resolution took 1.6112 s (  2.5x faster) 
![moon-1](https://user-images.githubusercontent.com/5488440/90388627-63a2a900-e088-11ea-81ae-fd51831aaf2b.png)

```

Should I also include this, or leave it as is?

**Comment by MaxNoe** [comment]
@bwinkel @mhvk I added a doc section below the general performance tips and a changelog entry.

I think I addressed all comments, please have a look again.

**Comment by StuartLittlefair** [comment]
@MaxNoe The speedup is excellent, but I'm tempted to leave it out, simply because it does not fit well with the current neatness - i.e everywhere the ```ScienceState``` is used it returns an ```erfa``` astrometry context, but a call to ```nut06a``` would be different. 

If included, I think clarity would demand a tweak to the names from ```erfa_astrom``` to ```erfa_wrappers``` or similar.

**Comment by MaxNoe** [comment]
I tend to agree, this is why I didnt immediately include it.

**Comment by MaxNoe** [comment]
Profiling the code again, nearly all time is now spent in `get_dut1utc`

File: astropy/coordinates/astrom_provider.py (hunk 1 above)
Comment (mhvk) [reply astropy/coordinates/astrom_provider.py]: Indeed, casting seems to be quite slow to ufunc (e.g., `x.astype(bool)` is slower than `x != 0`). Though as you wrote before, this part will not dominate execution time anyway, so clarity before anything else!

File: astropy/coordinates/astrom_provider.py (hunk 2 above)
Comment (mhvk) [reply astropy/coordinates/astrom_provider.py]: Agreed - this would seem to be a question of documentation.

**Comment by mhvk** [comment]
> Profiling the code again, nearly all time is now spent in `get_dut1utc`

For cases where the difference between UT1 and UTC does not matter, one can explicitly set `delta_ut1_utc=0` on the times.  I don't know if that case is ever relevant, but one could imagine setting a fake `IERS` table which always returns 0 for UT1-UTC and the polar motion offsets and then using that in https://docs.astropy.org/en/latest/api/astropy.utils.iers.earth_orientation_table.html#astropy.utils.iers.earth_orientation_table - perhaps for another performance trick?

**Comment by mhvk** [comment]
On the name (always the most difficult thing...) -- the `ErfaWrapper` might leave a bit more leeway to expand later. But the current scheme seems best for keeping readability/clarity, so maybe at some point there has simply to be the idea that if one wants fully-optimized performance, one has to replace the whole tranform graph...

For the documentation, would it be an idea to add an explicit warning that this is experimental, and that while we will continue to provide the basic mechanism for having an interpolating option, the implementation and even name may change?

**Comment by MaxNoe** [comment]
@mhvk Ok, I got the doctests figured out now, test should be passing.

How would I go about marking this as experimental? Just writing it in the docs / CHANGES.rst or is there a mechanism for marking certain code as experimental?

**Comment by MaxNoe** [comment]
There also is a failure with regard to `trim_zeros` which I think is unrelated to the changes here, how would I go about that?

File: astropy/coordinates/erfa_astrom.py [hunk 3]
```diff
@@ -0,0 +1,260 @@
...
+    '''
+
+    @staticmethod
+    def apci(frame):
+        '''
+        Wrapper for ``erfa.apci``, used in conversions CIRS <-> ICRS
+
+        Arguments
+        ---------
+        frame: ``astropy.coordinate.CIRS``
```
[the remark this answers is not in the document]
MaxNoe: At least one of these expected a frame attribute that was not present for `AltAz` frames.

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,260 @@
...
+        '''
+        mjd_scaled = np.ravel(obstime.mjd / self. mjd_resolution)
+
+        # ndmin=1 needed in case only a single coordinate is transformed
+        mjd_lower = np.array(np.floor(mjd_scaled), ndmin=1, copy=False)
+        mjd_upper = np.array(np.ceil(mjd_scaled), ndmin=1, copy=False)
+
+        # unique already does sorting
+        mjd_u = np.unique(np.concatenate([
+            [mjd_scaled.min(), mjd_scaled.max()],
```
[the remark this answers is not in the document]
MaxNoe: It's explained in the docstring. If the resolution is larger than the time range of the values, min and max provide better precision without harming performance.

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,57 @@ properties::
...
+
+Improving Performance for Arrays of ``obstime``
+===============================================
+
+The most expensive operations when transforming between observer-dependent coordinate
+frames (e.g. ``AltAz``) and sky-fixed frames (e.g. ``ICRS``) are the calculation
+of astrometric values like precession and nutation matrices.
+
+If |skycoord| instances with a large ``obstime`` array are transformed,
+these calculations can be speed up by factors up to 100 and still keeping micro-arcsecond resolution
```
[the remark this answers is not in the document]
MaxNoe: I would keep the up too, since for more conservative time resolutions it is a factor of 10, it really depends on the density of the obstimes and the chosen time resolution.

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,57 @@ properties::
...
+   >>> t0 = perf_counter()
+   >>> with erfa_astrom.set(ErfaAstromInterpolator(300 * u.s)):  # doctest:+REMOTE_DATA
+   ...     crab_altaz_interpolated = crab.transform_to(frame)  # doctest:+REMOTE_DATA +IGNORE_WARNINGS
+   >>> print(f'Transformation took {perf_counter() - t0:.2f} s')  # doctest:+ELLIPSIS
+   Transformation took ... s
+
+   >>> err = crab_altaz.separation(crab_altaz_interpolated)  # doctest:+REMOTE_DATA
+   >>> print(f'Mean error of interpolation: {err.to(u.microarcsecond).mean():.4f}')  # doctest:+ELLIPSIS +REMOTE_DATA
+   Mean error of interpolation: 0.0063 uarcsec
+
```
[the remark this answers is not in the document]
MaxNoe: I feel like this would encourage bad practices, doesn't it?

It's a pretty big change to how the transformations work and should be chosen appropriately at the point where it is used, right?

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,260 @@
...
+    >>> with erfa_astrom.set(ErfaAstromInterpolator(300 * u.s)):
+    ...    cirs = SkyCoord.from_name('Crab').transform_to(CIRS(obstime=obstime))  # doctest: +REMOTE_DATA
+    '''
+
+    @u.quantity_input(time_resolution=u.day)
+    def __init__(self, time_resolution):
+        if time_resolution.to_value(u.ms) < 1:
+            warnings.warn(
+                f'Using {self.__class__.__name__} with `time_resolution`'
+                ' below 1ms might not improve performance',
```
bwinkel, line 134: I'm not sure, if this warning should really be raised. First of all, the value of 1 ms seems arbitrary to me. Whether there is a speed-up depends entirely on how often the ERFA routines must be queried, but this is dependent on the number of support points, which itself depend on the ratio of the "spread" of the times and the resolution. If all observations were done within few milliseconds, then a resolution of ms will still lead to a speed-up. Whether it is necessary, is a different questions and depends on the required accuracy. So perhaps it would be better to move this discussion into the docs, and add a table or a plot, which demonstrates the achievable accuracy as a function of resolution?

MaxNoe: Agreed.

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,260 @@
...
+                AstropyWarning
+            )
+        if time_resolution.to_value(u.us) < 10:
+            warnings.warn(
+                f'Using {self.__class__.__name__} with `time_resolution`'
+                ' below 10 us is probably not practical since float64 for MJD '
```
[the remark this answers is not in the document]
bwinkel: I'd formulate the warning a bit differently: "Using ... with time_resolution below 10 us might lead to numerical inaccuracies, as MJD is internally represented by a 64-bit float." Has anyone tried already, what happens if one uses a very small resolution? (I didn't.) Could well be that the `np.interp` goes nuts.

astropy/coordinates/tests/test_erfa_astrom.py
```diff
@@ -0,0 +1,98 @@
...
+
+        altaz = AltAz(location=location, obstime=obstime)
+        gcrs = GCRS(obstime=obstime)
+
+        for frame, tcode in zip([altaz, altaz, gcrs], ['apio13', 'apci', 'apcs']):
+            without_interp = getattr(provider, tcode)(frame)
+            assert without_interp.shape == shape
+
+            with_interp = getattr(interp_provider, tcode)(frame)
+            assert with_interp.shape == shape
```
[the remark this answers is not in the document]
MaxNoe: done

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,57 @@ properties::
...
+Improving Performance for Arrays of ``obstime``
+===============================================
+
+The most expensive operations when transforming between observer-dependent coordinate
+frames (e.g. ``AltAz``) and sky-fixed frames (e.g. ``ICRS``) are the calculation
+of astrometric values like precession and nutation matrices.
+
+If |skycoord| instances with a large ``obstime`` array are transformed,
+these calculations can be speed up by factors up to 100 and still keeping micro-arcsecond resolution
+by utilizing interpolation instead of calculating astrometric values for each individual point.
```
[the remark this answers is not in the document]
MaxNoe: done

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,57 @@ properties::
 
 ..
   EXAMPLE END
+
+Improving Performance for Arrays of ``obstime``
+===============================================
+
+The most expensive operations when transforming between observer-dependent coordinate
+frames (e.g. ``AltAz``) and sky-fixed frames (e.g. ``ICRS``) are the calculation
+of astrometric values like precession and nutation matrices.
```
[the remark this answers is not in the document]
MaxNoe: done

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,260 @@
...
+    def apcs(frame):
+        '''
+        Wrapper for ``erfa.apci``, used in conversions CIRS <-> ICRS
+
+        Arguments
+        ---------
+        frame: ``astropy.coordinate.GCRS``
+        '''
+        jd1_tt, jd2_tt = get_jd12(frame.obstime, 'tt')
+        pv = pav2pv(
```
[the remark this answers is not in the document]
MaxNoe: done

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,260 @@
...
+        '''
+        jd1_tt, jd2_tt = get_jd12(frame.obstime, 'tt')
+        cip = get_cip(jd1_tt, jd2_tt)
+        earth_pv, earth_heliocentric = prepare_earth_position_vel(frame.obstime)
+        return erfa.apci(jd1_tt, jd2_tt, earth_pv, earth_heliocentric, *cip)
+
+    @staticmethod
+    def apcs(frame):
+        '''
+        Wrapper for ``erfa.apci``, used in conversions CIRS <-> ICRS
```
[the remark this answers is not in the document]
MaxNoe: done

docs/coordinates/performance.inc.rst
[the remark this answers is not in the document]
MaxNoe: I took the other part of your suggestion, though

astropy/coordinates/erfa_astrom.py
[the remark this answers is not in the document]
StuartLittlefair: My point here is that the link between float64 representation of Julian date and microseconds is not at all apparent without a lot of thought. My phrasing made the connection between the resolution with which we represent time and the interpolation size explicit.

astropy/coordinates/erfa_astrom.py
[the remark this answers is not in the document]
MaxNoe: the `obsgeoloc` and `obsgeovel` frame attributes of the `GCRS` frame are needed in `apcs` and those attributes are not present on an `AltAz` frame.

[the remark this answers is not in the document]
StuartLittlefair: Yes - I see that passing the wrong frame won’t work. 

That wasn’t the point I was trying to make. It’s that your doc string is too specific. The argument is either a frame or a skycoord. The docstring should specify that both are allowed. 

Since there’s no such class as a CIRSSkycoord so the explanation of what frame is expected needs to go below.

docs/coordinates/performance.inc.rst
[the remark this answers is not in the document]
StuartLittlefair: I don’t think so; I can imagine someone wanting to use interpolation inside a whole script or interpreter process. As long as they understand the consequences I don’t see why we should discourage that.

MaxNoe:
@bwinkel Something like this:

coords

astropy/coordinates/erfa_astrom.py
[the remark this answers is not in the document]
MaxNoe: done

docs/coordinates/performance.inc.rst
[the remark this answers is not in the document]
MaxNoe: ok, added an example for this

MaxNoe:
Where could such an image go into the documentation?

astropy/coordinates/erfa_astrom.py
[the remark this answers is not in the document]
bwinkel: Personally, I'd rather remove the `[mjd_scaled.min(), mjd_scaled.max()]` as it will lead to surprising results in some cases. A minimal example:
```python
t1 = Time('2020-01-01T20:00:00') + [1, 2, 3, 4, 5] * u.day
t2 = Time('2020-01-01T20:00:00') + [2, 3, 4, 5, 6] * u.day
location = EarthLocation.of_site('Roque de los Muchachos')
crab = SkyCoord.from_name("Crab Nebula")

frame1 = AltAz(obstime=t1, location=location)
frame2 = AltAz(obstime=t2, location=location)

with erfa_astrom.set(ErfaAstromInterpolator(10 * u.day)):
    azel1 = crab.transform_to(frame1)
    azel2 = crab.transform_to(frame2)
    # printing array entries, which are present in both cases
    print(f'{azel1[1:].az}\n{azel1[1:].alt}')
    print(f'{azel2[:-1].az}\n{azel2[:-1].alt}')
# Would expect the same results for both, which is not the case:
# [82.93128862 83.326738   83.72438029 84.12447817] deg
# [36.87963601 37.73749477 38.59604886 39.45526105] deg
# [82.93128338 83.32673447 83.72437851 84.1244782 ] deg
# [36.87963684 37.73749523 38.59604895 39.45526079] deg
```
It is true that the min-max injection will lead to more precise results, but I would prefer to get the same results for same input times. At the moment, this depends on which other numbers are in the arrays.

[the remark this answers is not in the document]
MaxNoe: Does that mean you would also like to have the interpolation for single values? Right now, that uses the non-interpolating case.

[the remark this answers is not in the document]
bwinkel: Good point. I'd say yes. If one doesn't want that behavior (interpolation even for single obstimes), one should not use the interpolated version anyways. But let's wait what others think about this.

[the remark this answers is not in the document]
mhvk: I'm a bit of two minds: ideally, the examples given should give the same answers, but it would also seem that if the code *knows* it is going to do more calculations for the interpolating case than for the regular case (as is certainly true for a single object), it might as well do the regular one. An in-between solution might be to always generate a complete MJD grid but discard items that for sure are never going to be used (i.e., that are not nearest right and left neighbours of requested MJD). That reasonably minimizes overhead to a factor 2 even in the worst case (single object or times separated by much more than the interpolation time), yet ensures the answers are always the same for a given interpolation time.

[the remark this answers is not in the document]
bwinkel: Thanks for your opinion @mhvk. I think, what you suggest is already the current behavior, as the `ceil` and `floor` function would result in a "sparse" support if the input obstimes are not homogeneously distributed. Or am I misunderstanding something here? The only problem, which I see here is that the min and max injection "destroys" the regular spacing of the support grid for the elements that fall into the first and last grid cell.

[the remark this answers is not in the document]
mhvk: Ah, yes, I clearly had a different mental picture of the routine than was its reality! I think overall, I'd tend to agree to just remove the special-casing near the edge, and probably also the scalar case. Just keep the code as simple as possible.

MaxNoe:
@mhvk Do you have an idea why the coverage is failing?

mhvk:
@MaxNoe - I think the coverage is done on a run without `remote_data` (since with `remote_data` we just get time-out failures too often). But I think you can have your tests be without that (@pllim removed most of the need for it!), as long as the times used are covered by our base IERS B table. I had only a quick look, but 2020 January should be included. Do the tests fail if you remove `remote_data`?

MaxNoe:
@mhvk Ok thanks, seems indeed it is unnecessary though. I switched to hard coded coordinates in between without relying on `from_name` or `of_site`, so I think for those it was needed before.

MaxNoe:
@StuartLittlefair Could you re-review? Are all your comments addressed?

StuartLittlefair (approved):

StuartLittlefair:
@MaxNoe sorry for the holiday induced delay! Looks good to me...

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,269 @@
...
+            For this function, a CIRS frame is expected.
+        '''
+        obstime = frame_or_coord.obstime
+        # no point in interpolating for a single value
+        if obstime.size == 1:
+            return super().apci(frame_or_coord)
+
+        support = self._get_support_points(obstime)
+
+        jd1_tt, jd2_tt = get_jd12(obstime, 'tt')
```
mhvk, line 221: Total nitpick, but I think this would be clearer just before the `erfa.apci` call (indeed, could move it inside the call with `*get_jd12(obstime, 'tt')`)

Alternatively, moving it *before* the `support = self....` line would also work to make clearer what belongs together.

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,269 @@
...
+            For this function, a GCRS frame is expected.
+        '''
+        obstime = frame_or_coord.obstime
+        # no point in interpolating for a single value
+        if obstime.size == 1:
+            return super().apci(frame_or_coord)
+
+        support = self._get_support_points(obstime)
+
+        jd1_tt, jd2_tt = get_jd12(obstime, 'tt')
```
mhvk, line 246: As above.

astropy/coordinates/tests/test_erfa_astrom.py
```diff
@@ -0,0 +1,140 @@
+import numpy as np
+import pytest
+
+
+def test_science_state():
+    import astropy.units as u
+    from astropy.coordinates.erfa_astrom import (
```
mhvk, line 7: Just move the imports up top (here and below)

astropy/coordinates/tests/test_erfa_astrom.py
```diff
@@ -0,0 +1,140 @@
...
+    coord = SkyCoord(ra=83.63308333, dec=22.0145, unit=u.deg)
+
+    # do the reference transformation, no interpolation
+    ref = coord.transform_to(altaz)
+
+    with erfa_astrom.set(ErfaAstromInterpolator(300 * u.s)):
+        interp_300s = coord.transform_to(altaz)
+
+    # make sure they are actually different
+    assert np.any(ref.separation(interp_300s) > u.Quantity(0.01, u.microarcsecond))
```
mhvk, line 68: Just `0.01 * u.microarcsecond` is clearer than creating a `Quantity` explicitly (same below).

astropy/coordinates/tests/test_erfa_astrom.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    from astropy.utils.exceptions import AstropyWarning
+
+    with pytest.warns(AstropyWarning):
+        with erfa_astrom.set(ErfaAstromInterpolator(9 * u.us)):
+            pass
+
+
+def test_erfa_astrom():
+    # I was having a pretty hard time in coming
```
mhvk, line 39: Agreed.

astropy/coordinates/tests/test_erfa_astrom.py
```diff
@@ -0,0 +1,140 @@
...
+def test_interpolation_broadcasting():
+    from astropy.coordinates.tests.utils import randomly_sample_sphere
+    from astropy.coordinates import SkyCoord, EarthLocation, AltAz
+    from astropy.time import Time
+    import astropy.units as u
+
+    from astropy.coordinates.erfa_astrom import erfa_astrom, ErfaAstromInterpolator
+
+    # 1000 random locations on the sky
+    ra, dec, _ = randomly_sample_sphere(1000)
```
mhvk, line 120: Can we use smaller numbers (say, 10)? This test will be quite slow otherwise (and will get run >10 times for every push for every PR).

(I'd slightly prefer just including a bit in the above test, also trying different time shapes, but not very important.)

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,157 @@ properties::
...
+
+
+Improving Performance for Arrays of ``obstime``
+-----------------------------------------------
+
+The most expensive operations when transforming between observer-dependent coordinate
+frames (e.g. ``AltAz``) and sky-fixed frames (e.g. ``ICRS``) are the calculation
+of the orientation and position of Earth.
+
+If |skycoord| instances with a large ``obstime`` array are transformed,
```
mhvk, line 68: Maybe "If |skycoord| instances are transformed for a large number of closely spaced ``obstime``"

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,157 @@ properties::
...
+these calculations can be sped up by factors up to 100, whilst still keeping micro-arcsecond precision,
+by utilizing interpolation instead of calculating Earth orientation parameters for each individual point.
+
+This can be achieved through the ``erfa_astrom`` state and the ``ErfaAstromInterpolator``
```
mhvk, line 72: I'd just delete this sentence (if not, do link the text properly with either the above or the below, and ensure that `erfa_astrom` and `ErfaAstromInterpolator` are links, as in `~astropy.coordinates.erfa_astrom.erfa_astrom`)

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,157 @@ properties::
...
+   >>> from astropy.coordinates import SkyCoord, EarthLocation, AltAz
+   >>> from astropy.coordinates.erfa_astrom import erfa_astrom, ErfaAstromInterpolator
+   >>> from astropy.time import Time
+   >>> from time import perf_counter
+   >>> import numpy as np
+   >>> import astropy.units as u
+
+
+   >>> # array with 10000 obstimes
+   >>> obstime = Time('2020-01-01T20:00') + np.linspace(0, 6, 10000) * u.hour
```
mhvk, line 90: I'd pick an earlier time so remote data is not necessary.

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,157 @@ properties::
...
+   >>> location = location = EarthLocation(lon=-17.89 * u.deg, lat=28.76 * u.deg, height=2200 * u.m)
+   >>> frame = AltAz(obstime=obstime, location=location)  # doctest:+REMOTE_DATA
+   >>> crab = SkyCoord.from_name('Crab')  # doctest:+REMOTE_DATA
```
mhvk, line 93: Similarly, use an explicit coordinate so we do not need remote data.

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,157 @@ properties::
...
+
+   >>> # transform with default transformation and print duration
+   >>> t0 = perf_counter()
+   >>> crab_altaz = crab.transform_to(frame)  # doctest:+REMOTE_DATA +IGNORE_WARNINGS
+   >>> print(f'Transformation took {perf_counter() - t0:.2f} s')  # doctest:+ELLIPSIS
```
mhvk, line 98: How about `+IGNORE_OUTPUT` here and then give the actual number? The point here is more to show how much it helps than to get the doctest to pass!

docs/coordinates/performance.inc.rst
```diff
@@ -56,3 +56,157 @@ properties::
...
+..
+  EXAMPLE END
+
+
+Here, we look into choosing an appropriate ``time_resolution``. 
+We will transform a single sky coordinate for lots of observation times from
+``ICRS`` to ``AltAz`` and evaluate precision and runtime for different values
+for ``time_resolution`` compared to the non-interpolating, default approach.
+
+.. plot::
```
mhvk, line 124: Very nice to include this plot - best possible way to help people determine how to check

mhvk (commented):
@MaxNoe - looked over carefully one more time, now more focussing on details - so, quite a few nitpicks...

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,269 @@
...
+This module contains a helper function to fill erfa.astrom struct and a
+ScienceState, which allows to speed up coordinate transformations at the
+expense of accuracy.
+"""
+import warnings
+
+import numpy as np
+import erfa
+
+from ..time import Time
```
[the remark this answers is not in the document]
MaxNoe: Just out of interest,  could you point me to the reasoning behind this?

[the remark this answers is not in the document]
mhvk: It was done in #8200, with most of the discussion being in #6452 (which started about imports for tests).

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,269 @@
...
+    >>> import numpy as np
+
+    >>> obstime = Time.now() + np.linspace(0, 4, 1000) * u.hour
+    >>> with erfa_astrom.set(ErfaAstromInterpolator(300 * u.s)):
+    ...    cirs = SkyCoord.from_name('Crab').transform_to(CIRS(obstime=obstime))  # doctest: +REMOTE_DATA
+    '''
+
+    @u.quantity_input(time_resolution=u.day)
+    def __init__(self, time_resolution):
+        if time_resolution.to_value(u.us) < 10:
```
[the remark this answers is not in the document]
MaxNoe: It's a factor of > 4 slower. I see that is not a performance bottleneck here, but we made the experience that one should use the best practices as people reading the code adapt usage they see. 

This is also the reason for using `u.Quantity(value, unit)` instead of `value * unit` below.

Again, I see this is not performance critical here but I don't see a real legibility improvement through the other, worse performing notation.

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,269 @@
...
+        mjd_scaled = np.ravel(obstime.mjd / self. mjd_resolution)
+
+        # ndmin=1 needed in case only a single coordinate is transformed
+        mjd_lower = np.array(np.floor(mjd_scaled), ndmin=1, copy=False)
+        mjd_upper = np.array(np.ceil(mjd_scaled), ndmin=1, copy=False)
+
+        # unique already does sorting
+        mjd_u = np.unique(np.concatenate([mjd_lower, mjd_upper]))
+
+        return Time(
```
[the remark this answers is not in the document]
MaxNoe: It's 83 columns, do we not observe the 80 columns limit?

astropy/coordinates/erfa_astrom.py
MaxNoe: I actually already had this change :)

astropy/coordinates/tests/test_erfa_astrom.py
MaxNoe: I made the experience that having imports in the test functions is a dramatic quality of life improvement when import breaks.

Having the imports inside the test will make the test fail instead of making the test collection fail.

This allows e.g. `pytest --last-failed` to work properly.

docs/coordinates/performance.inc.rst
MaxNoe: done

docs/coordinates/performance.inc.rst
MaxNoe: done

docs/coordinates/performance.inc.rst
MaxNoe: I thought the actual output would still be reported, but of course it can't be.

docs/coordinates/performance.inc.rst
MaxNoe: Even with a hard coded coordinate and a 2010 time it still needs remote data for the cirs / altaz transformation

docs/coordinates/performance.inc.rst
MaxNoe: I used a 2010 time and it still required remote data

docs/coordinates/performance.inc.rst
MaxNoe: The iers `finals2000A.a` file

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,269 @@
...
+    def _get_support_points(self, obstime):
+        '''
+        Calculate support points for the interpolation.
+
+        We divide the MJD by the time resolution (as single float64 values),
+        and calculate ceil and floor.
+        Then we take the unique and sorted values and scale back to MJD.
+        This will create a sparse support for non-regular input obstimes.
+        '''
+        mjd_scaled = np.ravel(obstime.mjd / self. mjd_resolution)
```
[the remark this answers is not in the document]
MaxNoe: fixed

astropy/coordinates/erfa_astrom.py
```diff
@@ -0,0 +1,269 @@
...
+        '''
+        jd1_tt, jd2_tt = get_jd12(frame_or_coord.obstime, 'tt')
+        cip = get_cip(jd1_tt, jd2_tt)
+        earth_pv, earth_heliocentric = prepare_earth_position_vel(frame_or_coord.obstime)
+        return erfa.apci(jd1_tt, jd2_tt, earth_pv, earth_heliocentric, *cip)
+
+    @staticmethod
+    def apcs(frame_or_coord):
+        '''
+        Wrapper for ``erfa.apci``, used in conversions GCRS <-> ICRS
```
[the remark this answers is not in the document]
MaxNoe: fixed

astropy/coordinates/erfa_astrom.py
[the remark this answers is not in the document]
MaxNoe: ok. Used absolute imports for everything but the imports from the same sub-module.

astropy/coordinates/erfa_astrom.py
MaxNoe: fixed

astropy/coordinates/tests/test_erfa_astrom.py
MaxNoe: I reduced all numbers in the tests

MaxNoe:
@mhvk I adressed most of your comments by simply taking you suggestions but left a few out for now. I have no real problems with following your suggestions anywhere but nitpicked back a bit :rofl:

astropy/coordinates/erfa_astrom.py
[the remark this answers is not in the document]
astrojuanlu: What about `value << unit` then?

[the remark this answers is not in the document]
MaxNoe: I don't see how the `<<` operator would help in this comparison.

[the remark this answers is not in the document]
MaxNoe: Also, `<<` is probably the worst option when it comes to readability.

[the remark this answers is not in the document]
astrojuanlu: I suggested it hoping that it would be slightly faster than `value * unit`, but the performance gain [of this trick]() is only visible when working with large arrays.

I can confirm that your version is faster:

```python
In [11]: %timeit time_resolution < 10 * u.us                                                                                                                                                                                                  
22 µs ± 357 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [12]: %timeit time_resolution < (10 << u.us)                                                                                                                                                                                               
22.4 µs ± 214 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [13]: %timeit time_resolution.to_value(u.us) < 10                                                                                                                                                                                          
1.11 µs ± 10.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

docs/coordinates/performance.inc.rst
mhvk: Hmm, that is strange - it should not need `finals2000A` if the time is well within the bundled `IERS_B` file. But definitely a different issue, so let's stick with what you have.

astropy/coordinates/tests/test_erfa_astrom.py
mhvk: I can see that being useful if the imports in all the tests are different, but that is not the case here... Also, this is something we try to do everywhere in astropy...

MaxNoe: Ok. But maybe something worth discussing, as a single failed import will stop the full test collection and then mess with the `--last-failed` and similar functions of pytest. I'll change here and maybe open an issue.

MaxNoe: done

MaxNoe:
@mhvk Ok, I think that's all comments adressed.

mhvk (approved):
@MaxNoe - OK, looks all fine now, so I'll go ahead and merge. Thanks so much for making this happen - it went from a feature I wasn't all that sure about to a really nice implementation that both solves the original purpose and sets us up a nice framework for possible future extensions.

On the imports, I do think that if we cannot import things from astropy, I think one would want test collection to fail immediately, as something is quite broken! But there are obviously more corner cases where an import depends on something else that is not available, and I'm not sure the tangle of `pytest.skipif(...)` decorators we have now is the best system.

Anyway, always good to discuss whether things can be improved, so definitely do open an issue with suggestions!

bwinkel:
@MaxNoe thank you so much for making this happen. Made my day!

PR #10649, opened 2020-08-13
Include 3-d list for full coverage of list init code

taldcroft:
### Description

This is a follow-up to #10636 since I noticed the inline code coverage comment that [one line of the code](https://github.com/astropy/astropy/pull/10636/files#diff-30ea8400f76ea2c3dfd24787113276f6R248) there was not tested. Indeed that was correct and this PR adds a 3-d list that will now test that line.

### DIFF
--- a/astropy/table/tests/test_masked.py
+++ b/astropy/table/tests/test_masked.py
@@ -183,14 +183,14 @@ class TestTableInit(SetupData):
     """Initializing a table"""
 
     @pytest.mark.parametrize('type_str', ('?', 'b', 'i2', 'f4', 'c8', 'S', 'U', 'O'))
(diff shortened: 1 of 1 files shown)


taldcroft:
This is a pretty trivial change - anybody watching could review.  @pllim, @mhvk?

taldcroft:
Yes, the inline comment by codecov actually made me see the problem. Previously I was just uniformly ignoring it.

taldcroft:
What I want to know is if codecov will give me a gold star for fixing a coverage problem.

mhvk:
That in-line comment does indeed seem nice - at least one knows it is related to one's own PR!

Issue #6828, opened 2017-11-10
TimeDelta includes some methods that don't apply


astrofrog:
It looks like ``TimeDelta`` includes some methods it probably shouldn't. For example ``TimeDelta.now()`` doesn't make sense, and maybe ``TimeDelta.light_travel_time``.

mhvk:
@taldcroft - these methods are inherited directly from `Time`. To avoid this, should we have a `TimeBase` from which both `Time` and `TimeDelta` inherit? It is rather painful in that then `isinstance(..., Time)` will no longer catch a `TimeDelta`. But I'm not sure how else to solve this.

Cadair:
Could some combo of ABCs and subclass hooks fix the `isinstance(..., TIme)` work?

taldcroft:
Good point.  The only question is if we are getting into "perfect is the enemy of good" territory (unless @Cadair can produce the appropriate magic).

astrofrog:
@mhvk - just to check, why should ``isinstance(..., Time)`` catch ``TimeDelta``? Those are two quite different things, so in fact ``isinstance(..., Time)`` returning ``False`` for ``TimeDelta`` seems like a feature to me :)

astrofrog:
Just to clarify my comment, here's what I get with the datetime module:

```python
In [3]: from datetime import datetime

In [5]: dt1 = datetime.now()

In [6]: dt2 = datetime.now()

In [7]: dt2 - dt1
Out[7]: datetime.timedelta(0, 4, 731509)

In [8]: isinstance(dt2 - dt1, datetime)
Out[8]: False

In [9]: isinstance(dt1, datetime)
Out[9]: True
```

There's really no reason ``TimeDelta`` should be as a subclass of ``Time``.

mhvk:
@astrofrog - I agree with you that conceptually they're different and it makes sense to have a `TimeBase`. It is just that this may break code, as it is not how it is implemented...

PR #10652, opened 2020-08-13
 Prevent comparisons between Time and TimeDelta from succeeding.

### LINKED ISSUE #10651: Time comparisons sometimes fail to fail
From https://github.com/astropy/astropy/pull/10646#discussion_r470067619
```
In [21]: TimeDelta(1*u.day, scale='tai') < Time('J2000')                                                                            
Out[21]: True
```

The reason this fails is that the comparisons first try to get both arguments to the same class and scale, and this works when it wasn't expected to work:
```
In [35]: Time(TimeDelta(1<<u.day, scale='tt'))                                                         
Out[35]: <TimeDelta object: scale='tt' format='jd' value=1.0>
```

It is not completely obvious whether generically a superclass passing back a subclass unchanged is OK, but for `Time` is shouldn't happen!

mhvk:
fixes #10651

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -166,6 +166,9 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- ``Time`` can no longer be initialized with subclasses such as ``TimeDelta``,
+  thus allowing one to count on getting a ``Time`` instance. [#10652]
+
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
 
@@ -246,6 +249,9 @@ astropy.tests
 
 astropy.time
 ^^^^^^^^^^^^
+
+- Prevent comparisons between ``Time`` and ``TimeDelta`` from sometimes succeeding.
+  [#10652]
 
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -395,6 +395,8 @@ class Time(ShapedLikeNDArray):
             update_leap_seconds()
 
         if isinstance(val, cls):
+            if not type(val) is cls:
+                raise TypeError(f"cannot instantiate {cls} with a {type(val)}.")
             self = val.replicate(format=format, copy=copy)
         else:
             self = super().__new__(cls)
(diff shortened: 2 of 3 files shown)


taldcroft:
This has made me think about what should happen when passing a Time subclass into Time. Previously the subclass would come out directly, which is definitely wrong. But my intuition is that you would get back out the same moment(s) of time but represented in a pure `Time` object.  E.g. we have a subclass of `Time` called `CxoTime` for Chandra ops work. One should be able to do:
```
Time(CxoTime('2020:001'))
```
Enabling this behavior would probably mean taking out the init handling in `__new__` (which I put into #523) and doing something more like a4b8041 in #523. And you would need to have some special handling to specifically exclude subclasses of `TimeDelta` in the `Time` initializer.

Anyway, I'm OK with merging this basically as-is, but with a caveat of adding an API change entry noting that `Time` and `TimeDelta` can no longer be initialized from sub-classes. We could add a placeholder issue to rework the initialization to allow init from an "appropriate" subclass.

mhvk:
@taldcroft - yes, I agree, one would expected `Time(TimeSubClass(...))` to work, just like `Quantity(Angle(...))` returns a `Quantity`. It would also be solved if we restructured to have a `TimeBase` class of which both `Time` and `TimeDelta` are subclasses - see https://github.com/astropy/astropy/issues/6828#issuecomment-343749941

One thing regarding the API change: that is nominally not OK for 4.0.2 -- though I guess the current behaviour for the initializer is somewhat of a bug.

taldcroft:
Yeah, that `TimeBase` idea is good.

I guess there are two options:
- Backport this and make the nominal bug fix available for 4.0.x, 4.1.x, introducing a change that might break existing code.
- Only put this into 4.2.

I feel like this bug is not so critical and would be inclined to only putting it into 4.2. The API says that you can always initialize a Time object with another Time object, so I think that fixing the current incorrect behavior for subclasses needs to be in a feature release.

mhvk:
Yes, let's go for 4.2; it is not like we've noticed the bug for a couple of years!

mhvk:
Rebased. And raised #10654 to remind us of the subclass issue and the idea of a `TimeBase` superclass.

mhvk:
No longer necessary, as the issue is fixed (by #10656)

PR #10656, opened 2020-08-14
Refactor `Time` and `TimeDelta` to inherit from common `TimeBase` class

### LINKED ISSUE #6828: TimeDelta includes some methods that don't apply

### LINKED ISSUE #10654: Time subclasses and using them to initialize Times
`Time.__new__` has special treatment for `Time` instances, essentially passing them through with perhaps a scale change. It turned out this included `TimeDelta`, so that `Time(TimeDelta(...))` returned a `TimeDelta` instance, which was surprising and caused #10651. In #10652, the bug was fixed by restricting the special case to instances of the class proper. But as @taldcroft [notes](https://github.com/astropy/astropy/pull/10652#issuecomment-673617676), in general one would like for `Time(TimeSubClass(...))` to work and to return a `Time` instance. It would be good to allow that.

Part of the problem is that `TimeDelta` should probably not be a subclass of `Time` - indeed, as noted in #6828 it shares methods such as `.now()` and `.light_travel_time()` that really make no sense. In #6828, it was suggested a better scheme might be to have both `Time` and `TimeDelta` inherit from a common `TimeBase`.

taldcroft:
### Description

This pull request is to address a longstanding issue that `TimeDelta` inherited from `Time` but many of the methods (e.g. `light_travel_time()` were really not applicable. Instead now there is a base class `TimeBase` that has the common methods and each independently inherits from `TimeBase`.

After this refactor there were only two problems that showed up in the entire astropy code base, both related to assuming `Time` as the base class. In one case `TimeBase.insert()` this was fixed using `self.__class__` and in the other `Table.to_pandas()` using `TimeBase`. There were no other code changes apart from moving methods around intact.

Fixes #6828
Fixes #10654.

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -1333,113 +1333,6 @@ class TimeBase(ShapedLikeNDArray):
 
         return val
 
-    def __sub__(self, other):
-        # T      - Tdelta = T
-        # T      - T      = Tdelta
-        other_is_delta = not isinstance(other, Time)
-        if other_is_delta:  # T - Tdelta
-            # Check other is really a TimeDelta or something that can initialize.
-            if not isinstance(other, TimeDelta):
-                try:
-                    other = TimeDelta(other)
-                except Exception:
-                    return NotImplemented
-
-            # we need a constant scale to calculate, which is guaranteed for
-            # TimeDelta, but not for Time (which can be UTC)
-            out = self.replicate()
-            if self.scale in other.SCALES:
-                if other.scale not in (out.scale, None):
-                    other = getattr(other, out.scale)
-            else:
-                if other.scale is None:
-                    out._set_scale('tai')
-                else:
-                    if self.scale not in TIME_TYPES[other.scale]:
-                        raise TypeError("Cannot subtract Time and TimeDelta instances "
-                                        "with scales '{}' and '{}'"
-                                        .format(self.scale, other.scale))
-                    out._set_scale(other.scale)
-            # remove attributes that are invalidated by changing time
-            for attr in ('_delta_ut1_utc', '_delta_tdb_tt'):
-                if hasattr(out, attr):
-                    delattr(out, attr)
-
-        else:  # T - T
-            # the scales should be compatible (e.g., cannot convert TDB to LOCAL)
-            if other.scale not in self.SCALES:
-                raise TypeError("Cannot subtract Time instances "
-                                "with scales '{}' and '{}'"
-                                .format(self.scale, other.scale))
-            self_time = (self._time if self.scale in TIME_DELTA_SCALES
-                         else self.tai._time)
-            # set up TimeDelta, subtraction to be done shortly
-            out = TimeDelta(self_time.jd1, self_time.jd2, format='jd',
-                            scale=self_time.scale)
-
-            if other.scale != out.scale:
-                other = getattr(other, out.scale)
-
-        jd1 = out._time.jd1 - other._time.jd1
-        jd2 = out._time.jd2 - other._time.jd2
-
-        out._time.jd1, out._time.jd2 = day_frac(jd1, jd2)
-
-        if other_is_delta:
-            # Go back to left-side scale if needed
-            out._set_scale(self.scale)
-
-        return out
-
-    def __add__(self, other):
-        # T      + Tdelta = T
-        # T      + T      = error
-        if isinstance(other, Time):
-            raise OperandTypeError(self, other, '+')
-
-        # Check other is really a TimeDelta or something that can initialize.
-        if not isinstance(other, TimeDelta):
-            try:
-                other = TimeDelta(other)
-            except Exception:
-                return NotImplemented
-
-        # ideally, we calculate in the scale of the Time item, since that is
-        # what we want the output in, but this may not be possible, since
-        # TimeDelta cannot be converted arbitrarily
-        out = self.replicate()
-        if self.scale in other.SCALES:
-            if other.scale not in (out.scale, None):
-                other = getattr(other, out.scale)
-        else:
-            if other.scale is None:
-                out._set_scale('tai')
-            else:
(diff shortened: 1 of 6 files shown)


taldcroft:
@mhvk - check out e8af382, which (I think) solves the whole subclass problem and would fix #10654.

I don't have time right now to write the tests, but basically I got expected results with:
```
t = Time.now()
class MyTime(Time):
    pass
mt = MyTime.now()

MyTime(t, format='iso')
Time(mt, format='iso')
```

Note that in __new__ and __init__ one now has hardwired references to `Time` or `TimeDelta`, which are the prerequisite base classes in order to work properly.

I can put this change in a separate PR if you prefer.

taldcroft:
Also note that last commit is a bit of a WIP, there are docstring updates and other possible issues which I'll fix.

taldcroft:
This is failing on one TimeSeries test, will check.

astropy/time/core.py
```diff
@@ -317,53 +317,8 @@ def new_like(self, cols, length, metadata_conflicts='warn', name=None):
...
-        length for geocentric coordinates, or contain a longitude, latitude,
-        and an optional height for geodetic coordinates.
-        Can be a single location, or one for each input time.
-    copy : bool, optional
-        Make a copy of the input values
-    """
+class TimeBase(ShapedLikeNDArray):
+    """Base time class from which Time and TimeDelta inherit."""
 
     SCALES = TIME_SCALES
```
mhvk, line 323: `SCALES` and `FORMATS` should move to `Time`

astropy/timeseries/sampled.py
```diff
@@ -92,7 +92,7 @@ def __init__(self, data=None, *, time=None, time_start=None,
         elif time is not None and time_start is not None:
             raise TypeError("Cannot specify both 'time' and 'time_start'")
 
-        if time is not None and not isinstance(time, Time):
+        if time is not None and not isinstance(time, (Time, TimeDelta)):
```
mhvk, line 95: Would it make sense to just import `TimeBase` and check for that?

mhvk:
@taldcroft - I pushed directly to your branch some changes (in separate commits, just in case you'd prefer to do things differently); I also amended your `TimeSeries` commit to do `isinstance(..., TimeBase)`, also in `binned.py`, which at least locally fixed the failing doctest.  The commit moving `__add__`, etc., is the most significant, as those did `isinstance` checks that could now be changed.

I think there is a bit more refactoring that could be done, given the code now in TimeBase that is looking for attributes like `_delta_*`, `location`, etc., but I think we might as well do that after, since it is a bit more intricate and keeping it there for now does not really change anything (the same could be argued for the move of `_make_value_equivalent`, which is after all simply overwritten in `TimeDelta`, but it is very `Time`-specific).

astropy/timeseries/sampled.py
taldcroft: I thought about this, but in the end I think that the code is more readable and explicit this way, as in it tells the non-expert reader immediately that the results apply to both absolute and delta times.

It's an arbitrary line, but I thought that code inside of the time subpackage should use `TimeBase` (internal) while the rest would be explicit. I guess there is some philosophy here -- if we or someone else adds a new subclass of `TimeBase`, do we want that to be in or out from consideration? One can make valid arguments both ways.

taldcroft:
@mhvk - thanks for pushing forward with a bunch more changes.  Alas I had made most of those changes already locally and was doing the last bit of tidy / double-checking but got stopped before I could push. I'm pretty sure the two versions of changes will be unreconcilable by automated means, but I'll have a look manually. One thing I did was look for all instances of `isinstance\(.*, Time\)` and evaluated how those should be.  I do want to go back to `(Time, TimeDelta)` for reasons mentioned.

taldcroft:
Reminder to open an issue about testing in TimeSeries @astrofrog. There were a couple of failures along the way that only showed up in the doc testing. IIRC the first was about initializing a TimeSeries with a TimeDelta.

astropy/timeseries/sampled.py
mhvk: @taldcroft - indeed, and the advantage of `Time, TimeDelta` was that both were imported already. Note that I changed it to `TimeBase` in the later commit, but that was really because I thought that might be the reason for the doctest failure (which was instead that the same change had to be made in `binned.py`).

mhvk:
@taldcroft - do feel free to just remove my commits!

taldcroft:
RTD fail seems unrelated:
```
WARNING: the readthedocs_ext.readthedocs extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
WARNING: doing serial read
```

pllim:
No, it is indeed not related. See #10668

taldcroft:
@mhvk - AFAIK all your comments have been addressed.  Also from what I can see this entirely fixes #10654 so I updated the description accordingly.
```
In [2]: class TimeSub(Time): 
   ...:     pass 
   ...:                                                                                                                      

In [3]: ts = TimeSub('2000:001')                                                                                             
In [4]: ts                                                                                                                   
Out[4]: <TimeSub object: scale='utc' format='yday' value=2000:001:00:00:00.000>

In [5]: t = Time(ts)                                                                                                         
In [6]: t                                                                                                                    
Out[6]: <Time object: scale='utc' format='yday' value=2000:001:00:00:00.000>

In [7]: td = TimeDelta(1)                                                                                                    
In [8]: Time(td)  # Fail                                                                                                             
```

mhvk (approved):
@taldcroft - this looks very good indeed! I feel a bit bad about the coverage misses, but all were there before and are of the "obviously correct code" type, so might as well leave that for another time, when we just aim go get coverage of time up to 100%. So, merging!

PR #10659, opened 2020-08-15
Account for removal of financial functions in numpy 1.20

mhvk:
For once, I may have made the fix before @pllim found the problem! (But I had warning from failures with numpy and astropy-dev in my own package...). Numpy is removing the financial functions and that causes failures because I accessed these just to then ignore them for anything `Quantity` related. So, the fix is just to ignore them more.

Should be easy to backport...

EDIT: measure of success is that tests pass, with the -dev run only giving the trim-zeroes error.

### DIFF
--- a/astropy/units/quantity_helper/function_helpers.py
+++ b/astropy/units/quantity_helper/function_helpers.py
@@ -42,7 +42,8 @@ import numpy as np
 
 from astropy.units.core import (
     UnitsError, UnitTypeError, dimensionless_unscaled)
-from astropy.utils.compat import NUMPY_LT_1_17, NUMPY_LT_1_18
+from astropy.utils.compat import (NUMPY_LT_1_17, NUMPY_LT_1_18,
+                                  NUMPY_LT_1_20)
 from astropy.utils import isiterable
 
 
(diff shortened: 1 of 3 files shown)


mhvk:
Actually, it looks like this is slightly premature - numpy-dev has been updated, but the nightly not yet.

pllim:
> but the nightly not yet

Yeah, we use the "nightly" wheel that's actually weekly.

pllim:
So, I see this error in `master` cron build now. Is this what you are trying to fix or is this something else?
```
AttributeError: In accordance with NEP 32, the function fv was removed from NumPy version 1.20.  A replacement for this function is available in the numpy_financial library: https://pypi.org/project/numpy-financial
ERROR: InvocationError for command .../pytest --pyargs astropy /home/travis/build/astropy/astropy/docs --remote-data=any (exited with code 1)
```

mhvk:
Yes, that's it! I've restarted the numpy-dev travis run to check that this PR indeed fixes the problem...

mhvk:
Just to be sure, the master cron job fails for numpy==1.20.0.dev0+986e533 (see https://travis-ci.org/github/astropy/astropy/jobs/718551567), with pytest not even getting to run. The same numpy version will not prevent runs here (though there are still the `trim_zeros` and a lot of URL timeouts).

pllim:
I'll take that as agreement. Thanks! 😸

mhvk:
OK, great! (I'm actually less than pleased that the numpy upgrade *broke* the pytest run, but will leave that for another PR... Fortunately, removal of functions is quite rare...)

pllim:
Module level `__getattr__` not possible here until we drop Python 3.6, see astropy/astropy-APEs#62

mhvk:
@pllim - the `__getattr__` is used inside numpy - for possible use with already released versions of astropy. For any new versions, the changes here should suffice. Nicely, the changes in numpy will not need any changes here, since the financial functions are still not present in the numpy namespace.

PR #10664, opened 2020-08-17
MNT: Drop Numpy 1.16

pllim:
### Description

This pull request is to bump minimum supported version of Numpy from 1.16 to 1.17. This also adds CI for Numpy 1.19.

This is a prerequisite for #10644 .


pllim:
Oooo... Failures with Numpy 1.19!

pllim:
@astrofrog or @Cadair , how do I upgrade Scipy version in the failed `mpl` jobs?

dhomeier:
Possibly of interest: SciPy is currently having a discussion about [dropping Python 3.6 and NumPy 1.14 / 1.15 support](https://mail.python.org/pipermail/scipy-dev/2020-August/024318.html) – so slightly offset to this plan, but some arguments e.g. about available versions in the various Linux distros might apply.

mhvk:
I think a good point being made in that thread is that when a user updates astropy, updating numpy at the same time is essentially trivial (if not already automatic), but for python itself one is much more dependent on OS, etc.
So, :+1: on this PR - after all, we do offer 4.0-lts for those who prefer sticking to older versions.

astropy/table/operations.py
```diff
@@ -1168,7 +1167,7 @@ def _join(left, right, keys=None, join_type='inner',
 
             out[out_name] = col_cls.info.new_like(cols, n_out, metadata_conflicts, out_name)
 
-            if not NUMPY_LT_1_17 or issubclass(col_cls, (Column, Time)):
```
mhvk, line 1171: `not NUMPY_LT_1_17` will now always be true, so I think we can de-indent what follows and remove the `else` clause! (I always wondered why we didn't use `NUMPY_GE_<something>`!)

astropy/units/quantity.py
```diff
@@ -428,7 +427,7 @@ def __array_wrap__(self, obj, context=None):
 
         raise NotImplementedError('__array_wrap__ should not be used '
                                   'with a context any more, since we require '
-                                  'numpy >=1.16.  Please raise an issue on '
+                                  'numpy >=1.17.  Please raise an issue on '
```
mhvk, line 430: Let's remove the version here - maybe "since all use should go through __array_function__. Please ..."

astropy/utils/iers/iers.py
```diff
@@ -484,7 +476,7 @@ def _combine_a_b_columns(cls, iers_a):
         # IERS B values in the table are consistent with the true ones.
         table = cls._substitute_iers_b(table)
 
-        q_where = cls._quantity_where if NUMPY_LT_1_17 else np.where
+        q_where = np.where
```
mhvk, line 479: Ideally, we'd just replace `q_where` with `np.where`... (sorry to have set up this extra work for you!)

pllim:
Thanks, @mhvk ! I'll address the review comments in the next round of edit. I'm waiting for CircleCI fix idea before proceeding.

astropy/table/operations.py
pllim: Good catch!

pllim:
I hacked around the CircleCI failures. The one remote data failure is unrelated and hopefully will go away when numpy nightly is refreshed.

mhvk:
Before I merge, I think there was consensus that dropping 1.16 was OK, right?

Issue #10356, opened 2020-05-15
CIRS is not at the ultimate accuracy


mhvk:
Copying @mkbrewer's https://github.com/astropy/astropy/issues/10230#issuecomment-629491534:

>It is unfortunate that you only implement geocentric position for your CIRS frame. You use apci() and apio() rather than the more rigorous apco() which builds the observer's topocentric position and velocity (if present) right in. This has two side effects. In going from ICRS to CIRS light deflection doesn't take into account a topocentric observer's position and velocity and in going from CIRS to AltAz the less rigorous simple first-order vector expression for diurnal aberration as mentioned in paragraph 4.7 of the SOFA Astrometry Tools document is used.

It sounds like we could improve things fairly simply since GCRS seems OK:

> On the other hand, your GCRS frame does implement topocentric position rigorously. Therefore, I find it curious that you don't derive your CIRS frame from GCRS. After all, with GCRS you are just one step of applying the CIO based BPN matrix from true topocentric CIRS intermediate (apparent) place. The call to apcs() in this transform properly sets up the ASTROM data structure for rigorously calculating gravitational light deflection, diurnal aberration and diurnal parallax including the observer's position and velocity (including taking the observer's velocity into account in beta - 1). I particularly like the way that diurnal aberration is naturally included in Ab() this way, so that the transformation from topocentric CIRS to AltAz can be done without resorting to the use of astrom->diurab.

mkbrewer:
Now that GCRS is pretty much all set, it's time to address the other elephant in the room which are the GCRS to CIRS and CIRS to observed transforms. The problem here is that instead of following the `atco` chain, you chose to use `atci` and `atio` instead. What SOFA fails to mention, is that the later is completely incompatible with solar system bodies where diurnal parallax comes into play (although, if you look closely, the documentation for `atci13()` does say that it is specifically for star data) This is especially true of the Moon, so that is what I've concentrated on here. There, I can't help but notice a rather large (to my standards) error in the returned Alt/Az coordinates. Here is the error between your Alt/Az and mine in the latest data file that I have from you. 
![AzElError](https://user-images.githubusercontent.com/10635444/90988111-3cebe300-e55e-11ea-8bf7-1e9388aa757f.png)
This is in coordinate space and on this date, the Moon reaches lower meridian transit at 13:05 UT. At that time the elevation is -80.5 degrees, so I also plotted the actual offset on the sky which looks better, but is still pretty bad.
[...]

PR #10666, opened 2020-08-18
Include a realistic treatment of solar light deflection for nearby objects

### LINKED ISSUE #10291: Account for distance properly in aticq and atciqz

StuartLittlefair:
### Description

This pull request addresses an issue where the Solar light deflection was always being applied assuming the source was an infinite distance away, so that the observer -> source vector equalled the Sun -> source vector. In the absolute worst case, this causes errors of up to 3 arc seconds. See #10291 for full details.

Implementation was more awkward than I expected, since we need special case treatments for objects which appear to actually be the Sun itself, and objects at the origin of their frames. I've also special cased the example where there is no distance, to reduce performance impact.

Despite the added complexity in ```astropy.coordinates.builtin_frames.utils.atciqz``` and ```astropy.coordinates.builtin_frames.utils.aticq``` there is no major performance impact. Obtaining the GCRS position of the Moon at 400 times and transforming to ICRS and thence to CIRS took no detectable extra time compared to astropy master - as determined using the test code below.

```python
from time import perf_counter
import numpy as np
from astropy.time import Time
import astropy.units as u
from astropy.coordinates import EarthLocation, CIRS, ICRS, get_body

times = Time('2020-08-21T00:00') + np.linspace(-3*u.hour, 3*u.hour, 400)
location = EarthLocation.of_site('lapalma')
t0 = perf_counter()
moon = get_body('moon', times, location=location)
moon_icrs = moon.transform_to(ICRS)
moon_cirs = moon.transform_to(CIRS(obstime=moon.obstime))
print(f'Transforms took {perf_counter() - t0:.2f} s')
```

Despite these changes, I have not been able to reproduce without any discrepancies the Moon positions supplied by @mkbrewer in #10291. The discrepancies are however reduced:

master_comparison
current_comparison

I've attached the output of @mkbrewer s test script and the script itself:

[moon_topo_astropy_thispr.txt](https://github.com/astropy/astropy/files/5090651/moon_topo_astropy_thispr.txt)
[skyfield_topo_test.py.txt](https://github.com/astropy/astropy/files/5090671/skyfield_topo_test.py.txt)

If @mkbrewer has any input as to the remaining discrepancies I'd be extremely grateful to receive it!

Fixes #10291

### DIFF
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -37,10 +37,7 @@ def icrs_to_cirs(icrs_coo, cirs_frame):
 
     if icrs_coo.data.get_name() == 'unitspherical' or icrs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just do the infinite-distance/no parallax calculation
-        usrepr = icrs_coo.represent_as(UnitSphericalRepresentation)
-        i_ra = usrepr.lon.to_value(u.radian)
-        i_dec = usrepr.lat.to_value(u.radian)
-        cirs_ra, cirs_dec = atciqz(i_ra, i_dec, astrom)
+        cirs_ra, cirs_dec = atciqz(icrs_coo.cartesian.without_differentials(), astrom)
 
         newrep = UnitSphericalRepresentation(lat=u.Quantity(cirs_dec, u.radian, copy=False),
                                              lon=u.Quantity(cirs_ra, u.radian, copy=False),
@@ -53,11 +50,8 @@ def icrs_to_cirs(icrs_coo, cirs_frame):
         astrom_eb = CartesianRepresentation(astrom['eb'], unit=u.au,
                                             xyz_axis=-1, copy=False)
         newcart = icrs_coo.cartesian - astrom_eb
-
         srepr = newcart.represent_as(SphericalRepresentation)
-        i_ra = srepr.lon.to_value(u.radian)
-        i_dec = srepr.lat.to_value(u.radian)
-        cirs_ra, cirs_dec = atciqz(i_ra, i_dec, astrom)
+        cirs_ra, cirs_dec = atciqz(newcart.without_differentials(), astrom)
 
         newrep = SphericalRepresentation(lat=u.Quantity(cirs_dec, u.radian, copy=False),
                                          lon=u.Quantity(cirs_ra, u.radian, copy=False),
@@ -68,14 +62,10 @@ def icrs_to_cirs(icrs_coo, cirs_frame):
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, ICRS)
 def cirs_to_icrs(cirs_coo, icrs_frame):
-    srepr = cirs_coo.represent_as(SphericalRepresentation)
-    cirs_ra = srepr.lon.to_value(u.radian)
-    cirs_dec = srepr.lat.to_value(u.radian)
-
     # set up the astrometry context for ICRS<->cirs and then convert to
     # astrometric coordinate direction
     astrom = erfa_astrom.get().apci(cirs_coo)
-    i_ra, i_dec = aticq(cirs_ra, cirs_dec, astrom)
+    i_ra, i_dec = aticq(cirs_coo.cartesian.without_differentials(), astrom)
 
     if cirs_coo.data.get_name() == 'unitspherical' or cirs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just use the coordinate direction to yield the
@@ -89,6 +79,7 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
 
         # the distance in intermedrep is *not* a real distance as it does not
         # include the offset back to the SSB
+        srepr = cirs_coo.represent_as(SphericalRepresentation)
         intermedrep = SphericalRepresentation(lat=u.Quantity(i_dec, u.radian, copy=False),
                                               lon=u.Quantity(i_ra, u.radian, copy=False),
                                               distance=srepr.distance,
@@ -124,10 +115,7 @@ def icrs_to_gcrs(icrs_coo, gcrs_frame):
 
     if icrs_coo.data.get_name() == 'unitspherical' or icrs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just do the infinite-distance/no parallax calculation
-        usrepr = icrs_coo.represent_as(UnitSphericalRepresentation)
-        i_ra = usrepr.lon.to_value(u.radian)
-        i_dec = usrepr.lat.to_value(u.radian)
-        gcrs_ra, gcrs_dec = atciqz(i_ra, i_dec, astrom)
+        gcrs_ra, gcrs_dec = atciqz(icrs_coo.cartesian.without_differentials(), astrom)
 
         newrep = UnitSphericalRepresentation(lat=u.Quantity(gcrs_dec, u.radian, copy=False),
                                              lon=u.Quantity(gcrs_ra, u.radian, copy=False),
@@ -142,9 +130,7 @@ def icrs_to_gcrs(icrs_coo, gcrs_frame):
         newcart = icrs_coo.cartesian - astrom_eb
 
         srepr = newcart.represent_as(SphericalRepresentation)
-        i_ra = srepr.lon.to_value(u.radian)
-        i_dec = srepr.lat.to_value(u.radian)
-        gcrs_ra, gcrs_dec = atciqz(i_ra, i_dec, astrom)
+        gcrs_ra, gcrs_dec = atciqz(newcart.without_differentials(), astrom)
 
         newrep = SphericalRepresentation(lat=u.Quantity(gcrs_dec, u.radian, copy=False),
                                          lon=u.Quantity(gcrs_ra, u.radian, copy=False),
@@ -156,15 +142,11 @@ def icrs_to_gcrs(icrs_coo, gcrs_frame):
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference,
                                  GCRS, ICRS)
 def gcrs_to_icrs(gcrs_coo, icrs_frame):
-    srepr = gcrs_coo.represent_as(SphericalRepresentation)
-    gcrs_ra = srepr.lon.to_value(u.radian)
-    gcrs_dec = srepr.lat.to_value(u.radian)
-
     # set up the astrometry context for ICRS<->GCRS and then convert to BCRS
     # coordinate direction
     astrom = erfa_astrom.get().apcs(gcrs_coo)
 
-    i_ra, i_dec = aticq(gcrs_ra, gcrs_dec, astrom)
+    i_ra, i_dec = aticq(gcrs_coo.cartesian.without_differentials(), astrom)
 
     if gcrs_coo.data.get_name() == 'unitspherical' or gcrs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just use the coordinate direction to yield the
@@ -178,6 +160,7 @@ def gcrs_to_icrs(gcrs_coo, icrs_frame):
 
         # the distance in intermedrep is *not* a real distance as it does not
         # include the offset back to the SSB
+        srepr = gcrs_coo.represent_as(SphericalRepresentation)
         intermedrep = SphericalRepresentation(lat=u.Quantity(i_dec, u.radian, copy=False),
                                               lon=u.Quantity(i_ra, u.radian, copy=False),
                                               distance=srepr.distance,
@@ -210,14 +193,10 @@ def gcrs_to_hcrs(gcrs_coo, hcrs_frame):
         frameattrs['obstime'] = hcrs_frame.obstime
         gcrs_coo = gcrs_coo.transform_to(GCRS(**frameattrs))
 
-    srepr = gcrs_coo.represent_as(SphericalRepresentation)
-    gcrs_ra = srepr.lon.to_value(u.radian)
-    gcrs_dec = srepr.lat.to_value(u.radian)
-
     # set up the astrometry context for ICRS<->GCRS and then convert to ICRS
     # coordinate direction
     astrom = erfa_astrom.get().apcs(gcrs_coo)
-    i_ra, i_dec = aticq(gcrs_ra, gcrs_dec, astrom)
+    i_ra, i_dec = aticq(gcrs_coo.cartesian.without_differentials(), astrom)
 
     # convert to Quantity objects
     i_ra = u.Quantity(i_ra, u.radian, copy=False)
@@ -233,6 +212,7 @@ def gcrs_to_hcrs(gcrs_coo, hcrs_frame):
 
         # Note that the distance in intermedrep is *not* a real distance as it
         # does not include the offset back to the Heliocentre
+        srepr = gcrs_coo.represent_as(SphericalRepresentation)
         intermedrep = SphericalRepresentation(lat=i_dec, lon=i_ra,
                                               distance=srepr.distance,
                                               copy=False)
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -37,7 +37,8 @@ def icrs_to_cirs(icrs_coo, cirs_frame):
 
     if icrs_coo.data.get_name() == 'unitspherical' or icrs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just do the infinite-distance/no parallax calculation
-        cirs_ra, cirs_dec = atciqz(icrs_coo.cartesian.without_differentials(), astrom)
+        srepr = icrs_coo.spherical
+        cirs_ra, cirs_dec = atciqz(srepr.without_differentials(), astrom)
 
         newrep = UnitSphericalRepresentation(lat=u.Quantity(cirs_dec, u.radian, copy=False),
                                              lon=u.Quantity(cirs_ra, u.radian, copy=False),
@@ -51,7 +52,7 @@ def icrs_to_cirs(icrs_coo, cirs_frame):
                                             xyz_axis=-1, copy=False)
         newcart = icrs_coo.cartesian - astrom_eb
         srepr = newcart.represent_as(SphericalRepresentation)
-        cirs_ra, cirs_dec = atciqz(newcart.without_differentials(), astrom)
+        cirs_ra, cirs_dec = atciqz(srepr.without_differentials(), astrom)
 
         newrep = SphericalRepresentation(lat=u.Quantity(cirs_dec, u.radian, copy=False),
                                          lon=u.Quantity(cirs_ra, u.radian, copy=False),
@@ -65,7 +66,8 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
     # set up the astrometry context for ICRS<->cirs and then convert to
     # astrometric coordinate direction
     astrom = erfa_astrom.get().apci(cirs_coo)
-    i_ra, i_dec = aticq(cirs_coo.cartesian.without_differentials(), astrom)
+    srepr = cirs_coo.represent_as(SphericalRepresentation)
+    i_ra, i_dec = aticq(srepr.without_differentials(), astrom)
 
     if cirs_coo.data.get_name() == 'unitspherical' or cirs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just use the coordinate direction to yield the
@@ -79,7 +81,6 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
 
         # the distance in intermedrep is *not* a real distance as it does not
         # include the offset back to the SSB
-        srepr = cirs_coo.represent_as(SphericalRepresentation)
         intermedrep = SphericalRepresentation(lat=u.Quantity(i_dec, u.radian, copy=False),
                                               lon=u.Quantity(i_ra, u.radian, copy=False),
                                               distance=srepr.distance,
@@ -115,7 +116,8 @@ def icrs_to_gcrs(icrs_coo, gcrs_frame):
 
     if icrs_coo.data.get_name() == 'unitspherical' or icrs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just do the infinite-distance/no parallax calculation
-        gcrs_ra, gcrs_dec = atciqz(icrs_coo.cartesian.without_differentials(), astrom)
+        srepr = icrs_coo.represent_as(SphericalRepresentation)
+        gcrs_ra, gcrs_dec = atciqz(srepr.without_differentials(), astrom)
 
         newrep = UnitSphericalRepresentation(lat=u.Quantity(gcrs_dec, u.radian, copy=False),
                                              lon=u.Quantity(gcrs_ra, u.radian, copy=False),
@@ -130,7 +132,7 @@ def icrs_to_gcrs(icrs_coo, gcrs_frame):
         newcart = icrs_coo.cartesian - astrom_eb
 
         srepr = newcart.represent_as(SphericalRepresentation)
-        gcrs_ra, gcrs_dec = atciqz(newcart.without_differentials(), astrom)
+        gcrs_ra, gcrs_dec = atciqz(srepr.without_differentials(), astrom)
 
         newrep = SphericalRepresentation(lat=u.Quantity(gcrs_dec, u.radian, copy=False),
                                          lon=u.Quantity(gcrs_ra, u.radian, copy=False),
@@ -146,7 +148,8 @@ def gcrs_to_icrs(gcrs_coo, icrs_frame):
     # coordinate direction
     astrom = erfa_astrom.get().apcs(gcrs_coo)
 
-    i_ra, i_dec = aticq(gcrs_coo.cartesian.without_differentials(), astrom)
+    srepr = gcrs_coo.represent_as(SphericalRepresentation)
+    i_ra, i_dec = aticq(srepr.without_differentials(), astrom)
 
     if gcrs_coo.data.get_name() == 'unitspherical' or gcrs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just use the coordinate direction to yield the
@@ -160,7 +163,6 @@ def gcrs_to_icrs(gcrs_coo, icrs_frame):
 
         # the distance in intermedrep is *not* a real distance as it does not
         # include the offset back to the SSB
-        srepr = gcrs_coo.represent_as(SphericalRepresentation)
         intermedrep = SphericalRepresentation(lat=u.Quantity(i_dec, u.radian, copy=False),
                                               lon=u.Quantity(i_ra, u.radian, copy=False),
                                               distance=srepr.distance,
@@ -196,7 +198,8 @@ def gcrs_to_hcrs(gcrs_coo, hcrs_frame):
     # set up the astrometry context for ICRS<->GCRS and then convert to ICRS
     # coordinate direction
     astrom = erfa_astrom.get().apcs(gcrs_coo)
-    i_ra, i_dec = aticq(gcrs_coo.cartesian.without_differentials(), astrom)
+    srepr = gcrs_coo.represent_as(SphericalRepresentation)
+    i_ra, i_dec = aticq(srepr.without_differentials(), astrom)
 
     # convert to Quantity objects
     i_ra = u.Quantity(i_ra, u.radian, copy=False)
@@ -212,7 +215,6 @@ def gcrs_to_hcrs(gcrs_coo, hcrs_frame):
 
         # Note that the distance in intermedrep is *not* a real distance as it
         # does not include the offset back to the Heliocentre
-        srepr = gcrs_coo.represent_as(SphericalRepresentation)
         intermedrep = SphericalRepresentation(lat=i_dec, lon=i_ra,
                                               distance=srepr.distance,
                                               copy=False)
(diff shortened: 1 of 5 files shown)


astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -51,11 +48,8 @@ def icrs_to_cirs(icrs_coo, cirs_frame):
...
         newcart = icrs_coo.cartesian - astrom_eb
-
         srepr = newcart.represent_as(SphericalRepresentation)
-        i_ra = srepr.lon.to_value(u.radian)
-        i_dec = srepr.lat.to_value(u.radian)
-        cirs_ra, cirs_dec = atciqz(i_ra, i_dec, astrom)
+        cirs_ra, cirs_dec = atciqz(newcart.without_differentials(), astrom)
 
         newrep = SphericalRepresentation(lat=u.Quantity(cirs_dec, u.radian, copy=False),
                                          lon=u.Quantity(cirs_ra, u.radian, copy=False),
```
mhvk, line 55: If, on the line below, we use `distance=newcart.norm()`, we can get rid of `srepr` altogether here.

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -91,6 +81,7 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
 
         # the distance in intermedrep is *not* a real distance as it does not
         # include the offset back to the SSB
+        srepr = cirs_coo.represent_as(SphericalRepresentation)
```
mhvk, line 84: Similarly, can avoid this by using `distance=cirs_coo.norm()` in the initializer below.

pllim:
I don't understand the RTD warnings... Perhaps not related?
```
WARNING: the readthedocs_ext.readthedocs extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
WARNING: doing serial read
```

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -236,25 +279,61 @@ def atciqz(rc, dc, astrom):
...
+    # ignore parallax effects if no distance, or far away
+    ignore_distance = object_observer_distance.unit == u.one
+
     # BCRS coordinate direction (unit vector).
-    pco = erfa.s2c(rc, dc)
+    if not ignore_distance and u.allclose(astrometric.xyz, 0*u.km):
+        # special casing for positions at origin of frame
+        # replicating the behaviour of erfa.s2c with 0, 0 input
+        # which for reasons unknown gives the vector (1, 0, 0)
+        xhat = astrometric.unit_vectors()['x']
```
mhvk, line 310: This seems a bit long. Couldn't one do `tmp = astrometric + CartesianRepresentation([1., 0., 0.], unit=astrometric.x.unit)`? Or just `pco = CartesianRepresentation([1., 0., 0.], unit=u.dimensionless_unscaled)`?

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -236,25 +279,61 @@ def atciqz(rc, dc, astrom):
...
+        Declination in radians
     """
+    # first of all get distance from object to observer
+    object_observer_distance = astrometric.norm()
+    # ignore parallax effects if no distance, or far away
+    ignore_distance = object_observer_distance.unit == u.one
+
     # BCRS coordinate direction (unit vector).
-    pco = erfa.s2c(rc, dc)
+    if not ignore_distance and u.allclose(astrometric.xyz, 0*u.km):
```
mhvk, line 306: This will not do the same as `erfa.s2c` if astrometric is a multidimensional with only some entries being near the origin.

It does suggest the representations should have a way to get unit vectors pointing in a coordinates position... Short of that, maybe
```
pco = astrometric.get_xyz(axis=-1)
at_origin = np.all(pco == 0, axis=-1, keepdims=True)
pco[at_origin] = [1., 0., 0.] << pco.unit
pco /= np.sqrt((pco**2).sum(-1, keepdims=True)
```

(need to think a bit more...)

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -236,25 +279,61 @@ def atciqz(rc, dc, astrom):
...
-    dc : float or `~numpy.ndarray`
-        declination, radians
+    astrometric : `~astropy.coordinates.CartesianRepresentation`
+        Astrometric ICRS position of object from observer
     astrom : eraASTROM array
         ERFA astrometry context, as produced by, e.g. ``eraApci13`` or ``eraApcs13``
 
     Returns
     --------
     ri : float or `~numpy.ndarray`
```
mhvk, line 295: Since we are changing the input to be a representation, should we also change the output? Maybe a `UnitSphericalRepresentation`?

StuartLittlefair: I think this would add a conversion step and make the results harder to deal with. Currently the ra, dec output is pretty easy to handle afterwards - i.e adding distance if necessary

astropy/coordinates/builtin_frames/utils.py
StuartLittlefair: I'm thinking it may be easier not to pass in a CartesianRepresentation at all, but a SphericalRepresentation and retain the call to erfa.s2c....

StuartLittlefair:
Thanks for the review @mhvk - it made me realise I was using the wrong type of representation to make this work more easily!

It took a while to get my head around this - maybe in reviewing you'll find the attached diagram helpful? It will certainly give you an insight into how bad my spatial visualisation is!

Jottings-103

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -179,22 +183,29 @@ def aticq(ri, di, astrom):
...
 
     Returns
     --------
     rc : float or `~numpy.ndarray`
+        Right Ascension in radians
     dc : float or `~numpy.ndarray`
+        Declination in radians
     """
+    # ignore parallax effects if no distance, or far away
+    ignore_distance = srepr.distance.unit == u.one
```
mhvk, line 205: I'd get the distance out here, since it is used below too:
```
srepr_distance = srepr.distance
```

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -211,7 +222,26 @@ def aticq(ri, di, astrom):
...
         before = norm(pnat-d)
-        after = erfa.ld(1.0, before, before, astrom['eh'], astrom['em'], 5e-8)
+        if ignore_distance:
+            # No distance to object, assume a long way away
+            q = before
+        else:
+            # Find BCRS direction of Sun to object.
+            # astrom['eh'] and astrom['em'] contain Sun to observer unit vector,
+            # and distance, respectively.
+            eh = astrom['em'] * CartesianRepresentation(astrom['eh'], unit=u.au, xyz_axis=-1, copy=False)
```
mhvk, line 232: I love representations and quantities, but I think here I would not use them, since everything is basically in the right format already. So, just
```
q = (astrom['em'][..., np.newaxis] * astrom['eh']
     + srepr.distance[..., np.newaxis].to_value(u.au) * before)
sundist = np.linalg.norm(q, axis=-1, keepdims=True)
q /= sundist
q[sundist < 1.e-10] = before
```
(and, as noted above, maybe get the distance out on top, perhaps even in AU?)

mhvk:
p.s. Bug fix for 4.0.2 does seem appropriate,

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -236,25 +279,61 @@ def atciqz(rc, dc, astrom):
...
+    # Find BCRS direction of Sun to object
+    if ignore_distance:
+        # No distance to object, assume a long way away
+        q = pco
+    else:
+        # astrom['eh'] and astrom['em'] contain Sun to observer unit vector,
+        # and distance, respectively.
+        eh = astrom['em'] * CartesianRepresentation(astrom['eh'], unit=u.au, xyz_axis=-1, copy=False)
+        # apply parallax to find unit vector from Sun to object
+        q = eh + astrometric
```
mkbrewer: Sorry to be so late here. I just noticed that you had already made an attempt at this. The problem here is that `eh` is the non-light delayed position of the Earth wrt the Sun. The Earth moves fairly quickly in its orbit, so it isn't valid to combine the the light delayed position of the target body wrt to the Earth and the non-light time delayed position of the Earth wrt to the Sun. For the proper way to do this please see my (late) comments at #10291.

mkbrewer: To do this properly from a topocentric perspective, you'd need the position of the observer wrt to the Sun time delayed by the time it takes the light to get from the target body to the observer. This can be large for a target body such as Jupiter near conjunction with the Sun, so it's much better to do this from a barycentric perspective. That takes the Earth's orbital velocity out of the equation.

mkbrewer: For other than the time delay, that is.

mkbrewer:
I'll take a look at the data that you posted. That was from the the `astropy_topo_test.py` script that I sent you rather than the Skyfield script that you posted along with it, correct? While the Moon was an excellent target for testing topocentric positions, it's not very good for testing the issue at hand. I'll try to set up some more appropriate tests. I'm thinking the Venus transit and the conjunction of Jupiter around the 27th of December 2019. Clearly, your schedule on this doesn't jibe with mine, but I'll try to get something together soon.

astropy/coordinates/builtin_frames/utils.py
StuartLittlefair: Thanks for the comments @mkbrewer - I must admit I'm getting quite confused by all the different possible time delays. I'm also confused by what you mean when you say a "topocentric perspective". All the vectors in ```erfa.ld``` are barycentric - it's just a question of at what time these positions correspond to.

My understanding is below. Assuming the light ray from the body reaches the observer (Earth) at time `t`:

- `p`: vector from observer to location of source *when it emitted ray* that reaches observer at `t`
- `e`: vector from Sun to observer *at the time of closest approach of the light ray to the Sun*
- `q`: vector from Sun to source *at the time when the source emitted ray* that reaches observer at `t`

In my code I'm ignoring the motion of the sun around the Barycentre and using the position at time `t`, but in the terms I use above, could you point out where I'm going wrong?

astropy/coordinates/builtin_frames/utils.py
StuartLittlefair: Thanks for the suggestion - and the solution! It really does take me three goes at least to get shapes right 🥇 

I stuck with ```np.where``` for the last line, since the direct mask struggled with trailing dimensions.

StuartLittlefair:
> I'll take a look at the data that you posted. That was from the the `astropy_topo_test.py` script that I sent you rather than the Skyfield script that you posted along with it, correct? While the Moon was an excellent target for testing topocentric positions, it's not very good for testing the issue at hand. I'll try to set up some more appropriate tests. I'm thinking the Venus transit and the conjunction of Jupiter around the 27th of December 2019. Clearly, your schedule on this doesn't jibe with mine, but I'll try to get something together soon.

Thank you - I really appreciate the guidance. The data I posted was from the `astropy_topo_test.py` script. I have implemented a 2012 Venus transit test in this PR, but if you have a precise and well known time and separation for the closest approach of Venus to the Sun's centre from a reliable source I could improve that test - the best I found was 554.4" from https://eclipse.gsfc.nasa.gov/OH/transit12.html

On schedule - even disregarding the UK/US timeshift I'm about to take 2 weeks holiday so please don't rush.

mkbrewer:
Of course, in the case of a star, the call to `atciqz()` would simply be: 
```
gcrs_ra, gcrs_dec = atciqz(i_ra, i_dec, i_ra, i_dec, astrom)
```

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -14,6 +14,7 @@
 from astropy.time import Time
 from astropy.utils import iers
 from astropy.utils.exceptions import AstropyWarning
+from ..representation import CartesianRepresentation
```
mhvk, line 17: Not needed anymore, I think.

astropy/coordinates/builtin_frames/utils.py
mkbrewer: Sorry, my comment concerning the Earth's orbital velocity isn't relevant. Both `eh` and `eb` are calculated without any time delay. I hadn't thought about this since May and was not fully back up to speed  when I said that. Regarding `e`, the time of closest approach is mainly for gravitational deflection by planets. Jupiter, for instance, can move 12 arcseconds wrt the solar system barycenter in the time that it takes for light to travel from it to Earth. The Sun moves much more slowly. In fact, it is stated in the Explanatory Supplement that just using `eh` without any time delay can only cause an error in the deflection of up to 0.1 mas which is less than the 0.5 mas accuracy of the formula itself. My main concern here is the lack of time delay in the position of the Sun used in `q`. Just scaling up that 0.1 mas from 8.3 minutes to the 4 hours that it takes light to travel from Neptune yields 2.9 mas.  What I meant by topocentric perspective is that in computing `q` you are using two vectors that involve the position of the Earth rather than the solar system barycenter. One of these, `eh` is geocentric, but the other `astrometric` (now `srepr`) can be topocentric which can displace the target body by up to 6378 km. For the Moon which has an orbital velocity of 1 km / s about the Earth, that translates into 106 minutes of time or about 53 arcminutes in position relative to the Earth. I don't know by how much these approximations affect the gravitational deflection, but you can find out yourself by implementing the algorithm that I suggested below. It is the algorithm that both the ES and I use, so comparing the results of that algorithm to the results from this PR should give you a good idea of the error.

mkbrewer:
I'm not sure what file of mine that you compared your output to. I was unable to reproduce your plot. I could come close by using the v4 version of my program though. The main difference between v4 and v5 is that v5 uses TDB whereas v4 uses TT. Using TT was an issue for Brandon Rhodes so I updated to TDB. Anyway here is what I get using v5 after making a slight adjustment to my EOP as your DUT1 seems to have changed. I also took out the polar motion adjustment to LST that I use from USNO Circular 179 as you don't use that and it makes for a slight difference in our LSTs which are not used in any of the calculations anyway. This new PR is a major improvement as there is only a small residual in RA left that may be due to your approximations. The residual is actually within the stated 0.5 mas accuracy of the algorithm. I still think that what you are doing is over complicated and that you should simplify it by using my suggested algorithm which would also be much more clear to the user.
AstropyNewPR_MKB_v5
Here is the new data file. 
[moon_topo_4_6_2017_mkb_ap_v5_new_pr.txt](https://github.com/astropy/astropy/files/5105142/moon_topo_4_6_2017_mkb_ap_v5_new_pr.txt)

mkbrewer:
Of course, the real tests would be the Venus transit and the Jupiter conjunction on December 27th 2019. If you can provide me with data for those two, I can get a much better idea of the error introduced by your approximations.

mkbrewer:
With the  errors in GCRS pretty much wrapped up, it's time to move on to CIRS. For that I'll post under issue #10356

mkbrewer:
When I set up a version of my program to duplicate the behavior of Astropy in transforming to Alt/Az for issue #10356, I was able to get agreement to 20 uas. This made me think some more about the apparent 0.3 mas offset in topocentric apparent RA. In order to get that, the script that I gave you calls `_apparent_position_in_true_coordinates()`. Looking at the code for that, I realized that you are currently using an older version of the BPN matrix provided by `pn00a()` to do the transform to true equator and equinox of date rather than `pnm06a()` that we both use normally. I figured that might be the problem, so I modified my program to use `pn00a()` also. Here is the result:
AstropyMaster_MKB_v5_SOFA_ld_pn00a
There is still a small offset with a non-diurnal slope that I still don't understand, but about 90% of the apparent offset was due to the difference in BPNs. You really should update `_apparent_position_in_true_coordinates()` to use `pnm06a()` in order to be compatible with the rest of your code. Do I need to open an issue on this? Or can you just take care of this? It'll only take a few minutes to fix.

btw,  those jumps every 2.5 hours are something on your side. I have no idea what, but nothing in my code has any periodicity like that.

mhvk:
@mkbrewer - thanks for the detailed sleuthing! I did a quick check on replacing `pn00a` with `pn06a` and there are no test failures, so at least we didn't bake in incorrect answers. Of course, if we make the correction, we should also include some comparisons with known-to-be-correct values (at a level where we would currently fail). Would you have a suggestion for a good test case?

mkbrewer:
Sorry, for some reason, I didn't receive notification of your comment. Unfortunately, I don't have a suggestion for a third party test of this. JPL Horizons isn't a good test as their apparent right ascension is off by about 53 mas and Skyfield doesn't currently handle polar motion correctly (I raised an issue there concerning this, but it has yet to be addressed). However, you can test for internal consistency by extracting the equation of the origins (`erfa.eors()`). Adding this to the apparent right ascension returned from `_apparent_position_in_true_coordinates()` should agree exactly with your CIRS right ascension. Of course, if you use `pn06a()` rather than `pnm06a()` as I strongly suggest (see `utils.get_cip()`), there will be a slight difference due to the different nutation model that is used. One caveat is that you must test geocentric coordinates only as your topocentric CIRS coordinates are currently incorrect (see issue #10356).

mkbrewer:
Note that testing this is purely a formality. If you use `pnm06a()` I guarantee that it will work.

mkbrewer:
I managed to get rid of the last tiny offset. This was on my side. My program is normally used to process data sampled at 200 Hz, so I only update the nutations once every 60 seconds. For these tests against astropy, that is the sampling interval, but the problem was that I use a 30 second look ahead so that the nutations are calculated at the midpoint of the interval. Eliminating the look ahead got rid of that offset.
AstropyNewPR_MKB_v6

mhvk:
Very nice to see such good agreement. But I must admit that I've been focussing too much on a couple of papers for the last week that I at least have lost track of whether this comparison is against the current PR, or against one that also makes changes like you recommended in https://github.com/astropy/astropy/pull/10666#issuecomment-677632970. Or was that recommendation just a suggested simplification?

mkbrewer:
The comparison is against the data that @StuartLittlefair posted along with this PR. However, while the Moon is an excellent source for checking most things, it's rather useless in the context of this PR. The light travel time from the Moon is so small that it makes no difference whatsoever. A much better test for this PR is the Jupiter conjunction on December 27th 2019. During that conjunction, Jupiter passed within 0.1 degrees of the center of the Sun. Here is my best projection of the error in the position of Jupiter using this PR and the outdated BPN that you currently use to calculate apparent position.
JupiterNoTimeDelayWrongBPN
Here is a script that you can run using this PR to check my prediction: [astropy_jupiter_topo_test.py.txt](https://github.com/astropy/astropy/files/5179854/astropy_jupiter_topo_test.py.txt)
If you run this script and post the output, I can generate a data file for you to use as a comparison. As you can see from the plot, most of the error is due to the incorrect BPN, but there is also a small signal that is due to the lack of time delay for the Sun. If you chose to correct the lack of time delay, then you can either use the algorithm that I posted in the comment that you quote (which I have edited to correct my error that I noted here: https://github.com/astropy/astropy/pull/10666#discussion_r474876167), or you can use the SOFA method as I described here: https://github.com/astropy/astropy/pull/10666#issuecomment-678528759. The back tracked position of the Sun (using `tlight_sun` as defined here: https://github.com/astropy/astropy/pull/10666#issuecomment-678535720, should be applied to the observer's heliocentric position (`astrom['eh']`) before calculation `q`, as you do in this PR as the light delay applies to both the source and the observer.

StuartLittlefair:
@mkbrewer - **many** thanks for your detailed investigations whilst I was away. Given that the stated accuracy of the algorithm used for light deflection is 0.5 mas, I'm tempted to not include the time delay calculations in this pull request. I'm open to being talked round though. Comments @mhvk  ?

mkbrewer:
Given the magnitude of the effect, I think it is perfectly reasonable to make that approximation. If you do, you should probably note that in the documentation in order to avoid future confusion, however.

mkbrewer:
I am curious as to what the holdup here is. If you are still considering doing this rigorously, now that `prepare_earth_position_vel()` has been encapsulated in `ErfaAstrom`, it is no longer possible to leverage that to get both the position and velocity of the Sun.  A separate call to `get_body_barycentric_posvel('sun', obstime)` would be required.

mkbrewer:
btw, I looked into how `Skyfield` handles this. It does do it rigorously. In fact, more rigorously than I suggest here. It also takes into account gravitational deflection by the major planets and the Earth itself. Gravitational deflection by the Earth can be up to 0.5 mas when the source is near the horizon. In that respect, it makes little sense to correct an error of less than that when only taking into account the deflection by the Sun.

mhvk:
@mkbrewer - `prepare_earth_position_vel` will still exist somewhere, so we can count on that...

On the hold-up, I suspect "start of term" - I'm lucky in having research leave but that is paid by a grant and astropy is mostly for spare time (it is Thanksgiving today in Canada). But at least my colleagues and thus likely most working at a university this year have been finding themselves in the unpleasant situation of having to switch to online teaching - which, if one has not done a course before, is a large amount of work to do even moderately well.

For this particular PR, my holdup was mostly not being sure whether there were remaining astrometric problems. Scanning through the comments above, it seems you found good agreement, and I guess the change to `pnm06a` should have remedied the main remaining problems. But I see two remaining questions:
1. Should the time delay be included in the calculation?
2. Should other objects be included (Jupiter at least should be easy with `ldn`, I think).

In principle, it seems those two are additional improvements that we could address in follow-up PRs, i.e., we could just get this improvement in and then build on it. Is this correct?

mkbrewer:
Of course. The major effect is that of the Sun and, even without including the time delay, this PR covers 99.99% of that. Happy Thanksgiving!

mhvk:
@StuartLittlefair - I took the liberty of fetching your PR and rebasing it on master (a bit tricky since you had merged master in - rebase seems preferable), so that I could add a changelog entry and get this in. I verified that the code was identical to what you had, so will approve and merge once tests pass (never be too sure!)

mhvk (approved):
@StuartLittlefair - I think this was already good to go, but I had missed that - sorry! Approving now, but ~unless~ unlike I wrote just above, I will leave the merging to you since you should at least have a chance to see whether you agree with the changelog entry! (You'll recognize the text from the top of the PR...)

mhvk:
I raised #10843 to remind us about the possible follow-up.

StuartLittlefair:
Thanks @mhvk! In turn, I hadn't noticed that this was missing a changelog... Thank you also for the nudge @mkbrewer

mkbrewer:
Here is a plot of the difference between the output of this PR that @StuartLittlefair posted and JPL Horizons. I have adjusted the difference in apparent right ascension for the ~52 mas difference in frame bias between SOFA and Horizons. Most of the residual is due to the outdated BPN that you were using plus a slight difference between your EOPs and those that JPL uses. However, you can clearly see the effect of the gravitational deflection by the Earth here. Most of this is in right ascension and attains a maximum at moonrise and moonset, then drops to zero when the Moon is below the horizon. 

AstropyNewPR_Horizons
 
I am not suggesting that you should include this gravitational deflection. The purpose of this plot is merely to show that the error in the deflection of Jupiter above due to neglecting the time delay is, in fact, less than this deflection.

mkbrewer:
Oh, I didn't notice that this had already been merged. Sorry for the after-post.

eteq:
I just encountered this while trying to finish up backports for 4.1.  Fantastic stuff implementation-wise here!  (Wish I'd seen it before to help the review, but glad it happened anyway without me!)

But I have one concern: is this really supposed to go in 4.0.x, or even 4.1.x ?

I'm asking because it's a pretty substantial implementation change, and *will* change results where it applies.  One can argue the previous code was wrong, but to be fair it's been wrong like that for a long time and code may have been written assuming the old behavior.  I agree we should updated it in the future, but I think this kind of change needs to be highlighted in a what's new or something, which means either we should consider it a "new feature" which then goes into 4.2, or backport to 4.1.x right before I crank through the release (in which case I'm happy to add it to the what's new).  

@StuartLittlefair @mhvk

eteq:
Oh, and @mkbrewer, re:

> Oh, I didn't notice that this had already been merged. Sorry for the after-post.

maybe you can create a follow-on issue with your above plot? That way we can at least not lose track of this in case someone wants to take on further refinement

StuartLittlefair:
I don’t have strong feelings about the milestone, but if I had to go either way I’d say that IMO this is a bug, and the fact that a bug has been around for a long time, and required a substantial change to fix are not reasons to hold back from issuing a bug fix...

mhvk:
Agreed with @StuartLittlefair - cases where people care about the answer to the level of precision discussed here are probably relatively rare, but for those we really don't want to be giving the wrong answers any longer than strictly necessary.

mhvk:
p.s. @mkbrewer - added a note to #10843 linking to your comment above about residual effects.

mkbrewer:
@eteq The only effect that we have addressed here is the gravitational deflection of light for solar system bodies. There is no change whatsoever for more distant objects. This is quite similar to the light time correction for the observer's location that was applied to solar system bodies in PR #10292. The idea here is to bring astropy into closer agreement with JPL Horizons and Skyfield for solar system bodies. The basic issue is that SOFA's utility functions `aticq()` and `atciqz()` don't include any provision for solar system bodies whatsoever. All objects are assumed to be at infinite distance for gravitational deflection effects. This is a bug in my view and we have addressed it here.

eteq:
Right to be clear I agree this is the correct behavior, and I'm glad to see the fix (it was definitely a long-standing inconsistency between sofa and the 3d coordinate approach in astropy)! 

My concern here is over LTS changing results (even if they're small) - in general LTS's are supposed to not change "science" results. Your arguments convince me it definitely should be in 4.1 (and I'll backport it accordingly), but I still think it would be good to be sure this gets into a what's new so that anyone who had encountered it before (which includes me wearing a "user" hat!) would have a chance to see the changes when they upgrade.

eteq:
I guess a compromise that would unblock me right now for 4.1 might be to put it into the 4.1 changelog and what's new section, but we can always fix that up in 4.1.1 or 4.2 or whatever's next once it's decided if it should go in 4.0.3?

mhvk:
Yes, let's get it in 4.1 for sure.

On 4.0, I think mis-interpreting a distance is not really that different from a constant being wrong, leap seconds being misapplied, or some calculation only going wrong in the scalar case. So, still seems a bug to me, and worth back-porting.

StuartLittlefair:
I agree with @mhvk I'm afraid

eteq:
Alright, as an immediate item I'll backport this to 4.1 and make a quick PR noting the change in the 4.1 what's new.

eteq:
(I'm also going to update the milestone while I'm doing that, but we can change it back to 4.0.x if that's the consensus!)

adrn:
I am fairly neutral on backporting to 4.0.x, but if I was forced to pick, I would agree with @mhvk and @StuartLittlefair!

eteq:
I'm definitievly outvoted, so I can accept these arguments!

That said, after backporting to 4.1.x I'm now *also* concerned there might be some serious backport issues.  But I'm happy saying our *plan* is to do 4.0.x and only back out of that if the backport is too complex to be practical.

mhvk:
If it does appear to be too complicated, do ping - I'm happy to re-apply the commits (the one rebase I did was quite unpleasant, but not difficult to keep the implementations apart).

eteq:
Thanks, @mhvk - actually I'm now seeing a bunch of failures on v4.1.x that I think may be due to #10647 conflicts? (see e.g. https://travis-ci.org/github/astropy/astropy/jobs/736086646) - the commit of relevance in that branch is  30d8abb64e787ebb3c63dd4c0b9d7929edaaaad9 .  If you see where the problem lies, a fix would be wonderful! (you can either PR straight into the v4.1.x branch or point me to a corrected backport in your own fork). I was just about to dive in but I'll give a bit of time for you to take a look then.

eteq:
(note this is now the sole blocker for v4.1... At least until someone finds a new bug I need to backport :wink:)

mhvk:
OK, I'll have a go...

mhvk:
@eteq - I have to remove the failed merge in 4.1.x; is it better to push that up and then make a new PR against it, or just to replace it with a single commit that does all the backporting and push directly?

mhvk:
@eteq - see #10885 for the backport. I'm happy to do it more bluntly in v4.1.x itself (i.e., rebase and get rid of your earlier merge instead of reverting it), but we may as well check whether the tests pass.

eteq:
#10885 was great, thanks @mhvk!

That said, it brought my original concern about this in an LTS into sharper focus - https://github.com/astropy/astropy/pull/10885/files is basically the diff of what would go into 4.0.x, and some of the subtle changes  (particularly I'm thinking of https://github.com/astropy/astropy/pull/10885/files#diff-cf7d1f9b0f2d0f47c0cfec77e1cffb8f417bbac1d46146c4e0dd78913077a0f1R226) that alter behavior in unexpected ways that might make downstream users' lives painful - e.g., I know I've worked around quirks of some of this code in the past, and I wouldn't want to find my workaround suddenly breaking in an LTS... And it might apply even for people who aren't anywhere close to the sun since `aticq`/`aticz` are used in several places (I think?)

So to me that's new information I didn't have any more, and renews my concern.  But @mhvk, @StuartLittlefair, and @adrn, if you still think I'm over-worrying feel free to just :-1: my comment here and we can stick with the backport-to-4.0.x plan since #10885 should make that easy for the v4.0.x releaser.

mhvk:
I'm still :+1: on backporting... (and think that just using the second commit from #10666 is easier, as it removes one layer of git cherry-pick just doing something arbitrary...)

StuartLittlefair:
I am still relatively ambivalent. This is a minor bug fix that will only affect a few calculations and only at a level that hardly any users care about, so I’m not desperate to backport. 

In principle I think we should but I guess the issue is what if there is some edge case we’ve missed where this code does something unexpected?

mkbrewer:
Again, I would remind you that this is not a small error. Here is the error for the transit of Mercury on November 11, 2019. The maximum error is 6", whereas the angular diameter of Mercury at the time is 10". I should think that people who study these transits would care about an error of this size.
MercuryTransit
Regarding unexpected behavior, it seems to me that the sooner this is distributed to the wider community, the sooner you will hear about something like that.

PR #10673, opened 2020-08-23
Meta PR: Extensible Cosmology

nstarman:
What this PR does already and how it should be split into smaller PRs for review. In no particular order:

- [x] #11345 : Separating all the cosmology realizations (and controller `default_cosmology`) to a separate file — `realizations`
~I've currently called this `builtin`, but `realization` might be better.~ 

2.  upgrade `default_cosmology` to be a registry (a la `galactocentric_frame_defaults`) so users can add their own cosmology realizations and set them as the default.

- [ ] change the parameters file to an ECSV (if pyyaml becomes a dependency, JSON otherwise). #11517

4.1 Add entrypoints to the realizations, so that third-party packages can add specific cosmology realizations and users can access them straight through astropy.

4.2 Also allow auto-import of user-defined realizations from a file(s).

  - [ ] JSON & ECSV I/O  #11559 
  - [ ] #11517 
  - [ ] configuration parameter for parameter files location.

4. Enhance the `Cosmology` class so it initializes the default cosmology, with any user-passed modifications. This centralizes the cosmology creation process for any "perturbations" about the default cosmology. Finally a use for the base class.

    - [x] : step 1 is #11515 
    - [ ] : step 2 is #11518 

5. [Not in this PR] Add entrypoints to `core`, where the `Cosmology` classes are defined, so that  third-party packages can add cosmologies and have them integrate well with the realizations machinery.

At the end, astropy will provide a core set of cosmology classes and realizations, but be highly extensible to allow user-defined realizations, and aggregate third-party cosmologies and realizations.

This PR might also might partially alleviate the lack-of-maintainer problem. Instead of programming more `Cosmology` classes, we let third-party packages demonstrate user need. If a Cosmology class becomes stable & popular enough then it could be added as a built-in.


pllim:
I am not sure if we have a `cosmology` maintainer who has time to review a `+1,727 −576` PR. 👀

nstarman:
I just looked on the Team website. I see there isn’t anyone leading the cosmology module.

nstarman:
@pllim. Can you add a milestone, even if it's 6.0, so I can try to achieve all passing checks.

pllim:
Done. @nstarman , for future reference, to ensure your work is not wasted, it is preferable to discuss your development plan for something that takes significant effort (such as refactoring a subpackage) with the core maintainers first before doing the work. Not all PRs are merged for a reason or another. Thanks for your effort!

nstarman:
Completely fair and I definitely should have done that. I had most of this code done for personal research and sitting around since last summer. I just added some of the cleanup from #10624 before submitting this PR.

nstarman:
What's the best way to reach out? There are a number of convenience functions, such as to calculate z matter-radiation equality, that I would like to add (as a separate PR from this one).

pllim:
There are a few channels listed at https://www.astropy.org/help.html . Maybe try the "developer email list"?

nstarman:
For this PR, whom can I contact or with whom can I work?

pllim:
😬

nstarman:
What about splitting this PR into a few more elemental steps? This should vastly reduce the maintainer overhead.

1. Separating all the cosmology realizations (and controller `default_cosmology`) to a separate file.
I've currently called this `builtin`, but `realization` might be better.

2.  upgrade realizations and `default_cosmology` to be a registry so users can add their own cosmology realizations and set them as the default.

3. Add the entrypoints to the realizations, so that third-party packages can add specific cosmologies and users can access them straight through astropy.

4. Enhance the `Cosmology` class so it initializes the default cosmology, with any user-passed modifications.

embray:
I think there should also be a way to provide new cosmology realizations without having to do anything with entrypoints.  I'm not really up on how this works currently though (I used to be but I haven't looked at it in a looong time).

nstarman:
> I think there should also be a way to provide new cosmology realizations without having to do anything with entrypoints. I'm not really up on how this works currently though (I used to be but I haven't looked at it in a looong time).

@embray 
I certainly agree. In PR 3 I have
> Also add a configuration parameter for a list of locations of JSON files of cosmology realizations to make along with the built-in options at import, allowing for users to set a directory of custom cosmo realizations.

On further thought, I'm kind of leaning towards an ECSV instead of JSON because of the built-in unit and metadata support and the cool possibility of converting an MCMC chain to an ECSV file and built able to work with sets of cosmology  realizations. Also, I think the astropy config folder would be a good place to auto-import user-provided realizations. Anything else would be done by the user.

PR #10674, opened 2020-08-24
Write keys to ASDF only if the value is present

eslavich:
### Description

This PR changes ASDF serialization support to write keys only if the corresponding value is present.  This anticipates a change in behavior of asdf library 2.8, which will stop automatically removing keys with `None` values.  See https://github.com/asdf-format/asdf/pull/863 for more information.

I ran the astropy tests with the asdf branch that implements the change and these were the two `ExtensionType` subclasses that broke.  The updated subclasses are compatible with prior releases of asdf.

### DIFF
--- a/astropy/io/misc/asdf/tags/transform/tabular.py
+++ b/astropy/io/misc/asdf/tags/transform/tabular.py
@@ -47,12 +47,12 @@ class TabularType(TransformType):
     @classmethod
     def to_tree_transform(cls, model, ctx):
         node = {}
-        node["fill_value"] = model.fill_value
+        if model.fill_value is not None:
+            node["fill_value"] = model.fill_value
         node["lookup_table"] = model.lookup_table
         node["points"] = [p for p in model.points]
         node["method"] = str(model.method)
         node["bounds_error"] = model.bounds_error
-        node["name"] = model.name
         return node
 
     @classmethod
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -50,6 +50,9 @@ astropy.io.fits
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
+- Write keys to ASDF only if the value is present, to account
+  for a change in behavior in asdf 2.8. [#10674]
+
 astropy.io.votable
 ^^^^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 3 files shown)


eslavich:
> Are these the only optional attributes not being serialized? Or the only ones that showed up because they happened to be tested?

That's a good question, to some extent you can detect these by looking at the `from_tree` methods and seeing what properties are assumed to be present, but in some cases the tag code calls out to other methods and it's not obvious what is a required property.  Most models only add required properties on top of the base transform properties, so those are probably safe (the tabular transform is a bit of an odd duck).

I'm not certain that I didn't miss something, but I did take a tour through the tag classes to look for other optional properties.  `TimeType` and `TransformType` are the only other two that jumped out at me and in those we're careful not to write `None` values.

astropy/io/misc/asdf/tags/transform/tabular.py
```diff
@@ -47,12 +47,12 @@ def from_tree_transform(cls, node, ctx):
     @classmethod
     def to_tree_transform(cls, model, ctx):
         node = {}
-        node["fill_value"] = model.fill_value
+        if model.fill_value is not None:
+            node["fill_value"] = model.fill_value
```
nden, line 51: I think the above two lines are not necessary because `fill_value=None` has and is an allowed value but is not the default value.
The `name` attribute is a bug.

eslavich: If we want to write an explicit `fill_value: null` to the ASDF file, then we'll need to update the schema:

https://github.com/asdf-format/asdf-standard/blob/master/schemas/stsci.edu/asdf/transform/tabular-1.2.0.yaml#L50

That will need to become something like:

```
fill_value:
  anyOf:
    - type: number
    - type: "null"
```

nden: Ah, right. So let's do it. 
And we'll keep the current PR as is. I'll only move the milestone to 4.0.2 as it can go in the LTS release.

nden: You'll have to move the change log to 4.0.2 but feel free to merge after that.

nden (approved):

PR #10681, opened 2020-08-27
DOC: Remove broken UCO/Lick link

pllim:
### Description

Cron job linkcheck reports that https://www.ucolick.org/~sla/leapsecs/timescales.html is broken. I confirmed that the site is unreachable. Can't tell if this is temporary or permanent. If the latter, we should remove it or find a replacement.

### DIFF
--- a/docs/time/index.rst
+++ b/docs/time/index.rst
@@ -364,7 +364,6 @@ local  Local Time Scale          (LOCAL)
 .. [#] Wikipedia `time standard <https://en.wikipedia.org/wiki/Time_standard>`_ article
 .. [#] SOFA_ Time Scale and Calendar Tools
        `(PDF) <http://www.iausofa.org/sofa_ts_c.pdf>`_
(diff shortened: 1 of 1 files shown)


docs/time/index.rst
```diff
@@ -364,7 +364,6 @@ local  Local Time Scale          (LOCAL)
 .. [#] Wikipedia `time standard <https://en.wikipedia.org/wiki/Time_standard>`_ article
 .. [#] SOFA_ Time Scale and Calendar Tools
        `(PDF) <http://www.iausofa.org/sofa_ts_c.pdf>`_
-.. [#] `<https://www.ucolick.org/~sla/leapsecs/timescales.html>`_
```
mhvk, line 367: You'd also need to remove the reference to this footnote on line 345 (which would become `both" [#]_, [#]_.`, I think.

pllim: Ops. Thanks!

PR #10685, opened 2020-08-27
TST: Handle non-deterministic utils test failure

pllim:
### Description

This pull request is to address this non-deterministic failure in CI:
```
        with pytest.warns(CacheMissingWarning) as warning_lines:
            fnout = download_file(TESTURL, cache=True)
>       assert len(warning_lines) == 2, os.linesep.join(
            [str(w.message) for w in warning_lines])
E       AssertionError: Remote data cache could not be accessed due to OSError
E         unclosed <socket.socket [closed] fd=13, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6>
E         unclosed <socket.socket [closed] fd=14, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6>
E         ('Cache directory cannot be read or created (), providing data in temporary file instead.', '/tmp/astropy-download-5853-saggci1x')
E       assert 4 == 2
E        +  where 4 = len(WarningsChecker(record=True))
../../.tox/py37-test-alldeps-numpy117-cov/lib/python3.7/site-packages/astropy/utils/tests/test_data.py:1058: AssertionError
```

Most of the time, it only has 2 warnings but sometimes it has extra 2 (unclosed socket). 🤷 

Follow up of #10437 and #10671

cc @aarchiba

### DIFF
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -1054,10 +1054,11 @@ def test_data_noastropy_fallback(monkeypatch):
 
     with pytest.warns(CacheMissingWarning) as warning_lines:
         fnout = download_file(TESTURL, cache=True)
-    assert len(warning_lines) == 2, os.linesep.join(
+    # Sometimes 2 extra "unclosed socket" pop up in CI
+    assert len(warning_lines) in (2, 4), os.linesep.join(
         [str(w.message) for w in warning_lines])
     assert 'Remote data cache could not be accessed' in str(warning_lines[0].message)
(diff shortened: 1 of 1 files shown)


aarchiba:
Unclosed socket sounds like an actual bug? Do we have a -Werror traceback?

pllim:
> Do we have a -Werror traceback?

No. I am hesitant to add it to a CI that would affect all PRs. Maybe to the cron job...?

pllim:
Superseded by #10686

PR #10686, opened 2020-08-28
Ignore cache warnings to catch non-deterministic socket warnings

pllim:
### Description

This pull request is to address investigate https://github.com/astropy/astropy/pull/10685#issuecomment-682463852 . I need to know more about the socket warnings before deciding on the best course of action. We use pytest, so we cannot simply use `-Werror` (if I am mistaken, please correct me).

Related: #10437 and #10671

### DIFF
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -1023,6 +1023,7 @@ def test_get_pkg_data_contents():
     assert contents1 == contents2
 
 
+# FIXME: Undo PR 10686 after socket warnings mystery is solved.
 @pytest.mark.remote_data(source="astropy")
 @pytest.mark.filterwarnings("ignore:Remote data cache could not be accessed")
 @pytest.mark.filterwarnings(r"ignore:.*Cache directory cannot be read or created .*")
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -1024,6 +1024,8 @@ def test_get_pkg_data_contents():
 
 
 @pytest.mark.remote_data(source="astropy")
+@pytest.mark.filterwarnings("ignore:Remote data cache could not be accessed")
+@pytest.mark.filterwarnings(r"ignore:.*Cache directory cannot be read or created .*")
 def test_data_noastropy_fallback(monkeypatch):
     """
     Tests to make sure the default behavior when the cache directory can't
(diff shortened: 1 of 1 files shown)


pllim:
@aarchiba , given the non-deterministic nature of the socket warnings, should I modify the test as such, merge this, and then see what traceback we get when socket warnings happen again?

aarchiba:
I did not write this monkeypatch-based test and it has been nothing but trouble; but I have been preserving it because it was presumably meant to test something important. And indeed the socket warning is important. But are you sure we can't just use -Werror with pytest? I could have sworn that worked. And in that case leaving the test as is should do the right thing - get a traceback when the wrong warnings are emitted. This PR actually disables the test's ability to check for cache warnings, which was the main point of it. (Though I think we have some non-networked tests that I wrote that do the same check.)

pllim:
Re: -Werror 

I am unable to find any pytest doc about it. Also, I am afraid it will have side effects for all the other 10k tests we have. There are already too many ways to handle test warnings as it is.

I can make a note to undo this change once we figure out the socket stuff. This is never meant to be permanent but I wasn't clear about it. Sorry.

astropy/utils/tests/test_data.py
```diff
@@ -1023,7 +1023,10 @@ def test_get_pkg_data_contents():
     assert contents1 == contents2
 
 
+# FIXME: Undo PR 10686 after socket warnings mystery is solved.
```
pllim, line 1026: Does this address your concern, @aarchiba ?

jdavies-st:
Using `-W error` is a general python interpreter flag.  Pytest uses it as is.  It's the equivalent of what the `astropy` `setup.cfg` is already doing:

https://github.com/astropy/astropy/blob/214a599c10c741360878758df4039a3bb1bb6237/setup.cfg#L110-L128

I.e. converting all warnings to errors, except for some explicit ignores.  So the `astropy` setup as it is should work exactly like `pytest -W error`.  They are equivalent.

pllim:
Come to think of this somemore... maybe this PR won't fix anything. The way I did this here, maybe the ignore list in `setup.cfg` will catch the 2 extra warnings instead and silence them. 🤔 

I haven't seen this in CI lately but I did see it in Windows 4.1rc2 testing. But the way RC is tested, `setup.cfg` is ignored. What a headache.

pllim:
I could consistently produce the socket warnings on Windows, but I just cannot track down where it is coming from. The weirdest thing is I see it with `pytest astropy/utils/tests/test_data.py --remote-data` but I don't see it with `pytest astropy/utils/tests/test_data.py::test_data_noastropy_fallback --remote-data`. I went through some code in `data.py` but cannot find obvious places where a socket won't be closed. If this is a bug, I don't think it is a bug in the code being tested.

Ignoring stuff like this does result in socket warnings being ignored due to `setup.cfg` settings. I think I'll just chalk this up as "we'll revisit if we ever have time".

tl;dr -- This PR isn't the correct way to go.

PR #10687, opened 2020-08-28
Fixes for weights with linear fitter

### LINKED ISSUE #7035: `LinearLSQFitter` does not handle 2D weights properly
When passing 2D weigths to `LinearLSQFitter` in 2D mode:
```
  [....]
  File "/Users/michal/sources/astropy/astropy/modeling/fitting.py", line 456, in __call__
    lhs *= weights[:, np.newaxis]
ValueError: operands could not be broadcast together with shapes (441,10) (21,1,21) (441,10) 
```
`LevMarLSQFitter` works properly.

saimn:
Fix #7035.
I still find the `modeling` code quite confusing between the data axes, model sets, parameters, so I hope that I did it right for all cases (and tests seem to show that it works).

### DIFF
--- a/astropy/modeling/fitting.py
+++ b/astropy/modeling/fitting.py
@@ -708,21 +708,39 @@ class LinearLSQFitter(metaclass=_FitterMeta):
 
         a = None  # need for calculating covarience
 
-        if weights is not None and weights.ndim > 1:
+        if ((masked and len(model_copy) > 1) or
+            (weights is not None and weights.ndim > 1)):
 
-            # separate weights for multiple models case: Numpy's lstsq
-            # supports multiple dimensions only for rhs, so we need to loop
-            # manually on the models.  This may be fixed in the future with
-            # https://github.com/numpy/numpy/pull/15777
+            # Separate masks or weights for multiple models case: Numpy's
+            # lstsq supports multiple dimensions only for rhs, so we need to
+            # loop manually on the models. This may be fixed in the future
+            # with https://github.com/numpy/numpy/pull/15777.
 
-            scl = scl.T
-            lacoef = np.zeros(lhs.shape[-2:], dtype=rhs.dtype)
+            # Initialize empty array of coefficients and populate it one model
+            # at a time. The shape matches the number of coefficients from the
+            # Vandermonde matrix and the number of models from the RHS:
+            lacoef = np.zeros(lhs.shape[1:2] + rhs.shape[-1:], dtype=rhs.dtype)
 
-            for model_lhs, model_rhs, model_lacoef in zip(lhs.T, rhs.T, lacoef.T):
+            # Arrange the lhs as a stack of 2D matrices that we can iterate
+            # over to get the correctly-orientated lhs for each model:
+            if lhs.ndim > 2:
+                lhs_stack = np.rollaxis(lhs, -1, 0)
+            else:
+                lhs_stack = np.broadcast_to(lhs, rhs.shape[-1:] + lhs.shape)
+
+            # Loop over the models and solve for each one. By this point, the
+            # model set axis is the second of two. Transpose rather than using,
+            # say, np.moveaxis(array, -1, 0), since it's slightly faster and
+            # lstsq can't handle >2D arrays anyway. This could perhaps be
+            # optimized by collecting together models with identical masks
+            # (eg. those with no rejected points) into one operation, though it
+            # will still be relatively slow when calling lstsq repeatedly.
+            for model_lhs, model_rhs, model_lacoef in \
+                zip(lhs_stack, rhs.T, lacoef.T):
 
                 # Cull masked points on both sides of the matrix equation:
                 good = ~model_rhs.mask if masked else slice(None)
-                model_lhs = model_lhs.T[good]
+                model_lhs = model_lhs[good]
                 model_rhs = model_rhs[good][..., np.newaxis]
 
                 # Solve for this model:
@@ -730,7 +748,7 @@ class LinearLSQFitter(metaclass=_FitterMeta):
                                                              model_rhs, rcond)
                 model_lacoef[:] = t_coef.T
 
-        elif len(model_copy) == 1 or not masked:
+        else:
 
             # If we're fitting one or more models over a common set of points,
             # we only have to solve a single matrix equation, which is an order
@@ -742,38 +760,11 @@ class LinearLSQFitter(metaclass=_FitterMeta):
             lacoef, resids, rank, sval = np.linalg.lstsq(lhs[good],
                                                          rhs[good], rcond)
 
-        else:
-
-            # Where fitting multiple models with masked pixels, initialize an
-            # empty array of coefficients and populate it one model at a time.
-            # The shape matches the number of coefficients from the Vandermonde
-            # matrix and the number of models from the RHS:
-            lacoef = np.zeros(lhs.shape[-1:] + rhs.shape[-1:], dtype=rhs.dtype)
-
-            # Loop over the models and solve for each one. By this point, the
-            # model set axis is the second of two. Transpose rather than using,
-            # say, np.moveaxis(array, -1, 0), since it's slightly faster and
-            # lstsq can't handle >2D arrays anyway. This could perhaps be
-            # optimized by collecting together models with identical masks
-            # (eg. those with no rejected points) into one operation, though it
-            # will still be relatively slow when calling lstsq repeatedly.
-            for model_rhs, model_lacoef in zip(rhs.T, lacoef.T):
-
-                # Cull masked points on both sides of the matrix equation:
-                good = ~model_rhs.mask
-                model_lhs = lhs[good]
-                model_rhs = model_rhs[good][..., np.newaxis]
-                a = model_lhs
-                # Solve for this model:
-                t_coef, resids, rank, sval = np.linalg.lstsq(model_lhs,
-                                                             model_rhs, rcond)
-                model_lacoef[:] = t_coef.T
-
         self.fit_info['residuals'] = resids
         self.fit_info['rank'] = rank
         self.fit_info['singular_values'] = sval
 
-        lacoef = (lacoef.T / scl).T
+        lacoef /= scl[:, np.newaxis] if scl.ndim < rhs.ndim else scl
         self.fit_info['params'] = lacoef
 
         _fitter_to_model_params(model_copy, lacoef.flatten())
--- a/astropy/modeling/fitting.py
+++ b/astropy/modeling/fitting.py
@@ -628,7 +628,7 @@ class LinearLSQFitter(metaclass=_FitterMeta):
                 if weights is not None:
                     # Same for weights
                     if weights.ndim > 2:
-                        weights = np.rollaxis(z, model_axis, z.ndim)
+                        weights = np.rollaxis(weights, model_axis, weights.ndim)
                         weights = weights.reshape(-1, weights.shape[-1])
                     else:
                         weights = weights.flatten()
@@ -674,7 +674,7 @@ class LinearLSQFitter(metaclass=_FitterMeta):
                 raise ValueError("x and weights should have the same length")
 
             if rhs.ndim == 2:
-                if weights is not None and weights.shape == rhs.shape:
+                if weights.shape == rhs.shape:
                     # separate weights for multiple models case: broadcast
                     # lhs to have more dimension (for each model)
                     lhs = lhs[..., np.newaxis] * weights[:, np.newaxis]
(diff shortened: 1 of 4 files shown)


jehturner:
Thanks, @saimn! I may be wrong, but I think this can be simplified significantly. Let me check...

jehturner:
No, that idea was just a misunderstanding of how you've done it and why -- but could we maybe combine the separate loops for `masked` & `weights`? I could work on that if you want (I also had one or two remaining uncertainties about the 2D weights behaviour that I could check at the same time).

I think it's good to have several tests where the weights aren't all 1, to make sure the results are correct and not just that the code runs. I see you do have one test that uses real weights for a single 2D model and also one for a 1D model set, which covers the main cases of looping & not looping. I wonder about adding some real weights for a 2D model set, to make sure the algebra is still correct in that case, but this is already pretty good and it's great that you're covering all these cases.

saimn:
> could we maybe combine the separate loops for masked & weights? I could work on that if you want

Sure, another pair of eyes would be useful ;).

> I think it's good to have several tests where the weights aren't all 1

Indeed. I needed tests to cover quickly all the different kind of inputs, so I basically copied the existing tests and added weights, but using something else than 1 requires to spend more time understanding each test case. We should certainly do that, if the algorithmic part is on the good path.

astropy/modeling/fitting.py
```diff
@@ -446,8 +461,18 @@ def __call__(self, model, x, y, z=None, weights=None, rcond=None):
...
                     # ~5-10x faster than using rollaxis.
                     rhs = z.T if model_axis == 0 else z
+
+                if weights is not None:
+                    # Same for weights
+                    if weights.ndim > 2:
+                        weights = np.rollaxis(z, model_axis, z.ndim)
+                        weights = weights.reshape(-1, weights.shape[-1])
+                    else:
+                        weights = weights.flatten()
```
saimn, line 471: This I'm not really sure about, because it doesn't match the else clause for the data just above, but I didn't really understand this case.

jehturner: Yes, that's one of the main things I didn't understand. It doesn't look right but I'll try to figure it out.

jehturner (commented):
I haven't actually finished reviewing this but I'll do "finish your review" because I think you won't see my couple of line comments otherwise? How do you want any changes I make? I would think a PR on your PR (like Curtis's) is best, since they'll probably be too complicated for github suggestions. At least this should be easier now we can check further changes against your tests...

saimn:
@jehturner - Yes, a PR on my PR would work :+1:

pllim:
> PR on my PR

Random thought: One of these days, we need to see how many PR to PR we can chain for a single PR. 😄

jehturner:
Maybe we could try a circular one?

astropy/modeling/fitting.py
```diff
@@ -446,8 +461,18 @@ def __call__(self, model, x, y, z=None, weights=None, rcond=None):
                     # but z has a second axis for the model set. NB. This is
                     # ~5-10x faster than using rollaxis.
                     rhs = z.T if model_axis == 0 else z
+
+                if weights is not None:
+                    # Same for weights
+                    if weights.ndim > 2:
+                        weights = np.rollaxis(z, model_axis, z.ndim)
```
jehturner, line 468: Uh oh, this says `z` instead of `weights`!

astropy/modeling/fitting.py
jehturner: I believe I've figured out what to do about this, but haven't done a lot more after a somewhat interrupted day (mostly by screaming kids, which is a bit of a nightmare when trying to keep all these different cases in one's head...). The current `else` here is handling multiple cases that need separating and there are one or two other changes needed to fix it, including an additional test or two for a case that currently fails.

astropy/modeling/fitting.py
saimn: Oops, good catch, fixed.

astropy/modeling/fitting.py
```diff
@@ -482,14 +507,20 @@ def __call__(self, model, x, y, z=None, weights=None, rcond=None):
...
-            if len(x) != len(weights):
+            if len(lhs) != len(weights):
                 raise ValueError("x and weights should have the same length")
+
             if rhs.ndim == 2:
-                lhs *= weights[:, np.newaxis]
-                # Don't modify in-place in case rhs was the original dependent
-                # variable array
-                rhs = rhs * weights[:, np.newaxis]
+                if weights is not None and weights.shape == rhs.shape:
```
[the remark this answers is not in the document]
saimn: Fixed.

astropy/modeling/tests/test_fitters.py
```diff
@@ -291,6 +326,35 @@ def test_linear_fit_model_set_fixed_parameter(self):
         assert_allclose(fitted_model.c1, [1., -2.], atol=1e-14)
         assert_allclose(fitted_model.c2, [0.5, 0.], atol=1e-14)
 
+    def test_linear_fit_2d_model_set_common_weight(self):
+        init_model = models.Polynomial2D(degree=2, c1_0=[1, 2], c0_1=[-0.5, 1],
+                                         n_models=2,
+                                         fixed={'c1_0': True, 'c0_1': True})
```
jehturner, line 337: Just wondering whether there's any special reason for fixing the parameters here and in the other test?

nden:
Is this ready for review?
@jehturner You are most familiar with this code, I'll wait for your approval before I review it.

astropy/modeling/fitting.py
```diff
@@ -499,7 +543,29 @@ def __call__(self, model, x, y, z=None, weights=None, rcond=None):
 
         masked = np.any(np.ma.getmask(rhs))
 
-        if len(model_copy) == 1 or not masked:
+        if weights is not None and weights.ndim > 1 and weights.shape == rhs.shape:
```
jehturner, line 546: I think you don't need the 3rd condition here (`shape`), because `weights` can only have a model set axis at this point if the `rhs` also has one (and their shapes should have been checked earlier).

jehturner:
I think the code is correct now, @nden, except that I was still checking what happens with `scl`. It would be good if the separate loops for `weights.ndim > 1` and `len(model_copy) > 1 and masked` could be consolidated in a reasonably efficient way though; I was hoping to look at that last week but didn't get to it. I think @saimn and I are both out next week and am not sure if one of us will get to it before then (or convince ourselves it would be too messy).

astropy/modeling/fitting.py
```diff
@@ -1299,16 +1369,23 @@ def _convert_input(x, y, z=None, n_models=1, model_set_axis=0):
...
+        else:
+            z_shape = z.shape
 
     if z is None:
+        if y_shape != x.shape:
+            raise ValueError("x and y should have the same shape")
         farg = (x, y)
     else:
+        if not (x.shape == y.shape == z_shape):
+            raise ValueError("x, y and z should have the same shape")
```
jehturner, line 1388: This check that I added is currently blocking the case of a length-1 model set axis, which otherwise works, so we should perhaps change that and add a test for it. This happens because the criterion for dealing with the model set axis in `_convert_inputs` is `n_models > 1` rather than the actual array dimensionality.

jehturner: In the existing code, however, providing y/z/weights with a length-1 model set axis **only works for 2D models**, not 1D ones, so this usage might not actually have been intended to work, @nden? But people could still be using it and it would seem more convenient if it does work (in both cases).

astropy/modeling/fitting.py
saimn: Ok, I removed the condition.

astropy/modeling/fitting.py
saimn: How is it useful to have length-1 weights ?

jehturner: Not `weights` specifically, also `y` or `z`. People might already be using this, eg. because they have a script that fits some range of image rows that can sometimes be a single row (and therefore ends up with a dimension of 1 unless they explicitly check for that case). I think we may as well make life easier by tolerating that case, rather than breaking people's stuff.

jehturner: I have a fix for this.

jehturner: I think that takes care of my concerns, except that it would be good to combine the loops I mentioned (but I doubt I'd be able to look at that now before the week after next).

saimn:
@nden @perrygreenfield - This is now ready for review, thanks to @jehturner for addressing the remaining items :)

astropy/modeling/tests/test_fitters.py
```diff
@@ -873,6 +1002,134 @@ def test_fitters_with_weights():
...
+    def test_linear_2d_separate_weights_axis_2(self):
+        model = models.Polynomial2D(1, model_set_axis=2)
+        fitter = LinearLSQFitter()
+        model = fitter(model, self.x2, self.y2, np.rollaxis(self.z2, 0, 3),
+                       weights=self.w2[..., np.newaxis])
+        assert_allclose(model.c0_0, 1., atol=1e-12)
+        assert_allclose(model.c1_0, -0.1, atol=1e-12)
+        assert_allclose(model.c0_1, 0.2, atol=1e-12)
+
+    # # The following 2 cases also happen to work, but model set fitting is
```
pllim, line 1106: Why not use `pytest.mark.skip`?

pllim (commented):
Any plans to address codecov comment about certain LOCs not covered? If it is not covered on purpose, can add `# pragma: no cover` directive too.

Need another rebase for the change log conflict.

p.s. When this is merge, the _weight_ is off your shoulder! 😉

astropy/modeling/tests/test_fitters.py
saimn: Not sure if we need to keep those tests, @jehturner ?

jehturner: I don't think we particularly need to keep them, but if they're there they could just be uncommented when those fitters support model sets. They do also work now, but I'm not sure it's reasonable to require those cases to continue working as long as the fitters aren't advertised to support model sets (so there's no reason for there to be an extra dimension).

saimn:
We (with @jehturner) are working on integrating this in our fitting code, it works well, and I just found a little issue when weights for one of the models are all 0 and the data is not a masked array. In that case the lapack routine in Numpy will crash with a rather obscure exception (`ValueError: On entry to DLASCL parameter number 4 had an illegal value`) so I added a check for this to raise a better exception. If the data is a masked array there will be a `RuntimeWarning: invalid value encountered in true_divide` but the code does not crash and just return a model with NaN parameters.

ping @nden @perrygreenfield - please review when you can :)

nden:
This is a substantial improvement! 
I have only one request - move the model set tests to `test_model_sets.py`.

Because of the late review the change log needs to be moved to `4.0.4`. 
And please squash commits. Thanks!

saimn:
@nden - I moved most of the model set tests to `test_model_sets.py`, but there were already some model set tests in `test_fitting.py`, and for some of the new model set tests we use the fixtures/class from `test_fitting.py`.
I also squashed some commits (e.g. changelog related), but I'm not sure what you want me to do here ? Do you want a single commit for this whole blob of changes ? IMO some of them deserve their own commit to be understandable (if one has to browse the history in the future) and some have comments that could be helpful as well...

saimn:
Ping @nden, would be super useful for us to have this in the next bugfix releases (i.e. 4.1.1).

astropy/modeling/tests/test_model_sets.py
```diff
@@ -311,3 +315,262 @@ def test_fitting_shapes():
...
+    def test_linear_2d_separate_weights_axis_2(self):
+        model = Polynomial2D(1, model_set_axis=2)
+        fitter = LinearLSQFitter()
+        model = fitter(model, self.x2, self.y2, np.rollaxis(self.z2, 0, 3),
+                       weights=self.w2[..., np.newaxis])
+        assert_allclose(model.c0_0, 1., atol=1e-12)
+        assert_allclose(model.c1_0, -0.1, atol=1e-12)
+        assert_allclose(model.c0_1, 0.2, atol=1e-12)
+
+    # # The following 2 cases also happen to work, but model set fitting is
```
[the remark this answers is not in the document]
saimn: I talked with @jehturner and I think it's better to remove them, since this is not a supported feature. If we can keep them in the history, maybe at some point we can find them again if it's useful.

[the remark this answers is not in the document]
saimn: I pushed a commit to remove these tests, and skipped CI since this is removing only commented lines.

pllim:
RTD warnings are real but unrelated. Merging then. Thanks, everyone!

PR #10690, opened 2020-08-30
Circular standard deviation

### LINKED ISSUE #10677: Circular standard deviation
### Description

The [circstats](https://docs.astropy.org/en/stable/stats/circ.html) module includes a function for the circular variance, but not for the standard deviation. It seems trivial to implement, given its definition as `sqrt(-2 * ln(R))` ([sd.circular: Circular Standard Deviation](https://rdrr.io/rforge/circular/man/sd.circular.html))

### Additional context

The implementation could be a simple as:

```
def circstd(data, axis=None, weights=None):
    """ Computes the circular standard deviation of an array of circular data.

    The standard deviation implemented here is based on the definition given by
    [1]_, which is also the same used by the R package 'Circular' [2]_.

    Parameters
    ----------
    data : numpy.ndarray or dimensionless Quantity
        Array of circular (directional) data, which is assumed to be in
        radians whenever ``data`` is ``numpy.ndarray``.
    axis : int, optional
        Axis along which circular variances are computed. The default is to
        compute the variance of the flattened array.
    weights : numpy.ndarray, optional
        In case of grouped data, the i-th element of ``weights`` represents a
        weighting factor for each group such that ``sum(weights, axis)``
        equals the number of observations. See [1]_, remark 1.4, page 22,
        for detailed explanation.

    Returns
    -------
    circstd : numpy.ndarray or dimensionless Quantity
        Circular standard deviation.

    Examples
    --------
    >>> import numpy as np
    >>> from astropy.stats import circvar
    >>> data = np.array([51, 67, 40, 109, 31, 358])*u.deg
    >>> circstd(data) # doctest: +FLOAT_CMP
    <Quantity 0.5976699855802333>

    References
    ----------
    .. [1] K.V. Mardia. "Statistics of Directional Data".
       Academic Press, London, sec. 26.5, p. 617, 1972.
    .. [2] U. Lund, C. Agostinelli, et al. "Circular Statistics from 'Topics in
       Circular Statistics (2001)'". 2017.
       <https://cran.r-project.org/web/packages/circular/circular.pdf>

    """

    return np.sqrt(-2.0 * np.log(_length(data, 1, 0.0, axis, weights)))
```

Gabriel-p:
### Description

Add two methods for estimating the circular standard deviation in a `circstd()` function.

Fixes #10677

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,6 +61,8 @@ astropy.samp
 astropy.stats
 ^^^^^^^^^^^^^
 
+- Added ``circstd`` function to obtain a circular standard deviation.
+
 astropy.table
 ^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,7 +61,7 @@ astropy.samp
 astropy.stats
 ^^^^^^^^^^^^^
 
-- Added ``circstd`` function to obtain a circular standard deviation.
+- Added ``circstd`` function to obtain a circular standard deviation. [#10690]
 
 astropy.table
 ^^^^^^^^^^^^^
--- a/astropy/stats/circstats.py
+++ b/astropy/stats/circstats.py
@@ -13,8 +13,8 @@ are based on reference [1]_, which is also the basis for the R package
 import numpy as np
 from astropy.units import Quantity
 
-__all__ = ['circmean', 'circvar', 'circmoment', 'circcorrcoef', 'rayleightest',
-           'vtest', 'vonmisesmle']
+__all__ = ['circmean', 'circstd', 'circvar', 'circmoment', 'circcorrcoef',
+           'rayleightest', 'vtest', 'vonmisesmle']
 __doctest_requires__ = {'vtest': ['scipy']}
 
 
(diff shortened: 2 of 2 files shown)


Gabriel-p:
Could anybody give me some pointers on how to address the two fails?

CHANGES.rst
```diff
@@ -61,6 +61,8 @@ astropy.samp
 astropy.stats
 ^^^^^^^^^^^^^
 
+- Added ``circstd`` function to obtain a circular standard deviation.
```
pllim, line 64: This might fix the change log failure.
```suggestion
- Added ``circstd`` function to obtain a circular standard deviation. [#10690]
```

Also need to add tests.

Gabriel-p:
There's now a weird issue with RTD

pllim:
Unfortunately, there is no "rebuild" button for RTD, so if you plan to push more commit, it will rebuild next time you do that. Not sure what is going on with Windows job, but we see that in other PRs too (let's wait and see). Remote data failure is unrelated (actually XPASSED).

Gabriel-p:
Should I push some small commit to force a rebuild (I could for example remove the dot at the end of `Added ``circstd`` function to obtain a circular standard deviation. [#10690]`)? Or just wait?

pllim:
@Gabriel-p , do you have plans to write new tests for this new function?

Gabriel-p:
@pllim should I? There are two tests already

pllim:
> should I?

Let's see what @larrybradley says. 😄

larrybradley:
`travis` tests in the "Final" stage were failing, so I can't see the coverage report.  I've restarted the travis tests.

larrybradley (approved):
Thanks, @Gabriel-p.  One of the restarted tests passed and travis is now green.  The other test with dev versions is an "allowed failure".  There's only 1 line not covered by tests and it's relatively minor (the check on the method type).  You could write a test for test, but we can also do it later.

@pllim What do we do about the failing RTD build?

pllim:
RTD should be fixed by now. I am going to close/open to restart it.

PR #10692, opened 2020-08-31
Notify slack on cron success

astrofrog:
Currently we only see when the cron doesn't work on slack. I think it would be good to also know when it does work as some green would be nice from time to time! 😆

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
 
 notifications:
   slack:
-    on_success: never
+    on_success: always
     on_failure: always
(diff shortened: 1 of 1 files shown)


pllim:
Not sure if @saimn wants that... 😆

astrofrog:
I think it’s be good to have otherwise we might miss that it is in fact sometimes green - there was a green cron job a couple of days ago but you wouldn’t know it from slack...

astrofrog:
Anyway green will be rare so won’t change much 😂

saimn:
Indeed for astropy success notifications should not pollute too much :joy:

saimn:
For the other repos (e.g pytest plugins), do we just need to copy the notifications block, with the same slack token ?

astrofrog:
@saimn good question I need to check if the token is repo specific

astrofrog:
But yeah otherwise the block can be the same

PR #10694, opened 2020-08-31
Fix reported module name of math_functions model classes

eslavich:
### Description

The current value of `__module__` for models in the math_functions module is "abc", which is an artifact of the dynamic class creation.  This PR sets `__module__` to "astropy.modeling.math_functions" instead.


jdavies-st:
Travis CI test failures look unrelated?

nden (approved):

PR #10697, opened 2020-09-01
Do not create models for ufuncs that are only aliases

eslavich:
### Description

`numpy.mod` and `numpy.divide` are alternate names for `numpy.remainder` and `numpy.true_divide`, respectively, but their corresponding model classes are distinct (`assert numpy.mod is numpy.remainder` succeeds, but `assert math_functions.ModUfunc is math_functions.RemainderUfunc` fails).  I think these should be the same class so that we can build a model using either name and end up with the same result.  If nothing else, the change will make model serialization to ASDF more consistent.

### DIFF
--- a/astropy/modeling/math_functions.py
+++ b/astropy/modeling/math_functions.py
@@ -14,14 +14,23 @@ trig_ufuncs = ["sin", "cos", "tan", "arcsin", "arccos", "arctan", "arctan2",
                "arctanh", "deg2rad", "rad2deg"]
 
 
-math_ops = ["add", "subtract", "multiply", "divide", "logaddexp", "logaddexp2",
+math_ops = ["add", "subtract", "multiply", "logaddexp", "logaddexp2",
             "true_divide", "floor_divide", "negative", "positive", "power",
-            "remainder", "mod", "fmod", "divmod", "absolute", "fabs", "rint",
+            "remainder", "fmod", "divmod", "absolute", "fabs", "rint",
             "exp", "exp2", "log", "log2", "log10", "expm1", "log1p", "sqrt",
-            "square", "cbrt", "reciprocal"]
+            "square", "cbrt", "reciprocal", "divide", "mod"]
 
 
 supported_ufuncs = trig_ufuncs + math_ops
+
+
+# These names are just aliases for other ufunc objects
+# in the numpy API.  The alias name must occur later
+# in the lists above.
+alias_ufuncs = {
+    "divide": "true_divide",
+    "mod": "remainder",
+}
 
 
 class _NPUfuncModel(Model):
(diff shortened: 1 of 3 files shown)


astropy/modeling/math_functions.py
```diff
@@ -14,16 +14,25 @@
...
 
 
 supported_ufuncs = trig_ufuncs + math_ops
 
 
+# These names are just aliases for other ufunc objects
+# in the numpy API.  The alias name must occur later
+# in the lists above.
+alias_ufuncs = {
+    "divide": "true_divide",
```
perrygreenfield, line 31: This is a clear case where divide is better than true_divide. True_divide is clearly a Python construct. I wouldn't even consider exposing true_divide at all as a name

astropy/modeling/math_functions.py
```diff
@@ -14,16 +14,25 @@
...
+    "mod": "remainder",
```
perrygreenfield, line 32: Modulo or remainder, hmmm. Is there a clear preference out there. Perhaps we should do a mini language survey. Well, to some these are not the same! https://rob.conery.io/2018/08/21/mod-and-remainder-are-not-the-same/

jdavies-st:
Looks good.  Again, the Travis CI failures seem to be unrelated.

astropy/modeling/math_functions.py
eslavich: I agree, but since the current schema is clear that these should be numpy ufunc names, this may be the "best" choice for now.  `"true_divide"` is how numpy describes the ufunc:

```
In [2]: np.divide
Out[2]: <ufunc 'true_divide'>
```

And that is how we're serializing the `DivideUfunc` model:

```
In [3]: with asdf.AsdfFile({"model": DivideUfunc()}) as af:
   ...:     af.write_to("test.asdf")
   ...:

In [4]: cat test.asdf
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 2.8.0.dev67+g2e2628f}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension.BuiltinExtension
    software: !core/software-1.0.0 {name: asdf, version: 2.8.0.dev67+g2e2628f}
  - !core/extension_metadata-1.0.0
    extension_class: astropy.io.misc.asdf.extension.AstropyAsdfExtension
    software: !core/software-1.0.0 {name: astropy, version: 4.2.dev560+g679a74a46}
model: !transform/math_functions-1.0.0 {func_name: true_divide}
...
```

perrygreenfield: Yeah, but this will have to be designated a python specific extension, and one I hope we eventually deep six and replace by one that uses numpy ufuncs for a Python library, but without anything in schemas that uses such nomenclature. And if this is kept as a python specific extension, I guess the mod/remainder thing can be made python specific, but it isn't universal.

jdavies-st:
This needs a changelog and milestone.

eslavich:
> This needs a changelog and milestone.

Changelog is already there, I added a milestone to match.

nden (approved):

PR #10700, opened 2020-09-01
TST: XPASS to not fail IERS test

pllim:
Sometimes the remote data job fails with only this message:

```
________________________________ test_iers_a_dl ________________________________
[XPASS(strict)] condition: TRAVIS
```

It is annoying because it should have given a green status.

Not sure on the milestone... If you think it needs backporting, please re-milestone.

### DIFF
--- a/astropy/utils/iers/tests/test_iers.py
+++ b/astropy/utils/iers/tests/test_iers.py
@@ -14,7 +14,6 @@ from astropy import units as u
 from astropy.table import QTable
 from astropy.time import Time, TimeDelta
 
-
 TRAVIS = os.environ.get('TRAVIS', False)
 
 FILE_NOT_FOUND_ERROR = getattr(__builtins__, 'FileNotFoundError', OSError)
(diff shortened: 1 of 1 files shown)


astropy/utils/iers/tests/test_iers.py
```diff
@@ -346,14 +345,18 @@ def test_IERS_B_parameters_loading_into_IERS_Auto():
 
 
 # Issue with FTP, rework test into previous one when it's fixed
-@pytest.mark.xfail('TRAVIS')
```
mhvk, line 349: How about just `skipif(TRAVIS, 'Flaky on travis CI')`? Seems to be no point even running the test if we cannot rely on it!

pllim: I think we debated about this before and that's why we have it fail on XPASS, but the failing on XPASS makes me sad.

Yeah, good point about not running it on Travis CI at all... 🤔

pllim: I made the change as you requested, @mhvk .

pllim:
RTD has been flaky today. Its failure is not related. Can probably ignore codecov too.

Issue #10699, opened 2020-09-01
TST: Windows job fails with unistd.h error


pllim:
Last successful run on `master` (cron): https://travis-ci.org/github/astropy/astropy/builds/722439530
First seen on `master` (cron): https://travis-ci.org/github/astropy/astropy/builds/722669911

They both attempt to run with astropy at 679a74a46379482 because nothing new was merged between the two scheduled runs. Somehow, one was successful but one failed with `fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory`.

A quick glance did not reveal any obvious upstream changes (like conda or MSVSC updates).

This is blocking unrelated PRs and the rest of the cron jobs.

pllim:
Notes for future selves: It was because of `setuptools` 50.x release. Also see pypa/setuptools#2353

PR #10698, opened 2020-09-01
MNT: Fix Windows and doc build issues due to setuptools 50

eslavich:
This PR fixes the Windows build by dropping quotes around MSVC compiler options, which setuptools 50 seems to be mangling somehow,.  It also disables system packages on readthedocs, which should resolve a different setuptools 50 issue in the doc build.  The docs are unfortunately still failing but I believe that is due to yet another problem introduced by readthedocs' own 5.4.1 release that came out yesterday in the middle of our troubleshooting.

Fix #10699

PR #10702, opened 2020-09-01
DOC: RTD no use sys pkg

pllim:
Ed suggests we don't use system packages.

Also see  #10698 #10699 #10701


pllim:
Yup. Will be part of #10698

PR #10705, opened 2020-09-02
MNT: Change config home path logic

### LINKED ISSUE #8731: config.paths._find_home() doesn't check HOMESHARE correctly
Importing Astropy on my system takes anywhere between 3-5 minutes due to checking the wrong network location for its Home directory.

The OS is Windows 10 1809 (the vast majority of users here are not comfortable with Linux environments), so `os.name = 'nt'`

The code in` config.paths._find_dir()` relies on the contents of `os.environs`. It searches for `'MSYSTEM'` (does not exist), then `'HOMESHARE'`, which does. It then attempts to find `%HOMESHARE%\.astropy`, which does not exist and never will. `%HOMESHARE% = \\server_root\`, and I obviously do not have write permissions to create `\\server-root\.astropy`. Python hangs for several minutes waiting to timeout as it attempts to access `\\server-root\.astropy\`

It should instead attempt to find `%HOMESHARE%\%HOMEPATH%\.astropy = \\server-root\user-name\.astropy`, which does exist - or at least, will have write permission available. This would be equivalent to how the next IF statement uses the environment variable `HOMEPATH`.

### LINKED ISSUE #10661: config folder/file not found
On a Windows 10 computer, when I issue the command from astropy.config import get_config_dir, I get the following error:

```
C:\ProgramData\Anaconda3\envs\py38\lib\site-packages\astropy\config\configuration.py:557:
ConfigurationMissingWarning: Configuration defaults will be used due to FileNotFoundError:2 on
None warn(ConfigurationMissingWarning(msg))
```

But, I have set XDG_CONFIG_HOME=C:\Users<USERID>.astropy\config, where USERID is my user ID, and there is a valid configuration file in that folder.

pllim:
### Description

This pull request is to address the problems Windows users encounter because `config` looks in `HOMESHARE` (either inaccessible or simply missing) too early. Now, it only looks for `HOMESHARE` if all the other options are exhausted.

I thought about removing it altogether but maybe some people do use `HOMESHARE` to store config; who knows. 🤷 

If this is too controversial, we can move the milestone down to 5.0.

Fixes #10661 
Fixes #8731

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -119,6 +119,9 @@ astropy.config
   it and relying on reloading it from the previous config file. This ensures
   that any programmatically made changes are preserved as well. [#10474]
 
+- Configuration path detection for Windows will now only look for ``HOMESHARE``
+  if all other options are exhausted. [#10705]
+
 astropy.constants
 ^^^^^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -119,8 +119,9 @@ astropy.config
   it and relying on reloading it from the previous config file. This ensures
   that any programmatically made changes are preserved as well. [#10474]
 
-- Configuration path detection for Windows will now only look for ``HOMESHARE``
-  if all other options are exhausted. [#10705]
+- Configuration path detection logic has changed: Now, it looks for ``~`` first
+  before falling back to older logic. In addition, ``HOMESHARE`` is no longer
+  used in Windows. [#10705]
 
 astropy.constants
 ^^^^^^^^^^^^^^^^^
@@ -1064,7 +1065,7 @@ astropy.stats
 
 - Fixed bug in ``funcs.poisson_conf_interval`` with
   ``interval='kraft-burrows-nousek'`` where certain combinations of source
-  and background count numbers led to ``ValueError`` due to the choice of 
+  and background count numbers led to ``ValueError`` due to the choice of
   starting value for numerical optimization. [#10618]
 
 
--- a/astropy/config/paths.py
+++ b/astropy/config/paths.py
@@ -15,7 +15,7 @@ __all__ = ['get_config_dir', 'get_cache_dir', 'set_temp_config',
 
 
 def _find_home():
-    """ Locates and return the home directory (or best approximation) on this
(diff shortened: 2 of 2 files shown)


embray:
As I wrote on #10661, I wonder if we shouldn't replace this logic entirely to just use `os.path.expanduser('~')` as the default on both POSIX and Windows systems.  By doing so we can delegate this logic to what Python already provides, which is probably better battle-tested and thought out by experts.  According to [the docs](https://docs.python.org/3.6/library/os.path.html#os.path.expanduser):

> On Windows, `HOME` and `USERPROFILE` will be used if set, otherwise a combination of `HOMEPATH` and `HOMEDRIVE` will be used. An initial `~user` is handled by stripping the last directory component from the created user path derived above.

It appears `HOMESHARE` is not used at all in this logic. According to [this page](https://renenyffenegger.ch/notes/Windows/development/environment-variables/index) which documents Windows environment variables better than anything I could find directly from Microsoft, `HOMESHARE` is only meaningful if `HOMEDRIVE` happens to be a UNC path (and I guess in this case they would be the same), in which case `HOMESHARE` points to directory under `HOMEDRIVE` that contains the user's home.  A lot of this probably exists for legacy purposes that aren't very meaningful to a typical Windows user outside some specialized enterprise deployment.

Also worth noting that as of Python 3.8:

> *Changed in version 3.8*: No longer uses `HOME` on Windows.

That change was motivated by [this issue](https://bugs.python.org/issue36264).

pllim:
@embray , I tried to implement your suggestion in the second commit but there is a chance I completely misunderstood, so please review. Does the doc at https://docs.astropy.org/en/latest/config/index.html need updating too?

embray:
I'm otherwise :+1: on this as-is.

pllim:
@embray , thanks for opening the meta issue, it was really illuminating. If you are okay with this, would you mind approving and merging? If you need write access, please let Kelle know.

embray (approved):
LGTM

embray:
Cool, looks like I have access to merge.  Just had to approve my review.

pllim:
Thanks! 😄

PR #10709, opened 2020-09-02
Fix reported module name of tabular model classes

eslavich:
### Description

Similar to #10694, the `__module__` attribute reports "abc" for `tabular.Tabular1D` and `tabular.Tabular2D`.  This PR sets the value to "astropy.modeling.tabular" instead.

### DIFF
--- a/astropy/modeling/tests/test_models.py
+++ b/astropy/modeling/tests/test_models.py
@@ -752,6 +752,16 @@ def test_tabular1d_inverse():
     assert_allclose(t(result), 100)
 
 
+@pytest.mark.skipif("not HAS_SCIPY_14")
+def test_tabular_module_name():
+    """
+    The module name must be set manually because
+    these classes are created dynamically.
+    """
+    for model in [models.Tabular1D, models.Tabular2D]:
+        assert model.__module__ == "astropy.modeling.tabular"
+
+
 class classmodel(FittableModel):
     f = Parameter(default=1)
     x = Parameter(default=0)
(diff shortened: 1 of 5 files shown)


astropy/modeling/tests/test_models.py
```diff
@@ -752,6 +752,16 @@ def test_tabular1d_inverse():
     assert_allclose(t(result), 100)
 
 
+@pytest.mark.skipif("not HAS_SCIPY_14")
```
pllim, line 755: `HAS_SCIPY_14` can simply be `HAS_SCIPY` now. Somehow missed this one in #10345 

https://github.com/astropy/astropy/blob/ee8de6ba65e2d265dcb1d6ee48d658f3564fc99e/setup.cfg#L61

eslavich: Ok, thanks, I'll update all of these in this file

eslavich:
@pllim can we milestone this for 4.0.2?  The new asdf-astropy package won't work with versions of astropy that don't have this change.

nden (commented):
Looks good to me, not sure why readthedocs is failing, may need to push a commit to restart it.

pllim:
RTD is unrelated, see #10711 . If you are worried about doc build failing because of this PR, then for now, you have to attempt to build the docs locally.

pllim:
RTD just fixed. I am going to close/open to kick it off again.

nden (approved):

PR #10710, opened 2020-09-03
Fix logic and improve error message of name_resolve

pllim:
### Description

This pull request is to address an error in remote data test that I can reproduce locally. It does not seem to be one of those transient internet errors. Either there is a bug in the server or the test setup. Will need some `coordinates` expert to review.

While I am at it, also fixed a logic bug (`and` -> `or`) and upgrade some messages to use f-string.

Example log: https://travis-ci.org/github/astropy/astropy/jobs/723713161

**Todo**

- [x] How to fix the test? (Upstream fixed required.)

**To reproduce this error locally**

```python
>>> from astropy.coordinates.name_resolve import sesame_database
>>> with sesame_database.set('vizier'):
...     icrs = SkyCoord.from_name('NGC 3642')
```
```
.../coordinates/sky_coordinate.py in from_name(cls, name, frame, parse, cache)
   1947         from .name_resolve import get_icrs_coordinates
   1948 
-> 1949         icrs_coord = get_icrs_coordinates(name, parse, cache=cache)
   1950         icrs_sky_coord = cls(icrs_coord)
   1951         if frame in ('icrs', icrs_coord.__class__):

.../coordinates/name_resolve.py in get_icrs_coordinates(name, parse, cache)
    192             err = f"Unable to find coordinates for name '{name}' in database {database} using {url}"
    193 
--> 194         raise NameResolveError(err)
    195 
    196     # Return SkyCoord object

NameResolveError: Unable to find coordinates for name 'NGC 3642' in database vizier
using http://cdsweb.u-strasbg.fr/cgi-bin/nph-sesame/V?NGC%203642
```

Indeed, when I enter http://cdsweb.u-strasbg.fr/cgi-bin/nph-sesame/V?NGC%203642 in my browser, this is what I get:

```
# NGC 3642	#Q148742
#! *** NNNothing found ***  (from cache)

#====Done (2020-Sep-03,16:09:44z)====
```

But from the test setup, seems like this is what is expected:

```
# NGC 3642    #Q22523677
#=V=VizieR (local):    1
%J 170.56 +59.08 = 11:22.2     +59:05
%I.0 {NGC} 3642



#====Done (2013-Feb-12,16:37:42z)====
```

What happened to the galaxy, Vizier? https://en.wikipedia.org/wiki/NGC_3642

### DIFF
--- a/astropy/coordinates/name_resolve.py
+++ b/astropy/coordinates/name_resolve.py
@@ -170,8 +170,8 @@ def get_icrs_coordinates(name, parse=False, cache=False):
             # There are some cases where urllib2 does not catch socket.timeout
             # especially while receiving response data on an already previously
             # working request
-            e.reason = "Request took longer than the allowed {:.1f} " \
-                       "seconds".format(data.conf.remote_timeout)
+            e.reason = ("Request took longer than the allowed "
+                        f"{data.conf.remote_timeout:.1f} seconds")
             exceptions.append(e)
             continue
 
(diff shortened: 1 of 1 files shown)


pllim:
Such a small logic change... Does this need a change log?

adrn:
I can reproduce locally as well, but this looks like a database issue with NGC objects (I tried some other NGC XX and get the same response), so I would guess that this an intermittent issue on the server-side.

adrn (approved):
But regardless of the server issues, these changes look good!

pllim:
RTD error is unrelated. Thanks for the review!

I also emailed the service provider. Let's see what they say... 🤞

PR #10718, opened 2020-09-08
DOC: Fix HST data handbook links

pllim:
### Description

This pull request is to address outdated links to STScI data handbooks for WFPC2 and WFC3. While the links are not broken, they currently also don't take you anywhere useful, as they re-route to STScI generic landing page. This patch links them to where they really should go.

Rendered page at https://astropy--10718.org.readthedocs.build/en/10718/nddata/bitmask.html


mwcraig (approved):
Thanks, merging in a moment

Issue #10409, opened 2020-05-28
Test failures in io.ascii in manylinux docker environment


astrofrog:
This is extracted from https://github.com/astropy/astropy/pull/10405 - basically in some circumstances there are io.ascii failures, which can be reproduced by doing:

```
docker run -it -v $PWD:/repo quay.io/pypa/manylinux2010_x86_64
```

(:warning: quay.io is down currently so this may not work right now)

followed by:

```
/opt/python/cp36-cp36m/bin/pip install virtualenv
/opt/python/cp36-cp36m/bin/python -m virtualenv --no-download /tmp/tmp.ISnfMUQtVv/venv
source /tmp/tmp.ISnfMUQtVv/venv/bin/activate
pip install .[test]
cd
pytest --pyargs astropy.io.ascii --astropy-header
```

The failures are:

```
=================================== FAILURES ===================================
_______________ test_data_out_of_range[False-fast_reader1-True] ________________

parallel = False, fast_reader = {'parallel': False, 'use_fast_converter': False}
guess = True

    @pytest.mark.parametrize("guess", [True, False])
    # fast_reader configurations: False| 'use_fast_converter'=False|True
    @pytest.mark.parametrize('fast_reader', [False, dict(use_fast_converter=False),
                                             dict(use_fast_converter=True)])
    @pytest.mark.parametrize("parallel", [False, True])
    def test_data_out_of_range(parallel, fast_reader, guess):
        """
        Numbers with exponents beyond float64 range (|~4.94e-324 to 1.7977e+308|)
        shall be returned as 0 and +-inf respectively by the C parser, just like
        the Python parser.
        Test fast converter only to nominal accuracy.
        """
        # Python reader and strtod() are expected to return precise results
        rtol = 1.e-30
    
        # Update fast_reader dict; adapt relative precision for fast_converter
        if fast_reader:
            fast_reader['parallel'] = parallel
            if fast_reader.get('use_fast_converter'):
                rtol = 1.e-15
            elif np.iinfo(np.int_).dtype == np.dtype(np.int32):
                # On 32bit the standard C parser (strtod) returns strings for these
                pytest.xfail("C parser cannot handle float64 on 32bit systems")
    
        if parallel:
            if not fast_reader:
                pytest.skip("Multiprocessing only available in fast reader")
            elif TRAVIS:
                pytest.xfail("Multiprocessing can sometimes fail on Travis CI")
    
[...]

PR #10719, opened 2020-09-09
Add wheel and sdist auto-release Azure pipeline

astrofrog:
*Note: this is a variant on https://github.com/astropy/astropy/pull/10405 - the present PR is for stable releases and I've tried to address some common concerns that were raised in #10405*

### Background

The [astropy-wheels](https://github.com/MacPython/astropy-wheels) machinery is currently broken and needs work to update its infrastructure to follow other projects. However no one from astropy has yet had the time to get it working fully (unlike conda-forge, there is no bot to update these things). It looks like other -wheels repos have moved to Azure so if that's the case why not just use Azure here?

In addition, the current wheel building process can be frustrating, because it is in a separate repo so it often takes a while to iterate on fixing issues. Further, the LTS, nightly and main builds are in different branches of the astropy-wheels repo which are a pain to sync. With this PR we have one fewer repo to maintain.

Finally the astropy-wheels approach is specific to the core package and can therefore not be used for coordinated packages (unless we create many -wheels repos).

This PR is a *suggestion* (emphasis) to switch to using the OpenAstronomy Azure template for building and auto-releasing wheels. I don't want to force this on people but I want to highlight how simple the configuration is here.

### Ugh, another CI service to look after?

We are probably going to have to deal with Azure even if we stick with astropy-wheels since other projects are doing that. Also the Azure configuration here is really minimal.

### Isn't this going to mean more things to maintain?

No. The OpenAstronomy template is simpler than the astropy-wheels repo and is being maintained already for other projects. The cibuildwheel tool that the template uses is actively maintained by others outside the project. In addition the wheels will be built in our repo so no dancing around between repos when issues come up.

@Cadair and I have commited to maintain the template and if this PR is merged I commit to maintaining the astropy use of the template.

### Is this actually reliable?

Yes. We already use this appoach for packages such as pyerfa, reproject, and astropy-healpix and it is also used by sunpy, glue, and fast-histogram. There have been zero complaints about the generated wheels. In addition we can test this first for 4.1rc2 to minimize any risks.

### Are there other benefits?

I'm glad you asked! This automatically uploads the wheels, and the source distribution, to PyPI when a ``v*`` tag is pushed. So that's fewer steps in the release process.

### Wait this is going to increase CI time and complexity for PRs... 🙄 

No - I want to make this only run on ``v*`` tags and ``v*`` branches (and optionally other staging branches if we want, for testing). Releases would only be uploaded to PyPI if all wheels build successfully, and only for ``v*`` tags. For now it is enabled for all PRs to allow this PR to be tested

### GitHub Actions FTW!

If someone is willing to write and maintain a GitHub Action that does all this, I'm on board. But that means fragmenting things more since a bunch of projects in the astropy org already use this. This template works right now.

### You are going to update the dev docs if we decide to go with this, right? 📖 

Of course!

### The CI on this PR is failing! Ha, i knew it we shouldn't use this!

If any issues come up they might be real issues with astropy not highlighte by the astropy-wheels process! I'll work on fixing issues that come up :)

### Wait do we discuss it here or in https://github.com/astropy/astropy/pull/10405?

Let's discuss things here and keep the discussion in https://github.com/astropy/astropy/pull/10405 specific to building **nightly** wheels. If we go ahead with this PR, https://github.com/astropy/astropy/pull/10405 is a trivial extension.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,6 +28,8 @@ jobs:
         pypi_connection_name : 'pypi_endpoint'
       targets:
       - sdist
-      - wheels_linux
+      - wheels_*linux_i686
+      - wheels_*linux_x86_64
       - wheels_macos
-      - wheels_windows
+      - wheels_*win32
+      - wheels_*win_amd64
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,8 +28,8 @@ jobs:
         pypi_connection_name : 'pypi_endpoint'
       targets:
       - sdist
-      - wheels_*linux_i686
-      - wheels_*linux_x86_64
-      - wheels_macos
-      - wheels_*win32
-      - wheels_*win_amd64
+      - wheels_cp3[678]*linux_i686
+      - wheels_cp3[678]*linux_x86_64
+      - wheels_cp3[678]*macosx_x86_64
+      - wheels_cp3[678]*win32
+      - wheels_cp3[678]*win_amd64
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,6 @@
-variables:
-  CIBW_BUILD: cp36-* cp37-* cp38-*
+# NOTE: this Azure configuration is used only for building the sdist and wheels.
+# It uses the OpenAstronomy azure template described in detail at
+# https://openastronomy-azure-pipelines.readthedocs.io/en/latest/
 
 resources:
   repositories:
@@ -9,10 +10,14 @@ resources:
     name: OpenAstronomy/azure-pipelines-templates
     ref: master
 
+# NOTE: for now we only use Azure Pipelines on v* branches and tags
+# for building the source and wheel distributions. If you want to
+# make changes to this configuration via a pull request, you can
+# *temporarily* change the branches include trigger to just '*'
 trigger:
   branches:
     include:
-    - '*'
+    - 'v*'
   tags:
     include:
     - 'v*'
@@ -20,12 +25,18 @@ trigger:
 jobs:
   - template: publish.yml@OpenAstronomy
     parameters:
+
       # FIXME: we exclude the test_data_out_of_range test since it
       # currently fails, see https://github.com/astropy/astropy/issues/10409
       test_command: pytest -p no:warnings --astropy-header -k "not test_data_out_of_range" --pyargs astropy
       test_extras: test
+
+      # NOTE: for v* tags, we auto-release to PyPI. See
+      # https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html
+      # for information on how to configure things on the Azure Pipelines side
       ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
         pypi_connection_name : 'pypi_endpoint'
+
       targets:
       - sdist
       - wheels_cp3[678]*linux_i686
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,7 @@ trigger:
 jobs:
   - template: publish.yml@OpenAstronomy
     parameters:
-      test_command: pytest -p no:warnings --pyargs astropy
+      test_command: pytest -p no:warnings --astropy-header --pyargs astropy
       test_extras: test
       ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
         pypi_connection_name : 'pypi_endpoint'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,31 @@
+variables:
+  CIBW_BUILD: cp36-* cp37-* cp38-*
+
+resources:
+  repositories:
+  - repository: OpenAstronomy
+    type: github
+    endpoint: astropy
+    name: OpenAstronomy/azure-pipelines-templates
+    ref: master
+
+trigger:
+  branches:
+    include:
+    - '*'
+  tags:
+    include:
+    - 'v*'
+
+jobs:
+  - template: publish.yml@OpenAstronomy
+    parameters:
+      test_command: pytest -p no:warnings --pyargs astropy
+      test_extras: test
+      ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
+        pypi_connection_name : 'pypi_endpoint'
+      targets:
+      - sdist
+      - wheels_linux
+      - wheels_macos
+      - wheels_windows
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,9 @@ trigger:
 jobs:
   - template: publish.yml@OpenAstronomy
     parameters:
-      test_command: pytest -p no:warnings --astropy-header --pyargs astropy
+      # FIXME: we exclude the test_data_out_of_range test since it
+      # currently fails, see https://github.com/astropy/astropy/issues/10409
+      test_command: pytest -p no:warnings --astropy-header -k "not test_data_out_of_range" --pyargs astropy
       test_extras: test
       ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
         pypi_connection_name : 'pypi_endpoint'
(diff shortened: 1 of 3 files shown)


Cadair:
One suggestion from my experience with SunPy, I suggest you enable building (but obviously not uploading) the wheels for all builds on non-master branches (release branches) so that you can find issues with the builds before you tag a release.

astrofrog:
Ah yes of course, having it on the ``v*`` branches in addition to ``v*`` tags would be the natural way to test the machinery before actually tagging a release. I'll update my description above.

astrofrog:
There are failures in the linux builds which are a real reproducible issue: https://github.com/astropy/astropy/issues/10409

pllim:
cc @embray 

> GitHub Actions FTW!

Haha! Why do I feel targeted?

Also, I am neutral on this. I'll just stay on the sideline and use whatever come out victorious.

astrofrog:
For due diligence, I will also open a PR that does the exact same thing but with Travis, so that we can compare options.

astrofrog:
This is ready for review! If/once approved, I need to do two things:

* Remove the last commit
* Add the PyPI token to Azure

I'll then delete the v4.1rc2 tag, backport this PR, and push the tag again.

Note that we can also backport this to v4.0.x but that will require a small edit to setup.py to work properly - I'll open up a follow up PR for that.

astrofrog:
The failure is just a SAMP timeout, might exclude those tests from the releasing testing since they are flakey

pllim:
p.s. When are we moving SAMP to PyVO? 😉

astrofrog:
> p.s. When are we moving SAMP to PyVO? 😉

I would gladly approve that PR!

eteq:
>> p.s. When are we moving SAMP to PyVO? wink

>I would gladly approve that PR!

Me too!

CHANGES.rst
```diff
@@ -1159,6 +1159,8 @@ Other Changes and Additions
 - ``extract_array`` raises a ``ValueError`` if the data type of the
   input array is inconsistent with the ``fill_value``. [#10602]
 
+- Add Azure Pipelines configuration for auto-releasing source and
```
eteq, line 1162: It's not clear to me this is needed, since this is about the CI infrastructure and docs, which don't normally require changelog entries?

azure-pipelines.yml
```diff
@@ -0,0 +1,47 @@
...
+
+# NOTE: for now we only use Azure Pipelines on v* branches and tags
+# for building the source and wheel distributions. If you want to
+# make changes to this configuration via a pull request, you can
+# *temporarily* change the branches include trigger to just '*'
+trigger:
+  branches:
+    include:
+    #- 'v*'
+    - '*'
```
eteq, line 21: I speculate this is here to make sure the tests run it as noted above, but I guess this comment is a reminder to switch it back before merging

docs/development/releasing.rst
```diff
@@ -157,19 +138,14 @@ packages that use the full bugfix/maintenance branch approach.)
...
-   Travis-CI interface at https://travis-ci.org/MacPython/astropy-wheels.
-   You'll need commit privileges to the ``astropy-wheels`` repo; ask Tom Kooij
-   or on the mailing list if you do not have them.
+  At this point if all goes well, the wheels and sdist will be build
+  in the `Azure core package pipeline`_ and uploaded to PyPI!
 
-#. If the tests do *not* pass, you'll have to fix whatever the problem is.
-   First you will need to back out the release procedure by dropping the commits
-   you made for release and removing the tag you created::
+#. In the unlikely event there are any issues with the wheel building for the tag
```
eteq, line 144:
```suggestion
#. In the event there are any issues with the wheel building for the tag
```
This may look like trolling, but it's in the docs style guide to try to avoid "extraneous"/judgemental  language (https://docs.astropy.org/en/stable/development/style-guide.html#a-note-about-style-and-tone) - this is debatable, but I feel it applies here.

azure-pipelines.yml
```diff
@@ -0,0 +1,47 @@
...
+    ref: master
+
+# NOTE: for now we only use Azure Pipelines on v* branches and tags
+# for building the source and wheel distributions. If you want to
+# make changes to this configuration via a pull request, you can
+# *temporarily* change the branches include trigger to just '*'
+trigger:
+  branches:
+    include:
+    #- 'v*'
```
eteq, line 20: Do we really want the trigger to happen on all v* branches? In the run-up to releases on LTS branches there's often a few commits for things like cleaning up changelogs, etc, which shouldn't get built, right?

eteq (changes requested):
Mostly I like this and am happy with it as a way to get around the current release troubles - I  think we should go with it provisionally at least for rc2 and if it works there try it for the final release.

See my inline comments about some more minor items.  But one "big picture" item: is there a reasonable way to have things run up to the pypi release but have a "check point"? I think it's good for the release to have a pause where the releaser has to do one final check to make sure things are good.  Otherwise (as the docs point out) we are stuck with backing out history and all the problems that leads to.

CHANGES.rst
astrofrog: Ok I'll remove

azure-pipelines.yml
astrofrog: Yep indeed I'll remove before merging

azure-pipelines.yml
astrofrog: To be clear they will be built but not uploaded - I think having them build on any push to a v* branch is a good idea as it gives advance indication of whether the wheels are broken, as opposed to just at the point of tagging. It's a compromise since at least we aren't building in master and all PRs.

astrofrog:
> See my inline comments about some more minor items. But one "big picture" item: is there a reasonable way to have things run up to the pypi release but have a "check point"? I think it's good for the release to have a pause where the releaser has to do one final check to make sure things are good. Otherwise (as the docs point out) we are stuck with backing out history and all the problems that leads to.

The main checkpoint at the moment is that all wheels have to successfully build for them to be uploaded, so that there should be no incomplete releases. If you want an additional check, we don't have to upload to the main PyPI server, we could upload to the test PyPI or our own if we wanted. So yes it's technologically possible to upload to any PyPI-type server that we want.

However, for rc2, why don't we upload direct to PyPI, and we can consider not doing that for subsequent releases if needed? (which gives us more time to think what other server we'd want to upload to)

Another variant on this is that we could imagine uploading to test.pypi.org for all v* branch builds and main PyPI for tags?

astrofrog:
I've now removed the temporary commit that forced the build in the PR since the wheels are building fine.

azure-pipelines.yml
eteq: Gotcha, understood, that sounds good then

eteq: Gotcha, that sounds good then.

eteq:
> However, for rc2, why don't we upload direct to PyPI, and we can consider not doing that for subsequent releases if needed? (which gives us more time to think what other server we'd want to upload to)

yeah, I like this idea! My thinking for the check point is not so much because of a concern about the builds failing, but rather a desire to time the PyPI release with announcements and the like. (and admittedly a possibly-pathological desire to want to be able to do one last check before going forward with it.)  So this will satsify this for now and we can think about whetether we need anything later.

eteq (approved):

eteq:
Alright, this LGTM now! I'm not merging on the basis that I'm not clear if a change needs to be made based on what @astrofrog suggested in https://github.com/astropy/astropy/pull/10719#issuecomment-692923905, but I'm good with this now as long as that's all set.

astrofrog:
This is ready, I'll merge, backport, and re-push the tag

Cadair:
I am also reasonably sure you can have azure pipelines have a human in press the go button before a deploy. I would have to look into it though.

eteq:
> I am also reasonably sure you can have azure pipelines have a human in press the go button before a deploy. I would have to look into it though.

@Cadair - yeah I know I saw a demo of this once during a talk by an Azure developer/salesperson... But that was like a  year or two ago so :shrug:

astrofrog:
This might be relevant: https://stackoverflow.com/a/61519132

Cadair:
While it can be done, I am interested in understanding what you think the human-in-the-loop would be checking about the built wheels / sdists which hasn't already (or couldn't be) automated?

PR #10726, opened 2020-09-11
add `spat` unit of solid angle

### LINKED ISSUE #10721: add the Spat unit for solid angles
### Description

Add the [Spat unit of solid angle](https://en.wikipedia.org/wiki/Spat_(unit)), where 1 Spat = 4 sterad is the angle subtended by the entire sphere. This would make it possible to give "fsky" sky fractions as quantities of solid angle in a not too contrived way.

(Although I do concede that I have also never heard of the "spat" until I came across it just now.)

ntessore:
### Description

This pull request is to address #10721. It adds the [`spat` unit (abbreviation `sp`, apparently)](https://en.wikipedia.org/wiki/Spat_(unit)) where `1 spat` corresponds to 4pi steradian, i.e. the solid angle of the entire sphere.

That makes it possible to convert `fsky` sky fractions to solid angle transparently:

```python
survey_area = fsky*u.spat
```

And also to convert solid angles to sky fractions:

```python
sky = survey_area.to_value(u.spat)
```

Fixes #10721

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -104,6 +104,9 @@ astropy.units
 - ``Quantity.to`` has gained a ``copy`` option to allow copies to be avoided
   when the units do not change. [#10517]
 
+- Added the ``spat`` unit of solid angle that represents the full sphere.
+  [#10726]
+
 astropy.utils
 ^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


pllim:
Hello and thank you! Please add a change log under "new features" under the given milestone.

@mhvk , does this need a "what's new"?

mhvk (approved):
Super! And funny to see that in the previous PR for cycle, we didn't seem to see a need to include it even in `CHANGES.rst` - we have made some progress! Don't think it is necessary to add to `what's new` - I hope units are such that when people want something, it is already there, and, if not, they'll add it (like here). Thanks, @ntessore!

ntessore:
Thanks everyone for the quick help! That was fun

Issue #10725, opened 2020-09-11
Bug in SkyCoord unit kwarg when distance is specified


parejkoj:
I believe I've found a bug with the `unit` argument of `SkyCoord`. Specifying `unit` and `distance` to the `SkyCoord` constructor results in a `UnitConversionError`, while providing the ra/dec values as `astropy.units.Quantity` objects does not:

```python
from astropy.coordinates import SkyCoord, Distance
import astropy.units as u

# no error
coord1=SkyCoord(ra=10*u.deg, dec=10*u.deg,
               obstime="J2000",
               pm_ra_cosdec=10*u.mas/u.yr,
               pm_dec=10*u.mas/u.yr,
               distance=Distance(parallax=10*u.mas))

# UnitConversionError: 'pc' (length) and 'deg' (angle) are not convertible
coord2=SkyCoord(ra=10, dec=10, unit="deg"
               obstime="J2000",
               pm_ra_cosdec=10*u.mas/u.yr,
               pm_dec=10*u.mas/u.yr,
               distance=Distance(parallax=10*u.mas))
```

I think this is due to something not getting the correct equivalency somewhere, based on the same error being described on the [equivalency page](https://docs.astropy.org/en/stable/units/equivalencies.html#how-to-convert-parallax-to-distance), but I don't know how.

`astropy.__version__` reports `4.0.1.post1`.

EDIT: Syntax highlighting.

astrojuanlu:
Complete traceback for future reference:

```
>>> from astropy.coordinates import SkyCoord, Distance
>>> import astropy.units as u
>>> coord2=SkyCoord(ra=10, dec=10, unit="deg",
...                obstime="J2000",
...                pm_ra_cosdec=10*u.mas/u.yr,
...                pm_dec=10*u.mas/u.yr,
...                distance=Distance(parallax=10*u.mas))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/sky_coordinate.py", line 312, in __init__
    skycoord_kwargs, components, info = _parse_coordinate_data(
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/sky_coordinate_parsers.py", line 245, in _parse_coordinate_data
    valid_components.update(_get_representation_attrs(frame, units, kwargs))
[...]

PR #10728, opened 2020-09-13
Disallow passing both unit= and Quantities in SkyCoord

astrojuanlu:
### Description

This closes gh-10725 by disallowing the use of both `unit=` and `Quantity`es as keyword parameters in the `SkyCoord` initializer, which is somewhat ambiguous.

Unsure if this deserves a changelog entry.


pllim:
Thanks! Technically a bug fix, so it does need a change log. I'll let reviewer move the milestone if I set it incorrectly.

astrojuanlu:
@mhvk has expressed in https://github.com/astropy/astropy/issues/10725#issuecomment-691745127 that he would just clarify the docstring though, so I will wait a bit for others to chime in before moving this forward.

adrn:
👋 Thanks for working on this @astrojuanlu! I left some comments in the original issue - I'm also in favor of keeping the error, fixing the docstring, but perhaps adding some more clarification to the error message.

astrojuanlu:
Superseded by gh-10749.

Issue #7572, opened 2018-06-18
Logger test failures when pytest-qt is installed


astrofrog:
There is a pytest plugin called pytest-qt that is used by some packages - one of the side effect it has on all packages is to fail if there is any output printed to stderr during the test phase. There are a couple of failures in astropy that happen when pytest-qt is installed - this is more annoying than serious but it'd still be good to track down why these happen, and whether they can be safely ignored. 

```
============================================================================== FAILURES ===============================================================================
_______________________________________________________________________ test_exception_logging ________________________________________________________________________
CALL ERROR: Qt exceptions in virtual methods:
________________________________________________________________________________
Traceback (most recent call last):
  File "/private/var/folders/zy/t1l3sx310d3d6p0kyxqzlrnr0000gr/T/astropy-test-1izlslk0/lib/python3.6/site-packages/astropy/tests/test_logger.py", line 205, in test_exception_logging
    raise Exception("This is an Exception")
Exception: This is an Exception

________________________________________________________________________________
Traceback (most recent call last):
  File "/private/var/folders/zy/t1l3sx310d3d6p0kyxqzlrnr0000gr/T/astropy-test-1izlslk0/lib/python3.6/site-packages/astropy/tests/test_logger.py", line 217, in test_exception_logging
    raise Exception("This is an Exception")
Exception: This is an Exception

________________________________________________________________________________
Traceback (most recent call last):
  File "/private/var/folders/zy/t1l3sx310d3d6p0kyxqzlrnr0000gr/T/astropy-test-1izlslk0/lib/python3.6/site-packages/astropy/tests/test_logger.py", line 232, in test_exception_logging
    raise Exception("This is an Exception")
Exception: This is an Exception

________________________________________________________________________________

------------------------------------------------------------------------ Captured stderr call -------------------------------------------------------------------------
Qt exceptions in virtual methods:
________________________________________________________________________________
Traceback (most recent call last):
  File "/private/var/folders/zy/t1l3sx310d3d6p0kyxqzlrnr0000gr/T/astropy-test-1izlslk0/lib/python3.6/site-packages/astropy/tests/test_logger.py", line 205, in test_exception_logging
    raise Exception("This is an Exception")
Exception: This is an Exception

________________________________________________________________________________
ERROR: Exception: This is an Exception [astropy.tests.test_logger]
Qt exceptions in virtual methods:
________________________________________________________________________________
Traceback (most recent call last):
[...]

PR #10730, opened 2020-09-14
TST: Ignore unknown config ini key (pytest 6.1)

pllim:
### Description

This pull request is to utilize pytest-dev/pytest#7700 to silence this warning below that does not break tests but is kinda annoying. It will not be proven work (or not) until pytest 6.1 is out (pytest-dev/pytest#7703), so there is no point in merging this too soon.

```
=============================== warnings summary ===============================
.../_pytest/config/__init__.py:1148
  .../_pytest/config/__init__.py:1148: PytestConfigWarning: Unknown config ini key: qt_no_exception_capture
  
    self._warn_or_fail_if_strict("Unknown config ini key: {}\n".format(key))

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

We need to keep the unused plugin config due to #7572

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -125,6 +125,7 @@ filterwarnings =
     ignore:::astropy.tests.disable_internet
     ignore:direct construction of Asdf:pytest.PytestDeprecationWarning
     ignore:Unknown pytest.mark.mpl_image_compare:pytest.PytestUnknownMarkWarning
+    ignore:Unknown config ini key:pytest.PytestConfigWarning
     ignore:matplotlibrc text.usetex:UserWarning:matplotlib
 doctest_norecursedirs =
     */setup_package.py
--- a/setup.cfg
+++ b/setup.cfg
@@ -125,7 +125,7 @@ filterwarnings =
     ignore:::astropy.tests.disable_internet
     ignore:direct construction of Asdf:pytest.PytestDeprecationWarning
     ignore:Unknown pytest.mark.mpl_image_compare:pytest.PytestUnknownMarkWarning
-    ignore:Unknown config ini key:pytest.PytestConfigWarning
+    ignore:Unknown config option:pytest.PytestConfigWarning
     ignore:matplotlibrc text.usetex:UserWarning:matplotlib
 doctest_norecursedirs =
     */setup_package.py


setup.cfg
```diff
@@ -125,6 +125,7 @@ filterwarnings =
     ignore:::astropy.tests.disable_internet
     ignore:direct construction of Asdf:pytest.PytestDeprecationWarning
     ignore:Unknown pytest.mark.mpl_image_compare:pytest.PytestUnknownMarkWarning
+    ignore:Unknown config ini key:pytest.PytestConfigWarning
```
dhomeier, line 128:
```suggestion
    ignore:Unknown config ini key:pytest.PytestConfigWarning
    ignore:Unknown config option:pytest.PytestConfigWarning
```
Seems we need both to get rid of all warnings, which otherwise is now with 6.1 causing a failure early on (due to our `xfail_strict = true`?).
So this would actually be needed quite urgently to fix #10773 .

pllim: Thanks for the comment. I see that they released pytest yesterday during a weekend. I'll investigate on Monday.

dhomeier: Yes, looks like all tests are failing right now. So you might give it a try with the `ignore: Unknown config option` added to see if that passes, or if we also need the
`ignore:The imp module is deprecated in favour of importlib:DeprecationWarning`

saimn: It seems they renamed the warning so the "Unknown config option" line would be sufficient.
@dhomeier - all warnings are turned into error with :  
```
filterwarnings =
    error
    ...
```

saimn:
```suggestion
    ignore:Unknown config option:pytest.PytestConfigWarning
```

saimn: I committed the change for the warning to see if it fixes the CIs.

dhomeier (approved):
Made it to the actual tests now; so once they've all succeeded this can go in, unless you want to still change the milestone.

saimn:
Yes, this would need to be backported to the other release branches.
@dhomeier - I ran the tests and did not see the `imp` warning, so I think this is ready to merge once all builds pass.

pllim:
Thanks, y'all!

dhomeier:
Thanks for spotting this early on and having a solution ready before the problem popped up!

pllim:
Kudos go to @nicoddemus and other `pytest` devs involved too to allow us to ignore this warning. 😄

PR #10733, opened 2020-09-15
POC: Actions to comment on draft PR

pllim:
### Description

This pull request implements our first GitHub Action! It leaves a friendly comment when a Draft PR is opened. The comment reminds PR author that they can skip CI if it is not ready. This is based on a discussion that happened, starting around https://github.com/astropy/astropy/issues/10729#issuecomment-692715045 . 

Examples of this same Action, uh, in action:

* Opened as a draft PR: pllim/playpen#29
* Normal turning into draft PR: pllim/playpen#31 (Turning into draft later will not re-run the Action, so no message; do we even want a message in this case?)
* Opened as a normal PR: pllim/playpen#28 (You can see that the Action executed as no-op)
* Draft turning into real PR: pllim/playpen#30 (Making it real did not re-run the Action, but it would be no-op anyway)

p.s. A draft PR about a draft PR. *Much Actions, so meta!*

# Alternative using astropy-bot

1. OpenAstronomy/baldrick#99 needs to get reviewed and merged.
2. `baldrick` needs to be re-deployed by Tom R or Stuart M (unless deploy is automatic somehow?).
3. astropy/astropy-bot#118 needs to be tested and merged.
4. `astropy-bot` needs to be re-deployed by Tom R (unless deploy is automatic somehow?).
5. Integration test of new bot feature with this repo. This step can be moved up a little if Tom R is willing to deploy astropy/astropy-bot#118 on `experimental` branch first and hook it up to a test repo.


.github/workflows/draft_pr_comment.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  pull_request_target:
+    types: ['opened']
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    steps:
+    - name: 'Comment Draft PR'
+      uses: actions/github-script@v3
```
mhvk, line 12: Maybe add a comment where the script comes from?

mhvk (commented):
That looks nice! I think having a message also for the case when someone turns a regular PR into a draft PR is useful (though would be nice if it happened only once).

.github/workflows/draft_pr_comment.yml
pllim: It is from https://github.com/actions/github-script

pllim: Thanks for the feedback! I'll refrain from making this too fancy, though, until we have a buy-in. I just want to show that it is possible right now. For a more complicated workflow, like checking if the bot has commented before, I'll have to investigate more and maybe writing a script to go with this action.

bsipocz (approved):
Looks good, and does the job.

pllim:
As @astrofrog requested, I also updated my original post above to include a route for "Alternative using astropy-bot".

.github/workflows/draft_pr_comment.yml
mhvk: It certainly gives a sense of what's possible; agreed that we should first discuss what to do, since ideally we do not have two different systems.

pllim:
A lot has happened since this was opened, so I removed the standalone YAML and put labeler and this new Action in the same YAML instead.

pllim:
If there are no objections, I would like to get this in and see how things go. We can think about how to handle the use case when someone opens a non-draft PR but then turn it into draft in a future PR, if there is enough demand.

embray (approved):
Makes sense to me =)

pllim:
Thanks for the reviews! Merging.

Issue #7766, opened 2018-08-23
Create radio UVW frame


mhvk:
An attempt was made in #6882, but that may have been incomplete - see https://github.com/astropy/astropy/pull/6882#issuecomment-411385141 in particular

adrn:
Just to re-raise my comment from #6882 with a suggestions: perhaps we could name the frame something like `RadioUVW` to distinguish it from the Galactic velocity components (also UVW)?

mhvk:
@adrn - OK, I see, it is the cartesian representation of `Galactic` you are worried about. So, let's aim for `InterferometricUVW` (unless a better name appears); I've changed the title accordingly.

eteq:
What about just `Interferometric`?  That actually seems cleaner to me - after all we don't call it `GalacticUVW` or even `GalacticLatLon`... Really `AltAz` is the only one with the names (and that's already gotten us into trouble... see #7708)

cmurray-astro:
As the UVW coordinate system for interferometers is used to describe complex visibilities, how about `InterferometricVisibility`?

eteq:
I like `InterferometricVisibility` - agree, @adrn and @mhvk?

Joshuaalbert:
@mhvk If you want to still implement this, I have a reference that will speed things up. The Event Horizon Telescope (VLBI) calculates their UV coordinates [here](https://github.com/achael/eht-imaging/blob/master/ehtim/observing/obs_helpers.py#L55). They only define the U and V coordinates, and the W can be found easily using projU and projV in that source as ```projW = eye(3) - projU - projV```. However, I feel there is actually an easier and more robust way to do UVW with astropy's builtin coords.

Also, I'm not 100% certain the EHT's implementation is correct. I don't think it takes into account the motion of the observer which is different for antennas in different places on Earth.

Joshuaalbert:
@caseyjlaw I found a good person to bring in, who will help get this implemented properly. Frits (@tikk3r) is a PhD student working with LOFAR including the international baselines. We don't call it VLBI because the baselines are only a few 1000km but it's still in the realm of VLBI. Anyways, Frits offers his attention to this. I think we can get it going in a few weeks. @tikk3r [this](https://github.com/astropy/astropy/pull/9869) is the PR related to it and we'll do our conversation there.

Joshuaalbert:
@mhvk a few questions are needed to assert the accuracy of this implementation, based on astropy coordinates.
[...]

PR #10737, opened 2020-09-16
Renewed attempt to add the InterferometricVisibility frame

Joshuaalbert:
This PR reopens #9869 and is a solution for issue #7766. 

Individuals that should follow and give input: @caseyjlaw, @eteq, @mhvk, @tikk3r

I think given the prior PR we can largely break this PR into the following tasks:

- [ ] Derive the exact solution
- [ ] Investigate setting up the frame using astropy.coordinates.SkyOffset
- [ ] If possible, set up the frame using astropy.coordinates.SkyOffset or else find other solution
- [ ] Set up CALC11 server to do testing
- [ ] Investigate setting up CALC11 server as part of continuous integration unittesting

## The exact solution

In order to compute the coordinates in this frame we follow the approach of computing the proper time for a given
geodesic (which forms the `w` coordinate) and then computing the `u` and `v` coordinates from the directional derivatives of
proper time w.r.t the `l` and `m` coordinates.

```
w = c * proper_time
u = dw/dl
v = dw/dm
```

Note: In order to perform VLBI-precise calcuations we must take into account the velocity of the local reference frame of each
geodesic origin.

Computing the coordinates can be expensive, which is why VLBI folks compute them over a temporal abscissa of about 120 seconds and do spline interpolation between. I think we should provide the exact coordinates, and let the user choose if they want to use spline interpolation.

### DIFF
--- a/astropy/coordinates/interferometric_visibility.py
+++ b/astropy/coordinates/interferometric_visibility.py
@@ -1,0 +1,22 @@
+"""
+Provide the interferometric visibility frame, also known as the UVW frame, of radio astronomy to a precision
+suitable for VLBI. The accuracy is comparable to CALC11.
+
+In order to compute the coordinates in this frame we follow the approach of computing the proper time for a given
+geodesic (which forms the w coordinate) and then computing the u and v coordinates from the directional derivatives of
+proper time w.r.t the l and m coordinates.
+
+w = proper-time
+u = dw/dl
+v = dw/dm
+
+In order to perform VLBI-precise calcuations we must take into account the velocity of the local reference frame of each
+geodesic origin.
+
+TODO(@tikk3r, @joshuaalbert): derive the exact solution
+TODO(@tikk3r, @joshuaalbert): Investigate setting up the frame using astropy.coordinates.SkyOffset
+TODO(@tikk3r, @joshuaalbert): If possible, set up the frame using astropy.coordinates.SkyOffset
+TODO(@tikk3r, @joshuaalbert): Set up CALC11 server to do testing
+TODO(@tikk3r, @joshuaalbert): Investigate setting up CALC11 server as part of continuous integration
+"""
+pass


pllim:

mhvk:
Thanks very much - the description of the tasks looks good. Agreed also to start with just getting as-exact-as-possible solutions, even if they are expensive, and worry about interpolation later (though I can resist noting that we just merged a PR that allows it for the current transforms - #10646; even though that focuses on `erfa` routines, I think the infrastructure would be useful here too).

caseyjlaw:
I'd also encourage consideration of an approximate solution that is appropriate for connected interferometers. Computational intensity is one concern. It would be good to know if the SkyOffset approach will allow a fast, vectorized calculation in time (use case: calculate uvw for 8 hours of data with 1 second samples).

Joshuaalbert:
I'm going to look at the SkyOffset code to get acquainted with it.

Joshuaalbert:
Ok after looking at `SkyOffset` we may not  use it for the exact solution. From `SkyOffset` docs:
```
    Coordinates in this frame are both centered on the position specified by the
    ``origin`` coordinate, *and* they are oriented in the same manner as the
    ``origin`` frame.  E.g., if ``origin`` is `~astropy.coordinates.ICRS`, this
    object's ``lat`` will be pointed in the direction of Dec, while ``lon``
    will point in the direction of RA.
```
This won't work for two antennas in different reference frames.

mhvk:
@Joshuaalbert - have you checked the `rotation` argument to the initializer? Also, if it is close, I think it would be fine to adapt or subclass `SkyOffsetFrame`. The question may be the extent to which actual transformations are similar to what we need: https://github.com/astropy/astropy/blob/55c18d11e78117ed75a77e1e9b75f51814d4795e/astropy/coordinates/builtin_frames/skyoffset.py#L87-L97

mhvk:
p.s. Just to be sure: I think a conclusion from the previous discussion was that the `SkyOffsetFrame` *idea* was quite similar to what's needed for `InterferometricVisibility`, in things being based around a particular pointing direction. If it is not obvious and easy, it would seem totally fine to start with a set of classes that work and verify they reproduce `calc`, before worrying too much about whether `SkyOffsetFrame` is directly usable itself.

Joshuaalbert:
Hey @mhvk I mean due to relativistic effects the notion of a pointing offset is incorrect in the exact formulation suitable for VLBI. In the exact solution the UVW coordinates are a function of both the antenna position and the direction of the radio source. The `w` coordinate is the speed of like times the group delay,
Screenshot from 2020-09-19 01-36-07

and `u` and `v` are the partial derivatives with respect to the first two components of `s` in a Cartesian system.

Joshuaalbert:
p.s. that's not exact still because it's lacking some terms, but is still very accurate. The full group delay is in section 11.4.1 of https://www.iers.org/SharedDocs/Publikationen/EN/IERS/Publications/tn/TechnNote36/tn36.pdf

mhvk:
Ah, thanks! (I fear you had in fact told me this before...) So, it sounds like, if anything, `SkyOffsetFrame` would be a special case. The idea of `SkyOffsetFrame` to set oneself up with a meta-class might still be a good one, though.

Joshuaalbert:
To get an idea of the importance of the relativistic kinematic (non gravitational) corrections here I show the difference in delay between including and not including them:
```python
For baseline of  1.0 km magnitude of relativistic corrections 0.3337616158773926 ns
For baseline of  2.7825594022071245 km magnitude of relativistic corrections 0.928711522354206 ns
For baseline of  7.742636826811269 km magnitude of relativistic corrections 2.5841949784698954 ns
For baseline of  21.544346900318832 km magnitude of relativistic corrections 7.190676034471835 ns
For baseline of  59.94842503189409 km magnitude of relativistic corrections 20.008483207959216 ns
For baseline of  166.81005372000593 km magnitude of relativistic corrections 55.674793074140325 ns
For baseline of  464.15888336127773 km magnitude of relativistic corrections 154.91841893456876 ns
For baseline of  1291.549665014884 km magnitude of relativistic corrections 431.06970318127424 ns
For baseline of  3593.813663804626 km magnitude of relativistic corrections 1199.4770555924624 ns
For baseline of  10000.0 km magnitude of relativistic corrections 3337.61615877226 ns
```

Joshuaalbert:
What we can do is provide an option that computes all components with delays above a certain threshold. E.g. Most telescopes will get away with 0.1ns precision.

pllim:

Jan-Willem:
I have created a python package that wraps CALC11 (https://github.com/casangi/radio_telescope_delay_model/blob/main/README.md). Currently, it only wraps the almacalc function which calls the main CALC function aDRIVR. The Fortran files of interest are calc11/src/almacalc.f and calc11/src/adrvr.f 

Here is a link to the example notebook that can be run on Google Colab (https://colab.research.google.com/github/casangi/radio_telescope_delay_model/blob/main/example_notebooks/Radio_Telescope_Delay_Model_Example.ipynb). It compares CALC, CASA, Astropy, and data from a measurement set. The Astropy and CASA calculations are done using code from @demorest. There is no exact agreement and the ms uvw coordinates show a strange difference pattern with the other methods. If I understand correctly the uvw coordinates in an Alma ms (converted from asdm) are done in https://open-bitbucket.nrao.edu/projects/CASA/repos/casa6/browse/casa5/code/alma/apps/asdm2MS/UvwCoords.cc . I have not yet done a careful comparison between UvwCoords.cc and CASA code in the notebook. In the "Run and Compare Functions" section of the notebook, you can change the pointing direction by changing use_user_pointing and user_pointing_direction. 

Let me know if I should expose any of the other CALC functions.

mhvk:
@Jan-Willem - thanks so much! Will need some time to really go through it all...

Joshuaalbert:
@Jan-Willem great job! I'm taking a look at the notebook now.

Joshuaalbert:
Quite strange the difference between MS and the others. Perhaps something related to precision? Thanks for making this notebook. I think it'll be good for comparing against the astropy version which should be as precise or better than CALC.

demorest:
Hi all, after reviewing some of the older discussion, I'm still not sure a new frame is needed.  It still seems to me that the existing `SkyOffsetFrame` does the job of defining the "l" and "m" directions in the usual way relative to a specified sky position.  It would be great to have VLBI-level UVW calculation implemented using astropy, possibly using the derivative method as mentioned above, but I'm not seeing how this needs a new coordinate frame or what that frame would be.

Thanks @Jan-Willem for posting the new comparisons!  When I looked at this before, it seemed that the difference between my astropy and CASA UVWs was a small rotation that got larger for sky positions near the poles.  But I never really tracked down the cause.  Also I looked a bit at CASA code but I think you have much more experience with it than I do.. if UvwCoords.cc is doing something different that would be good to know.

caseyjlaw:
It seems like it would be easy to implement a high-level function that could call either the "astropy-like" function or "calc-like" function, based on the application and precision required. If we took that approach, we could create a PR now that uses the astropy-like function and document the fact that it has known variances from CASA and calc.
I agree that we haven't actually fully characterized those variances, though.

mhvk:
Indeed, there is precedence for that idea of allowing users to chose in how one sets, e.g., the solar system ephemeris. But as you note, we first have to be sure we understand the variances.

Joshuaalbert:
Hi @demorest, the point is that for the precise formulation, the UVW coordinates are a function of position, direction (of the geodesic), and velocity with respect to the solar system barycenter. I agree `SkyOffsetFrame` could be useful if delay is linearised in direction, as it just does an arbitrary rotation of a right-handed-system:
```python
origin = skyoffset_frame.origin.spherical
mat1 = rotation_matrix(-skyoffset_frame.rotation, 'x')
mat2 = rotation_matrix(-origin.lat, 'y')
mat3 = rotation_matrix(origin.lon, 'z')
```
However, even in that case the `origin` and `rotation` arguments would need to be calculated per antenna, direction, and time. I also agree that we can provide varying levels of precision. I personally, don't see why it would be difficult to provide the VLBI-precise transform. Astropy has all the ingredients already.

mhvk:
Slightly randomly but I started another way to access CALC11, one that uses `numpy.f2py` so one can access all subroutines, over at https://github.com/mhvk/pycalc11. For now, mostly just tests.

This happened to be picked up by @aelanman who mentioned his own efforts to see how consistent astropy and CALC11 were (and more interface routines at https://github.com/aelanman/difxcalc_astropy). He also mentioned the https://github.com/RadioAstronomySoftwareGroup/pyuvdata package that does non-relativistic UVW (and apparently uses astropy throughout).

No chance yet to look into any of this in detail...

theXYZT:
> p.s. that's not exact still because it's lacking some terms, but is still very accurate. The full group delay is in section 11.4.1 of https://www.iers.org/SharedDocs/Publikationen/EN/IERS/Publications/tn/TechnNote36/tn36.pdf

@Joshuaalbert I can't seem to access this reference. Can you provide a mirror/alternate link?

pllim:

Joshuaalbert:
@caseyjlaw I have written a JAX+astropy based VLBI-accurate UVW engine here https://github.com/Joshuaalbert/DSA2000-Cal/blob/main/dsa2000_cal/dsa2000_cal/common/uvw_utils.py for the DSA2000 project and wonder if there is any desire still to have this functionality in astropy.  #9869

mhvk:
Definitely still would like to have UVW in astropy (though not a dependency on JAX; is it optional?)

Joshuaalbert:
The implementation relies on the auto-diff functionality of JAX to compute the I've coordinates. Without it derivatives would have to be written out by hand.

Joshuaalbert:
This is the core part that relies on auto diff https://github.com/Joshuaalbert/DSA2000-Cal/blob/4dfd9d2d85c215850de2ef01fc6efef940f3ce7f/dsa2000_cal/dsa2000_cal/common/uvw_utils.py#L312

mhvk:
We do have transformations that do automatic differentation too -- `FunctionTransformWithFiniteDifference` -- so hopefully this is not a showstopper!

caseyjlaw:
Nice to see this work coming back into astropy.
@Joshuaalbert Would you suggest we open a PR to start discussing the question of dependencies, etc. with astropy developers?

Joshuaalbert:
Sure we can open another PR for that. @mhvk that might be an option, do you have some examples of how that might work?

mhvk:
I think I wrote a bit too quickly, `FunctionTransformWithFiniteDifference` does the finite difference only to transform velocities, so that is not directly relevant here. But I think we really should discuss in #7766, since this is a closed PR, and hence will not be easily found/tracked!!

mhvk:
I started an entry at https://github.com/astropy/astropy/issues/7766#issuecomment-2178104410 to transfer us over to #7766.

PR #10741, opened 2020-09-17
[4.1rc2] Fix permissions for a FITS test file

olebole:
### Description

FITS files are not executable and therefore their permissions should be set to 0644. This was not the case for one file, which triggered one of Debian's "lintian" warnings.

This trivial PR fixes the permissions for this file. A changelog entry is probably not needed.


pllim (approved):
Looks uncontroversial enough. Thanks!

p.s. @eteq and @astrofrog can update the milestone if this cannot make it to final 4.1.

pllim:
Test failure is unrelated.

PR #10743, opened 2020-09-17
Improve HDF5 I/O documentation

### LINKED ISSUE #10736: HDF5 metadata handling
### Description
 
PR #7007 fundamentally changed how astropy creates hdf5 files.

Before, the native way of storing metadata to hdf5, hdf5 attributes were used.
This was a good way for interoperability.

The new way stores a custom, second dataset along the normal dataset with the metadata.
This is a very non standard approach that hurts the interoperability of hdf5 files created through astropy.
It also makes it necessary to copy these metadata datasets when copying datasets, while attributes are automatically included when copying a single dataset.

I think it should be possible to at least optionally use the approach taken before and use normal hdf5 attributes for metadata to improve interoperability.

taldcroft:
### Description

Discussion in #10736 highlighted some room for improvement in the documentation related to writing HDF5 files with metadata.

- I noticed that the `compatibility_mode` argument was in the docs but is no longer found in an actual code. I don't recall what happened, but since this was marked as being deprecated it makes sense now to take this out of the docs.
- Improve wording ("as" => "if")
- Add more detail about how the metadata are actually serialized.

Fixes #10736

### DIFF
--- a/docs/io/unified.rst
+++ b/docs/io/unified.rst
@@ -971,7 +971,7 @@ and ``meta``.
 
 By default, when writing a table to HDF5 the code will attempt to store each
 key/value pair within the table ``meta`` as HDF5 attributes of the table
-dataset. This will fail as the values within ``meta`` are not objects that can
+dataset. This will fail if the values within ``meta`` are not objects that can
 be stored as HDF5 attributes. In addition, if the table columns being stored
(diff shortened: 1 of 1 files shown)


pllim:
You can ignore Travis failure. RTD passed, so should be okay.

mhvk (approved):
Looks good!

PR #10745, opened 2020-09-17
DOC: Update ci skip doc

### LINKED ISSUE #10521: DOC: Update doc testing doc to reflect switch to RTD PR builder
### Description

We have stopped using CircleCI + Giles to test doc build in a PR. Now, we use RTD PR builder. The main difference is that, now `[skip travis]` isn't really the right directive anymore to skip CI *except* for doc build. This is because that would still run CircleCI, which is now unnecessary.

In fact, even `[ci skip]` will not stop RTD PR builder from running. However, this might change in the future, see readthedocs/readthedocs.org#871. But until it is implemented over there, `[ci skip]` is now the right thing to do to skip CI on a PR *except* the doc build.

pllim:
### Description

This pull request is to address confusion about usage of `[ci skip]` when building documentation.

Will have to be updated again when readthedocs/readthedocs.org#871 is implemented. Maybe even have to fall back to `[skip travis]` if `[ci skip]` would also skip RTD, but that would kick off CircleCI unnecessarily. We'll revisit when that bridge exists.

Fixes #10521

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,19 +111,20 @@ Other Tips
         $ git commit --amend
 
 - To skip only the tests running on Travis CI use ``[skip travis]``.
+  This will still execute CircleCI.
 
 - If your commit makes substantial changes to the documentation but no code
-  changes, then you can use ``[skip travis]``, which will skip Travis CI
-  because documentation build is done on CircleCI. The exception to this rule
+  changes, then you can use ``[ci skip]``, which will all CI except RTD,
+  where the documentation is built. The exception to this rule
   is when your changes to documentation include code snippets that need to
   be tested using ``doctest``.
 
 - When contributing trivial documentation fixes (i.e., fixes to typos, spelling,
   grammar) that don't contain any special markup and are not associated with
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -111,19 +111,20 @@ Other Tips
         $ git commit --amend
 
 - To skip only the tests running on Travis CI use ``[skip travis]``.
+  This will still execute CircleCI.
 
 - If your commit makes substantial changes to the documentation but no code
-  changes, then you can use ``[skip travis]``, which will skip Travis CI
-  because documentation build is done on CircleCI. The exception to this rule
+  changes, then you can use ``[ci skip]``, which will all CI except RTD,
```
taldcroft, line 117:
```suggestion
  changes, then you can use ``[ci skip]``, which skip will all CI except RTD,
```

taldcroft (approved):
Looks good to me with just the one skipped word fixed.

pllim:
For some reason, I see that you suggested a change in email, but I cannot find it here. But I see where the mistake is, I'll push out a fix. Thanks for checking!

pllim:
What about now, @taldcroft ?

PR #10749, opened 2020-09-19
Raise more helpful error message in SkyCoord for mixed components

astrojuanlu:
### Description

This pull request fixes the outdated docstring for the `units` parameter of the `SkyCoord` initializer and provides a more helpful error message when an inconsistent mix of quantities and unit-less components is used (credit to @adrn for writing the actual message). This supersedes gh-10728, which was too strict.

Fix gh-10725.

### DIFF
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -222,9 +222,9 @@ class SkyCoord(ShapedLikeNDArray):
     unit : `~astropy.units.Unit`, string, or tuple of :class:`~astropy.units.Unit` or str, optional
         Units for supplied coordinate values.
         If only one unit is supplied then it applies to all values.
-        Notice that passing only one unit might lead to unit conversion errors
+        Note that passing only one unit might lead to unit conversion errors
         if the coordinate values are expected to have mixed physical meanings
-        (e.g. angles and distances).
+        (e.g., angles and distances).
     obstime : valid `~astropy.time.Time` initializer, optional
         Time(s) of observation.
     equinox : valid `~astropy.time.Time` initializer, optional
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -220,9 +220,11 @@ class SkyCoord(ShapedLikeNDArray):
         Type of coordinate frame this `SkyCoord` should represent. Defaults to
         to ICRS if not given or given as None.
     unit : `~astropy.units.Unit`, string, or tuple of :class:`~astropy.units.Unit` or str, optional
-        Units for supplied ``LON`` and ``LAT`` values, respectively.  If
-        only one unit is supplied then it applies to both ``LON`` and
-        ``LAT``.
+        Units for supplied coordinate values.
+        If only one unit is supplied then it applies to all values.
+        Notice that passing only one unit might lead to unit conversion errors
+        if the coordinate values are expected to have mixed physical meanings
+        (e.g. angles and distances).
     obstime : valid `~astropy.time.Time` initializer, optional
         Time(s) of observation.
     equinox : valid `~astropy.time.Time` initializer, optional
--- a/astropy/coordinates/sky_coordinate_parsers.py
+++ b/astropy/coordinates/sky_coordinate_parsers.py
@@ -591,16 +591,14 @@ def _get_representation_attrs(frame, units, kwargs):
             try:
                 valid_kwargs[frame_attr_name] = repr_attr_class(value, unit=unit)
             except u.UnitConversionError as err:
-                error_message = "Unit '{unit}' ({type}) could not be applied to '{name}'. ".format(
-                    unit=unit, type=unit.physical_type, name=frame_attr_name
-                )
-                error_message += (
+                error_message = (
+                    f"Unit '{unit}' ({unit.physical_type}) could not be applied to '{frame_attr_name}'. "
                     "This can occur when passing units for some coordinate components "
                     "when other components are specified as Quantity objects. "
                     "Either pass a list of units for all components (and unit-less coordinate data), "
                     "or pass Quantities for all components"
                 )
-                raise ValueError(error_message) from err
+                raise u.UnitConversionError(error_message) from err
 
     # also check the differentials.  They aren't included in the units keyword,
     # so we only look for the names.
--- a/astropy/coordinates/sky_coordinate_parsers.py
+++ b/astropy/coordinates/sky_coordinate_parsers.py
@@ -588,7 +588,19 @@ def _get_representation_attrs(frame, units, kwargs):
     for frame_attr_name, repr_attr_class, unit in zip(frame_attr_names, repr_attr_classes, units):
         value = kwargs.pop(frame_attr_name, None)
         if value is not None:
-            valid_kwargs[frame_attr_name] = repr_attr_class(value, unit=unit)
+            try:
+                valid_kwargs[frame_attr_name] = repr_attr_class(value, unit=unit)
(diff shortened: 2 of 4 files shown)


astropy/coordinates/sky_coordinate.py
```diff
@@ -220,9 +220,11 @@ class or the corresponding string alias.  The frame classes that are built in
         Type of coordinate frame this `SkyCoord` should represent. Defaults to
         to ICRS if not given or given as None.
     unit : `~astropy.units.Unit`, string, or tuple of :class:`~astropy.units.Unit` or str, optional
-        Units for supplied ``LON`` and ``LAT`` values, respectively.  If
-        only one unit is supplied then it applies to both ``LON`` and
-        ``LAT``.
+        Units for supplied coordinate values.
+        If only one unit is supplied then it applies to all values.
+        Notice that passing only one unit might lead to unit conversion errors
```
adrn, line 225:
```suggestion
        Note that passing only one unit might lead to unit conversion errors
```

astropy/coordinates/sky_coordinate.py
```diff
@@ -220,9 +220,11 @@ class or the corresponding string alias.  The frame classes that are built in
...
+        if the coordinate values are expected to have mixed physical meanings
+        (e.g. angles and distances).
```
adrn, line 227:
```suggestion
        (e.g., angles and distances).
```

adrn (approved):
Only minor nitpicks from me!

pllim:
#10728 was milestoned to 4.0.2 -- Either milestone should be changed, or change log has to be moved. Thanks!

astropy/coordinates/sky_coordinate_parsers.py
```diff
@@ -588,7 +588,19 @@ def _get_representation_attrs(frame, units, kwargs):
     for frame_attr_name, repr_attr_class, unit in zip(frame_attr_names, repr_attr_classes, units):
         value = kwargs.pop(frame_attr_name, None)
         if value is not None:
-            valid_kwargs[frame_attr_name] = repr_attr_class(value, unit=unit)
+            try:
+                valid_kwargs[frame_attr_name] = repr_attr_class(value, unit=unit)
+            except u.UnitConversionError as err:
+                error_message = "Unit '{unit}' ({type}) could not be applied to '{name}'. ".format(
```
mhvk, line 594: Just use an `f-string`.

astropy/coordinates/sky_coordinate_parsers.py
```diff
@@ -588,7 +588,19 @@ def _get_representation_attrs(frame, units, kwargs):
...
+                    unit=unit, type=unit.physical_type, name=frame_attr_name
+                )
+                error_message += (
+                    "This can occur when passing units for some coordinate components "
+                    "when other components are specified as Quantity objects. "
+                    "Either pass a list of units for all components (and unit-less coordinate data), "
+                    "or pass Quantities for all components"
+                )
+                raise ValueError(error_message) from err
```
mhvk, line 603: I'd stick with a `UnitConversionError` (which is a subclass of `ValueError`). You could also consider just adding to the error message itself, with
```
err.args += (f".....",)
raise err
```

astropy/coordinates/tests/test_sky_coord.py
```diff
@@ -1809,3 +1809,22 @@ class MultipleAliasesFrame(BaseCoordinateFrame):
...
+    (
+        {"rho": 1 * u.m, "phi": 1, "z": 1 * u.m, "unit": "deg", "representation_type": "cylindrical"},
+        "Unit 'deg' (angle) could not be applied to 'rho'",
+    ),
+])
+def test_passing_inconsistent_coordinates_and_units_raises_helpful_error(kwargs, error_message):
+    # https://github.com/astropy/astropy/issues/10725
+    with pytest.raises(ValueError) as excinfo:
+        SkyCoord(**kwargs)
+    assert (
```
mhvk, line 1828: It is nitpicking but avoid the blacken parenthesis on its own line - the general goal is to stick to the style of a given file. This through-out.

Though here you can avoid the whole assert by doing `with pytest.raises(ValueError, match=error_message):`

mhvk (commented):
Thanks!  Mostly stylistic nitpicks, except a suggestion to stick with the same exception. The latter also ensures we do not need a changelog entry, as it is only the error message which changes.

astropy/coordinates/sky_coordinate_parsers.py
astrojuanlu: Right, I am so used to write Python 3.4 code I suppose :sweat_smile: will do

astropy/coordinates/tests/test_sky_coord.py
astrojuanlu: This is not because of black, it's a leftover because `error_message` used to be a long string, and I forgot to change it.

Will use the `match=` parameter instead though, much nicer!

astropy/coordinates/sky_coordinate_parsers.py
astrojuanlu: Adding more elements to the `.args` tuple ends up being shown like this:

```
astropy.units.core.UnitConversionError: ("'pc' (length) and 'deg' (angle) are not convertible", "Unit 'deg' (angle) could not be applied to 'distance'. ", 'This can occur when passing units for some coordinate components when other components are specified as Quantity objects. Either pass a list of units for all components (and unit-less coordinate data), or pass Quantities for all components')
```

which I find a tad ugly. I think I will raise a `UnitConversionError` but leave the custom message.

mhvk: Sure, the `raise ... from err` you had is fine too!

astropy/coordinates/sky_coordinate_parsers.py
```diff
@@ -588,7 +588,17 @@ def _get_representation_attrs(frame, units, kwargs):
...
-            valid_kwargs[frame_attr_name] = repr_attr_class(value, unit=unit)
+            try:
+                valid_kwargs[frame_attr_name] = repr_attr_class(value, unit=unit)
+            except u.UnitConversionError as err:
+                error_message = (
+                    f"Unit '{unit}' ({unit.physical_type}) could not be applied to '{frame_attr_name}'. "
+                    "This can occur when passing units for some coordinate components "
+                    "when other components are specified as Quantity objects. "
+                    "Either pass a list of units for all components (and unit-less coordinate data), "
+                    "or pass Quantities for all components"
```
parejkoj, line 599: Period at the end of the final sentence?

parejkoj (approved):
One punctuation comment.

Thank you, the new docs are almost enough themselves, but the improved exception seals it.

astropy/coordinates/sky_coordinate_parsers.py
astrojuanlu: I just checked and the use of periods at the end of error messages in Astropy is rather inconsistent :sweat_smile: However, in this case there are two sentences, so I will add it

<sub>And leave no period at the end of this comment because I just want to see the world burn</sub>

astrojuanlu:
The coverage report says `90.30% (+-0.01%) compared to 9e02760`, so... I suppose the failure is not relevant.

astrojuanlu:
There are two approvals and I have addressed @mhvk comments and nitpicks, so I am merging :rocket:

astropy/coordinates/sky_coordinate_parsers.py
pllim: Haha! There was a discussion about adding or not adding period to errors/warnings at some point, but I cannot find it now. I wonder if @MSeifert04 remembers, if he is still around.

PR #10751, opened 2020-09-20
ScienceState states

nstarman:
## Description

ScienceStates are great. One shortcoming I've ran into is that they operate on a single-value paradigm. Workarounds are apparent in both ``galactocentric_frame_defaults`` (see #10624)  and ``solar_system_ephemeris``, which custom manage a '_references' and '_kernel' attribute, respectively.
This PR is to change ScienceStates to a multi-valued paradigm where the state (now "_state") is a dictionary. The name of the 'main' value can be set by an attribute on the ScienceState (see new ``galactocentric_frame_defaults`` implementation).

In addition, this PR also separates out the contextmanager returned by ``ScienceStates.set``, allows for custom contextmanagers to be used, and allows for the contextmanager to be used in `with` statements.

All these changes were introduced in  #10624, before being delayed for this PR.

## Examples

```python
class ExState(ScienceState):
    _state_value_name_ = "main"  # changes it from "value"
    _state = {"meta": "some more info"}
    _value = 2  # moved into `_state` as "main" by metaclass. Supported for backward compatibility.

ExState.main == ExState._value == 2
ExState.meta == "some more info"
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -113,6 +113,21 @@ astropy.utils
 
 - ``ShapedLikeNDArray`` has gained the capability to use numpy functions
   that broadcast, change shape, or index. [#10337]
+- ``ScienceState`` adopts multiple-attribute states, now stored in dict
+  ``_state``. [#10751]
+
+  + Items in ``_state`` can be accessed as attributes.
+  + ``ScienceState.validate`` returns the whole state, not just primary value
+  + Central state value (previously "_value") now determined by attribute
+    ``_state_value_name_``, defaulting to "value".
+
+- ``ScienceStateContext`` is the new contextmanager returned by
+  ``ScienceState.set`` [#10751]
+
+  + Now works in ``with`` statements.
+  + ``ScienceState`` controls which ``ScienceStateContext`` or subclass is
+    returned, using the ``_context_`` attribute.
+
 
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


nstarman:
One thing I have not done, but which would simplify a lot of code, is change `ScienceState.validate` to return the whole state, not just the 'central' value. This is a larger API change, so I wanted some discussion first.

nstarman:
Despite travis saying it passed, it is actually failing in the Python 3.7 + remote data check since it cannot serialize a ``_io.BufferedReader`` object. This arises from a deepcopy that happens in ``ScienceState.set``. This next push avoids deep copies by reusing the indexed read-only dict introduced in #10624 and using a try/except on validation errors. Now setting a ScienceState should never drop the old state if there is an error in ``validate``. 

The remaining problem is ``validate`` methods — like ``solar_system_ephemeris`` — that refer to the previous state, after it has been cleared in ``set``. This relate to a broader design issue. Since ``validate`` is the main overwritten method, I've been trying to make it as safe as possible — minimizing opportunities to unintentionally mutate the state. On the other hand, I've avoided making `_state` an IndexedMappingView (readonly by normal access but writable through ``._mapping``) so that  it could be modified. 
Choosing one paradigm over the other would make things simpler. I'm inclined towards the former: Make ``validate`` as easy  to write as it is currently and make `_state` an IndexedMappingView. Any method that wants to mutate the state would use ``ScienceState._state._mapping``. I think this will avoid the need to clear states.

adrn:
Overall, I think I'm in favor of the idea here! I plan to take a look ASAP, but could be a few days. Thanks for thinking about this, @nstarman!

Issue #9619, opened 2019-11-18
TST: Investigate ResourceWarning in tests (currently ignored)


pllim:
### Description

Un-ignore the following warnings and investigate whether they can be mitigated without being ignored whole-sale.

https://github.com/astropy/astropy/blob/6864219167a402cb2e574c45ff91e3bb7a23861f/setup.cfg#L109-L111

### Additional context

This is a follow-up of #7928 

Also see #8673

pllim:
These 2 still a problem.

```yaml
 ignore:unclosed <socket:ResourceWarning 
 ignore:unclosed <ssl.SSLSocket:ResourceWarning 
```

I think mostly `utils.data` and `samp` tests. I wonder if we should localize the ignore. I took a peek but couldn't see any trivial fix.

PR #10754, opened 2020-09-21
TST: Accept socket warning for now in test_data_noastropy_fallback

pllim:
### Description

This pull request is to add in logic to account for 2 extra warnings sometimes in `test_data_noastropy_fallback` so it won't fail.

Is this acceptable for now, @aarchiba ?

Also see #9619 and #10686

Follow up of  #10437 and #10671

### DIFF
--- a/astropy/utils/tests/test_data.py
+++ b/astropy/utils/tests/test_data.py
@@ -1054,10 +1054,17 @@ def test_data_noastropy_fallback(monkeypatch):
 
     with pytest.warns(CacheMissingWarning) as warning_lines:
         fnout = download_file(TESTURL, cache=True)
-    assert len(warning_lines) == 2, os.linesep.join(
-        [str(w.message) for w in warning_lines])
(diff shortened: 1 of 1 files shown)


eteq:
@pllim - can you clarify if this actually appears in some cases "live", or is it only appearing in the test suite?

Put another way: is this a 4.1 blocker, or more of a test suite cleanup? (I've gotten lost in the cross-linkings to other issues).

pllim:
I have seen the socket warning "live" but I couldn't trace it down. Either way, not a blocker for release. Feel free to remilestone.

eteq (approved):
I'm good with this as it stands but I wonder if it might be safer to not assume the socket warnings are between the two other warnings?  I *think* I've seen it on CI in a different order...

pllim:
Hmm... if the warning order is non-deterministic, I'll have to rewrite the logic. Don't merge yet.

pllim:
Getting unrelated errors. I will rebase, just in case.

pllim:
Thanks for the help and the review, @eteq !

PR #10757, opened 2020-09-22
TST: Skip flaky utils test on Travis

pllim:
### Description

This pull request is to skip a flaky `utils` test on Travis CI. This is similar to #10700 , as @aarchiba added this test based on that other test, which is now skipped already.


mhvk (approved):
Makes sense.

PR #10758, opened 2020-09-22
add wcs slices as argument to functions in wcsaxes 

cshanahan1:
This fix is needed for https://github.com/aplpy/aplpy/pull/457

The contour plot function in aplpy doesn't allow >2D WCS, this fixes that by allowing WCS slices to be passed in to some of the functions in wcsaxes.

### DIFF
--- a/astropy/visualization/wcsaxes/core.py
+++ b/astropy/visualization/wcsaxes/core.py
@@ -536,7 +536,7 @@ class WCSAxes(Axes):
 
         return coords
 
-    def get_transform(self, frame):
+    def get_transform(self, frame, slices=None):
         """
         Return a transform from the specified frame to display coordinates.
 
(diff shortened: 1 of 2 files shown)


pllim:
I am leaning towards 4.2 until it is absolutely critical this needs to be backported. Need change log and test. Thanks!

astropy/visualization/wcsaxes/core.py
```diff
@@ -567,16 +567,16 @@ def get_transform(self, frame):
...
-    def _get_transform_no_transdata(self, frame):
+    def _get_transform_no_transdata(self, frame, slices=None):
         """
         Return a transform from data to the specified frame
         """
 
         if isinstance(frame, WCS):
 
-            transform, coord_meta = transform_coord_meta_from_wcs(frame, self.frame_class)
+            transform, coord_meta = transform_coord_meta_from_wcs(frame, self.frame_class, slices)
```
astrofrog, line 579: This should be ``slices=slices`` to be safe

astrofrog (changes requested):
This makes sense, thanks! Could you add a changelog entry in the 4.2 section?

Ideally this would require a test, but let me know if you don't have time to add one.

Cadair (changes requested):
I am not sure this change is needed. I don't think I see the advantage of adding new kwargs over applying the slice directly to the WCS object before it is passed to the method. (See my comment on the aplpy PR).

Let me know if I am missing something :)

eteq:
Since this hasn't made the feature freeze cutoff and is not a bug/docfix I'm re-milestoning to 5.0.

But @cshanahan1 might want to have a look at https://github.com/astropy/astropy/pull/10758#pullrequestreview-514522521 since it appears we have two votes, one for and one against... (which means @cshanahan1 might be able to be vote 3 :wink:)

PR #10759, opened 2020-09-22
Update wcs fit

yoachim:
### Description

When fitting WCS solutions, updated the code so longitude differences near the poles are scaled properly. Removed some lines that seemed to be doing angle wrapping corrections on variables in pixel coordinates.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -491,11 +491,6 @@ astropy.visualization
 astropy.wcs
 ^^^^^^^^^^^
 
-- Add .upper() to ctype or ctype names to wcsapi/fitwcs.py to mitigate bugs from
-  unintended lower/upper case issues [#10557]
-- Added bounds to ``fit_wcs_from_points`` to ensure CRPIX is on
-  input image. [#10346]
-
 - Objective functions called by ``astropy.wcs.fit_wcs_from_points`` were
   treating longitude and latitude distances equally. Now longitude scaled properly. [#10759]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -491,6 +491,15 @@ astropy.visualization
 astropy.wcs
 ^^^^^^^^^^^
 
+- Add .upper() to ctype or ctype names to wcsapi/fitwcs.py to mitigate bugs from
+  unintended lower/upper case issues [#10557]
+- Added bounds to ``fit_wcs_from_points`` to ensure CRPIX is on
+  input image. [#10346]
+
+- Objective functions called by ``astropy.wcs.fit_wcs_from_points`` were
+  treating longitude and latitude distances equally. Now longitude scaled properly. [#10759]
+
+
 Other Changes and Additions
 ---------------------------
 
--- a/astropy/wcs/tests/test_utils.py
+++ b/astropy/wcs/tests/test_utils.py
@@ -1112,11 +1112,29 @@ BP_1_1  =  -1.711876336719E-05
 A_DMAX  =    44.72893589844534
 B_DMAX  =    44.62692873032506
 """
+    # A known header that failed before
+    header_str_prob = """
+NAXIS   =                    2 / number of array dimensions
+WCSAXES =                    2 / Number of coordinate axes
+CRPIX1  =               1024.5 / Pixel coordinate of reference point
+CRPIX2  =               1024.5 / Pixel coordinate of reference point
+CD1_1   = -1.7445934400771E-05 / Coordinate transformation matrix element
+CD1_2   = -4.9826985362578E-08 / Coordinate transformation matrix element
+CD2_1   = -5.0068838822312E-08 / Coordinate transformation matrix element
+CD2_2   =  1.7530614610951E-05 / Coordinate transformation matrix element
+CTYPE1  = 'RA---TAN'           / Right ascension, gnomonic projection
+CTYPE2  = 'DEC--TAN'           / Declination, gnomonic projection
+CRVAL1  =      5.8689341666667 / [deg] Coordinate value at reference point
+CRVAL2  =     -71.995508583333 / [deg] Coordinate value at reference point
+"""
+
     header_linear = fits.Header.fromstring(header_str_linear, sep='\n')
     header_sip = fits.Header.fromstring(header_str_sip, sep='\n')
+    header_prob = fits.Header.fromstring(header_str_prob, sep='\n')
 
     true_wcs_linear = WCS(header_linear, relax=True)
     true_wcs_sip = WCS(header_sip, relax=True)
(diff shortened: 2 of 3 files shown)


pllim:
Thanks! Is this bug fix or enhancement? Does this need a new test case?

When someone sets the milestone, will also need a change log, maybe, but not yet.

yoachim:
I guess we can call it an enhancement. I think the current unit tests have it pretty well covered.

ceb8 (approved):
Looks good to me!

pllim:
If it is enhancement, then perhaps a change log under 4.2. 😄

astropy/wcs/utils.py
```diff
@@ -891,9 +891,13 @@ def _linear_wcs_fit(params, lon, lat, x, y, w_obj):
     w_obj.wcs.crpix = crpix
     lon2, lat2 = w_obj.wcs_pix2world(x, y, 0)
 
-    resids = np.concatenate((lon-lon2, lat-lat2))
-    resids[resids > 180] = 360 - resids[resids > 180]
-    resids[resids < -180] = 360	+ resids[resids < -180]
+    lat_resids = lat-lat2
+    lon_resids = lon-lon2
```
mcara, line 895: PEP8:
```suggestion
    lat_resids = lat - lat2
    lon_resids = lon - lon2
```

astropy/wcs/utils.py
```diff
@@ -891,9 +891,13 @@ def _linear_wcs_fit(params, lon, lat, x, y, w_obj):
...
+    # In case the longitude has wrapped around
+    lon_resids[lon_resids > 180] = 360 - lon_resids[lon_resids > 180]
+    lon_resids[lon_resids < -180] = 360 + lon_resids[lon_resids < -180]
```
mcara, line 898:
```suggestion
    lon_resids = np.mod(lon_resids - 180.0, 360.0) - 180.0
```

astropy/wcs/utils.py
```diff
@@ -891,9 +891,13 @@ def _linear_wcs_fit(params, lon, lat, x, y, w_obj):
...
+
+    resids = np.concatenate((lon_resids*np.cos(np.radians(lat)), lat_resids))
```
mcara, line 900: PEP8:
```suggestion
    resids = np.concatenate((lon_resids * np.cos(np.radians(lat)), lat_resids))
```

astropy/wcs/utils.py
mcara: Shouldn't the same be applied to `lat_resids`?

mcara (changes requested):
This PR actually is a bug fix and it fixes a bug that I was struggling with when I tried to use this function (before reviewing this PR). My comments are very minor but I would like to know whether same wrapping transformation should be applied to `lat_resids`.

yoachim:
> This PR actually is a bug fix and it fixes a bug that I was struggling with when I tried to use this function (before reviewing this PR). My comments are very minor but I would like to know whether same wrapping transformation should be applied to `lat_resids`.

We don't need to wrap the latitude. For longitude coordinates, 355 is 10 degrees away from 5. There is no analogous discontinuity in latitude coordinates since they run from -90 to 90.

astropy/wcs/utils.py
yoachim: This assumes the sign of the residual isn't relevant, and since we're passing it to a least squares, presumably it gets squared and doesn't matter.

mcara (approved):

mcara:
Thank you @yoachim

mcara:
One more thing though. Could you, please, add a test that covers what was previously failing?

mcara:
I was able to get this code to fail with the following "test" WCS (used to produce RA, DEC from some x, y pairs):
```
WCSAXES =                    2 / Number of coordinate axes
CRPIX1  =               1024.5 / Pixel coordinate of reference point
CRPIX2  =               1024.5 / Pixel coordinate of reference point
CD1_1   = -1.7445934400771E-05 / Coordinate transformation matrix element
CD1_2   = -4.9826985362578E-08 / Coordinate transformation matrix element
CD2_1   = -5.0068838822312E-08 / Coordinate transformation matrix element
CD2_2   =  1.7530614610951E-05 / Coordinate transformation matrix element
CTYPE1  = 'RA---TAN'           / Right ascension, gnomonic projection
CTYPE2  = 'DEC--TAN'           / Declination, gnomonic projection
CRVAL1  =      5.8689341666667 / [deg] Coordinate value at reference point
CRVAL2  =     -71.995508583333 / [deg] Coordinate value at reference point
```

mcara:
@yoachim Thank you so much!

nden:
@yoachim One last request : could you move the changelog from 4.2 to 4.1.1, rebase the PR to resolve the conflicts and squash the commits. Thanks!

astropy/wcs/tests/test_utils.py
```diff
@@ -1235,6 +1288,8 @@ def test_fit_wcs_from_points():
     assert (fit_wcs.wcs.crpix.astype(int) == [1100, 1005]).all()
     assert fit_wcs.pixel_shape == (200, 10)
 
+    
+
```
mcara, line 1292: Please remove these extra blank lines

mcara:
@yoachim Please rebase the PR to resolve the conflicts and squash the commits. Thanks!

pllim:
@yoachim , the merge commit will not do. Please do something like the following:

```
git fetch upstream master
git rebase upstream/master
```

Thanks!

nden:
I can rebase and push to this branch.

yoachim:
oops, might have been simultaneously rebasing there. I'll let you take it away

nden:
@yoachim Sorry about that, looks OK now. I'll leave it to @mcara to merge.

CHANGES.rst
```diff
@@ -491,6 +491,15 @@ astropy.visualization
 astropy.wcs
 ^^^^^^^^^^^
 
+- Add .upper() to ctype or ctype names to wcsapi/fitwcs.py to mitigate bugs from
+  unintended lower/upper case issues [#10557]
+- Added bounds to ``fit_wcs_from_points`` to ensure CRPIX is on
+  input image. [#10346]
+
```
mcara, line 498: @nden This somehow got left from the merge commit or maybe in some other way. The resulting file will have two entries for the same bug fix in `wcsapi`.

```suggestion
```

mcara:
I do not understand the failure of Travis CI for Python 3.8 build but it does not look to be related to this PR. Also, docs build seems to be stuck and I think it was was passing in previous commits. So, I think this PR can be merged.

mcara:
Thank you @yoachim , @nden

PR #10763, opened 2020-09-23
Update link to USNO circular 179

adrn:
This updates the (currently broken) link to the USNO Circular 179 from the `astropy.coordinates` documentation. It seems like the USNO astronomy special publications site is just...[gone](aa.usno.navy.mil/publications)?! The arxiv link should be more stable.

_doh - I forgot to skip ci ..._

### DIFF
--- a/docs/coordinates/index.rst
+++ b/docs/coordinates/index.rst
@@ -501,7 +501,7 @@ See Also
 Some references that are particularly useful in understanding subtleties of the
 coordinate systems implemented here include:
 
-* `USNO Circular 179 <https://aa.usno.navy.mil/publications/docs/Circular_179.php>`_
+* `USNO Circular 179 <https://arxiv.org/abs/astro-ph/0602086>`_
     A useful guide to the IAU 2000/2003 work surrounding ICRS/IERS/CIRS and
     related problems in precision coordinate system work.
 * `Standards Of Fundamental Astronomy <http://www.iausofa.org/>`_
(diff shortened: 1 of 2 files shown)


pllim:
Didn't they take down their server for "maintenance" and then somehow the maintenance never stopped? Anyways, replacing it is definitely better. Thanks!

docs/coordinates/index.rst
```diff
@@ -521,7 +521,6 @@ coordinate systems implemented here include:
 
 Built-in Frame Classes
 ======================
-
```
pllim, line 524: Why removing the blank line here? I guess if RTD doesn't complain, it is fine.

pllim (approved):

pllim:
p.s. I cancelled the CI but we still need to wait for RTD check.

docs/coordinates/index.rst
adrn: Uh, weird, not sure why this happened - my text editor is getting more chaotic!

adrn:
> Didn't they take down their server for "maintenance" and then somehow the maintenance never stopped?

That does ring a bell!

pllim:
astropy/astropy#9453 -- Looks like they still maintaining...

pllim:
`[docs only]` doesn't work anymore because it is in `ci-helpers` and `tox` doesn't use `ci-helpers` anymore. So I cancelled the CI again. Will just merge now that the blank line is gone. Thanks!

adrn:
Oh, oops! Ok thanks!

PR #10766, opened 2020-09-24
Ignore a toolz warning triggered by a dask import

Cadair:
see pytoolz/toolz#500


pllim (approved):
Thanks! Assuming they fix it soon enough, we won't have to backport.

pllim:
Remote data job failure is unrelated.

PR #10768, opened 2020-09-24
Fix handling of 1d table elements with single entry in fits

### LINKED ISSUE #10765: Table reading/writing to/from fits does not support writing 1d arrays with a single element
### Description

Astropy `Table` writing and reading to/from fits looses the information about dimensionality,
if the cell data written is a 1d array with a single entry.
Then, on reading back the data, it will be read as a number, so ndim=0, instead of ndim=1.

### Expected behavior

Preserve this informatino

### Actual behavior

Wrong dimensionality

### Steps to Reproduce

```python

❯ python
Python 3.8.5 (default, Sep  5 2020, 10:50:12) 
[GCC 10.2.0] on linux
>>> t = Table({'x': [[1], [2]]})
>>> t.write('/tmp/test.fits', overwrite=True)
>>> t['x'].shape
(2, 1)
>>> Table.read('/tmp/test.fits')['x'].shape
(2,)

```

The error seems to be in both reading and writing.

The writing part fails to add the `TDIM` keyword and the reading part has a regex that does not match a tuple with just one entry.

### System Details

```
Linux-5.8.10-1-MANJARO-x86_64-with-glibc2.2.5
Python 3.8.5 (default, Sep  5 2020, 10:50:12) 
[GCC 10.2.0]
Numpy 1.19.2
astropy 4.0.1.post1
Scipy 1.5.2
Matplotlib 3.3.2
```

MaxNoe:
### Description

Add test for and fix writing and reading of fits columns that contain single values in 1d columns
This pull request is to address #10765 

Fixes #10765

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,8 +47,6 @@ astropy.io.ascii
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
-- Fix handling of 1-dimensional arrays with single elements in BinaryTableHDUs [#10768]
-
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
@@ -1033,6 +1031,8 @@ astropy.io.fits
   [#10640]
 
 - Fix compilation of cfitsio with Xcode 12. [#10772]
+
+- Fix handling of 1-dimensional arrays with a single element in ``BinTableHDU`` [#10768]
 
 astropy.io.misc
 ^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,8 @@ astropy.io.ascii
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
+- Fix handling of 1-dimensional arrays with single elements in BinaryTableHDUs [#10768]
+
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
--- a/astropy/io/fits/tests/test_table.py
+++ b/astropy/io/fits/tests/test_table.py
@@ -1772,7 +1772,7 @@ class TestTableFunctions(FitsTestCase):
 
     def test_oned_array_single_element(self):
         # a table with rows that are 1d arrays of a single value
-        data = np.array([1, 2]).view([('x', 'int64', (1, ))])
+        data = np.array([(1, ), (2, )], dtype=([('x', 'i4', (1, ))]))
         thdu = fits.BinTableHDU.from_columns(data)
 
         thdu.writeto(self.temp('onedtable.fits'))
--- a/astropy/io/fits/tests/test_table.py
+++ b/astropy/io/fits/tests/test_table.py
@@ -1768,6 +1768,19 @@ class TestTableFunctions(FitsTestCase):
         assert t.field(1).dtype.str[-1] == '5'
         assert t.field(1).shape == (3, 4, 3)
 
+    def test_oned_array_single_element(self):
+        # a table with rows that are 1d arrays of a single value
+        data = np.array([1, 2]).view([('x', 'int64', (1, ))])
+        thdu = fits.BinTableHDU.from_columns(data)
+
+        thdu.writeto(self.temp('onedtable.fits'))
+
+        with fits.open(self.temp('onedtable.fits')) as hdul:
+            thdu = hdul[1]
+
+            c = thdu.data.field(0)
+            assert c.shape == (2, 1)
+
     def test_bin_table_init_from_string_array_column(self):
         """
         Tests two ways of creating a new `BinTableHDU` from a column of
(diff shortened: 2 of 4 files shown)


MaxNoe:
regex101 links for the new and the old regex:

new: https://regex101.com/r/jlQzyT/2
old: https://regex101.com/r/YseaAK/1

CHANGES.rst
```diff
@@ -47,6 +47,8 @@ astropy.io.ascii
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
+- Fix handling of 1-dimensional arrays with single elements in BinaryTableHDUs [#10768]
```
dhomeier, line 50:
```suggestion
- Fix handling of 1-dimensional arrays with single elements in ``BinTableHDU``s [#10768]
```
Suggest to use the class name; otherwise looks good once it's moved to the new milestone.

dhomeier (changes requested):

CHANGES.rst
MaxNoe: Both done

MaxNoe:
@dhomeier The build seems to fail due to an unrelated problem to my changes. Any idea what happened there?

dhomeier (approved):

dhomeier:
> @dhomeier The build seems to fail due to an unrelated problem to my changes. Any idea what happened there?

This appears to be caused by a simultaneous pytest update on both CI platforms, just reported in #10773.

Otherwise looks good to me; with only changelog changes could probably even have gone in with a `[ci skip]`!

MaxNoe:
> Otherwise looks good to me; with only changelog changes could probably even have gone in with a [ci skip]!

Yes, sorry forgot about that

MaxNoe:
Thanks for getting this into the next  point release

astropy/io/fits/tests/test_table.py
```diff
@@ -545,7 +545,7 @@ def test_appending_a_column(self):
         for i in range(len(t1[1].columns)):
             hdu.data.field(i)[nrows1:] = t2[1].data.field(i)
 
-        hdu.writeto(self.temp('newtable.fits'))
+        hdu.writeto(self.temp('newtable.fits'), overwrite=True,)
```
saimn, line 548: I guess you needed this for debugging, but could you remove it now ?

astropy/io/fits/tests/test_table.py
```diff
@@ -1702,12 +1702,14 @@ def test_multi_dimensional_columns(self):
...
-        thdu.header['TDIM1'] = '(2,3)'
-        thdu.header['TDIM2'] = '(4,2)'
 
         thdu.writeto(self.temp('newtable.fits'))
 
+        with fits.open(self.temp('newtable.fits'), mode='update') as hdul:
+            # Modify the TDIM fields to my own specification
+            hdul[1].header['TDIM1'] = '(2,3)'
+            hdul[1].header['TDIM2'] = '(4,2)'
+
```
saimn, line 1712: Why is this change needed ?

astropy/io/fits/tests/test_table.py
```diff
@@ -1753,6 +1755,19 @@ def test_multi_dimensional_columns(self):
...
+        thdu = fits.BinTableHDU.from_columns(data)
+
+        thdu.writeto(self.temp('onedtable.fits'))
+
+        with fits.open(self.temp('onedtable.fits')) as hdul:
+            thdu = hdul[1]
+
+            c = thdu.data.field(0)
+            assert c.shape == (2, 1)
+
```
saimn, line 1770: Could you also add a check for `TDIM` ?

saimn: Also it could be useful to add a test with `astropy.table.Table` (in `test_connect.py`) to make sure the use case of your original issue is fixed.

saimn (changes requested):
Thanks for the fix @MaxNoe, and for the regex101 links ;)
Just a few minor comments, otherwise looks good !

astropy/io/fits/tests/test_table.py
MaxNoe: yes

astropy/io/fits/tests/test_table.py
MaxNoe: At least when running the tests locally, it seemed that setting these headers *before* writing the file was overwritten by the actual dimensions of the data.

So I had to change to modify them after the file was initially written.

MaxNoe: This did not trigger before, as TDIM was only added `if len(shape) > 1`, which is no longer the case.

astropy/io/fits/tests/test_table.py
MaxNoe: done

MaxNoe:
@saimn I think I addressed all your comments, please have a look again

MaxNoe:
@dhomeier The tests are passing again

saimn:
Looks good now, thanks @MaxNoe !

PR #10772, opened 2020-09-26
Fix compilation of cfitsio with Xcode 12

### LINKED ISSUE #10770: Astropy does not build on MacOSX with Xcode 12.0.0
### Description

When I try to build astropy on my Mac (Catalina + Xcode 12.0) it fails with the following error:

```
  cextern/cfitsio/lib/group.c:5664:8: error: implicit declaration of function 'getcwd' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    if (!getcwd(buff,FLEN_FILENAME))
         ^
  1 warning and 1 error generated.
  error: command 'gcc' failed with exit status 1

(astropy) ➜  astropy-tmp git:(master) gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 12.0.0 (clang-1200.0.32.2)
Target: x86_64-apple-darwin19.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Xcode 12.0 was just released Sep. 16 and I'm guessing that it was automatically updated on my Mac since I did not request an update. Astropy is apparently not alone, e.g. https://gitlab.com/graphviz/graphviz/-/issues/1826.

There is a long slack thread with helpful inputs from @saimn, @dhomeier and @manodeep here: https://astropy.slack.com/archives/C067V74GK/p1600984862007700

A workaround is:
```
CFLAGS=-Wno-error=implicit-function-declaration pip install -e .
```

### Expected behavior

Astropy builds and runs tests.

### Actual behavior

Build failed as shown.

### Steps to Reproduce

1. Checkout astropy at master, `git clean -fxd`
2. `tox -e test` or `pip install -e .`

### System Details

```
Darwin-19.5.0-x86_64-i386-64bit
Python 3.7.7 (default, Mar 26 2020, 10:32:53) 
Numpy 1.18.1
astropy 4.2.dev715+gfaccb8b41
Scipy 1.4.1
Matplotlib 3.1.3
```

saimn:
Fix compilation of cfitsio with Xcode 12

Fix #10770: Xcode 12 defines `-Werror=implicit-function-declaration`
flag by default which breaks the compilation of cfitsio. To fix this
we could shut off this flag but it's better to define `HAVE_UNISTD_H`
to really fix the warning.

### DIFF
--- a/astropy/io/fits/setup_package.py
+++ b/astropy/io/fits/setup_package.py
@@ -42,6 +42,8 @@ def _get_compression_extension():
                 '-Wno-declaration-after-statement'
             ])
 
+            cfg['define_macros'].append(('HAVE_UNISTD_H', None))
+
             if not debug:
                 # these switches are to silence warnings from compiling CFITSIO
                 # For full silencing, some are added that only are used in
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1027,6 +1027,8 @@ astropy.io.fits
 - Fix repr for commentary cards and strip spaces for commentary keywords.
   [#10640]
 
+- Fix compilation of cfitsio with Xcode 12. [#10772]
+
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 


taldcroft (approved):
Works for me!  :tada:

astropy/io/fits/setup_package.py
```diff
@@ -42,6 +42,8 @@ def _get_compression_extension():
                 '-Wno-declaration-after-statement'
             ])
 
+            cfg['define_macros'].append(('HAVE_UNISTD_H', None))
```
pllim, line 45: I thought this setting is supposed to be for `darwin` only, but the placement here seems to imply otherwise. Can you please clarify?

dhomeier: The conditional is
[`#if defined(unix) || defined(__unix__)  || defined(__unix) || defined(HAVE_UNISTD_H)`](https://github.com/astropy/astropy/blob/master/cextern/cfitsio/lib/group.c#L27)
so on Linux this should be inconsequential, but not sure about Windows – is the windows 1809 test on Travis running with `ASTROPY_USE_SYSTEM_CFITSIO=0`?

saimn: Yeah, for Linux this should make no difference, and for Windows there is another include:

https://github.com/astropy/astropy/blob/392d2213475650657d82886b708150355ba01b96/cextern/cfitsio/lib/group.c#L23-L29

dhomeier: Ah, I had missed the comment that `direct.h` defines `getcwd` as well.
I was rather concerned if it would try to include a `unistd.h` that does not exist on Windows; seeing now this is actually explicitly disabled:
https://github.com/astropy/astropy/blob/392d2213475650657d82886b708150355ba01b96/astropy/io/fits/setup_package.py#L29-L39

dhomeier (approved):

saimn:
Ok, let's merge then, thanks for the reviews.

astropy/io/fits/setup_package.py
pllim: Thanks for the clarifications!

PR #10774, opened 2020-09-27
Never use setuptools_scm except when importing astropy from a git repository

### LINKED ISSUE #10760: [4.1RC2] Astropy module may return the wrong version
### Description

When I have a `PKG-INFO` file in the current working dir, astropy shows the version number from this file instead of its hown one. This is something what I observed while packaging the new "photutils" version, where the wrong path is taken during pytest.

### Expected behavior

Astropy should show its own version number under all circumstances

### Actual behavior

The version number is wrong if a `PKG-INFO` file is present.

### Steps to Reproduce

Shell script:
```shell
$ python3 -c "import astropy; print(astropy.__version__)"
4.1-rc2
$ echo Version: 1.2.3 > PKG-INFO
$ python3 -c "import astropy; print(astropy.__version__)"
1.2.3
```

### System Details

Linux-5.8.0-1-amd64-x86_64-with-glibc2.29
Python 3.8.5 (default, Aug  2 2020, 15:09:07) 
Numpy 1.19.1
astropy 1.2.3   :grin: :grin: :grin:
Scipy 1.5.2
Matplotlib 3.3.2

I didn't observe this with 4.0.1.post1.

embray:
This adds a special package `astropy._dev` which is never installed, and only exists when developing Astropy from within the source repo.  It is ommitted from distributions such as tarballs and wheels.

`astropy/version.py` becomes a file checked into the repo, and the file generated by `setuptools_scm` is now `astropy/_version.py`, and is greatly simplified.

`astropy/version.py` will use the version from `astropy._dev._scm_version` if it exists (generated at runtime by `setuptools_scm`).  Otherwise it will use the static version from `astropy/_version.py`.

Resolves #10760, though IIUC if some version of this approach is accepted the package-template will need to be updated as well.


olebole (approved):
I can confirm that this solves my problem #10760, and I like the solution, because it no longer tries to import setuptools_scm in the installed package.
Thank you very much @embray!

Cadair:
I really like this, it's easy to understand and will clearly work. This would need porting to both the astropy and openastronomy package templates.

Cadair (approved):

mhvk:
Looks nice. Just to be sure, does it need any update for the release instructions?

Cadair:
Shouldn't do, all the files controlling the packaging have been updated.

pllim:
I didn't expect the fix to be so complicated. Is our way of versioning simply incompatible with how `setuptools_scm` work? Should we rethink the way we keep track of dev revision?

Cadair:
I don't think this is very complex. But yes, setuptools_scm explicitly doesn't support this, but some people really wanted it when we were doing the APE 17 stuff.

astrofrog (approved):
Very nice solution and I like that is simplifies the ``setup.py``

I don't think this means we shouldn't use setuptools_scm - I don't think there's going to be any way to avoid writing a little custom code to get the behavior developers here want (auto-updating version numbers when in develop mode).

astropy/_dev/_scm_version.py
```diff
@@ -0,0 +1,8 @@
+# Try to use setuptools_scm to get the current version; this is only used
```
Cadair, line 1: I had a thought about this overnight. Is there a reason why this needs to be a submodule rather than a single file in the root of the repo? It feels like that would be less obvious and prone to confusing people?

astrofrog: I was wondering that too and thought that if it’s not in the package then there is a non zero chance that file could exist one level up in an installed version of astropy?

Cadair: did you mean to say "if it's in the package"? otherwise I am not sure I understand what you are saying.

astrofrog: I meant that if ``_scm_version.py`` is at the root of the repo, and we check for ``../_scm_version.py`` in ``version.py``, then if there happens to be a ``_scm_version.py`` file in ``site-packages``, one could run into issues. Having the file be inside the package somewhat guarantees there won't be collisions with other files in the site-packages directory?

mhvk: Maybe have it as `./_scm_version.py`, i.e., same directory as `version.py`? It would seem a bit cleaner.

Cadair: > if there happens to be a _scm_version.py file in site-packages

oh yeah you are correct this will trigger the original

Cadair: Having it in the astropy namespace ensures it's always the astropy one and not another one getting picked up via relative import. That being said, it could just be a file in the astropy folder not a folder?

astrofrog: Yes I agree, maybe no need for the _dev namespace

embray:
> I don't think this means we shouldn't use setuptools_scm - I don't think there's going to be any way to avoid writing a little custom code to get the behavior developers here want (auto-updating version numbers when in develop mode).

Indeed, as discussed I think in #10760 there have been discussions on the setuptools_scm project on ways to better handle this problem, but no one official solution has emerged.  This solution is actually, I think, novel.  While working on this I tried some of my other previous ideas, and while they do work, the idea came to me of having some files that are only ever present when working out of the repository, but are never installed.  This seemed like the most "explicit" approach to me since it's explicitly excluding functionality from installed versions of the package that are only relevant for development.

embray:
Rebased onto (hopefully non-broken) main branch and squashed fixups.

astropy/_dev/_scm_version.py
embray: Originally I had it in `astropy/_scm_version.py` but unless you can tell me a way to exclude a single file from wheels (see https://github.com/pypa/setuptools/issues/511) this has to be an entire sub-package.  In any case there might be other uses found for it.

One thing I should do is rename it from `astropy._dev._scm_version` to `astropy._dev.scm_version`.  Since the subpackage is already "private" there's no reason for the module name to be as well.

embray:
Should I make a PR to the package-template before merging this?

astrofrog:
@embray no, fixing the package template and coordinated packages can be done after this is merged IMHO

astropy/_dev/_scm_version.py
mhvk: Ah, that's an oddity!  So, seems fine to stick to what you have then!

astrofrog (approved):

mhvk:
Hmm, with latest master, if I now do `tox -e test --develop`, I get a failure - maybe expected as I should now take care of installing it myself, but thought I would note it anyway....

```
=============================================== ERRORS ================================================
____________________________ ERROR collecting astropy/_dev/scm_version.py _____________________________
../../astropy/_dev/scm_version.py:5: in <module>
    from setuptools_scm import get_version
E   ModuleNotFoundError: No module named 'setuptools_scm'

../../astropy/_dev/scm_version.py:8: in <module>
    raise ImportError('setuptools_scm broken or not installed')
E   ImportError: setuptools_scm broken or not installed
======================================= short test summary info =======================================
ERROR ../../astropy/_dev/scm_version.py - ImportError: setuptools_scm broken or not installed
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================== 60 skipped, 1 error in 223.11s (0:03:43) ===============================
ERROR: InvocationError for command /home/mhvk/data/mhvk/packages/astropy/.tox/test/bin/pytest --pyargs astropy /home/mhvk/data/mhvk/packages/astropy/docs (exited with code 2)
_______________________________________________ summary _______________________________________________
ERROR:   test: commands failed
```

astrofrog:
@mhvk we need to add a line in the setup.cfg file under the pytest section to exclude _dev from the test collection - think it is called ignore glob or norecursedir (can check later)

PR #10778, opened 2020-09-28
Changes for pylance/pyright correctness

taldcroft:
### Description

This pull request is to fix warnings that show up in static code analysis with `pyright` via the `Pylance` VS code plugin. Most of them are "possibly undefined variables" that are probably OK, but there were some actual problems in logic that are good to fix.

A few of the changes impact the code logic. I've put in some comments to help navigate through these.

### DIFF
--- a/astropy/io/ascii/daophot.py
+++ b/astropy/io/ascii/daophot.py
@@ -238,8 +238,8 @@ class DaophotInputter(core.ContinuationLinesInputter):
         # this case we have to figure out how many apertures there are based on
         # the file structure.
 
-        comment, special, cont = zip(*(self.re_multiline.search(l).groups()
-                                       for l in lines[:depth]))
+        comment, special, cont = zip(*(self.re_multiline.search(line).groups()
+                                       for line in lines[:depth]))
 
         # Find first non-comment line
         data_start = first_false_index(comment)
--- a/astropy/table/operations.py
+++ b/astropy/table/operations.py
@@ -1040,10 +1040,12 @@ def _apply_join_funcs(left, right, keys, join_funcs):
     right = right.copy(copy_data=False)
     for key, join_func in join_funcs.items():
         ids1, ids2 = join_func(left[key], right[key])
-        for ii in itertools.count(1):
-            id_key = key + '_' * ii + 'id'
-            if id_key not in left.columns and id_key not in right.columns:
-                break
+        # Define a unique id_key name, and keep adding underscores until we have
+        # a name not yet present.
+        id_key = key + '_id'
+        while id_key in left.columns or id_key in right.columns:
+            id_key = id_key[:-2] + '_id'
+
         keys = tuple(id_key if orig_key == key else orig_key for orig_key in keys)
         left.add_column(ids1, index=0, name=id_key)  # [id_key] = ids1
         right.add_column(ids2, index=0, name=id_key)  # [id_key] = ids2
--- a/astropy/table/serialize.py
+++ b/astropy/table/serialize.py
@@ -329,14 +329,18 @@ def _construct_mixin_from_columns(new_name, obj_attrs, out):
     # example would be a formatted time object that would have (e.g.)
     # "time_col" and "value", respectively.
     for name, data_attr in data_attrs_map.items():
-        col = out[name]
-        obj_attrs[data_attr] = col
+        obj_attrs[data_attr] = out[name]
         del out[name]
 
     info = obj_attrs.pop('__info__', {})
     if len(data_attrs_map) == 1:
         # col is the first and only serialized column; in that case, use info
-        # stored on the column.
+        # stored on the column. First step is to get that first column which
+        # has been moved from `out` to `obj_attrs` above.
+        data_attr = next(iter(data_attrs_map.values()))
+        col = obj_attrs[data_attr]
+
+        # Now copy the relevant attributes
         for attr, nontrivial in (('unit', lambda x: x not in (None, '')),
                                  ('format', lambda x: x is not None),
                                  ('description', lambda x: x is not None),
--- a/astropy/table/soco.py
+++ b/astropy/table/soco.py
@@ -82,7 +82,7 @@ class SCEngine:
         Add a key, value pair.
         '''
         if self._unique and (key in self._nodes):
-            message = f'duplicate {key:!r} in unique index'
+            message = f'duplicate {key!r} in unique index'
             raise ValueError(message)
         self._nodes.add(Node(key, value))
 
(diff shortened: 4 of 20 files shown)


astropy/table/table.py
```diff
@@ -3023,10 +3025,10 @@ def _is_mapping(obj):
         else:
             raise TypeError('Vals must be an iterable or mapping or None')
 
+        # Insert val at index for each column
         columns = self.TableColumns()
-        try:
```
taldcroft, line 3027: Putting the `try` into the loop is logically the same but ensures correctness.

astropy/time/formats.py
```diff
@@ -85,26 +85,26 @@ class TimeFormatMeta(type):
 
     _registry = TIME_FORMATS
 
-    def __new__(mcls, name, bases, members):
-        cls = super().__new__(mcls, name, bases, members)
+    def __new__(cls, name, bases, members):
```
taldcroft, line 88: Use standard idiom `__new__(cls, ...)`.

astropy/time/formats.py
```diff
@@ -1844,7 +1841,7 @@ def set_jds(self, val1, val2):
                 if epoch_type.upper() != epoch_prefix:
                     raise ValueError
             except (IndexError, ValueError, UnicodeEncodeError):
-                raise ValueError(f'Time {time_str} does not match {self.name} format')
+                raise ValueError(f'Time {val} does not match {self.name} format')
```
taldcroft, line 1844: This was just wrong in the original.

taldcroft:
@mhvk @hamogu - I ended up spending way more time on this little PR than I intended given that it mostly doesn't not do anything besides "improve" the code. But it does fix a couple of actual bugs. Treat this as a low-priority thing to review at some point.

astropy/table/soco.py
```diff
@@ -82,7 +82,7 @@ def add(self, key, value):
         Add a key, value pair.
         '''
         if self._unique and (key in self._nodes):
-            message = f'duplicate {key:!r} in unique index'
+            message = f'duplicate {key!r} in unique index'
```
hamogu, line 85: I have never seen `!` in string formatting, and don't know what it does, but I'm sure @taldcroft knows what he's doing.

hamogu (approved):
As usual, PRs by @taldcroft are either perfect or beyond my abilities. I only have a slightly higher-level question: Should we really define initial values just to keep the static checker happy? Is there a risk of hiding real errors, i.e. is there is some downstream problem in the old version that code would fail at some point with "variable not defined" while in the new one it would move on with that initial value? I did not see places where that will happen, but then again the whole point is to deal with cases we did not see.

astropy/table/soco.py
taldcroft: That gives the `repr(key)`, and one could equivalently write `'{repr(key)}'`.

taldcroft:
> Is there a risk of hiding real errors, i.e. is there is some downstream problem in the old version that code would fail at some point with "variable not defined" while in the new one it would move on with that initial value? 

Good question (both in practice and in philosophy). In at least one case I did add a new check for `None` and raise an exception. I think that in the other cases the code would fail with a `None` value since the expected processing generates a type that has attributes / behaviors that are inconsistent with `None`. But I'll double check.

astropy/table/serialize.py
```diff
@@ -336,7 +336,12 @@ def _construct_mixin_from_columns(new_name, obj_attrs, out):
     info = obj_attrs.pop('__info__', {})
     if len(data_attrs_map) == 1:
         # col is the first and only serialized column; in that case, use info
-        # stored on the column.
+        # stored on the column. First step is to get that first column which
+        # has been moved from `out` to `obj_attrs` above.
+        data_attr = next(iter(data_attrs_map.values()))
```
mhvk, line 341: This worked before because `data_attr` is already defined correctly. Somewhat annoying to re-do it just because of the type checker. Maybe again define a default on top instead?

astropy/table/table.py
mhvk: Yes, clearer.

astropy/time/formats.py
```diff
@@ -15,7 +15,7 @@
 
 from astropy.utils.decorators import lazyproperty, classproperty
 from astropy.utils.exceptions import AstropyDeprecationWarning
-from astropy import units as u
+import astropy.units as u
```
mhvk, line 18: Still a bit surprised about this: I've been changing it in the other direction in my own code...

astropy/time/formats.py
mhvk: But this actually is a metaclass, no? I'd keep it as is.

mhvk (approved):
Nice that it caught some real bugs! Though I'm with @hamogu that having to initialize variables just to please the checker is not great. For one case, I had another suggestion. Otherwise, it is all super-nitpicks and OKs.

astropy/table/operations.py
```diff
@@ -1027,6 +1027,7 @@ def _apply_join_funcs(left, right, keys, join_funcs):
     right = right.copy(copy_data=False)
     for key, join_func in join_funcs.items():
         ids1, ids2 = join_func(left[key], right[key])
+        id_key = None
```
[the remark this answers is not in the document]
taldcroft: Done.

astropy/time/formats.py
taldcroft: Done.

astropy/io/ascii/connect.py
```diff
@@ -4,7 +4,7 @@
 
 import re
 
-from astropy.io import registry as io_registry  # noqa
+import astropy.io.registry as io_registry  # noqa
```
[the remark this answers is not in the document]
taldcroft: Reverted.

astropy/io/ascii/daophot.py
```diff
@@ -238,8 +238,8 @@ def search_multiline(self, lines, depth=150):
         # this case we have to figure out how many apertures there are based on
         # the file structure.
 
-        comment, special, cont = zip(*(self.re_multiline.search(l).groups()
-                                       for l in lines[:depth]))
+        comment, special, cont = zip(*(self.re_multiline.search(ii).groups()
```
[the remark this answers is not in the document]
taldcroft: Done.

astropy/table/serialize.py
taldcroft: In the code above (line 331 loop) `col` is created as a "side-effect" of the actual goal. I actually changed the code to not make that side-effect variable, so it is clean and understandable now. 
```
    for name, data_attr in data_attrs_map.items():
        obj_attrs[data_attr] = out[name]
        del out[name]
```

taldcroft (commented):
Trying to revive this. Next up, rebase.

PR #10779, opened 2020-09-28
TST: Pin matplotlib to older version in pyinstaller job

pllim:
### Description

This pull request is to address `The MATPLOTLIBDATA environment variable was deprecated in Matplotlib 3.1 and will be removed in 3.3.` error in cron job for pyinstaller.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -213,7 +213,7 @@ matrix:
           python: 3.8
           dist: bionic
           name: bundling with pyinstaller
-          stage: Cron tests
+          stage: Initial tests
           env: TOXENV="pyinstaller"
 
     allow_failures:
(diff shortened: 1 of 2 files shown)


pllim:
Nope. Now I get numpy warnings, which look familiar, so maybe we're regressing here. Maybe @astrofrog knows of the proper pyinstaller fix.

embray:
I don't fully understand how cron jobs in Travis work (they are new since I last used Travis).  Was your change in the travis.yml of "stage: Initial tests" just a workaround to force the job to run in a normal PR build?  (Which I assume would then be removed once it's confirmed to pass?)

embray:
Nevermind, f6a436308 seems to answer that question.

Issue #8094, opened 2018-11-08
Moffat1D and Moffat2D derivatives


ycopin:
The derivatives in Moffat1D are wrong:

https://github.com/astropy/astropy/blob/2b8fe1305b2689d8b2715f98985e36f2c7d5c0d9/astropy/modeling/functional_models.py#L2223

They should be replaced by:
```
    @staticmethod
    def fit_deriv(x, amplitude, x_0, gamma, alpha):
        """One dimensional Moffat model derivative with respect to parameters"""

        fac = (1 + (x - x_0) ** 2 / gamma ** 2)
        d_A = fac ** (-alpha)
        d_x_0 = (2 * amplitude * alpha * (x - x_0) * d_A / (fac * gamma ** 2))
        d_gamma = (2 * amplitude * alpha * (x - x_0) ** 2 * d_A /
                   (fac * gamma ** 3))
        d_alpha = -amplitude * d_A * np.log(fac)
        return [d_A, d_x_0, d_gamma, d_alpha]
```

In Moffat2D:

https://github.com/astropy/astropy/blob/2b8fe1305b2689d8b2715f98985e36f2c7d5c0d9/astropy/modeling/functional_models.py#L2301

only derivative wrt gamma is wrong. It should be
```
        d_gamma = 2 * amplitude * alpha * d_A * rr_gg / ((1 + rr_gg) * gamma ** 3)
```

pllim:
These were originally added by @adonath as part of #1251 and has since been renamed a few times.

adonath:
I remember that for the `Beta1D` and `Beta2D` models I computed the derivatives and reference values in a notebook using using Sympy (http://nbviewer.jupyter.org/gist/adonath/6088712). So I'm pretty sure it used to be correct at that time. Maybe when the rename to `Moffat1D` and `Moffat2D` happened, the parametrization changed as well and the derivatives where not adapted? From a quick look  in the code it seems there are no unit tests for the model derivatives.

@ycopin I didn't check whether your proposed formulae are correct, please check against Sympy or some other algebra library and post the result as a reference. In case you still find the current formulae are incorrect, would you mind preparing a minimal pull request with the correct ones? Ideally one should add unit tests for the model derivatives, but this is a somewhat larger project...

ycopin:
[...]

PR #10784, opened 2020-09-29
undoes part of #8108 that was incorrect

chris-simpson:
# Description

This pull request is to address #8094 which was not a bug, but a bug was introduced by PR #8108

The `Moffat2D` gamma derivative is now (once again) correctly calculated, as shown in comments at the end of #8094

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1081,6 +1081,8 @@ astropy.modeling
 - Do not create new ``math_functions`` models for ufuncs that are
   only aliases (divide and mod). [#10697]
 
+- Fix calculation of the ``Moffat2D`` derivative with respect to gamma. [#10784]
+
 astropy.nddata
 ^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


pllim:
Thanks! cc @ycopin 

Needs a change log.

Does this need a test as well? Why wasn't this bug caught the last time?

chris-simpson:
It's not technically a bug because the code does what it's designed to do, it's just that design is wrong. So a test won't really help (IMHO) because if there had been a test that the change would cause to fail, the claim would have been made that the test was wrong because it was testing the wrong thing. In the original issue, there was a request for a proof of the revised (incorrect) formula, which was never presented so I guess there was a bit of misplaced trust.

I've put the PR# in the changelog (in the right place at the second time of asking) and squashed commits so this ought to be good to go if you're happy with the algebraic proof presented in #8094

nden:
Thanks @chris-simpson! I think the confusion came from `1/gamma**2` being included in `rr_gg`.

PR #10727, opened 2020-09-13
Allow **kwargs in `realize_frame`

astrojuanlu:
### Description

This pull request closes gh-7784 by allowing `**kwargs` in `realize_frame`, for simmetry with `replicate_without_data` and especially to allow `representation_type`, which makes it possible to realize a frame and explicitly set the representation type in one line.

This also paves the way to changes discussed in gh-6435 about making a frame instantiated with a `BaseRepresentation` object automatically infer the `representation_type`, but allow the user to override it.

PR #10788, opened 2020-10-01
Set representation type using argument class when appropriate

### LINKED ISSUE #6435: Frame representation is not respected when instanced using a representation
Yes, the title of this issue is awful... Please feel free to edit for clarity.

When I instance a frame using a subclass of `BaseRepresentation` as the container of the data (instead of passing the data directly to the initializer of the frame), I would expect the representation of the frame to be the one I used, but it's not:

```python
In [3]: GCRS(CartesianRepresentation(*iss.r))
Out[3]: 
<GCRS Coordinate (obstime=J2000.000, obsgeoloc=( 0.,  0.,  0.) m, obsgeovel=( 0.,  0.,  0.) m / s): (ra, dec, distance) in (deg, deg, km)
    ( 281.73052421,  51.41284322,  6774.76995296)>

In [4]: _.representation
Out[4]: astropy.coordinates.representation.SphericalRepresentation
```

Is there a reason for this? I'm aware that there are other ways to instance a frame though, which behave differently:

```python
In [5]: GCRS(x=iss.r[0], y=iss.r[1], z=iss.r[2], representation="cartesian")
Out[5]: 
<GCRS Coordinate (obstime=J2000.000, obsgeoloc=( 0.,  0.,  0.) m, obsgeovel=( 0.,  0.,  0.) m / s): (x, y, z) in km
    ( 859.07256, -4137.20368,  5295.56871)>

In [6]: GCRS(x=iss.r[0], y=iss.r[1], z=iss.r[2], representation=CartesianRepresentation)
Out[6]: 
<GCRS Coordinate (obstime=J2000.000, obsgeoloc=( 0.,  0.,  0.) m, obsgeovel=( 0.,  0.,  0.) m / s): (x, y, z) in km
    ( 859.07256, -4137.20368,  5295.56871)>
```

astrojuanlu:
### Description

This pull request fixes #6435 by setting the `representation_type` of the frame from the args if they are `BaseRepresentation` objects. This avoids tedious code like `GCRS(CartesianRepresentation(...), representation="cartesian")`.

This made me realize that there might be places where unnecessary round-tripping conversion is done between representations. For example, many transformations starting from spherical coordinates will convert to cartesian, and then back to spherical because it is assumed everywhere that the spherical attributes of the transformed frame can be accessed without `.spherical` intermediate steps. This is related to [@ayshih comment here](https://github.com/astropy/astropy/issues/6435#issuecomment-688567347):

> It's admittedly odd that `.data` can be a different representation than what shows in the `repr`, but what I tell SunPy users is to never use `.data`. If they want to access the component data and the type of representation matters, I tell them to be explicit: e.g., use `.spherical`.

Nothing is said in `SkyCoord.transform_to` or `BaseCoordinateFrame.transform_to` about this assumption, but I decided to retain it anyway. This required setting explicit `representation_type=` in several `realize_frame` calls in the transformation functions, as well as initializers, which will require downstream users to adapt, as was already discussed in #6435 (SunPy will surely be affected). All coordinates tests passed locally, but I will give another closer look.

I made changes in 7 tests to adapt to this new behaviour.

### DIFF
--- a/astropy/coordinates/builtin_frames/fk4.py
+++ b/astropy/coordinates/builtin_frames/fk4.py
@@ -163,7 +163,7 @@ def fk4_to_fk4_no_e(fk4coord, fk4noeframe):
     # if no obstime was given in the new frame, use the old one for consistency
     newobstime = fk4coord._obstime if fk4noeframe._obstime is None else fk4noeframe._obstime
 
-    fk4noe = FK4NoETerms(rep, equinox=fk4coord.equinox, obstime=newobstime)
+    fk4noe = FK4NoETerms(rep, equinox=fk4coord.equinox, obstime=newobstime, representation_type=fk4coord.representation_type)
     if fk4coord.equinox != fk4noeframe.equinox:
         # precession
         fk4noe = fk4noe.transform_to(fk4noeframe)
@@ -206,4 +206,4 @@ def fk4_no_e_to_fk4(fk4noecoord, fk4frame):
     if isinstance(fk4noecoord.data, UnitSphericalRepresentation):
         rep = rep.represent_as(UnitSphericalRepresentation)
 
-    return fk4frame.realize_frame(rep)
+    return fk4frame.realize_frame(rep, representation_type=fk4noecoord.representation_type)
--- a/astropy/coordinates/sky_coordinate.py
+++ b/astropy/coordinates/sky_coordinate.py
@@ -624,6 +624,10 @@ class SkyCoord(ShapedLikeNDArray):
                      set(frame_kwargs.keys())):
             frame_kwargs.pop(attr)
 
+        # Ensure transformed coordinates have representation and differential types
+        # as set by the destination frame
+        new_coord.set_representation_cls(frame.default_representation, frame.default_differential)
+
         return self.__class__(new_coord, **frame_kwargs)
 
     def apply_space_motion(self, new_obstime=None, dt=None):
--- a/docs/coordinates/skycoord.rst
+++ b/docs/coordinates/skycoord.rst
@@ -1051,6 +1051,7 @@ As the next step, those coordinates are transformed into an
 Again, as in the last example, we need to convert the coordinates in radians
 and make sure they are between :math:`-\pi` and :math:`\pi`:
 
+    >>> c_gal_icrs.representation_type = "spherical"
     >>> ra_rad = c_gal_icrs.ra.wrap_at(180 * u.deg).radian
     >>> dec_rad = c_gal_icrs.dec.radian
 
--- a/docs/coordinates/velocities.rst
+++ b/docs/coordinates/velocities.rst
@@ -225,7 +225,7 @@ the existing frame using
         (1., 2., 3.)>
     >>> vel_to_add = CartesianDifferential(4*u.km/u.s, 5*u.km/u.s, 6*u.km/u.s)
     >>> newdata = icrs.data.to_cartesian().with_differentials(vel_to_add)
-    >>> icrs.realize_frame(newdata) # doctest: +FLOAT_CMP
+    >>> icrs.realize_frame(newdata, representation_type="spherical", differential_type="sphericalcoslat") # doctest: +FLOAT_CMP
     <ICRS Coordinate: (ra, dec, distance) in (deg, deg, kpc)
         (1., 2., 3.)
      (pm_ra_cosdec, pm_dec, radial_velocity) in (mas / yr, mas / yr, km / s)
@@ -243,7 +243,7 @@ specifying the differentials because of the lack of explicit unit information::
         (1., 2.)>
     >>> rv_to_add = RadialDifferential(500*u.km/u.s)
     >>> data_with_rv = icrs_no_distance.data.with_differentials({'s':rv_to_add})
-    >>> icrs_no_distance.realize_frame(data_with_rv) # doctest: +FLOAT_CMP
+    >>> icrs_no_distance.realize_frame(data_with_rv, differential_type="sphericalcoslat") # doctest: +FLOAT_CMP
     <ICRS Coordinate: (ra, dec) in deg
         (1., 2.)
      (radial_velocity) in km / s
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,11 @@ astropy.coordinates
 - Method ``.realize_frame`` from coordinate frames now accepts ``**kwargs``,
   including ``representation_type``. [#10727]
 
+- Initializing a ``BaseCoordinateFrame`` from a ``BaseRepresentation`` now sets its ``representation_type``
+  to the class of the argument. This means that the default representation of the frame might be overridden,
+  and therefore the readily available attributes might change. To retain the previous behaviour,
+  pass a explicit ``representation_type=`` to the initializer and to methods like ``.realize_frame```. [#10788]
+
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -463,6 +463,13 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
                     'and other positional arguments')
 
             if representation_data is not None:
+                if not hasattr(self, '_representation'):
+                    # Representation has not been set yet, do it now
+                    # This saves the need to specify the type twice in cases like
+                    # `ICRS(CartesianRepresentation(...), representation_type="cartesian")`
+                    # and effectively overrides the default representation of the frame,
+                    # see https://github.com/astropy/astropy/issues/6435
+                    self.set_representation_cls(representation_data.__class__)
                 diffs = representation_data.differentials
                 differential_data = diffs.get('s', None)
                 if ((differential_data is None and len(diffs) > 0) or
--- a/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
@@ -91,7 +91,7 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
         loccirs = altaz_coo.location.get_itrs(altaz_coo.obstime).transform_to(cirs_frame)
 
         astrometric_rep = SphericalRepresentation(lon=cirs_ra, lat=cirs_dec,
-                                                  distance=altaz_coo.distance)
+                                                  distance=altaz_coo.spherical.distance)
         newrepr = astrometric_rep + loccirs.cartesian
         cirs_at_aa_time = CIRS(newrepr, obstime=altaz_coo.obstime)
 
--- a/astropy/coordinates/builtin_frames/ecliptic_transforms.py
+++ b/astropy/coordinates/builtin_frames/ecliptic_transforms.py
@@ -70,14 +70,14 @@ def gcrs_to_geoecliptic(gcrs_coo, to_frame):
 
     rmat = _mean_ecliptic_rotation_matrix(to_frame.equinox)
     newrepr = gcrs_coo2.cartesian.transform(rmat)
-    return to_frame.realize_frame(newrepr)
+    return to_frame.realize_frame(newrepr, representation_type=gcrs_coo.representation_type)
 
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GeocentricMeanEcliptic, GCRS)
 def geoecliptic_to_gcrs(from_coo, gcrs_frame):
     rmat = _mean_ecliptic_rotation_matrix(from_coo.equinox)
     newrepr = from_coo.cartesian.transform(matrix_transpose(rmat))
-    gcrs = GCRS(newrepr, obstime=from_coo.obstime)
+    gcrs = GCRS(newrepr, obstime=from_coo.obstime, representation_type=from_coo.representation_type)
 
     # now do any needed offsets (no-op if same obstime and 0 pos/vel)
     return gcrs.transform_to(gcrs_frame)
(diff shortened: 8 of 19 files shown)


astropy/coordinates/builtin_frames/fk4.py
```diff
@@ -206,4 +206,4 @@ def fk4_no_e_to_fk4(fk4noecoord, fk4frame):
     if isinstance(fk4noecoord.data, UnitSphericalRepresentation):
```
mhvk, line 206: Here, it seems more logical to just do `rep = rep.represent_as(fk4noecoord.representation_type)` for all cases. And perhaps one should do the same elsewhere too?
Alternatively, I think the `if` clause can just be deleted as you now give `representation_type` below.

mhvk (commented):
I like where this is going. One question is where exactly to do the transformation back to the input type. Options:
1. Passing in `representation_type` (as you have it).
2. Explicitly doing `.represent_as(old_type)` before.
3. Make a new general method `BaseRepresentation.transform(matrix)` that goes through `Cartesian` and transforms back. (Least code, but perhaps a bit strange that the matrix then is for cartesian).

Right now, I think what you have may be cleanest...

astropy/coordinates/builtin_frames/fk4.py
astrojuanlu: There are several instances of `if isinstance\(.*, UnitSpherical` where different logic is applied if there is no distance information, so I only replaced this one that you pointed out as well as the inverse one.

astrojuanlu:
Working on the doctests now, there are quite a few failures

astrojuanlu:
Provisionally milestoning this as 4.2.

astropy/coordinates/builtin_frames/fk4.py
astrojuanlu: ...I misunderstood the comment, now I see that I can actually remove the `if` clause. Doing so.

docs/coordinates/skycoord.rst
```diff
@@ -652,8 +652,8 @@ coordinates are in the exact same reference frame::
 
   >>> sc2 = SkyCoord(3, 4, frame='fk4', unit='deg', obstime='J1978.123', equinox='B1960.0')
   >>> sc.transform_to(sc2)  # doctest: +FLOAT_CMP
-  <SkyCoord (FK4: equinox=B1960.000, obstime=J1978.123): (ra, dec) in deg
-      (0.48726331, 1.77731617)>
+  <SkyCoord (FK4: equinox=B1960.000, obstime=J1978.123): (ra, dec, distance) in (deg, deg, )
+      (0.48726331, 1.77731617, 1.)>
```
astrojuanlu, line 656: Not sure if we should investigate further to avoid this...

docs/coordinates/skycoord.rst
```diff
@@ -1127,7 +1128,7 @@ The example below illustrates the floating point issue using the exact
 equality comparison, where we do a roundtrip transformation
 FK4 => ICRS => FK4 and then compare::
 
-  >>> sc1 = SkyCoord(1*u.deg, 2*u.deg, frame='fk4')
+  >>> sc1 = SkyCoord(1*u.deg, 2*u.deg, representation_type="unitspherical", frame='fk4')
```
astrojuanlu, line 1131: Otherwise, the comparison fails because `sc1.icrs.fk4` and `sc1` have different representation types. Similarly to what happens with the doctest above, not sure if we should try to avoid this.

docs/coordinates/velocities.rst
```diff
@@ -225,11 +225,11 @@ the existing frame using
         (1., 2., 3.)>
     >>> vel_to_add = CartesianDifferential(4*u.km/u.s, 5*u.km/u.s, 6*u.km/u.s)
     >>> newdata = icrs.data.to_cartesian().with_differentials(vel_to_add)
-    >>> icrs.realize_frame(newdata) # doctest: +FLOAT_CMP
+    >>> icrs.realize_frame(newdata, representation_type="spherical") # doctest: +FLOAT_CMP
     <ICRS Coordinate: (ra, dec, distance) in (deg, deg, kpc)
         (1., 2., 3.)
-     (pm_ra_cosdec, pm_dec, radial_velocity) in (mas / yr, mas / yr, km / s)
-        (0.34662023, 0.41161335, 4.29356031)>
+     (pm_ra, pm_dec, radial_velocity) in (mas / yr, mas / yr, km / s)
```
astrojuanlu, line 231: I don't understand the reason for this still, investigating.

docs/coordinates/velocities.rst
```diff
@@ -225,11 +225,11 @@ the existing frame using
...
+        (0.34683151, 0.41161335, 4.29356031)>
```
astrojuanlu, line 232: ...and this numerical change is perplexing.

astrojuanlu:
Regarding `examples/coordinates/plot_galactocentric-frame.py`, even after this diff:

```diff
diff --git a/examples/coordinates/plot_galactocentric-frame.py b/examples/coordinates/plot_galactocentric-frame.py
index 7e30d00fa..bde78eabc 100644
--- a/examples/coordinates/plot_galactocentric-frame.py
+++ b/examples/coordinates/plot_galactocentric-frame.py
@@ -64,6 +64,7 @@ c1 = coord.SkyCoord(ra=89.014303*u.degree, dec=13.924912*u.degree,
 # can simply do:
 
 gc1 = c1.transform_to(coord.Galactocentric)
+gc1.representation_type = "cartesian"
 
 ##############################################################################
 # From here, we can access the components of the resulting
@@ -94,6 +95,7 @@ gc_frame = coord.Galactocentric(
 # We can then transform to this frame instead, with our custom parameters:
 
 gc2 = c1.transform_to(gc_frame)
+gc2.representation_type = "cartesian"
 print(gc2.v_x, gc2.v_y, gc2.v_z)
 
 ##############################################################################
@@ -116,6 +118,7 @@ gc_frame2 = coord.Galactocentric(galcen_distance=galcen_distance,
                                  galcen_v_sun=v_sun2,
                                  z_sun=0*u.pc)
 gc3 = c1.transform_to(gc_frame2)
+gc3.representation_type = "cartesian"
 print(gc3.v_x, gc3.v_y, gc3.v_z)
 
 ##############################################################################
```

I still get this intriguing error:

```
$ python examples/coordinates/plot_galactocentric-frame.py 
30.254684717897074 km / s 171.29916086104885 km / s 18.19390627095307 km / s
28.427958360720748 km / s 169.69916086104888 km / s 17.70831652451455 km / s
28.427958360720748 km / s 167.61484955608267 km / s 18.118916793584443 km / s
  File "examples/coordinates/plot_galactocentric-frame.py", line 184, in <module>
    gal_rings = gc_rings.transform_to(coord.Galactic)
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/sky_coordinate.py", line 627, in transform_to
    return self.__class__(new_coord, **frame_kwargs)
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/sky_coordinate.py", line 330, in __init__
    self._sky_coord_frame = frame_cls(copy=copy, **frame_kwargs)
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/baseframe.py", line 592, in __init__
    self._data = representation_data.with_differentials(
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/representation.py", line 922, in with_differentials
    new_rep._validate_differentials(differentials))
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/representation.py", line 720, in _validate_differentials
    diff._check_base(self)
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/representation.py", line 2284, in _check_base
    raise TypeError("Differential class {} is not compatible with the "
TypeError: Differential class <class 'astropy.coordinates.representation.CartesianDifferential'> is not compatible with the base (representation) class <class 'astropy.coordinates.representation.SphericalRepresentation'>
```

astrojuanlu:
The problem above is fixed by doing `gc_rings.representation_type = "spherical"` just before `gc_rings.transform_to(coord.Galactic)`, but I find that extremely annoying. Probably it should just work.

docs/coordinates/skycoord.rst
astrojuanlu: Scratch my comment above, this is still failing.

mhvk:
@astrojuanlu - hmm, strange what you are finding.  I'll try to look but likely only later this week...

adrn:
Hmm, I need to look at this more carefully, but so far am not convinced on this! (and apologies: I should have commented on in #6435 but missed that issue and only caught it from the discussion here!)

In particular, I also rely heavily in my code on the fact that the output of a `.transform_to()` comes out in the default representation of the new frame. This comes up a lot in an example that you had to update in this PR: something like

    c = coord.SkyCoord(ra=..., dec=..., distance=..., frame='icrs')
    galcen = c.transform_to(coord.Galactocentric)

I have quite a bit of code that assumes that I can then take this object and do, e.g., `galcen.x`. If I understand correctly, with the changes here, I would have to add a line

    galcen.representation_type = 'cartesian' 

after every transformation? 

I actually like the current behavior in which 

    c = coord.ICRS(coord.CartesianRepresentation(...))

adopts the frame's default `representation_type` unless explicitly stated. I do see the argument that this could be unintuitive to some, and I am probably "too close" to have a useful opinion, but I find the current behavior more intuitive :)

mhvk:
@adrn - I can definitely see your argument for the current default for `.transform_to()`, since the user has not really stated a particular intent. Since often the names of the representation components change anyway, I think there shouldn't be too much of a surprise that the representation can change too. So, in the context of this PR, perhaps the default for `representation_type` in `.transform_to()` could be the default of the new class? (i.e., it is still overridable, but one has to pass in `representation_type='cartesian'`, etc.). This may also solve some of the more annoying doctest problems (which in themselves signal that perhaps this is something users rely on).

I think for direct initialization, the case for the present behaviour is much less strong. Indeed, in the issue that led to the PR showed that this was surprising behaviour that seemed quite generally considered to be a bug. Hence, perhaps we can still change it such that for `ICRS(CartesianRepresentation(...))` the default `representation_type` becomes `cartesian`?

adrn:
OK, yes, good points. I'm more open to the latter if we can keep the current `transform_to()` behavior.

astrojuanlu:
Point well taken about retaining the current behavior of `transform_to`, I agree this change was being too disruptive. Notice though that `BaseCoordinateFrame.transform_to` does not accept `representation_type`, so I will go and add `**kwargs` like I did for `realize_frame` in #10727.

astrojuanlu:
...although it ends up being a `.replicate(**kwargs)` call after `trans(self, new_frame)`, perhaps it doesn't add that much value. In any case, I will rework the PR ensuring that `.transform_to` uses the destination frame default representation as the representation type.

astropy/coordinates/sky_coordinate.py
```diff
@@ -624,6 +624,10 @@ def transform_to(self, frame, merge_attributes=True):
                      set(frame_kwargs.keys())):
             frame_kwargs.pop(attr)
 
+        # Ensure transformed coordinates have representation and differential types
+        # as set by the destination frame
+        new_coord.set_representation_cls(frame.default_representation, frame.default_differential)
+
```
astrojuanlu, line 630: I'm a bit confused here.

- Setting the new coordinates `representation_type` and `differential_type` to the **default of the new frame** makes some tests fail.
- Setting the new coordinates `representation_type` and `differential_type` to the **values of the original frame** makes some other tests fail.

Therefore, I understand that the logic is a bit more complex for `SkyCoord.transform_to` and that the magic happens at the "attributes merging" stage. Am I right?

astrojuanlu:
Current status:

cat_playing_whac-a-mole_gif_9985306640

I feel like either (a) there is a lot of implicit behavior that is only documented in the form of doctests or (b) the behavior is perfectly explicit but I am not paying enough attention trying to rush a solution for #6435.

Couple of random thoughts:

- `SkyCoord.transform_to` does not call `self.frame.transform_to(frame)`, but repeats the `frame_transform_graph.get_transform` call plus some custom attribute merging logic through a `GenericFrame`. This was unexpected for me and I need to inspect it more closely.
- I need to reflect on how can I achieve at the same time that (1) direct instantiation respects the representation class, if given explicitly (#6435), (2) `.transform_to` keeps the default representation of the destination frame and (3) `SkyCoord.transform_to` does not break.
- It's a bit of a burden having to modify all the transforms to do this. At some point I tried `.replicate(representation_type=...)` after the transform but it didn't work, probably because I was also missing `differential_type=...`. Again, I need to revisit this.

I think I will step back, grab some pen and paper, and mark this as draft in the meanwhile.

mhvk:
Yes, the `SkyCoord.transform_to` needs to make sure that things that are not defined on the frame (such as `obstime` for an ICRS coordinate) do get propagated as needed. It is quite a dance indeed. But nevertheless a bit disappointing that the `representation_type` is not an easy addition with default - it means that proper "separation of concerns" has failed a bit in those methods.

pllim:
Given feature freeze is tomorrow, I am moving the milestone for now.

nstarman:
#11444 might simplify things ?
It certainly applies to https://github.com/astropy/astropy/pull/10788#pullrequestreview-500308756 point 3

eteq:
@astrojuanlu, do you want to try to push on this in the next week to try to get it in for 4.3? Or would you rather we re-milestone to 5.0?

astrojuanlu:
Darn it 😓  I don't think I will have time to work on this, it was quite difficult to get right when I tried and it requires me a lot of concentration. Let's re-milestone to 5.0.

mhvk:
No worries, we'll get back to this!

mhvk:
@astrojuanlu - this still seems nice even if tricky. Do you want to pursue it further? Right now, it definitely needs at least a rebase so that we're sure we're testing the right code!

pllim:

nstarman:
What double post?

nstarman:
What double post?

mhvk:
Not sure we actually converged, though. Possibly, it is best to have a separate issue that starts with a summary and the problems associated with various solutions.

PR #10789, opened 2020-10-01
moved units from astrophys.py to misc.py 

### LINKED ISSUE #10769: Move non-astronomy units out of astrophys.py to new misc.py or so
Currently, most units are defined in si.py, cgs.py, and astrophys.py. As noted by @adrn in https://github.com/astropy/astropy/pull/10586#issuecomment-698653617, the latter has become a bit of a miscellaneous bunch, including units that are not particular to astronomy at all. It would make sense to move those to a new `misc.py`. Indeed, some are sufficiently obscure that it might be an idea to allow these to be removed from the name space (though they would of course be present by default).

addddd123:
### Description

This pull request is to address issue  #10769 . i moved all units deffinitions  (bar, barn, bin, bit, byte, cycle, dalton, pixel, spat, torr ) from /units/astrophy.py to /units/misc.py  file

Fixes #10769


adrn:
Hi @addddd123 - Thanks for starting on this! A few comments:
- Could you please fill out some more information in the pull request description? 
- It looks like you moved the unit definitions to `astropy/utils/misc.py`. I realize now that the description of the solution in the issue #10769 is a little vague. The unit definitions need to be moved to a new file in `astropy/units/misc.py` instead.
- Please remove the extra blank lines you have added.
- Please review my comment [here](https://github.com/astropy/astropy/issues/10769#issuecomment-698926144) to make sure you moved all of the units I listed.

addddd123:
Hi @adrn I have done the changes appropriately as you said .

astropy/units/astrophys.py
```diff
@@ -203,4 +163,4 @@
 # standard units defined here.
 from .utils import generate_unit_summary as _generate_unit_summary
 if __doc__ is not None:
-    __doc__ += _generate_unit_summary(globals())
+    __doc__ += _generate_unit_summary(globals())
```
mhvk, line 166: You'll need a similar stanza at the end of your new `misc.py` as well, including the `CLEANUP` part above. And you need to import from `misc.py` in `__init__.py` (just add everywhere `astrophys` is done). Finally, also add a line in `docs/units/index.rst` (again look for `astrophys`).

Please also run tests locally before pushing - all tests are failing now since the units are not imported properly. Please do include the pep8 tests - you have many missing CR at the end of files (see instructions for astropy development).

mhvk:
Superceded by #11142 - sorry that there was a mess-up with reviews and no realization of the duplication.

Issue #10794, opened 2020-10-04
Slow coordinate example in documentation build


saimn:
Looking at pages that could make the documentation build slower than necessary, the coordinate page is the slowest by far (thanks to the `sphinx.ext.duration` extension). This is with a clean repository:
```
====================== slowest reading durations =======================
70.252 coordinates/index
14.204 visualization/wcsaxes/ticks_labels_grid
8.617 timeseries/index
7.544 visualization/wcsaxes/overlays
6.673 stats/ripley
```
And then running again, so with the cache and just a small change to `coordinates/index.rst`:
```
====================== slowest reading durations =======================
73.007 coordinates/index
0.552 api/astropy.coordinates.EarthLocation
```
This lead me to the [Performance tips](https://docs.astropy.org/en/latest/coordinates/index.html#performance-tips) (!), in particular the last example which is running some benchmark to produce a plot (_Transformation of SkyCoord with 100.000 obstimes over 12 hours_). If I remove `performance.inc.rst` then the page takes ~2.5s to build.

Running the last example manually gives:
```
No Interpolation took 0.3208 s
Interpolation with        0 s resolution took 33.4733 s (  0.0x faster) 
Interpolation with        1 s resolution took 9.1881 s (  0.0x faster) 
Interpolation with       10 s resolution took 1.2020 s (  0.3x faster) 
Interpolation with      100 s resolution took 0.3375 s (  1.0x faster) 
Interpolation with     1000 s resolution took 0.2250 s (  1.4x faster) 
Interpolation with    10000 s resolution took 0.1892 s (  1.7x faster) 
Interpolation with   100000 s resolution took 0.1719 s (  1.9x faster) 
CPU times: user 45.4 s, sys: 61.8 ms, total: 45.5 s
Wall time: 45.6 s
```

Could we avoid running this benchmark in the docs build (>1min over the 7/8 min of a build on my laptop) ? Or not run the first two steps, or run this in the benchmark repo ?

taldcroft:
@saimn - excellent sleuthing and I am definitely 👍  on improving docs build times. Could this example be moved out of the core docs and into astropy learn examples and then linked? Or just use a static png of the plot?  

Maybe the simplest of all is just reducing the number of points. Would the example still demonstrate the performance lesson using 10_000 or even 1000 points?

astrofrog:
@saimn how did you find the build times?

astrofrog:
Yeah we could just remove:

```
[...]

PR #10795, opened 2020-10-04
Run Sphinx with the duration extension

saimn:
Do not merge ! Ref #10794.

This is to see on RTD how long it makes for the slowest pages.


saimn:
It seems the extension doesn't work on RTD :man_shrugging:

pllim:
If it is necessary, we could open issue with RTD.

PR #10797, opened 2020-10-04
Fix WCSAxes bbox calculation

dstansby:
Fixes https://github.com/astropy/astropy/issues/10796.

This includes the axes bounding box in the caclulation; previously only the tick label bboxes were included.

PR #10799, opened 2020-10-04
WCSAxes AxisLabel refactoring

dstansby:
As I dive deep into the rabbit hole that is https://github.com/astropy/astropy/pull/10797, I have refactored some code along the way. This is probably easier to review one commit at a time

- Move the code to calculate the distance halfway along a spine to the spine class itself.
- Move code outside a loop that doesn't use the loop variable
- De-duplicate a check in an if statement that occurs 4 times
- Cleanup the logic used in calculating the position of tick labels

### DIFF
--- a/astropy/visualization/wcsaxes/axislabels.py
+++ b/astropy/visualization/wcsaxes/axislabels.py
@@ -85,7 +85,7 @@ class AxisLabels(Text):
             # Find position of the axis label. For now we pick the mid-point
             # along the path but in future we could allow this to be a
             # parameter.
-            xcen, ycen, normal_angle = self._frame[axis]._halfway_x_y_angle()
+            x, y, normal_angle = self._frame[axis]._halfway_x_y_angle()
 
             label_angle = (normal_angle - 90.) % 360.
             if 135 < label_angle < 225:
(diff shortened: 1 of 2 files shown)


astropy/visualization/wcsaxes/axislabels.py
```diff
@@ -105,37 +105,30 @@ def draw(self, renderer, bboxes, ticklabels_bbox,
                     coord_ticklabels_bbox[axis] = [None]
 
                 visible = axis in visible_ticks and coord_ticklabels_bbox[axis][0] is not None
+                x = xcen
+                y = ycen
```
Cadair, line 109: Why redfine these here? why not just call them `x` & `y` from the outset?

Cadair (approved):
Makes sense to me.

pllim:
I am going to close/re-open to kick off RTD again to make sure the timeout isn't related.

pllim:
@Cadair , if you are not waiting for someone else to review, feel free to merge when CI passes.

Cadair:
hmmm, it timed out again.

dstansby:
Rebased to see if the tests will pass; I also dropped the first commit, since on a second viewing it looks like that might not be a valid change to make (even though the tests passed).

Cadair:
@dstansby should this be a bugfix as well? is it needed for the bbox quest?

dstansby:
I think the bbox quest is kind of a bugfix, but in reality will require so much refactoring (and is a beaviour change, albeit from an incorrect to a correct behaviour) it's probably best to put it in a major release?

PR #10800, opened 2020-10-04
Keep reference to parent_axes in TickLabels and AxisLabels

dstansby:
The motivation for this is that `draw()` methods should not really depend on any parameters apart from `renderer`, as all the information needed to draw an `Artist` should be contained within the artist object. Making these changes will make it substantially easier to re-factor out the code in `draw()` that is used to update properties that are needed for other methods (currently not present), such as `get_tightbbox()` and `get_window_extent()`.

Instead of passing `bboxes` down a few layers of function calls, just to have it appended to, it is much clearer to retain a reference to `parent_coords`, and directly append to `self.parent_coords.parent_axes._bboxes`. I have retained function signatures here for API compatibility. I will leave inline notes explaining this, but maybe there is a better way to do this.

### DIFF
--- a/astropy/visualization/wcsaxes/coordinate_helpers.py
+++ b/astropy/visualization/wcsaxes/coordinate_helpers.py
@@ -104,6 +104,7 @@ class CoordinateHelper:
         self.dpi_transform = Affine2D()
         self.offset_transform = ScaledTranslation(0, 0, self.dpi_transform)
         self.ticks = Ticks(transform=parent_axes.transData + self.offset_transform)
+        self.ticks.parent_coords = self
 
         # Initialize tick labels
         self.ticklabels = TickLabels(self.frame,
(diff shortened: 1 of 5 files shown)


astropy/visualization/wcsaxes/coordinate_helpers.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, parent_axes=None, parent_map=None, transform=None,
         self.ticklabels = TickLabels(self.frame,
                                      transform=None,  # display coordinates
                                      figure=parent_axes.get_figure())
+        self.ticklabels.parent_coords = self
```
dstansby, line 112: Ideally one would hand this as an argument to `TickLabels`, but I can't add a parameter to `TickLabels.__init__` whilst maintaining backwards compatibility.

astropy/visualization/wcsaxes/coordinate_helpers.py
```diff
@@ -591,25 +593,26 @@ def _draw_grid(self, renderer):
...
-    def _draw_axislabels(self, renderer, bboxes, ticklabels_bbox, ticks_locs, visible_ticks):
+    def _draw_axislabels(self, renderer, ticklabels_bbox, ticks_locs, visible_ticks):
         # Render the default axis label if no axis label is set.
         if self._auto_axislabel and not self.get_axislabel():
             self.set_axislabel(self._get_default_axislabel())
 
         renderer.open_group('axis labels')
 
-        self.axislabels.draw(renderer, bboxes=bboxes,
+        self.axislabels.draw(renderer, bboxes=None,
```
dstansby, line 615: Ideally one would remove the `bboxes` parameter, but this would break the signature. Not sure how best to handle this, ideally with a deprecation warning?

astrofrog:
Thanks, I'll take a look! We have to be careful not to create circular references that would prevent garbage collection though so we might have to use weakref. I'll try and review soon.

dstansby:
Good point - I think there may be simpler ways to do the same thing without having circular references hanging around. At least for storing the bboxes of everything, I have found a nicer way that I will open a new PR for. I'll convert this one to a draft for now.

PR #10802, opened 2020-10-05
Reduce number in points in coordinate example, fixes #10794

### LINKED ISSUE #10794: Slow coordinate example in documentation build

maxnoe:
Reduce number of points. New runtime on my machine: 7 seconds:

```
No Interpolation took 2.2064 s
Interpolation with       0.1 s resolution took 2.4585 s (  0.9x faster) 
Interpolation with       1.0 s resolution took 0.5930 s (  3.7x faster) 
Interpolation with      10.0 s resolution took 0.0804 s ( 27.4x faster) 
Interpolation with     100.0 s resolution took 0.0262 s ( 84.3x faster) 
Interpolation with    1000.0 s resolution took 0.0187 s (118.2x faster) 
Interpolation with   10000.0 s resolution took 0.0173 s (127.2x faster) 
Interpolation with  100000.0 s resolution took 0.0164 s (134.7x faster) 
python test.py  7,11s user 0,08s system 99% cpu 7,251 total
```

EDIT: Follow up of #10647, fix #10794

### DIFF
--- a/docs/coordinates/performance.inc.rst
+++ b/docs/coordinates/performance.inc.rst
@@ -135,7 +135,7 @@ for ``time_resolution`` compared to the non-interpolating, default approach.
     np.random.seed(1337)
 
     # 100_000 times randomly distributed over 12 hours
-    t = Time('2020-01-01T20:00:00') + np.random.uniform(0, 12, 100_000) * u.hour
+    t = Time('2020-01-01T20:00:00') + np.random.uniform(0, 1, 10_000) * u.hour
 
     location = location = EarthLocation(
         lon=-17.89 * u.deg, lat=28.76 * u.deg, height=2200 * u.m
@@ -167,7 +167,7 @@ for ``time_resolution`` compared to the non-interpolating, default approach.
             duration = perf_counter() - t0
 
         print(
-            f'Interpolation with {resolution.value: 8.0f} {str(resolution.unit)}'
+            f'Interpolation with {resolution.value: 9.1f} {str(resolution.unit)}'
             f' resolution took {duration:.4f} s'
             f' ({reference / duration:5.1f}x faster) '
         )
--- a/docs/coordinates/performance.inc.rst
+++ b/docs/coordinates/performance.inc.rst
@@ -156,7 +156,7 @@ for ``time_resolution`` compared to the non-interpolating, default approach.
 
 
     # now the interpolating approach for different time resolutions
-    resolutions = 10.0**np.arange(-1, 6) * u.s
+    resolutions = 10.0**np.arange(-1, 5) * u.s
     times = []
     seps = []
 


mhvk (commented):
This looks good. Am I correct that part of this example needs `remote_data`? If so, could we mark the whole example as `remote_data`? In the end, all we really care here is that if people copy & paste the example, it will work for them. It is not something that has to be tested on every documentation build.

saimn:
~10s (on my laptop) is much better than >1min, but I still think that we should avoid running this kind of example in the documentation build. But maybe that's a more general discussion to have, e.g. another example in #10794 is `stats/ripley` where a single example takes ~7s. If we want a faster doc build this is the kind of things we should avoid/limit.

pllim:
If astropy/pytest-astropy#21 moves forward towards a solution, we could possible mark these as "slow" and only run them in cron?

mhvk:
:+1: on some way to mark some doctests (and regular tests) as `slow` - ideally, we do not limit what we show in documentation just because of runtime, so it would be great to be able to skip some tests/examples normally but check once in a while that they still work.

But for now maybe we can finish this one by marking things `remote_data` (if indeed the main test needs remote data anyway) and merge?

StuartLittlefair:
Got to agree with @mhvk here - the perfect solution is for these tests not to be run on every doc build, but I think marking them as remote_data effectively does that, no?

saimn:
https://github.com/astropy/pytest-astropy/issues/21 and `remote_data` are only used when running doctests with pytest, not for Sphinx. And pytest does not run the ``.. plot::`` directives, I think.
Anyway, it's fine for me to merge this as it's already much better, and we can have another discussion about what should be run in the docs in another ticket.

mhvk:
OK, let's leave it at this then!  Thanks for the quick fix, @maxnoe!

mhvk (approved):

pllim:
Oh... right, my bad! Looking at https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/ext/duration.py , I don't see anyway to skip it. Maybe this is a feature request for Sphinx?

PR #10803, opened 2020-10-05
Store bboxes in child objects of WCSAxes

dstansby:
Previously `WCSAxes` had a `_bboxes` attribute. This list was passed down the call stack to various child `draw()` methods, which after drawing appended their `bboxes` to this list.

This has been changed here so that the child objects (axis labels and ticklabels) store their own bbox list, and `WCSAxes` now just takes all of those bboxes and adds them into one list to get all the `bboxes`. This avoids having to hand a list by reference down the call stack, and makes it much clearer where all of the bboxes are coming from.

### DIFF
--- a/astropy/visualization/wcsaxes/axislabels.py
+++ b/astropy/visualization/wcsaxes/axislabels.py
@@ -7,6 +7,8 @@ from matplotlib import rcParams
 from matplotlib.text import Text
 import matplotlib.transforms as mtransforms
 
+from astropy.utils.exceptions import AstropyDeprecationWarning
+from astropy.utils.misc import _NONE
 from .frame import RectangularFrame
 
 
@@ -30,6 +32,7 @@ class AxisLabels(Text):
         self.set_va('center')
         self._minpad = minpad
         self._visibility_rule = 'labels'
+        self._bboxes = []
 
     def get_minpad(self, axis):
         try:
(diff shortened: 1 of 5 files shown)


astropy/visualization/wcsaxes/ticklabels.py
```diff
@@ -227,7 +231,7 @@ def draw(self, renderer, bboxes, ticklabels_bbox, tick_out_size):
                 # that has a key starting bit such as -0:30 where the -0
                 # might be dropped from all other labels.
 
-                if not self._exclude_overlapping or bb.count_overlaps(bboxes) == 0:
```
dstansby, line 230: I don't believe the bit I have removed ever returned anything but `True`, as at this point in the drawing `bboxes` was always just an empty list (so there were never any overlaps with `bb`).

astropy/visualization/wcsaxes/axislabels.py
```diff
@@ -59,7 +60,7 @@ def set_visibility_rule(self, value):
     def get_visibility_rule(self):
         return self._visibility_rule
 
-    def draw(self, renderer, bboxes, ticklabels_bbox,
```
pllim, line 62: Is this API change? I'll let the subpackage maintainers decide.

If it is, this needs a change log, and possibly some sort of deprecation?

dstansby: Yes, possibly. I can always put the function signature back to what it was, and issue a deprecation warning that the parameter I've changed doesn't do anything.

dstansby:
Could someone with the right powers restart the circleCI build?

pllim:
CircleCI restarted.

astropy/visualization/wcsaxes/axislabels.py
dstansby: Not sure if there's another canonical way to deprecate a parameter, but I couldn't find anything in the docs so I've brewed my own solution. Let me know if there's a better way!

pllim: We have https://docs.astropy.org/en/latest/api/astropy.utils.decorators.deprecated_renamed_argument.html but @larrybradley found bugs with it before, so not sure if I should recommend it or not.

dstansby:
Heh, so it turns out this won't work because of some clever machienery needed to avoid overlapping ticks, sorry for wasting people's time.

pllim:
Thanks for trying!

PR #10805, opened 2020-10-05
TST: Fix error message regex due to upstream change

pllim:
### Description

This pull request is to address numpy-dev job failure due to a slight change in expected error message from

```
could not broadcast input array from shape (2,2) into shape (2)
```

to

```
could not broadcast input array from shape (2,2) into shape (2,)
```

Example log: https://travis-ci.org/github/astropy/astropy/jobs/733096867

**Note to reviewer:** Check the numpy-dev/remote-data job that is allowed to fail.

p.s. I am guessing this needs to be backported to LTS so its CI would pass with a future numpy release (1.20?). But if that is not the case, feel free to re-milestone.

### DIFF
--- a/astropy/coordinates/tests/test_sky_coord.py
+++ b/astropy/coordinates/tests/test_sky_coord.py
@@ -576,7 +576,7 @@ def test_insert_exceptions():
 
     # Bad shape
     with pytest.raises(ValueError, match='could not broadcast input array from '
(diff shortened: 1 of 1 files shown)


mhvk (approved):
Thanks, @pllim! Dev tests passed, so will merge. Agree with backport.

astrofrog:
I think ``test_insert_exceptions`` doesn't exist in v4.0.x so updating milestone

Issue #8703, opened 2019-05-14
tight_layout() doesn't work properly on plots with a WCSAxes projection


dstansby:
A Matpltolib `Axes` created with `projection=wcs` does not play well with Matplotlib's `tight_layout()` method, that is supposed to re-size elements of a figure such that there are no overlapping labels/axes etc.

Consdier the following example:
```python
import matplotlib.pyplot as plt

from astropy.wcs import WCS
from astropy.io import fits
from astropy.utils.data import get_pkg_data_filename

filename = get_pkg_data_filename('galactic_center/gc_msx_e.fits')

hdu = fits.open(filename)[0]
wcs = WCS(hdu.header)

fig = plt.figure(figsize=(8, 6))
n = 2
for i in range(n):
    projection = wcs
    # projection = None
    plt.subplot(n, 1, i + 1, projection=projection)
    # plt.imshow(hdu.data, vmin=-2.e-5, vmax=2.e-4, origin='lower')
    plt.grid(color='white', ls='solid')
    plt.xlabel('Galactic Longitude')
    plt.ylabel('Galactic Latitude')
    plt.title('WCSAxes plot')

fig.tight_layout()
plt.show()
```
**`projection=None`, no tight_layout:**
![Figure_1](https://user-images.githubusercontent.com/6197628/57730604-ba088700-7690-11e9-92b1-689c40cc3f76.png)
**`projection=None`, with tight_layout:**
![Figure_1](https://user-images.githubusercontent.com/6197628/57730646-d73d5580-7690-11e9-8875-63e92e5506ce.png)
(note how this re-organises things nicely)
**`projection='wcs'`, no tight_layout:**
![Figure_1](https://user-images.githubusercontent.com/6197628/57730714-fe942280-7690-11e9-98b2-c827b64aabaa.png)
**`projection='wcs'`, with tight_layout:**
![Figure_1](https://user-images.githubusercontent.com/6197628/57730759-123f8900-7691-11e9-9bbd-2e2fae288151.png)
Note how things actually get worse!

- astropy 3.1.2
- Matplotlib 3.0.3

pllim:
Is this related to https://github.com/astropy/astropy/issues/8678#issuecomment-491523170 ?

dstansby:
I don't think so, that is only a problem on Matplotlib dev (post version `3.0.3`), whereas this is an issue on a version of Maptlotlib (`3.0.3`) that predates that change.

astrofrog:
[...]

PR #10806, opened 2020-10-05
Refactor TickLabels drawing code in WCSAxes

dstansby:
xref https://github.com/astropy/astropy/issues/8703. This refactors code with the following aims:

- More methods with descriptive names makes the code more readable
- Move the logic for calculating x/y/alignments out of `draw()`, so eventually it can be used by a `get_tight_bbox()` method to calculate the bounding box without doing a draw
- Only recompute x/y/alignments when needed, checked using the `_stale` flag. This has the nice side effect that if any text parameters are changed (color, font weight, size etc.) that don't effect the text position, the positions are not re-computed (they were before)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -455,6 +455,10 @@ astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
 - Fixed an issue where ticks were sometimes not drawn at the edges of a spherical
   projection on a WCSAxes. [#10442]
+- The position of tick labels are now only calculated when needed. If any text
+  parameters are changed (color, font weight, size etc.) that don't effect the
+  tick label position, the positions are not recomputed, improving performance.
+  [#10806]
 
 astropy.wcs
 ^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -148,6 +148,10 @@ astropy.visualization
   quadrangle.  Unlike ``matplotlib.patches.Rectangle``, the edges of this
   patch will be rendered as curved lines if appropriate for the WCS
   transformation. [#10862]
+- The position of tick labels are now only calculated when needed. If any text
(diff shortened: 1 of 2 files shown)


Cadair:
@dstansby What's the status of this with regards to the other PRs now? is it just some more cleanups?

Cadair:
Otherwise this seems fine, needs a changelog mentioning the perf improvement?

dstansby:
This PR is just a refactor/cleanup; I should have mentioned that it's a pre-requisite for something like https://github.com/astropy/astropy/commit/e7c122aca3e8d21fad63480bfb8474592ae8fdab, which adds the functionality to correct calculate the bounding boxes of tick labels (I haven't quite had enough time to PR that yet though).

CHANGES.rst
```diff
@@ -425,6 +425,10 @@ astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
 - Fixed an issue where ticks were sometimes not drawn at the edges of a spherical
   projection on a WCSAxes. [#10442]
+- The position of tick labels are now only calculated when needed. If any text
```
Cadair, line 428:
```suggestion

- The position of tick labels are now only calculated when needed. If any text
```

Cadair:
would like @astrofrog to have a quick look at this as well.

astrofrog (approved):
Looks great, thanks!

astrofrog:
@dstansby @Cadair - what is the motivation for the 4.1.1 milestone? Why not 4.0.4 or 4.2?

astrofrog:
(I would assume it should be 4.0.4 if required for the bbox fixes that would be backported to 4.0.4)

dstansby:
🤷 doesn't matter to me, I think @Cadair milestoned

Cadair:
Yeah I think this should be 4.2 I think I was getting it confused with the other PRs.

PR #10809, opened 2020-10-07
ENH: Access table column as attribute

pllim:
### Description

This pull request is to allow accessing column as table attribute. @barentsen asked about this on Slack and I got curious. Not sure if this is too hacky or not. And somehow an expired leap seconds table prevented me from running tests locally on Windows, so I will have to rely on the CI.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,6 +91,9 @@ astropy.table
   a ``dtype`` from a numpy structured array. The list of ``names`` overrides the
   names specified in the ``dtype``. [#10419]
 
+- Allow accessing column through ``getattr``; i.e., ``t.a`` would work just as
+  ``t['a']``. [#10809]
+
 astropy.tests
 ^^^^^^^^^^^^^
 
@@ -435,7 +438,7 @@ astropy.wcs
 - Add .upper() to ctype or ctype names to wcsapi/fitwcs.py to mitigate bugs from
   unintended lower/upper case issues [#10557]
 - Added bounds to ``fit_wcs_from_points`` to ensure CRPIX is on
-  input image. [#10346] 
+  input image. [#10346]
 
 Other Changes and Additions
 ---------------------------
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -626,6 +626,8 @@ class Table:
         # user-supplied meta directly.
         if meta is not None:
             self.meta = deepcopy(meta) if copy else meta
+        else:
+            self.meta = {}
 
         # Whatever happens above, the masked property should be set to a boolean
         if self.masked not in (None, True, False):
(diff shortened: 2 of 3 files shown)


taldcroft:
I am vigorously opposed to this idea. Using item access improves:

- Readability: if you look at something like `t['name']` you know instantly from the syntax that is a column access operation.  However, `t.name` might be a column or it might be an actual attribute or method.
- Reliability: imagine that we adopt this convention and many codes start using it. There are lots of data tables with common words like `name` or `id`, etc. Then we decide that `Table` should get an attribute `name` or `id`, and this suddenly breaks all those codes. In other words, just because there is not a namespace clash today does not mean there won't be a namespace clash later.

Saving typing four characters is not worth reducing code readability and reliability.

dhomeier:
Looks like you might relatively easily subclass `Table` to get that functionality, but that still leaves the problem of breakage whenever new class attributes are added.

Should the leap second problem sort itself out with time, or will it require action on this side? There was a similar issue on the list, which seemed to solve itself as the day proceeded.

pllim:
We could create a special subclass, yes, but I doubt people would use it much, as `Table` or `QTable` is the advertised usage. Then, it just adds maintenance cost with little return, plus all the breakage that @taldcroft mentioned. Shall we close without merge then?

pllim:
`astropy.utils.exceptions.AstropyWarning: leap-second auto-update failed due to the following exception: IERSStaleWarning('leap-second file is expired.')`

Ah, so it is not just Windows. @mhvk , I forgot how to fix this. Do we need to update the package leap seconds table or something?

taldcroft:
I'm in favor of closing this, having now had the discussion for the record. If people want to make their own subclass that's fine, but I am likewise strongly opposed to having this subclass be part of astropy core.

barentsen:
For background: I raised the question on Slack because we recently refactored the `LightCurve` object in Lightkurve v2.0 to become a subclass of `TimeSeries`, and ended up overriding `__getattr__` to support attribute access to columns for the purpose of backwards-compatibility. (Lightkurve v1.x stored all data in properties, and we didn't want to break all existing code.)

I hear the readability and reliability concerns raised above.  They are valid!

Just to record an argument in favor of attribute access: I think there is value in having the API of `Table` mimic the API of pandas `DataFrame` as much as reasonably possible, to reduce the times a user needs to worry about whether they are using one or the other.

I am personally addicted to the convenience of attribute access, but that may tell you more about the laziness of my fingers than about the wisdom of that design decision.

mhvk:
@pllim - last update was #10617 - I'm surprised it is needed already!?

pllim:
Or maybe this patch broke something in a vague way. Let's wait and see if `master` fails...

PR #10814, opened 2020-10-08
Simplify prepare_earth_position_vel, removes calling epv00 twice, fixes #10813

### LINKED ISSUE #10813: prepare_earth_position_vel calls erfa.epv00 twice unnecessarily
### Description

When transforming CIRS to ICRS, the most expensive operation is calculating earth position vector using
`erfa.epv00`.

This function is called twice with the same input as result of how `prepare_earth_position_vel` works.

### Expected behavior

Don't do this expensive calculation twice.

### Steps to Reproduce

Run in ipython:

```
In [1]: from astropy.coordinates import SkyCoord, CIRS

In [2]: from astropy.time import Time

In [3]: import numpy as np

In [4]: import astropy.units as u

In [5]: t = Time('2020-10-01T20:00') + np.random.uniform(0, 1, size=100000) * u.min

In [6]: icrs = SkyCoord.from_name('Crab')

In [7]: %%prun -s tottime
    ...: icrs.transform_to(CIRS(obstime=t))
```

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        2    8.897    4.448    8.897    4.448 core.py:4976(epv00)
        1    5.157    5.157    5.157    5.157 core.py:10571(pnm06a)
        1    0.891    0.891    0.891    0.891 core.py:15917(dtdb)
        1    0.166    0.166    0.166    0.166 core.py:11280(s06)
        2    0.051    0.026    0.051    0.026 core.py:17216(utctai)
        1    0.049    0.049    0.049    0.049 core.py:16342(taiutc)
        1    0.012    0.012    0.012    0.012 core.py:972(apci)
        1    0.005    0.005    0.005    0.005 core.py:19849(c2s)
        9    0.005    0.001    0.005    0.001 utils.py:118(two_sum)
```

Observe that `epv00` is called twice.

maxnoe:
### Description

This simplifies `prepare_earth_position_vel` to simply call `erfa.epv00`, which already provides everything needed.

Fixes #10813

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,9 @@ astropy.coordinates
 - Method ``.realize_frame`` from coordinate frames now accepts ``**kwargs``,
   including ``representation_type``. [#10727]
 
+- Avoid an unnecessary call to ``erfa.epv00`` in transformations between
+  ``CIRS`` and ``ICRS``, improving performance by 50 %. [#10814]
+
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
--- a/astropy/coordinates/builtin_frames/utils.py
+++ b/astropy/coordinates/builtin_frames/utils.py
@@ -343,18 +343,38 @@ def prepare_earth_position_vel(time):
         Heliocentric position of Earth in au
     """
     # this goes here to avoid circular import errors
-    from astropy.coordinates.solar_system import (get_body_barycentric, get_body_barycentric_posvel)
+    from astropy.coordinates.solar_system import (
+        get_body_barycentric,
+        get_body_barycentric_posvel,
+        solar_system_ephemeris,
+    )
     # get barycentric position and velocity of earth
-    earth_p, earth_v = get_body_barycentric_posvel('earth', time)
 
-    # get heliocentric position of earth, preparing it for passing to erfa.
-    sun = get_body_barycentric('sun', time)
-    earth_heliocentric = (earth_p -
-                          sun).get_xyz(xyz_axis=-1).to_value(u.au)
+    ephemeris = solar_system_ephemeris.get()
 
-    # Also prepare earth_pv for passing to erfa, which wants it as
-    # a structured dtype.
-    earth_pv = pav2pv(
-        earth_p.get_xyz(xyz_axis=-1).to_value(u.au),
-        earth_v.get_xyz(xyz_axis=-1).to_value(u.au/u.d))
+    # if we are using the builtin erfa based ephemeris,
+    # we can use the fact that epv00 already provides all we need.
+    # This avoids calling epv00 twice, once
+    # in get_body_barycentric_posvel('earth') and once in
+    # get_body_barycentric('sun')
+    if ephemeris == 'builtin':
+        jd1, jd2 = get_jd12(time, 'tdb')
+        earth_pv_heliocentric, earth_pv = erfa.epv00(jd1, jd2)
+        earth_heliocentric = earth_pv_heliocentric['p']
+
+    # all other ephemeris providers probably don't have a shortcut like this
+    else:
+        earth_p, earth_v = get_body_barycentric_posvel('earth', time)
+
+        # get heliocentric position of earth, preparing it for passing to erfa.
+        sun = get_body_barycentric('sun', time)
+        earth_heliocentric = (earth_p - sun).get_xyz(xyz_axis=-1).to_value(u.au)
+
+        # Also prepare earth_pv for passing to erfa, which wants it as
+        # a structured dtype.
+        earth_pv = pav2pv(
+            earth_p.get_xyz(xyz_axis=-1).to_value(u.au),
+            earth_v.get_xyz(xyz_axis=-1).to_value(u.au / u.d)
+        )
+
     return earth_pv, earth_heliocentric


mkbrewer:
You need to check to see if a JPL ephemeris is in use before doing this. When a user sets the solar system ephemeris to be a JPL ephemeris, they expect the chosen ephemeris to actually be used. Not only will this likely take longer, but it will spoil the accuracy also. I have a program that I use for debugging astropy and we are currently in agreement on the distance to the Moon to within 5 cm using the DE430 ephemeris. Using this will likely cause an error of several kilometers.

mhvk:
This is a nice catch, but as @mkbrewer notes it is only OK for the built-in case. That said, it may well be that also the JPL case can be sped up by looking more carefully at what is done in the various `get_body` functions. Indeed, perhaps it is better to move the function to `solar_system` - that would make it clearer that it can depend on ephemeris and gives more freedom to use/define possible private methods there.

maxnoe:
I am not sure I understand the point here.

Where should what be checked?

mhvk:
@maxnoe - in `get_body_barycentric_posvel`, which is used in `prepare_earth_position_vel`, the position and velocity of the Earth get looked up either via `epv00`, if the solar system ephemeris is the built-in one (default), or via `jplephem` using (much more accurate) JPL ephemerides - see code around https://github.com/astropy/astropy/blob/89580eee52dbe06e54fd51abe69ba0b0ecefc33a/astropy/coordinates/solar_system.py#L223 (`kernel is None` is the built-in case).

So, the problem with your short-cut is that for people who are using the JPL ephemerides - such as is the case, e.g., for pulsar timing, the answer would change as suddenly the less accurate built-in ephemeris is used.

So, at some place the ephemeris in use should be checked. My sense is that this is more logically done inside `solar_system.py`, so that would suggest to create a new function there that just returns either the output of `epv00` if one is using the built-in ephemeris, or the equivalent from the JPL ephemerides (for a new function, I would return both heliocentric pos and vel). But it might also work to do it on your `erfa_astrom` classes, since those are the only ones that actually use `prepare_earth_position_vel`.

mkbrewer:
Well the simplest thing to do is: 
```
ephemeris = solar_system_ephemeris.get()
if ephemeris == 'builtin':
    #execute your new algorithm
else:
    #execute the original algorithm
```

maxnoe:
Ok, seems simple enough for a 50 % performance gain, will change

mhvk:
@mkbrewer's suggestion would definitely work too - though I'd prefer to move the functionality out of `utils` to either `solar_system` or `erfa_astrom` so that we spread it over 2 rather than 3 places. I think `solar_system` would be most appropriate if it turns out that also for the JPL ephemeris case, there is a short-cut for getting earth bary and heliocentric together; otherwise, `erfa_astrom` may be more appropriate. But we can also do a move in a subsequent refactoring PR.

mhvk:
And, yes, we should get that 50% performance gain!!

mkbrewer:
There is no short cut for the JPL ephemeris that I know of. One must get the position of the Earth and Sun wrt the solar system barycenter, then subtract the position of the Sun from that of the Earth to get the heliocentric position of the Earth.

mkbrewer:
btw, I took a look at your `ErfaAstromInterpolator` class which is new to me. There is definitely a bug in the `apcs()` method here: 
```
        if obstime.size == 1:
            return super().apci(frame_or_coord)
```
That should be `super().apcs` rather than `super().apci`.

mkbrewer:
Also, that should really be: `if obstime.size < 3`

mhvk:
Well, goes to show the risk of copy & paste and the quality of my review... And the risk of special-casing items that may hence not be tested as thoroughly.

pllim:
So... Is this PR still necessary? Thanks!

maxnoe:
I would say it is. Since there is an easy performance gain of 50 % for the default astropy behaviour.

pllim:
In that case, probably need a change log and also fix the CI. Thanks!

mhvk:
Indeed, a 50% gain is worth noting in the changelog!!

maxnoe:
@mhvk since this method returns unitless arrays in the structured format only helpful for calling erfa functions, I don't really see it helpful as a public method in `solar_system`. Would you agree on that?

maxnoe:
I think is more appropriate in `erfa_astrom` since it basically proces `epv00` or an equivalent implementation using a different ephemerides provider.

mhvk:
@maxnoe - yes, your argument for `erfa_astrom` makes sense; the only real reason I could think for `solar_system` was that in that case the check of the ephemeris would be more local. 

I think generally we can change things that are not exported up (i.e., not in an `__all__` and taken up to `astropy.coordinates`), but in an ideal world we would still have it be present in `utils` as well, but with a deprecation warning.

That said, this is clearly something that can be split in two steps: get the performance improvement, and move to a more logical place. I'm happy to go with that scheme too.

maxnoe:
@mhvk so should we for now keep it like this or should I already move to `erfa_astrom`?

I added the changelog entry

mkbrewer:
If `prepare_earth_position_vel()` is moved to `erfa_astrom`, then it would make sense to move `get_cip()` also.

mkbrewer:
Oh, never mind. I just realized that if CIRS is fixed along the lines of TETE, then the call to `get_cip()` will no longer be needed in `erfa_astrom`.

mkbrewer:
Oh darn, here I am being OT again. Sorry.

mhvk (approved):
@maxnoe - the PR looks good now. Since it does, I think we might as well merge - we can do the move in a subsequent PR, maybe as part of a bit larger cleanup. Thanks!!

PR #10818, opened 2020-10-08
Numpy dev fixes for subarray dtype

### LINKED ISSUE #10815: numpy-dev using a dtype with a subarray field is deprecated
- [x] Remember to undo #10863 ! See #10966

### Description

@astrofrog reported `DeprecationWarning: using a dtype with a subarray field is deprecated.` warnings using numpy-dev in v4.0.x ([example log](https://travis-ci.org/github/astropy/astropy/jobs/734002575)). Since v4.0.x is installing numpy from master, while master installs from "nightly" (i.e., weekly) wheel, we won't see it until numpy's wheel is updated.

Caused by numpy/numpy#17419

### Steps to Reproduce

I think this might trigger the warning:

```python
>>> import numpy as np
>>> field = tree.Field(None, name='c', datatype='floatComplex', arraysize='2x3*', config=config)
>>> c._base.format
'(2,)c8'
>>> default = np.array([], dtype=c._base.format)  # Warning?
```
```
E       DeprecationWarning: using a dtype with a subarray field is deprecated.
This can lead to inconsistent behaviour due to the resulting dtype being different from the input dtype.
You may try to use `dtype=dtype.base`, which should give the same result for most inputs,
but does not guarantee the output dimensions to match the subarray ones.  (Deprecated NumPy 1.20)
```

In this case, `c._base.format` is a string, so I am not sure what `dtype.base` means in this context. @mhvk , any advise?

mhvk:
fixes #10815 

Not very happy with the fix for `Column`, but could not think of a better way. Ideally, this gets fixed better by numpy. See https://github.com/numpy/numpy/issues/17511.

### DIFF
--- a/astropy/io/votable/converters.py
+++ b/astropy/io/votable/converters.py
@@ -493,7 +493,7 @@ class VarArray(Array):
         Array.__init__(self, field, config)
 
         self._base = base
-        self.default = np.array([], dtype=self._base.format)
+        self.default = np.empty(0, dtype=self._base.format)
 
     def output(self, value, mask):
         output = self._base.output
--- a/astropy/table/column.py
+++ b/astropy/table/column.py
@@ -383,6 +383,13 @@ class BaseColumn(_ColumnGetitemShim, np.ndarray):
             if meta is None:
                 meta = data.info.meta
 
+        elif isinstance(data, list) and data == []:
+            # We should special case the empty list since np.array([], dtype)
+            # does not return an empty array with the requested dtype, and in
+            # numpy 1.20-dev this is giving a DeprecationWarning.
+            # TODO: Adjust when/if the numpy behaviour behaviour changes.
+            # Hopefully, this clause can again be removed.  NUMPY_LT_1_20.
+            self_data = np.empty(0, dtype=dtype)
         else:
             if np.dtype(dtype).char == 'S':
                 data = cls._encode_str(data)


pllim:
I thought I saw multiple occurrence of the thingy in votable but I'll double check again to be sure. If nothing is amiss, I'll approve. Thank you so much!

astropy/table/column.py
```diff
@@ -383,6 +383,13 @@ def __new__(cls, data=None, name=None,
             if meta is None:
                 meta = data.info.meta
 
+        elif isinstance(data, list) and data == []:
+            # We should special case the empty list since np.array([], dtype)
+            # does not return an empty array with the requested dtype, and in
+            # numpy 1.20-dev this is giving a DeprecationWarning.
+            # TODO: Adjust when/if the numpy behaviour behaviour changes.
+            # Hopefully, this clause can again be removed.  NUMPY_LT_1_20.
+            self_data = np.empty(0, dtype=dtype)
```
seberg, line 392: One other solution that is possible if you know that the list represents effectively a 1-D dataset is to use:
`np.empty(len(data), dtype=dtype)` and then assign to that result.  That does have slightly different semantics though.

mhvk: Yes, that might indeed work. But I think the standard case should just go through `np.array(...)`, since `Column` is an ndarray subclass.

The whole thing does make one wonder whether the new behaviour in numpy is in fact an improvement, since it doesn't seem to move to the "more logical" state - re-opening the numpy issue.

mhvk:
I've temporarily converted this to a draft, pending further discussion over at the numpy issue.

astropy/io/votable/converters.py
```diff
@@ -493,7 +493,7 @@ def __init__(self, field, base, arraysize, config=None, pos=None):
         Array.__init__(self, field, config)
 
         self._base = base
-        self.default = np.array([], dtype=self._base.format)
+        self.default = np.empty(0, dtype=self._base.format)
```
pllim, line 496: Looks like this is the only case where `[]` is used with `self._base.format`. I think we should wait till Numpy wheel picks up that patch (and hence fails our numpy-dev job on `master`) and then make sure the numpy-dev job here passes with these changes. Otherwise, LGTM on the `io.votable` side. Thank you!

astropy/table/column.py
```diff
@@ -383,6 +383,13 @@ def __new__(cls, data=None, name=None,
             if meta is None:
                 meta = data.info.meta
 
+        elif isinstance(data, list) and data == []:
+            # We should special case the empty list since np.array([], dtype)
+            # does not return an empty array with the requested dtype, and in
```
taldcroft, line 388: I guess I'm behind the curve on this issue, but this statement confuses me.
```
In [2]: np.array([], dtype=np.int32)                                            
Out[2]: array([], dtype=int32)
```
Is that not an empty array with the requested dtype?

seberg: @taldcroft it is "subarray" dtypes, which are a bit strange, e.g.:

```
In [4]: dt = np.dtype("(3)i,")                                                                                          

In [5]: dt.shape                                                                                                        
Out[5]: (3,)

In [6]: dt.ndim                                                                                                         
Out[6]: 1

In [7]: dt.base                                                                                                         
Out[7]: dtype('int32')

In [8]: np.array([], dtype=dt)                                                                                          
Out[8]: array([], shape=(0, 3), dtype=int32)
```

taldcroft: OK, that makes sense. I think for this PR then we just need the comment to be even longer, e.g. " .. does not return an empty array with the requested dtype in certain cases like `dtype="(3)i,"` (see https://github.com/numpy/numpy/issues/17511).  In numpy 1.20-dev ...".

pllim:
The failure is on `master` now (example log: https://travis-ci.org/github/astropy/astropy/jobs/734945598). I restarted the numpy-dev job here. FYI.

pllim:
On `master`:
`Numpy: 1.20.0.dev0+07a54df` failed

Here:
`Numpy: 1.20.0.dev0+07a54df` passed 🎉

seberg:
Well, I will try to fix up NumPy to reduce the noise level of this change (for astropy, probably that effectively means no noise), but if the noise comes it might be more annoying :) (but also indicate buggy behaviour).

pllim:
@seberg , thank you! Please keep us updated. 😄

pllim:
@mhvk , does this mean we can close this without merge?

mhvk:
Yes, hopefully, but let's do it only once our numpy-dev tests pass again

PR #10819, opened 2020-10-08
Fix empty Column initialization for structured dtype.

mhvk:
This failed when a shape was passed in as well.

Found while writing #10818


mhvk:
Codecov failure is a red herring.

taldcroft (approved):
Looks good!

PR #10822, opened 2020-10-09
Made convolve work with units, closing issue #10811

### LINKED ISSUE #10811: Make convolve work with units
### Description

It would be great if one could properly convolve arrays with units. Currently, the units are stripped off:
```python
>>> import numpy as np
>>> import astropy.units as u
>>> from astropy.convolution import convolve
>>> convolve(np.array([1, 4, 5, 6, 5, 7, 8])*u.ph, [0.2, 0.6, 0.2])
array([1.4, 3.6, 5. , 5.6, 5.6, 6.8, 6.2])
```
As the workaround here is simple (`convolve(a.value, kernel) * a.unit`), it is not urgent at all. User code would ofcourse look better if this would be handled by convolve itself.

george-galvin:
### Description

Both convolve and convolve_fft now output a Quantity array, if the user
 input is one.

EDIT: Fixes #10811

### DIFF
--- a/astropy/convolution/convolve.py
+++ b/astropy/convolution/convolve.py
@@ -353,6 +353,10 @@ def convolve(array, kernel, boundary='fill', fill_value=0.,
         result[initially_nan] = np.nan
 
     # Convert result to original data type
+    array_unit = getattr(passed_array, "unit", None)
+    if array_unit is not None:
+        result <<= array_unit
+
     if isinstance(passed_array, Kernel):
         if isinstance(passed_array, Kernel1D):
             new_result = Kernel1D(array=result)
@@ -567,6 +571,9 @@ def convolve_fft(array, kernel, boundary='fill', fill_value=0.,
     if nan_treatment not in ('interpolate', 'fill'):
         raise ValueError("nan_treatment must be one of 'interpolate','fill'")
 
+    #Get array quantity if it exists
+    array_unit = getattr(array, "unit", None)
+
     # Convert array dtype to complex
     # and ensure that list inputs become arrays
     array = _copy_input_if_needed(array, dtype=complex, order='C',
@@ -758,6 +765,9 @@ def convolve_fft(array, kernel, boundary='fill', fill_value=0.,
         raise ValueError("Encountered NaNs in convolve.  This is disallowed.")
 
     fftmult *= kernel_scale
+
+    if array_unit is not None:
+        fftmult <<= array_unit
 
     if return_fft:
         return fftmult
--- a/astropy/convolution/tests/test_convolve.py
+++ b/astropy/convolution/tests/test_convolve.py
@@ -43,6 +43,7 @@ except ImportError:
 
 
 class TestConvolve1D:
+
     def test_list(self):
         """
         Test that convolve works correctly when inputs are lists
@@ -64,6 +65,26 @@ class TestConvolve1D:
         z = convolve(x, y, boundary=None)
         assert_array_almost_equal_nulp(z,
             np.array([0., 3.6, 5., 5.6, 5.6, 6.8, 0.]), 10)
+
+    @pytest.mark.parametrize(('boundary', 'nan_treatment',
+                              'normalize_kernel', 'preserve_nan', 'dtype'),
+                             itertools.product(BOUNDARY_OPTIONS,
+                                               NANHANDLING_OPTIONS,
+                                               NORMALIZE_OPTIONS,
+                                               PRESERVE_NAN_OPTIONS,
+                                               VALID_DTYPES))
+    def test_quantity(self, boundary, nan_treatment,
+                      normalize_kernel, preserve_nan, dtype):
+        """
+        Test that convolve works correctly when input array is a Quantity
+        """
+
+        x = np.array([1, 4, 5, 6, 5, 7, 8], dtype=dtype) * u.ph
+        y = np.array([0.2, 0.6, 0.2], dtype=dtype)
+        z = convolve(x, y, boundary=boundary, nan_treatment=nan_treatment,
+                          normalize_kernel=normalize_kernel, preserve_nan=preserve_nan)
+
+        assert x.unit == z.unit
 
     @pytest.mark.parametrize(('boundary', 'nan_treatment',
                               'normalize_kernel', 'preserve_nan', 'dtype'),
--- a/astropy/convolution/tests/test_convolve_fft.py
+++ b/astropy/convolution/tests/test_convolve_fft.py
@@ -8,6 +8,7 @@ from numpy.testing import assert_allclose, assert_array_almost_equal_nulp
 
 from astropy.convolution.convolve import convolve_fft, convolve
 from astropy.utils.exceptions import AstropyUserWarning
+from astropy import units as u
 from astropy.utils.compat.context import nullcontext
 
 VALID_DTYPES = ('>f4', '<f4', '>f8', '<f8')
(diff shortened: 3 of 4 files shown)


astropy/convolution/convolve.py
```diff
@@ -190,6 +190,12 @@ def convolve(array, kernel, boundary='fill', fill_value=0.,
     passed_kernel = kernel
     passed_array = array
 
+    #Get array quantity if exists
```
pllim, line 193:
```suggestion
    # Get array quantity if exists
```

astropy/convolution/convolve.py
```diff
@@ -364,16 +370,17 @@ def convolve(array, kernel, boundary='fill', fill_value=0.,
         new_result._separable = passed_array._separable
         if isinstance(passed_kernel, Kernel):
             new_result._separable = new_result._separable and passed_kernel._separable
-        return new_result
+        return new_result * passed_quantity
     elif array_dtype.kind == 'f':
         # Try to preserve the input type if it's a floating point type
         # Avoid making another copy if possible
+        result *= passed_quantity
```
pllim, line 377: Does this make a copy, and thus contradict the comment above it?

pllim (commented):
Thanks! Would need test(s).

I am also not very sure about the unnecessary multiplication when `passed_quantity` is 1. Perhaps some benchmark on performance comparing this patch and `master`? I'll defer to `convolution` maintainer(s)'s decisions.

astropy/convolution/convolve.py
george-galvin: Thanks for the review! I agree the multiplication by 1 looks redundant - the only alternative I can see is adding a conditional to each return statement, unless there's another way? As for the *=, I wasn't aware that created another copy?

pllim: I am not sure either. I think that depends on the implementation details in `__rmul__`.

keflavich: `*=` should be inplace, so I'm not worried about this, but line 373 above might create a copy.

Returning a ternary (`new_result * passed_quantity if <passed_quantity is a unit> else new_result`) might be better than this unneeded multiplication operation

astropy/convolution/convolve.py
```diff
@@ -364,16 +370,17 @@ def convolve(array, kernel, boundary='fill', fill_value=0.,
         new_result._separable = passed_array._separable
         if isinstance(passed_kernel, Kernel):
             new_result._separable = new_result._separable and passed_kernel._separable
-        return new_result
+        return new_result * passed_quantity
```
keflavich, line 373: this likely does create a copy

mhvk (commented):
The present method is not bad, but it does needlessly multiply an array with `1` if a non-quantity is passed in.

My sense would be something like:
```
unit = getattr(passed_array, "unit", None)
...
if unit is not None:
    result = result << unit
```
The latter just creates a view and attaches the unit.

Alternatively, since we support `__array_function__`, one could follow what is already done for `np.convolve` (https://github.com/numpy/numpy/blob/2262a7d7a961b1e9322ac1088a18283ed8ec32a0/numpy/core/numeric.py#L754), i.e., introduce a wrapper and define the quantity handler in `units/quantity_helpers/function_helpers.py`, following https://github.com/astropy/astropy/blob/625c6e6257af40e4b5c1070c9b3f707e7b0953dd/astropy/units/quantity_helper/function_helpers.py#L578-L583

The latter has the advantage that if, e.g., someone downstream wants to use `convolve` with their own array-like class (say, dask), then they can easily override the behaviour.

Looking at the code for `np.convolve` also reminds me: in principle, the kernel can have a unit too...

george-galvin:
Okay, I've implemented @mhvk's solution, which is way nicer looking and solves all the problems above.

pllim (commented):
Still need test(s). And perhaps squash the commits? Thanks!

george-galvin:
Ok, added two tests and squashed into one commit

astropy/convolution/tests/test_convolve_fft.py
```diff
@@ -69,6 +70,18 @@ def assert_floatclose(x, y):
 
 class TestConvolve1D:
 
+    def test_quantity(self):
```
pllim, line 73: Does this need a `@pytest.mark.parametrize(option_names, options)` like the other tests?

astropy/convolution/tests/test_convolve.py
```diff
@@ -65,6 +65,18 @@ def test_tuple(self):
         assert_array_almost_equal_nulp(z,
             np.array([0., 3.6, 5., 5.6, 5.6, 6.8, 0.]), 10)
 
+    def test_quantity(self):
```
pllim, line 68: Does this need parameterization like the other tests?
```
    @pytest.mark.parametrize(('boundary', 'nan_treatment',
                              'normalize_kernel', 'preserve_nan', 'dtype'),
                             itertools.product(BOUNDARY_OPTIONS,
                                               NANHANDLING_OPTIONS,
                                               NORMALIZE_OPTIONS,
                                               PRESERVE_NAN_OPTIONS,
                                               VALID_DTYPES))
```

george-galvin: I used the first couple of functions on test_convolve.py as a template, and so thought parameterizing it would be unnecessary - but this could be added instead of the equality check.

pllim: @keflavich , what do you think?

keflavich: I'd default to parametrizing the same way.  In principle, there's no interaction between what's being tested and the other parameters, but that's sort of the point of the test - make sure that's true and we never regress.

pllim:
Thanks for your patience! RTD and CircleCI failures don't seem related. I restarted Travis CI, just in case.

@keflavich and @mhvk , would you be able to do a final review and approve if it looks fine?

astropy/convolution/convolve.py
```diff
@@ -190,6 +190,9 @@ def convolve(array, kernel, boundary='fill', fill_value=0.,
     passed_kernel = kernel
     passed_array = array
 
+    # Get array quantity if exists
+    passed_quantity = getattr(passed_array, "unit", None)
```
mhvk, line 194: Nitpicky, but I think it would be clearer in this variable is called `array_unit` (or something similar)

astropy/convolution/convolve.py
```diff
@@ -353,6 +356,9 @@ def convolve(array, kernel, boundary='fill', fill_value=0.,
         result[initially_nan] = np.nan
 
     # Convert result to original data type
+    if passed_quantity is not None:
```
mhvk, line 359: Looking at this again, I think since we keep on purpose a reference to the input array, we might as well look for the unit only here, i.e., something like
```
unit = getattr(passed_array, 'unit', None)
if unit is not None:
    result <<= unit
```
(and the same for the other routine)

mhvk (commented):
Looks good modulo a suggestion to put things together (which I think is possible; if that is wrong, then a suggestion for a name change to make things a bit clearer)

mhvk:
p.s. The travis windows test failure also seems unrelated - what is happening!?!?

pllim:
Re: Windows CI -- No idea! Started happening today or so. Trying things out at #10842

astropy/convolution/convolve.py
george-galvin: Good points! In convolve_fft a reference to the input array isn't kept, so I didn't change the positioning of the line which looks for the unit. I've changed everything else you pointed out though.

mhvk (approved):
Thanks! Looks good to me.

george-galvin:
Thanks! Does this get merged now?

mhvk:
Yes, it will get merged! I just thought I'd give others a chance to have a last look, and then forgot... Thanks!!

olebole:
Thank you very much @george-galvin! I am very impressed by how quick my feature request is resolved! That is great work!

george-galvin:
Great, thanks for helping with my first PR guys :)

PR #10825, opened 2020-10-09
Fix Azure sdist job

astrofrog:
I backported the Azure configuration to build the wheels to v4.0.x - however things don't work out of the box because v4.0.x doesn't use pyproject.toml for example. This PR makes some small infrastructure updates to allow things to work, including adding a simple pyproject.toml file. This means source installations will now be built with build isolation (which is a good thing) but none of  this should have any impact on users.


Cadair (approved):
Seems sane to me :)

PR #10826, opened 2020-10-10
Remove unnecessary step in release and fix rst syntax

astrofrog:
Installing the build package is not necessary since the source distribution is never built locally with the current workflow


astrofrog:
(I need to check the RTD rendering before this is merged)

pllim:
So, this step doesn't need `build`? (I don't `tox`, so just wanna make sure.)
```
$ pip install tox --upgrade
$ TEST_READ_HUGE_FILE=1 tox -e test-alldeps -- --remote-data=any
```

astrofrog:
Nope, that step doesn't require ``build`` to be installed

pllim (approved):
LGTM then. Thanks!

PR #10828, opened 2020-10-10
Use IAU 2006 matrix for converting to true coordinates.

### LINKED ISSUE #10827: Inconsistent BPN matrix used in _apparent_position_in_true_coordinates()
The BPN matrix used in `coordinates.solar_system._apparent_position_in_true_coordinates()` is inconsistent with that used in `builtin_frames.utils.get_cip()`. This causes a small error in the apparent position with respect to the true equator and equinox of date when comparing it with the same coordinates obtained by subtracting the equation of the origins from astropy's CIRS right ascension. Now I know that this is an internal function, but it is currently the only way to get the apparent position with respect to the true equator and equinox of date until the appropriate frame is implemented (see issue #10230). Besides, fixing this is trivial. It involves changing only one line of code. From: 
```
    _, _, _, _, _, _, _, rbpn = erfa.pn00a(jd1, jd2)
```
to:
```
    rbpn = erfa.pnm06a(jd1, jd2)
```
This bug is so obvious that I feel that it needs no further documentation. It can easily be verified by mere inspection of `builtin_frames.utils.get_cip()`

mhvk:
This is a private function, used only for tests inside astropy, but also used at times outside of astropy.  As noted by @mkbrewer, it should be made consistent with what is used elsewhere in astropy, in particular in `coordinates.builtin_frames.utils.get_cip`.

I milestoned it 4.0.3, since one might as well get consistency for all versions we still support, but as this is a private function, I don't think a changelog entry is appropriate. Its use outside of astropy will become unnecessary when #10230 is addressed.  

fixes #10827

### DIFF
--- a/astropy/coordinates/solar_system.py
+++ b/astropy/coordinates/solar_system.py
@@ -518,6 +518,8 @@ def _apparent_position_in_true_coordinates(skycoord):
     are defined w.r.t to the true equinox and poles of the Earth
     """
     jd1, jd2 = get_jd12(skycoord.obstime, 'tt')
-    _, _, _, _, _, _, _, rbpn = erfa.pn00a(jd1, jd2)
+    # Classical NPB matrix, IAU 2006/2000A
+    # (same as in builtin_frames.utils.get_cip).
+    rbpn = erfa.pnm06a(jd1, jd2)
     return SkyCoord(skycoord.frame.realize_frame(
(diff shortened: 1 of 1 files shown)


StuartLittlefair (approved):
Thanks @mhvk . I’d been holding up on addressing that until I could implement the TETE frame, but it’s good to be consistent in earlier releases too.

PR #10831, opened 2020-10-11
Ensure possible astropy/_dev gets ignored in tests

mhvk:
Follow-up on #10774, to ensure tests also run with `tox -e test --develop`.
As suggested by @astrofrog in https://github.com/astropy/astropy/pull/10774#issuecomment-706701886
I confirmed that this worked.

I guess a question is whether we should change one of the travis test runs to use `--develop` so we are sure this mode of working remains supported?

Milestoning 4.1.1 rather than 4.1 since it does not affect the release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,6 +118,7 @@ matrix:
           name: Python 3.8 with all optional dependencies
           stage: Initial tests
           env: TOXENV="py38-test-alldeps"
+               TOXARGS="-v --develop"
                TOXPOSARGS="--durations=50"
           compiler: clang
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -100,6 +100,7 @@ norecursedirs =
     "docs[\/]generated"
     "astropy[\/]extern"
     "astropy[\/]_erfa"
+    "astropy[\/]_dev"
 astropy_header = true
 doctest_plus = enabled
 text_file_format = rst


astrofrog (approved):

mhvk:
Moved one travis test to use `--develop`. Should be OK to merge if all tests pass.

mhvk:
@astrofrog - I added a `--develop` test to travis. Things passed, so I think it is all OK, but letting you push the merge button just in case you want to have another look.

saimn:
The problem is that `--develop` does not really work. It works once but then will crash at the second tox execution (#10303). So this is not a problem for CI but I'm not sure if we should advertise the use of `--develop`.

mhvk:
@saimn - it may not be really foolproof but I do use it all the time - I can live with `git clean -fxd` before starting on a new branch, but am really keen to avoid the extra overhead for running `tox` instead of direct `pytest` in edit and test cycles in a given branch...

So, I'd love to avoid breakage in the future.

That said, it is more important to have the fix than to also have a regression test, so if you really prefer, I can remove the edit to `.travis.yml`.

mhvk:
p.s. And of course you had linked to an issue I raised myself, so I hardly needed to explain again!

saimn:
@mhvk - I have no strong feelings on this, it may be useful to have a build with `--develop` to make sure it works even if not perfectly.

astrofrog:
I think it does make sense to test --develop so we can know it at least works on the first execution for now?

astrofrog (approved):
With that in mind I'll approve this! Will leave open for a bit in case anyone disagrees.

pllim (approved):

Cadair (approved):
Seems reasonable to me, the test will at least catch potential issues.

PR #10833, opened 2020-10-12
Fix interpolator

maxnoe:
Adresses a copy paste error and a general comment by @mhvk and @mkbrewer by completely removing the special casing in the erfa interpolator.

See also the initial discussion in #10647 on this special casing.

### DIFF
--- a/astropy/coordinates/erfa_astrom.py
+++ b/astropy/coordinates/erfa_astrom.py
@@ -213,10 +213,6 @@ class ErfaAstromInterpolator(ErfaAstrom):
             For this function, a CIRS frame is expected.
         '''
         obstime = frame_or_coord.obstime
-        # no point in interpolating for a single value
-        if obstime.size == 1:
-            return super().apci(frame_or_coord)
-
         support = self._get_support_points(obstime)
 
         cip = self._get_cip(support, obstime)
@@ -239,9 +235,6 @@ class ErfaAstromInterpolator(ErfaAstrom):
         '''
         obstime = frame_or_coord.obstime
         # no point in interpolating for a single value
-        if obstime.size == 1:
-            return super().apcs(frame_or_coord)
-
         support = self._get_support_points(obstime)
 
         # get the position and velocity arrays for the observatory.  Need to
--- a/astropy/coordinates/erfa_astrom.py
+++ b/astropy/coordinates/erfa_astrom.py
@@ -240,7 +240,7 @@ class ErfaAstromInterpolator(ErfaAstrom):
         obstime = frame_or_coord.obstime
         # no point in interpolating for a single value
         if obstime.size == 1:
-            return super().apci(frame_or_coord)
+            return super().apcs(frame_or_coord)
 
         support = self._get_support_points(obstime)
 


astropy/coordinates/erfa_astrom.py
```diff
@@ -239,9 +235,6 @@ def apcs(self, frame_or_coord):
         '''
         obstime = frame_or_coord.obstime
         # no point in interpolating for a single value
-        if obstime.size == 1:
-            return super().apci(frame_or_coord)
```
maxnoe, line 243: This should have been apcs

mkbrewer:
This new `ErfaAstrom` class is somewhat unfortunate in that it adds another level of complexity to getting CIRS right. `apci` should never be called. The proper way to transform from ICRS to CIRS (both topocentric and geocentric) is exactly the same as ICRS to GCRS with the addition of two lines of code after the call to `apcs`:
```
x, y, s  = get_cip(jd1_tt, jd2_tt)
astrom['bpn'] = erfa.c2ixys(x, y, s)
```
In that way, the resulting CIRS frame will be all set for correct transformation to AltAz without the need for observatory coordinates as they will be already built in and the proper diurnal aberration containing both velocity and position components will be calculated by `atciqz` (see issue #10356).

Since you are interested in performance, try profiling a call to `get_body().transform_to(AltAz())` using the same `EarthLocation` for both to see all of the ugliness that currently occurs.

maxnoe:
@mkbrewer I don't think your comment is relevant to this PR or the addition of this class.

You make a general point on how the astropy transformations currently work, which is not dependent on this implementation.
If you are correct, the only thing that needs to change then is the calls in the transformation functions. 
Nothing in this class makes that more complicated.

maxnoe:
Also, the documentation of `apci` states:

> For a terrestrial observer, prepare star-independent astrometry parameters for transformations between ICRS and geocentric CIRS coordinates.

Why should this never be called? It sounds perfectly appropriate.

mkbrewer:
In the long run, what I'd like to see is: 
```
                                     -> TETE                                            
ICRS/BCRS -> astrometric -> GCRS -> |            -> HADec
                                     -> CIRS -> |
                                                 -> AltAz
```
But I'm beginning to think that may never happen in my lifetime.

mkbrewer:
The operative word there is "geocentric". As you can easily see, `apci` just calls `apcg`which then calls `apcs` after setting the observer's position and velocity to zero. This precludes the possibility of a topocentric CIRS frame which is required in order to properly calculate diurnal aberration including the effect of diurnal parallax.

StuartLittlefair:
@mkbrewer theres a couple of points to make about your comments. 

The first is that @maxnoe is right, this isn’t  really the PR for these discussions. 

The second is that at the moment you perhaps aren’t aware there is no one route from ICRS to CIRS. There is a frame transform graph and a direct route or a route through GCRS is possible. 

The final point is that astropy is largely written by volunteers all of whom are extremely busy with other things. I for one am trying to find time to address the really helpful comments you’ve made and the issues you’ve raised. We’ll get there but please show some patience.

mkbrewer:
@StuartLittlefair I'm very glad to see that you are considering the things that I bring up. As there has never been a single comment, let alone an extended discussion that I'd be happy to join in on regarding issue #10356, it is easy for me to think that no one is even looking at that. That's why I bring it up as often as I can. I'm sorry if doing so is deemed inappropriate here, so I'll shut up.

StuartLittlefair:
@mkbrewer that is a perfectly valid point. We are working on it. TETE is probably coming first, then more precise CIRS. 

Astrometric is the hardest as it requires us to sort out how to include space motion for a melange of solar system bodies and stars and miscellaneous others, although we could implement the frame in a fudged manner at present - but we can discuss that in the PR when it shows up.

mkbrewer:
I thought PR #10404 was a good start on that.

maxnoe:
Sorry, this is again a meta comment, but I think it not an appropriate reaction that if one is unhappy with the attention an issue is receiving to bring it up in other issues or pull request that seem tangentially related.

Keep discussions on topic.

mkbrewer:
Your and Stuart's objections have been noted. I'm sorry if I have caused offense. That was in no way my intention. I shall henceforth keep my thoughts to myself, unless my comments are solicited.

mhvk (approved):
@maxnoe - yes, I think this is the best route! For a single point, everything is so slow already that a bit of an extra overhead really doesn't matter. And now we know it will be fine too!

Thanks for the quick turn-around!

mhvk:
@mkbrewer - as @StuartLittlefair noted,  your comments and suggestions are very helpful in getting astropy to reach full precision (and for me to understand astrometry better!), and we will get there. So, please do continue to ping us if it seems nothing is happening for a while and you have a suggestion. But the best place is probably in either the issue you raised already or one of the slowly-getting-there PRs #10404 or #10666 (I use the github mode where once I comment on an issue/PR, I keep getting subsequent messages), if only so that the suggestions do not get lost. Thanks!

mhvk:
p.s. @mkbrewer - on your specific [comment[https://github.com/astropy/astropy/pull/10833#issuecomment-707197447], I think the design of the interpolator class that we ended up with allows for easy extension to dealing with a new or improved branch of the transform graph, so in that sense there is nothing lost. Anyway, discussion for elsewhere!

mkbrewer:
OK, will do. I didn't mean to sound critical with that comment (although I probably did). I just thought Max might be interested in looking at the bigger picture as it pertained to his class. It seems that he would rather stay strictly focused on the issue at hand though, which I completely understand. I have a bad habit of going all OT on things. Something that I need to work on being better about.

PR #10755, opened 2020-09-21
TST: URLError has different message on Windows

pllim:
### Description

This pull request is to address slightly different error message seen on Windows.

```
____________ test_download_ftp_file_properly_handles_socket_error _____________

    @pytest.mark.remote_data
    def test_download_ftp_file_properly_handles_socket_error():
        faulty_url = "ftp://anonymous:mail%40astropy.org@nonexisting/pub/products/iers/finals2000A.all"
        with pytest.raises(urllib.error.URLError) as excinfo:
            download_file(faulty_url)
>       assert "Name or service not known" in excinfo.exconly()
E       AssertionError: assert 'Name or service not known' in 'urllib.error.URLError: <urlopen error [Errno 11001] getaddrinfo failed>'
E        +  where 'urllib.error.URLError: <urlopen error [Errno 11001] getaddrinfo failed>' = <bound method ExceptionInfo.exconly of <ExceptionInfo URLError(gaierror(11001, 'getaddrinfo failed')) tblen=9>>()
E        +    where <bound method ExceptionInfo.exconly of <ExceptionInfo URLError(gaierror(11001, 'getaddrinfo failed')) tblen=9>> = <ExceptionInfo URLError(gaierror(11001, 'getaddrinfo failed')) tblen=9>.exconly

```

This is a direct follow-up of #10750 .

PR #10750, opened 2020-09-19
Properly handle FTP socket error

astrojuanlu:
## Description

This pull request address a minor coding error that was unvealed in https://travis-ci.org/github/astropy/astropy/jobs/728533005.

PR #10840, opened 2020-10-12
TST: yet more URLError alternate names

eteq:
This is a follow-on to #10755 which is a follow-in to #10750 - it turns out on "some platforms" (which means... mine - Ubuntu 20.04 with who knows what odd network setup I may have done to it), instead of the `URLError` messages currently in ``test_download_ftp_file_properly_handles_socket_error``, I get a "Temporary failure in name resolution".  So this PR updates the error messages to allow that.


pllim (approved):
The more the merrier! 😹

pllim:
CI failures are unrelated.

PR #10842, opened 2020-10-12
EXP: Investigate Windows failure on Travis CI

pllim:
### Description

Windows job on Travis CI suddenly cannot find `python.exe`. Larry saw same problem over at `photutils`. Not sure what had changed.

p.s. My job is queued. Looks like this gonna take a while.

PR #10846, opened 2020-10-13
TST: Allow Windows CI to fail for now

pllim:
### Description

This pull request is to temporary allow Windows CI job to fail until we can do one of the following:

* Have time and/or expertise to fix the obscure failure on Travis CI
* Move the CI to Azure (#8445)
* Move the CI to Actions (#10388)

Close #10842
Close #10844

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -225,6 +225,13 @@ matrix:
           env: TOXENV="py37-test-devdeps"
                TOXPOSARGS="--remote-data=any"
 
+        # Mysterious failure, cannot find python.exe
+        - os: windows
+          stage: Final tests
+          name: Python 3.7 with all optional dependencies
+          env: PYTHON_VERSION=3.7
+               TOXENV='py37-test-alldeps'
+
 before_install:
 
     # We need to use CCOMPILER otherwise Travis overwrites CC if we define it


pllim:
Hmm, Travis CI didn't kick off. Trying the close-open method.

pllim:
I cancelled CircleCI because it is not necessary. Travis CI somehow didn't report back but it is running at https://travis-ci.org/github/astropy/astropy/builds/735410532

eteq (approved):
The PR itself looks good, but just to be clear, but I don't think this actually closes #10842 since we still *might* want to fix it, right?

pllim:
Let's wait a bit though. @Cadair might just have a revelation.

pllim:
Still broken, so I'll push ahead with this after a rebase.

adrn:
Oops I just replied on slack, but I'll post here too. I don't think this is a travis-specific issue: I run windows tests for an affiliated package (gala, based on the package template) using github actions and am seeing the same error. It seems to have started ~1–2 weeks ago. Tox just had a new release a few days ago, so I thought that it could be the culprit, but even pinning tox doesn't seem to fix it. Mentioning this here in case that's useful for debugging! For example: https://github.com/adrn/gala/pull/190 and a failure: https://github.com/adrn/gala/runs/1249214462?check_suite_focus=true

pllim:
@Cadair did trace it to `could not satisfy requires PackageNotFoundError('tox-pypi-filter')` but not sure why it is happening.

pllim:
Merging this for now.

Issue #10836, opened 2020-10-12
np.full for Quantities


nbiederbeck:
### Description

I try to do
```python
my_quantity = u.Quantity(1, u.m)
np.full(10, my_quantity)
```
but get
```python
UnitConversionError: 'm' (length) and '' (dimensionless) are not convertible
```

What I want is what I get via
```python
u.Quantity(np.full(10, my_quantity.value), unit=my_quantity.unit)
>>> [1, 1, 1, 1, 1, 1, 1, 1, 1, 1] m
```
but the first version is so much cleaner.

Is there a (better) way to do this, or a chance to implement this?

pllim:
You could also do:
```python
np.full(10, 1) << u.m 
```

But for the original request, perhaps @mhvk has an idea. If it is indeed a "won't fix," we should probably add it to https://docs.astropy.org/en/latest/known_issues.html .

mhvk:
@nbiederbeck - thanks for raising the issue! Unfortunately, this is one of the very few numpy functions which we still cannot override (numpy allows only to override functions that take an array as an argument, so `np.full_like` *is* possible, though there is some work on trying to change this). I think @pllim's suggestion is probably the cleanest that we can do at the moment.

@pllim - you are right that we should add this to the known issues.  One similarly cannot do `np.zeros`, `np.ones`, and `np.empty`.

nbiederbeck:
Ok, thanks for your quick answers!

astrojuanlu:
> we should add this to the known issues. One similarly cannot do np.zeros, np.ones, and np.empty.

If you are okay then, until a proper fix is in place, this would be simple enough for new contributors.

astrojuanlu:
#10845

mhvk:
@astrojuanlu - thanks for raising the new issue. I marked the present one 'upstream fix required' since there are developments to make it work for numpy >=1.20 (though it would need an additional `like=` argument), but discussion is still ongoing - https://github.com/numpy/numpy/issues/17075

pentschev:
I just saw this due to the xref in https://github.com/numpy/numpy/issues/17075, it seems like `np.full_like` would be of use (there's no need for `like=`), and that's available since NumPy 1.16 (or maybe earlier). I have zero experience with astropy so I can't really comment on specifics for it and perhaps can be implemented/used here. `np.full_like` is very useful in other libraries such as CuPy, for example:

```python
In [1]: import numpy as np, cupy as cp

[...]

PR #10848, opened 2020-10-13
DOC: Update list of functions that lose units 

### LINKED ISSUE #10845: Complete list of functions that lose units
> you are right that we should add [`np.full`] to the known issues.  One similarly cannot do `np.zeros`, `np.ones`, and `np.empty`.

_Originally posted by @mhvk in https://github.com/astropy/astropy/issues/10836#issuecomment-707302194_

The list in question: https://docs.astropy.org/en/latest/known_issues.html#quantities-lose-their-units-with-some-operations

rpsfonseca:
### Description

This pull request is to address the missing functions on the list that details which operations lead to Quantities losing units.

Fixes #10845

### DIFF
--- a/docs/known_issues.rst
+++ b/docs/known_issues.rst
@@ -119,6 +119,20 @@ Both will throw an exception if units do not cancel, e.g.::
 
 See: https://github.com/astropy/astropy/issues/7582
 
+Numpy array creation functions cannot be used to initialize Quantity
+--------------------------------------------------------------------
+Trying the following example will throw an UnitConversionError::
+
+    >>> my_quantity = u.Quantity(1, u.m)
+    >>> np.full(10, my_quantity)
+    UnitConversionError: 'm' (length) and '' (dimensionless) are not convertible
+
+A workaround for this at the moment would be to do::
+    >>> np.full(10, 1) << u.m
+
+As well as with `~numpy.full` one cannot do `~numpy.zeros`, `~numpy.ones`, and `~numpy.empty`.
+
+
 Quantities Lose Their Units When Broadcasted
 --------------------------------------------
 


mhvk (commented):
@rpsfonseca - thanks! But these should not just be added, as the text just above notes the list is for "prior to astropy 4.0 and numpy 1.17". So, there should instead be a new header entry (maybe below this one is best), along the lines of "Numpy array creation functions cannot be used to initialize Quantity". For that, I'd include the explicit example with `np.full` that doesn't work (as it is the one most likely to surprise anyone)

rpsfonseca:
@mhvk So you say to add a new section like [this one](https://github.com/astropy/astropy/blob/dc84fa5d62d0eb6b758d33859d65fb469add43ae/docs/known_issues.rst#quantities-lose-their-units-with-some-operations) with the title being the one you mentioned?

And the explicit example you also mention, is the one given by the author of #10836 ?

mhvk:
@rpsfonseca - Yes to both!

mhvk (approved):
Thanks, looks good now!

pllim:
Since the issue that this PR fixed was labeled as Hacktoberfest, I applied "hacktoberfest-accepted" label here. I am still not fulling understand how the new fest rules work... cc @astrojuanlu 

Thank you, all!

astrojuanlu:
thanks @pllim ! I was about to do it myself :blush:

pllim:
For future reference, please do not skip CI when there is real code in the docs. This caused real failures, which I am patching it over at #10874 . FYI.

rpsfonseca:
Thank you for the help!
And sorry about skipping CI, I checked the CONTRIBUTING guide, and it mentioned that on documentation we could skip CI. But I will keep that in mind for future contributions.

mhvk:
@rpsfonseca - no worries, when I reviewed I noticed the few checks but that `readthedocs` was done, and I also thought that was OK. Having been around, I should have known better...

Question: would you be willing to adjust the `CONTRIBUTING` guide, adding a line about not skipping if there are doctests? (And for that one we *can* actually skip CI!)

pllim:
Perhaps this line in `CONTRIBUTING.md` confusing: `If your commit makes substantial changes to the documentation but no code changes` (... the key is "no code changes" but changing code in doc is basically code changes because we test it.)

I also added something in https://docs.astropy.org/en/latest/development/docguide.html#adding-a-git-commit .

If any of those could be less confusing, please feel free to propose changes as PR. Thanks!

rpsfonseca:
@mhvk I'd love to help on that for sure!
@pllim Yes, that was the line the lead me into the "mistake".

I will look into that and propose a change and check your opinion.

pllim:
@mhvk , @eteq and I just found out that the `UnitConversionError` doesn't happen in Numpy 1.16.x, as with in Numpy 1.20.x. Is this expected? We dropped Numpy 1.16 on `master` but still tests against it on older branches.

```python
>>> import numpy as np
>>> import astropy
>>> np.__version__
'1.16.5'
>>> astropy.__version__
'4.0.3'
>>> x = 1. * u.m
>>> np.full(10, x)
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])
```

mhvk:
@pllim - I think 1.16 just completely ignored the unit, without even realizing it.
But I guess that means the text of the docs is still not quite right...

pllim:
> that means the text of the docs is still not quite right

We dropped numpy 1.16 in #10664 , so it is non-issue for `master` but for the sake of making the release out, Erik and I decided to just skip that doctest in 4.0.x and 4.1.x branches.

rpsfonseca:
Sorry for taking this long, but I opened a PR to address the confusing tips on `CONTRIBUTING.md`, as we talked about.

PR #10850, opened 2020-10-13
Use manylinux1 for LTS wheels

astrofrog:
Before this is merged we need to check the Azure logs to make sure the correct manylinux1 wheels are produced.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,7 +38,9 @@ jobs:
 
       # FIXME: we exclude the test_data_out_of_range test since it
       # currently fails, see https://github.com/astropy/astropy/issues/10409
-      test_command: pytest -p no:warnings --astropy-header -k "not test_data_out_of_range" --pyargs astropy
+      # FIXME: we exclude the test_data_at_range_limit test since it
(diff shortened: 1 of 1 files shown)


astrofrog:
Looks like this is correctly producing manylinux1 wheels!

saimn:
Note that Numpy is considering dropping support for manylinux1 (https://mail.python.org/pipermail/numpy-discussion/2020-August/080974.html), so at some point we could the same for the next major releases. 
This is only for 4.0.x so 4.1 will be manylinux10 or you plan to do the same ?

astrofrog:
To be clear, I'm happy to have the discussion about 4.1 though if people feel we should use one or the other.

saimn:
Not sure which is best between 4.1 and 4.2. If we drop support for Python 3.6 in 4.2 that may be a better choice, but since we chose to use more modern packaging in 4.1 it seems also reasonable to require a recent pip for 4.1, as long as this clearly stated in the docs/release notes.

astrofrog:
Ok - I'll go ahead and merge this and will open a separate issue about 4.1.

PR #10861, opened 2020-10-15
EXP: Bump Python on Windows job

### LINKED ISSUE #10851: TST: Fix Windows job on Travis CI

pllim:
### Description

This pull request is to see if the real problem was simply using Python 3.7 in a Windows job. Will a pure upgrade work? If it works, do we favor status quo over #10860 (which bumps Python and also replaces ci-helpers with choco)?

Fixes #10851

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,10 +133,10 @@ matrix:
 
         # Try on Windows
         - os: windows
-          stage: Final tests
-          name: Python 3.7 with all optional dependencies
-          env: PYTHON_VERSION=3.7
-               TOXENV='py37-test-alldeps'
+          stage: Initial tests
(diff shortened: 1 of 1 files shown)


pllim:
Nope, doesn't work.

PR #10863, opened 2020-10-15
TST: Ignore dtype with a subarray field deprecation warning

pllim:
### Description

This pull request is to temporarily ignore deprecation warnings from #10815 because they make the log really hard to read while we await resolution upstream. We will ignore the warning for now so we can see if any new warnings crop up.

- [ ] Make sure numpy-dev job passes with this patch. It is allowed to fail.
- [ ] On merge, edit #10815 with a reminder to undo this later when we can.


mhvk (approved):
Yes, let's do that. Maybe add a note to the other issue that we should remove this when the upstream fix is in?

pllim:
Thanks! Added a check box at the top of #10815

PR #10865, opened 2020-10-15
Make pixel_shape have integer values and correct its computation in fit_wcs_from_points

mcara:
This PR fixes several issues with the computation of `pixel_shape` for the fitted WCS in `fit_wcs_from_points`:

1. IMO `pixel_shape` should be a list of integers. Having the meaning of image size (number of pixels along each axis) and also being related to `array_shape` I can't see how `pixel_shape` could take fractional values. In addition, these values get assigned to `NAXISn` and according to Wells (1981) paper these keywords take **integer values**.

2. I believe `pixel_shape` should not be computed as _difference_ between `max` and `min` of coordinates, but rather as `max` of coordinates. Assuming `x` and `y` coordinates are pixel coordinates from an image, the maximum of these coordinates would the best estimate for the size of the image from which these coordinates have been extracted.

### DIFF
--- a/astropy/wcs/utils.py
+++ b/astropy/wcs/utils.py
@@ -971,7 +971,7 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
     keywords in the input might cause unexpected behavior.
 
     Notes
-    ------
+    -----
     - The fiducial point for the spherical projection can be set to 'center'
       to use the mean position of input sky coordinates, or as an
       `~astropy.coordinates.SkyCoord` object.
@@ -1059,8 +1059,12 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
     if (type(sip_degree) != type(None)) and (type(sip_degree) != int):
         raise ValueError("sip_degree must be None, or integer.")
 
+    # compute bounding box for sources in image coordinates:
+    xpmin, xpmax, ypmin, ypmax = xp.min(), xp.max(), yp.min(), yp.max()
+
     # set pixel_shape to span of input points
-    wcs.pixel_shape = np.clip(np.ceil(np.max(xy, axis=1)).astype(int), 1, None)
+    wcs.pixel_shape = (1 if xpmax <= 0.0 else int(np.ceil(xpmax)),
+                       1 if ypmax <= 0.0 else int(np.ceil(ypmax)))
 
     # determine CRVAL from input
     close = lambda l, p: p[np.argmin(np.abs(l))]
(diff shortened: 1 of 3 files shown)


astropy/wcs/utils.py
```diff
@@ -1056,7 +1056,7 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
         raise ValueError("sip_degree must be None, or integer.")
 
     # set pixel_shape to span of input points
-    wcs.pixel_shape = (xp.max()+1-xp.min(), yp.max()+1-yp.min())
+    wcs.pixel_shape = np.ceil(np.max(xy, axis=1)).astype(np.int) + 1
```
nden, line 1059: Suppose the `xy` locations are near the center. If you really want to limit the solution to the area in the image including the sources, then you need to take xy.min() into account. However, in this case the FITS WCS does not have a mechanism to set a domain where the solution is valid, something equivalent to the `bounding_box` in gwcs.

Presumably `xy` come from an image. Ideally `wcs.pixel_shape` should be set to the size of that image but the information is missing.

I wonder if this is one of those cases when `wcs.pixel_shape` should be left as `None`.

nden:
Should this go in 4.1.1?

mcara:
> Should this go in 4.1.1?

I do not know...

astropy/wcs/utils.py
mcara: > Suppose the `xy` locations are near the center. If you really want to limit the solution to the area in the image including the sources, then you need to take xy.min() into account.

That would be neede for the `bounding_box` but 
`wcs.pixel_shape` does not have the same meaning. 

> Presumably `xy` come from an image. Ideally `wcs.pixel_shape` should be set to the size of that image but the information is missing.
> 
> I wonder if this is one of those cases when `wcs.pixel_shape` should be left as `None`.

I think we should do our best to estimate this size. All we can do is to estimate minimal size by taking `max(xi)` (actual image may be larger but not have sources beyond `max(xi)`). I think that's fine. Current algorithm would under-estimate severily `wcs.pixel_shape` if sources are all bunched together even if all sources are near the xmax, ymax,... corner. Plus, CRPIX probably would be outside of the "image" defined by `wcs.pixel_shape` (which also should be taken at the center of the `wcs.pixel_shape` instead of being mean of positions, I think).

pllim:
> Should this go in 4.1.1?

Yes. I moved the milestone. Please move the change log accordingly. Thanks!

mcara:
@nden Done

mcara:
Could anyone review this, please?

ceb8 (approved):
Looks good to me.

mcara:
Many thanks @ceb8 !

cshanahan1:
This looks good to me, apart from one small thing (that may be a non issue).

What happens if your maximum x/y source is less than a pixel away from the edge of the detector - would the '+1' make pixel_shape larger than the actual image size? Why is this +1 in there?

mcara:
> This looks good to me, apart from one small thing (that may be a non issue).
> 
> What happens if your maximum x/y source is less than a pixel away from the edge of the detector - would the '+1' make pixel_shape larger than the actual image size? Why is this +1 in there?

I do not think it matters. This is just an estimate. Moreover, what would be the danger in overestimating the size? The issue that I am trying to solve here is that we will underestimate the size in most common cases. The size is not critical for WCS transformations (hence +1 pixel is not an issue) but since we are providing it, it would be nice to do our best to estimate it.

cshanahan1 (approved):

mcara:
Ah, @cshanahan1 I am sorry, I was thinking about `ceil`. I just realized that your question was specifically about `+1`. The reason it is there is because coordinates `x, y` are 0-based. However, because `ceil` is used, I think you are right: `+1` is not necessary. However, this leads me to another issue/bug in this code: `CRPIX` in FITS FWCS is 1-based and so 1 should be added to values computed from 0-based coordinates.

nden (approved):

PR #10866, opened 2020-10-15
Fix overly pessimistic polar motion warning, fixes #10645

### LINKED ISSUE #10645: Warning about polar motion of dates in the future seems rather high
### Description
When transforming coordinates far in the future, e.g. 2040, a warning is raised that
coordinates will be imprecise to a level of "10s of arcseconds".

This seems rather much for missing polar motion information, since polar motion seems
to be < 1 arcsecond, see e.g. page 241ff of the Expl. Suppl. to the Astronomical Almanac or the 
IERS online information here: https://www.iers.org/IERS/EN/Science/EarthRotation/PolarMotion.html

### Steps to Reproduce

```python

>>> from astropy.coordinates import SkyCoord, AltAz, EarthLocation
>>> location = EarthLocation.of_site('Roque de los Muchachos')                                                                            
>>> time = Time('2040-01-01T00:00')                                                                                                       
WARNING: ErfaWarning: ERFA function "dtf2d" yielded 1 of "dubious year (Note 6)" [astropy._erfa.core]
>>> altaz = SkyCoord(alt='70d', az='0d', frame=AltAz(obstime=time, location=location))                                                    
>>> cirs = altaz.transform_to('cirs')                                                                                                     
WARNING: Tried to get polar motions for times after IERS data is valid. Defaulting to polar motion from the 50-yr mean for those. This may affect precision at the 10s of arcsec level [astropy.coordinates.builtin_frames.utils]
WARNING: ErfaWarning: ERFA function "apio13" yielded 1 of "dubious year (Note 2)" [astropy._erfa.core]
```

I might be misunderstanding the effect of polar motion, but the approximate algorithms in the Expl. Suppl. also make no mention of polar motion and claim to be accurate to the 1 arcsecond level (p. 298ff, section 7.4.5.1)

maxnoe:
Just removes the `10s` and fixes duplication of the warning message.

### DIFF
--- a/astropy/coordinates/builtin_frames/utils.py
+++ b/astropy/coordinates/builtin_frames/utils.py
@@ -42,25 +42,23 @@ def get_polar_motion(time):
     iers_table = iers.earth_orientation_table.get()
     xp, yp, status = iers_table.pm_xy(time, return_status=True)
 
-    wmsg = None
+    wmsg = (
+        'Tried to get polar motions for times {} IERS data is '
+        'valid. Defaulting to polar motion from the 50-yr mean for those. '
+        'This may affect precision at the arcsec level'
+    )
     if np.any(status == iers.TIME_BEFORE_IERS_RANGE):
-        wmsg = ('Tried to get polar motions for times before IERS data is '
-                'valid. Defaulting to polar motion from the 50-yr mean for those. '
-                'This may affect precision at the 10s of arcsec level')
         xp[status == iers.TIME_BEFORE_IERS_RANGE] = _DEFAULT_PM[0]
         yp[status == iers.TIME_BEFORE_IERS_RANGE] = _DEFAULT_PM[1]
 
-        warnings.warn(wmsg, AstropyWarning)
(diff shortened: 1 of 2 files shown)


maxnoe:
@pllim I rebased vs. master, let's see if the unrelated test error goes away

maxnoe:
Ok, now the failing test works but a later stage completely fails to start on windows

StuartLittlefair:
@maxnoe can you rebase this on master again now? It should pass and it's an uncontroversial change so it would be nice to get it merged.

For completeness, we should also include a test that the warning actually happens as expected for times before and after the valid range. That would fix the coverage issue.

maxnoe:
done

maxnoe:
@StuartLittlefair I also added a test for the warnings.

StuartLittlefair:
lovely. will merge when passing..

StuartLittlefair (approved):

StuartLittlefair:
Thanks @maxnoe. This fixes #10645

Issue #8996, opened 2019-07-14
Should self-transforms go through ICRS by default?


Juanlu001:
Recently we noticed that we had forgotten to define self transformations for ecliptic frames:

https://github.com/astropy/astropy/issues/8835

Perhaps the default behavior should be to go through ICRS if nothing else is defined? This is what some basic frames already do:

https://github.com/astropy/astropy/blob/master/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py#L216-L217

(At least CIRS, GCRS and HCRS already do that)

ayshih:
See also #7207 for a related, but alternative, suggestion from @Cadair

mhvk:
Looking at the transformation graph (which led to #8999 and then #9001 ;-), I do wonder why one should treat "given frame" to "same frame" any different from "given frame" to "other frame". So, in that sense going through ICRS makes sense.

That said, it would be good if we had some mechanism to ensure that if it is easy to add a to-same transform, it isn't forgotten...

PR #10867, opened 2020-10-15
Add True Equator, True Equinox (TETE) coordinate system and tests

StuartLittlefair:
As requested in #10230, this PR implements a True Equator, True Equinox frame, also known as "apparent place", although strictly speaking CIRS has taken over that role. This is the same apparent frame used by JPL horizons and is useful for comparison with coordinates from that service. In fact, we've had routines to convert GCRS to this frame hiding in solar system code for some time, for the sole sake of testing against JPL. This is also the correct frame to use if you want to use the equation LST = RA + HA...

I'm not sure which milestone to give this, which is also why this PR still requires a changelog.

I'm pinging @mkbrewer in case he has comments on the implementation, although the transform is rather straightforward I hope!

Below is the comparison in Moon position with JPL Horizons with this PR (after correction for the ~51 mas frame bias seen between astropy and JPL apparent right ascension)

moon


StuartLittlefair:
Note this pr **does not fix** #10230, since we also need to implement an astrometric, or BCRS, frame. See #10404 for an attempt at that.

astropy/coordinates/tests/test_solar_system.py
```diff
@@ -125,16 +120,17 @@ class TestPositionsGeocentric:
     def setup(self):
         self.t = Time('1980-03-25 00:00')
         self.frame = GCRS(obstime=self.t)
```
mhvk, line 122: Don't think `self.frame` is used any more (but may have missed it...)

astropy/coordinates/tests/test_solar_system.py
```diff
@@ -210,14 +206,16 @@ def setup(self):
         obsgeoloc, obsgeovel = kitt_peak.get_gcrs_posvel(self.t)
         self.frame = GCRS(obstime=self.t,
```
mhvk, line 207: Same, doesn't seem used any more.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -61,13 +62,101 @@ def cirs_to_itrs_mat(time):
...
     gamb, phib, psib, epsa = erfa.pfw06(*get_jd12(equinox, 'tt'))
     return erfa.fw2m(gamb, phib, psib, epsa)
 
 
 # now the actual transforms
 
+@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GCRS, TETE)
+def gcrs_to_tete(gcrs_coo, tete_frame):
+    # first apply GCRS self transform to correct time and observatory position/velocity
+    gcrs_coo2 = gcrs_coo.transform_to(GCRS(obstime=tete_frame.obstime, obsgeoloc=tete_frame.obsgeoloc,
```
mhvk, line 92: Maybe break such that `obsgeoloc` and `obsgeovel` align? (sorry, total nitpick...)

mhvk (commented):
This looks great (though I'll have to defer to you and @mkbrewer for the critical transformation part) - I really just have nitpicks.

As for the release to aim for: I think 4.2 is still totally fine.

mkbrewer:
Two comments: 

1. Strictly speaking, CIRS has not taken over the role of apparent place. Apparent place is TETE, CIRS is properly termed intermediate place. Similarly, for an observer on the surface of the Earth, topocentric apparent place or just topocentric place is TETE, whereas the proper term in CIRS is topocentric intermediate place. 

2. While I had envisioned TETE just inheriting its attributes from GCRS, it seems that it is also possible to set up a TETE frame with its own unique attributes. In that case, you might note that the observer's location in the TETE frame is realized in the GCRS frame. While one could imagine an observer's location realized in the TETE frame, there is no practical application for this. Conversion from geocentric apparent place to topocentric apparent place, for instance, does require reversion to ICRS/BCRS in order to take into account the difference in light time delay in the case of a solar system body, then passing though GCRS to take diurnal aberration into account. So again, the observer's location in TETE is ultimately inherited from the GCRS frame from which it is derived. That said, it wouldn't take much work to set up an observer's location in TETE. See Note 3 in `pvtob()`. This location could be transformed to/from GCRS using the same BPN matrix that is used to transform the source coordinates.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -61,13 +62,101 @@ def cirs_to_itrs_mat(time):
...
+    # compute the polar motion p-matrix
+    xp, yp = get_polar_motion(time)
+    sp = erfa.sp00(*get_jd12(time, 'tt'))
+    pmmat = erfa.pom00(xp, yp, sp)
+
+    # now determine the greenwich apparent siderial time for the input obstime
+    # we use the 2006A model for consistency with RBPN matrix use in GCRS <-> TETE
+    ujd1, ujd2 = get_jd12(time, 'ut1')
+    jd1, jd2 = get_jd12(time, 'tt')
+    gast = erfa.gst06a(ujd1, ujd2, jd1, jd2)
```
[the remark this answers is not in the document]
StuartLittlefair: I will leave this as is, since it is significantly faster (318 microsecs vs 1.4ms on my laptop)

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
StuartLittlefair: done

astropy/coordinates/tests/test_solar_system.py
StuartLittlefair: good catch!

astropy/coordinates/tests/test_solar_system.py
StuartLittlefair: good catch!

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -24,6 +24,7 @@
 from .gcrs import GCRS
 from .cirs import CIRS
 from .hcrs import HCRS
+from .equatorial import TETE
```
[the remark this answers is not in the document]
StuartLittlefair: gone

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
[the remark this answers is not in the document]
mhvk: OK, that's fine. Bit sad that the difference is so large!

mhvk:
Thanks, just a changelog entry needed now...

p.s. In the graph with the differences, are the RA, Dec offsets due to gravitational light bending by the earth? It is good that we do not have precision neutrino telescopes, otherwise they might be confused by the sharp jumps at the horizon :smile:

mkbrewer:
I actually discussed that very point with Brandon Rhodes. One way to handle that is to calculate the vector from the observer to the source passing through the Earth and then shrink the volume (and hence the mass) of the Earth to correspond to that of a sphere with a radius defined by the closest approach of a ray along that vector to the geocenter.

astropy/coordinates/builtin_frames/equatorial.py
```diff
@@ -65,21 +65,20 @@ class TETE(BaseRADecFrame):
...
 
-    TETE coordinates are often referred to as "apparent" coordinates, although CIRS
-    has replaced TETE as the modern system to use for apparent coordinates. TETE is the
-    apparent coordinate system used by JPL Horizons and is the correct coordinate
-    system to use when combining the right ascension with local apparent sidereal time
-    to calculate the hour angle.
+    TETE coordinates are often referred to as "apparent" coordinates, or
+    "apparent place". TETE is the apparent coordinate system used by JPL Horizons
+    and is the correct coordinate system to use when combining the right ascension
+    with local apparent sidereal time to calculate the hour angle.
```
mkbrewer, line 71: Oh, one more point. Hour angle there should read TIRS hour angle. The observed (ITRS) hour angle is slightly different due to the effects of polar motion.

mhvk:
@mkbrewer - I guess one would need to assume constant density or have a proper enclosed mass profile...

astropy/coordinates/builtin_frames/equatorial.py
StuartLittlefair: Given that polar motion is ~0.3 arcsecs I think I will leave out this distinction. My motivation is to keep the docstring as simple as possible so that astropy users don't try LST = HA + RA with ICRS coordinates (which I have seen in the wild several times).

StuartLittlefair:
tests are failing due to an unrelated problem with known_issues.rst...

StuartLittlefair:
these tests are failing due to an issue with some code in ```known_issues.rst```. 

Would a rebase address this and get the tests passing @pllim ?

pllim:
> Would a rebase address this and get the tests passing

Theoretically, yes. 😅 🤞

StuartLittlefair:
Wha! what the hell happened there? I followed the 

```
git fetch astropy master
git rebase astropy/master
```

magic incantation and ended up with 47 extra commits, including merge commits! Sorry to bug you @pllim but any advice as to how to fix?

astrojuanlu:
@StuartLittlefair I tried this locally on your branch and it worked:

```
$ git reset --hard 846a493a0
$ git push --force
```

Some weird merge happened and your commits appear twice, this should fix it.

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -288,3 +289,15 @@ def hcrs_to_hcrs(from_coo, to_frame):
...
+
+
+@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, TETE, TETE)
+def tete_to_tete(from_coo, to_frame):
+    if (np.all(from_coo.obstime == to_frame.obstime) and
+            np.all(from_coo.obsgeoloc == to_frame.obsgeoloc) and
+            np.all(from_coo.obsgeovel == to_frame.obsgeovel)):
+        return to_frame.realize_frame(from_coo.data)
+    else:
+        # can't do self-transform via ITRS due to infinite recursion
```
mhvk, line 301: Why is this? Bug in the code or physical reason? Would be good to add explanation (and perhaps TODO if just a bug), especially as the self-transform is used in `tete_to_itrs`

mhvk (approved):
Looks very good! Only a request for clarification of a comment...

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
ayshih: This is due to a flaw in how the TETE<->ITRS transformations are written (see https://github.com/astropy/astropy/pull/10909#issuecomment-714213513 for the same problem with the TEME<->ITRS transformations).  The TETE<->ITRS transformations should be rewritten to not rely on an existing TETE<->TETE loopback, and then the loopback can be written to depend on the TETE<->ITRS transformations without triggering infinite recursion.

ayshih: In short, an `obstime` change needs to be handled in those transformations as an ITRS loopback rather than a TETE loopback.

StuartLittlefair: @ayshih I actually think the way it's done here is preferable. Your re-written TEM[T]E<->ITRS transformations **still** require a trip through ICRS, it's just not explicit anymore.

This is because an ITRS<->ITRS loopback for different ```obstimes``` goes back through CIRS, which in turn goes back through ICRS. So the **actual transform** called still goes through CIRS and ICRS, you've just added a step in the lookup chain.

In the existing scenario:

1) Code sees that it needs a TETE->TETE' self transform
2) That transform goes through ICRS
3) Code looks up TETE->ICRS transform
4) That transform is TETE->CIRS->ICRS.
5) Code performs TETE->CIRS->ICRS-CIRS'-TETE' 

If I implemented the changes in your PR:

1) Code sees that it needs a TETE->TETE' self transform
2) That transform goes through ITRS
3) Code looks up TETE->ITRS transform
4) The TETE->ITRS involves an ITRS self-transform
5) The ITRS->ITRS' transform is ITRS->CIRS->ICRS->CIRS'->ITRS'
6) Code does TETE->ITRS->CIRS->ICRS->CIRS'->ITRS'->TETE'

I don't see the second scenario is better. As equatorial frames, TEME and TETE are brothers of CIRS and shouldn't sit "downstream" of ITRS in the transform chain.

StuartLittlefair: Hi @mhvk - I've added a hopefully clearer comment to the self-transform.

I've also added a "needs-discussion" label to **prevent anyone merging** even though it is approved and passing.

The reason is that I have a branch implementing improved accuracy in CIRS by way of adding a topocentric position to that frame. Given the similarities between CIRS and TETE, the intermediate rotation transforms in that branch are similar to the ones here.

However, although that branch passes almost all tests, there is some odd behaviour in ITRS->CIRS->ALTAZ that I don't currently understand. It's making me nervous there's a flaw in the transform I haven't seen.

ayshih: > In the existing scenario:
> 
> 1. Code sees that it needs a TETE->TETE' self transform
> 2. That transform goes through ICRS
> 3. Code looks up TETE->ICRS transform
> 4. That transform is TETE->CIRS->ICRS.
> 5. Code performs TETE->CIRS->ICRS->CIRS'->TETE'

I believe your comparison is misleading because you don't have defined TETE<->CIRS transformations.  So, it's actually:

4. That transform is TETE->GCRS->CIRS->ICRS.
5. Code performs TETE->GCRS->CIRS->ICRS->CIRS'->GCRS'->TETE'

That is, the difference between your current implementation and my suggested re-jigger is that the intermediate frame is ITRS instead of GCRS.  Importantly, the fact that the TETE->ICRS composite transformation goes through GCRS instead of through ITRS is *luck*: they traverse the same number of edges in the graph, so it comes down to the order they happen to be stored in the graph.

My perspective is that you should decide which frame you want to consider as "upstream" – GCRS or ITRS – and handle a changing `obstime` there, rather than handling it in the "downstream" frame.

> As an aside, I hate the way that the transform graph shows, for example, an ITRS->ITRS' loopback, since it's a lie! There's an argument for removing all of these shortcut functions and just making everything Earth-related go through ICRS if we don't write real self-transforms...

It's been suggested before (#8996).  If you're worried about hidden transformations, it pervades beyond just the loopbacks.  For example, the TETE->ITRS transformation is a single edge in the graph, but internally it will handle a change in `obstime` by a loop through ICRS using at least four other transformations and two other frames.

mhvk: @StuartLittlefair, @ayshih - thanks for the explanation, makes sense! And the comment now helps.

On the larger picture, I agree it is misleading to have self-transformations shown if they do not in fact exists. If we were to remove all self-transformations, would the graph automatically do the right thing, i.e., go to the nearest frame that does have a self-transformation and go back? Independently, we should remove the arrow...  Anyway, probably worth a separate issue, as we're not likely to change it for 4.2!

mhvk:
A propos the status of this PR for 4.2, is it a good fall-back option is to get TETE in for 4.2 as is, and then deal with CIRS separately (for 4.3 if need be)?

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
ayshih: > On the larger picture, I agree it is misleading to have self-transformations shown if they do not in fact exists.

Self-transformations have to be implemented at some level to show up on the graph.  The issue here is the how to implement the loopback if a "real" self-transform doesn't exist (i.e., one that is just calculations rather than relying on the existing transformation graph.)

> If we were to remove all self-transformations, would the graph automatically do the right thing, i.e., go to the nearest frame that does have a self-transformation and go back?

No.  Unless #8996 is addressed, the default behavior is that the transform will see the same frame as source and destination and perform a no-op.

StuartLittlefair: Yes - what @ayshih says.

Thanks for correcting my example, although really it's TETE->GCRS->ICRS->GCRS'->TETE. Which is conceptually correct, as shown in this figure from P6 of the "Sofa tools for earth attitude doc"

Screenshot 2020-10-22 at 16 27 00

> My perspective is that you should decide which frame you want to consider as "upstream" – GCRS or ITRS – and handle a changing obstime as a loopback there, rather than handling it in the "downstream" frame (TETE).

I *think* this is how I've done it here, no? In that the self transform goes through ICRS, which is really TETE->GCRS->ICRS. Have I missed your meaning?

ayshih: Whoops, you're correct.  I misread the graph.  You're correct.  TETE->ICRS goes through GCRS instead of through ITRS->CIRS.

StuartLittlefair: We both did, and I wrote the code. I really hate the graph.

mhvk: Yes, at least the no-op transform should only happen if the frames are equal...

StuartLittlefair:
> A propos the status of this PR for 4.2, is it a good fall-back option is to get TETE in for 4.2 as is, and then deal with CIRS separately (for 4.3 if need be)?

I'm happy with that - this doesn't cause any issues and I can re-base my CIRS branch on this one once merged. Once I work out where the flaw in the transform is, I can fix it in both frames. Will remove the "needs discussion" flag

mkbrewer:
> A propos the status of this PR for 4.2, is it a good fall-back option is to get TETE in for 4.2 as is, and then deal with CIRS separately (for 4.3 if need be)?

I would just note here that, although I haven't seen a PR yet, the fix for #10356 that @StuartLittlefair has implemented works fine. This is the last bug leading to inaccuracy in the positions of solar system bodies and so should be added to a release as soon as is practical.

StuartLittlefair:
@mhvk I'm happy for this to be merged so will hit the button unless you have objections?

PR #10869, opened 2020-10-15
Add 10666 what's new entry for 4.1

eteq:
This adds a mention of #10666 to the what's new, mainly as a "warning" given that this bug has long existed such that some users may have worked around it, and this will give a bit more visibility so that they know to un-do any workaround they might have.

Not I also moved the CHANGES.rst entry to 4.1 - this is mainly to be consistent with the What's new, but as discussed in #10666 we can certainly move it back if we agree it should go in the LTS branch!

cc @StuartLittlefair @mhvk

### DIFF
--- a/docs/whatsnew/4.1.rst
+++ b/docs/whatsnew/4.1.rst
@@ -282,9 +282,10 @@ Realistic treatment of solar light deflection for nearby objects
 
 The machinery underlying the `astropy.coordinates` transformations was updated
 to correct an issue where coordinates were always assumed to be at infinite
-distance when computing the effect of light deflection by the sun.  While this
-change is undeniably more accurate, this will change previous results that may
-have already corrected for this effect.
+distance when computing the effect of light deflection by the sun.  With this
+change, light deflection by the Sun is accurate also for solar system objects.
+Even for those, the results will change only slightly (by 3 arcsec in the
+worst case).
 
 For more context and discussion on this topic see
 `Github issue #10666 <https://github.com/astropy/astropy/pull/10666>`_.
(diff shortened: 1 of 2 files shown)


docs/whatsnew/4.1.rst
```diff
@@ -274,6 +274,21 @@ Support for ASDF serialization of models
...
+
+.. _whatsnew-4.1-light-defelction-fix:
+
+Realistic treatment of solar light deflection for nearby objects
+================================================================
+
+The machinery underlying the `astropy.coordinates` transformations was updated
+to correct an issue where coordinates were always assumed to be at infinite
+distance when computing the effect of light deflection by the sun.  While this
+change is undeniably more accurate, this will change previous results that may
```
mhvk, line 286: The "While this change..." sentence reads a bit weirdly. How about: "With this change, light deflection by the Sun is accurate also for solar system objects. Even for those, the results will change only slight be slightly (by 3 arcsec in the absolute worst case).
```
I don't think we have to warn about "may have already corrected" - anybody who knowingly did do a post-astropy correction will understand they need to change their code (and, really, they should have pinged us upstream in the first place!)

StuartLittlefair:
Agree with @mhvk here - that one sentence reads a bit oddly. Marten's suggestion (minus typos) looks OK to me though.

eteq:
@StuartLittlefair @mhvk - good points and I updated the wording - can one of you please approve/merge?

eteq:
(also as an aside for @StuartLittlefair  - I really wanted to include the image in https://github.com/astropy/astropy/pull/10666#issuecomment-675967367 but couldn't figure out an excuse.  Maybe someday we can sneak it into the docs!)

StuartLittlefair (approved):

PR #10872, opened 2020-10-15
ENH:  Calculate WCS pixel area and scales as Quantity

### LINKED ISSUE #6523: Should proj_plane_pixel_area return a Quantity?
At the moment, the ``astropy.wcs.proj_plane_pixel_area`` function returns a value in square degrees, but just as a value. I think we really should be returning a Quantity to avoid confusion. However, breaking backward compatibility may be problematic if we suddenly start returning Quantity. Maybe we could deprecate and rename the function to avoid nasty surprises? Or we could add a variant of the function that returns a quantity? The same applies to ``proj_plane_pixel_scales``.

pllim:
### Description

This pull request is to implement new `WCS` methods that return pixel scales and area as `Quantity`.

I am not convinced that the low-level functions in `utils` should be deprecated. If people like the non-Quantity stuff, they are free to use them still.

I am also not convinced that we should force the outputs to be in arcsec. If `cunit` says `'deg'`, I think it is reasonable to return degrees. People can easily convert to arcsec at their own leisure.

I also took the liberty to fix some PEP 8 stuff, so I know if I accidentally introduced new violation or not (I try not to). In that process, I also found a duplicate test function, which I deleted here.

If you disagree, we can discuss during the review process.

Fixes #6523

### DIFF
--- a/astropy/wcs/wcs.py
+++ b/astropy/wcs/wcs.py
@@ -2452,13 +2452,13 @@ reduce these to 2 dimensions using the naxis kwarg.
 
     def proj_plane_pixel_scales(self):
         """Same as :func:`astropy.wcs.utils.proj_plane_pixel_scales`
-        but returns a `~astropy.units.Quantity`.
+        but returns a list of `~astropy.units.Quantity`.
 
         """
         from astropy.wcs import utils as wcs_utils  # Avoid circular import
         val = wcs_utils.proj_plane_pixel_scales(self)
         units = [u.Unit(x) for x in self.wcs.cunit]
-        return u.Quantity(list(map(operator.mul, val, units)))
+        return list(map(operator.mul, val, units))  # Can have different units
 
     def proj_plane_pixel_area(self):
         """Same as :func:`astropy.wcs.utils.proj_plane_pixel_area`
--- a/astropy/wcs/wcs.py
+++ b/astropy/wcs/wcs.py
@@ -2477,7 +2477,7 @@ reduce these to 2 dimensions using the naxis kwarg.
 
         """  # noqa: E501
         from astropy.wcs.utils import proj_plane_pixel_scales  # Avoid circular import
-        values = proj_plane_pixel_scales(self.celestial)
+        values = proj_plane_pixel_scales(self)
         units = [u.Unit(x) for x in self.wcs.cunit]
         return [value * unit for (value, unit) in zip(values, units)]  # Can have different units
 
--- a/astropy/wcs/wcs.py
+++ b/astropy/wcs/wcs.py
@@ -77,8 +77,8 @@ if _wcs is not None:
 
     if not _wcs._sanity_check():
         raise RuntimeError(
-        "astropy.wcs did not pass its sanity check for your build "
-        "on your platform.")
+            "astropy.wcs did not pass its sanity check for your build "
+            "on your platform.")
 
     WCSBase = _wcs._Wcs
     DistortionLookupTable = _wcs.DistortionLookupTable
@@ -370,7 +370,7 @@ class WCS(FITSWCSAPIMixin, WCSBase):
        be raised by the constructor, not when subsequently calling a
        transformation method.
 
-    """
+    """  # noqa: E501
 
     def __init__(self, header=None, fobj=None, key=' ', minerr=0.0,
                  relax=True, naxis=None, keysel=None, colsel=None,
@@ -498,7 +498,7 @@ class WCS(FITSWCSAPIMixin, WCSBase):
             self.naxis = wcsprm.naxis
 
             if (wcsprm.naxis != 2 and
-                (det2im[0] or det2im[1] or cpdis[0] or cpdis[1] or sip)):
+                    (det2im[0] or det2im[1] or cpdis[0] or cpdis[1] or sip)):
                 raise ValueError(
                     """
 FITS WCS distortion paper lookup tables and SIP distortions only work
@@ -597,7 +597,8 @@ reduce these to 2 dimensions using the naxis kwarg.
 
         # Construct a list of dimensions from the original WCS in the order
         # in which they appear in the final WCS.
-        keep = [cname_uuid.index(cname) if cname in cname_uuid else None for cname in copy.wcs.cname]
+        keep = [cname_uuid.index(cname) if cname in cname_uuid else None
+                for cname in copy.wcs.cname]
 
         # Restore the original CNAMEs
         copy.wcs.cname = ['' if i is None else self.wcs.cname[i] for i in keep]
@@ -654,7 +655,7 @@ reduce these to 2 dimensions using the naxis kwarg.
 
         self.wcs.set_pv([])
         warnings.warn("Removed redundant SCAMP distortion parameters " +
-            "because SIP parameters are also present", FITSFixedWarning)
+                      "because SIP parameters are also present", FITSFixedWarning)
 
     def fix(self, translate_units='', naxis=None):
         """
@@ -749,7 +750,8 @@ reduce these to 2 dimensions using the naxis kwarg.
                     # do not require a header parameter
                     naxis1, naxis2 = self.pixel_shape
                 except (AttributeError, TypeError):
-                    warnings.warn("Need a valid header in order to calculate footprint\n", AstropyUserWarning)
+                    warnings.warn(
+                        "Need a valid header in order to calculate footprint\n", AstropyUserWarning)
                     return None
             else:
                 naxis1 = header.get('NAXIS1', None)
@@ -807,8 +809,9 @@ reduce these to 2 dimensions using the naxis kwarg.
                 dis = header[distortion].lower()
                 if dis == 'lookup':
                     del header[distortion]
-                    assert isinstance(fobj, fits.HDUList), ('An astropy.io.fits.HDUList'
-                                'is required for Lookup table distortion.')
+                    assert isinstance(fobj, fits.HDUList), (
+                        'An astropy.io.fits.HDUList'
+                        'is required for Lookup table distortion.')
                     dp = (d_kw + str(i)).strip()
                     dp_extver_key = dp + '.EXTVER'
                     if dp_extver_key in header:
@@ -842,10 +845,11 @@ reduce these to 2 dimensions using the naxis kwarg.
             return (tables.get(1), tables.get(2))
 
     def _read_d2im_old_format(self, header, fobj, axiscorr):
-        warnings.warn("The use of ``AXISCORR`` for D2IM correction has been deprecated."
-                      "`~astropy.wcs` will read in files with ``AXISCORR`` but ``to_fits()`` will write "
-                      "out files without it.",
-                      AstropyDeprecationWarning)
+        warnings.warn(
+            "The use of ``AXISCORR`` for D2IM correction has been deprecated."
+            "`~astropy.wcs` will read in files with ``AXISCORR`` but ``to_fits()`` will write "
+            "out files without it.",
+            AstropyDeprecationWarning)
         cpdis = [None, None]
         crpix = [0., 0.]
         crval = [0., 0.]
@@ -886,7 +890,6 @@ reduce these to 2 dimensions using the naxis kwarg.
             return
         dist = 'D2IMDIS'
         d_kw = 'D2IM'
-        err_kw = 'D2IMERR'
 
         def write_d2i(num, det2im):
             if det2im is None:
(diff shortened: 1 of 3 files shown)


astropy/wcs/wcs.py
```diff
@@ -886,7 +891,6 @@ def _write_det2im(self, hdulist):
             return
         dist = 'D2IMDIS'
         d_kw = 'D2IM'
-        err_kw = 'D2IMERR'
```
pllim, line 889: Note: Not used.

astropy/wcs/wcs.py
```diff
@@ -1008,10 +1012,10 @@ def _write_distortion_kw(self, hdulist, dist='CPDIS'):
 
         if dist == 'CPDIS':
             d_kw = 'DP'
-            err_kw = 'CPERR'
+            # err_kw = 'CPERR'
```
pllim, line 1015: Also not used. But somehow here I commented instead of deleting. I should be consistent, but I want to hear what others say first.

astropy/wcs/wcs.py
```diff
@@ -2446,6 +2450,26 @@ def sip_foc2pix(self, *args):
...
+        return u.Quantity(list(map(operator.mul, val, units)))
+
+    def proj_plane_pixel_area(self):
+        """Same as :func:`astropy.wcs.utils.proj_plane_pixel_area`
+        but returns a `~astropy.units.Quantity`.
+
+        """
+        from astropy.wcs import utils as wcs_utils  # Avoid circular import
+        val = wcs_utils.proj_plane_pixel_area(self)
+        unit = u.Unit(self.wcs.cunit[0]) * u.Unit(self.wcs.cunit[1])  # 2D only
```
astrofrog, line 2470: This should probably error if not 2D?

astrofrog (dismissed):
I remember why these were separate functions - it is because for the area at least I think this was mainly intended for celestial WCS. We should test this with non-celestial WCS to see if it produces nonsensical results. If so, then we might want to raise an error if the WCS is not celestial?

astropy/wcs/wcs.py
pllim: `astropy.wcs.utils.proj_plane_pixel_area` already checks it.

astropy/wcs/wcs.py
mhvk: Just delete it!

pllim:
I added a test, but I am no WCS expert, so if there are other cases to consider, please let me know.

astropy/wcs/wcs.py
```diff
@@ -2446,6 +2450,26 @@ def sip_foc2pix(self, *args):
...
 
+    def proj_plane_pixel_scales(self):
+        """Same as :func:`astropy.wcs.utils.proj_plane_pixel_scales`
+        but returns a list of `~astropy.units.Quantity`.
+
+        """
+        from astropy.wcs import utils as wcs_utils  # Avoid circular import
+        val = wcs_utils.proj_plane_pixel_scales(self)
+        units = [u.Unit(x) for x in self.wcs.cunit]
+        return list(map(operator.mul, val, units))  # Can have different units
```
mhvk, line 2461: I know it is the same thing, but to me the following is clearer:
```
return [value * unit for (v, unit) in zip(values, units)]
```

astropy/wcs/wcs.py
```diff
@@ -2446,6 +2450,26 @@ def sip_foc2pix(self, *args):
...
+
+    def proj_plane_pixel_area(self):
+        """Same as :func:`astropy.wcs.utils.proj_plane_pixel_area`
+        but returns a `~astropy.units.Quantity`.
+
+        """
+        from astropy.wcs import utils as wcs_utils  # Avoid circular import
+        val = wcs_utils.proj_plane_pixel_area(self)
```
mhvk, line 2469: Maybe use `value`? (total nitpick)

astropy/wcs/wcs.py
```diff
@@ -2446,6 +2450,26 @@ def sip_foc2pix(self, *args):
         """.format(docstrings.TWO_OR_MORE_ARGS('2', 8),
                    docstrings.RETURNS('pixel coordinates', 8))
 
+    def proj_plane_pixel_scales(self):
+        """Same as :func:`astropy.wcs.utils.proj_plane_pixel_scales`
+        but returns a list of `~astropy.units.Quantity`.
+
+        """
+        from astropy.wcs import utils as wcs_utils  # Avoid circular import
+        val = wcs_utils.proj_plane_pixel_scales(self)
```
mhvk, line 2459: Maybe use `values`? (More total nitpicking...)

mhvk (commented):
This looks good to me! Agreed with @astrofrog about the pixel area, though I think the code as written should give sensible results (even if the units are "Å arcsec" or so...). So, left are only the most picky of nits!

And agree very much not to enforce a particular unit for the output but rather assume the writers of the file knew what the preferred unit was.

astropy/wcs/wcs.py
pllim: Is this check insufficient?

https://github.com/astropy/astropy/blob/4eef52a1c95471b837949c5a8f862a01e4ca1bfa/astropy/wcs/utils.py#L390-L392

If we need to add more check, I feel like it has to be added inside the `utils` function, not the high-level method wrapper.

astropy/wcs/wcs.py
pllim: Done. I thought I was pretty progressive in using 3 chars, but apparently not progressive enough. 😸

pllim:
p.s. Failures are unrelated. Will need to rebase after #10874 is merged.

pllim:
Update: Rebased and tests passed. Please let me know if I overlooked any comments. Thanks for the reviews!

astropy/wcs/wcs.py
```diff
@@ -2446,6 +2448,26 @@ def sip_foc2pix(self, *args):
         """.format(docstrings.TWO_OR_MORE_ARGS('2', 8),
                    docstrings.RETURNS('pixel coordinates', 8))
 
+    def proj_plane_pixel_scales(self):
+        """Same as :func:`astropy.wcs.utils.proj_plane_pixel_scales`
+        but returns a list of `~astropy.units.Quantity`.
+
+        """
+        from astropy.wcs import utils as wcs_utils  # Avoid circular import
```
mcara, line 2456: Why not move the code from `utils.py` here as `_ proj_plane_pixel_scales()` and then call it from both functions? Doing this, I think, would avoid the need for this import within the method.

astropy/wcs/wcs.py
```diff
@@ -2446,6 +2448,26 @@ def sip_foc2pix(self, *args):
...
+        values = wcs_utils.proj_plane_pixel_scales(self)
+        units = [u.Unit(x) for x in self.wcs.cunit]
+        return [value * unit for (value, unit) in zip(values, units)]  # Can have different units
+
+    def proj_plane_pixel_area(self):
+        """Same as :func:`astropy.wcs.utils.proj_plane_pixel_area`
+        but returns a `~astropy.units.Quantity`.
+
+        """
+        from astropy.wcs import utils as wcs_utils  # Avoid circular import
```
mcara, line 2466: Same as in https://github.com/astropy/astropy/pull/10872/files#r506769107

astropy/wcs/wcs.py
```diff
@@ -2446,6 +2448,26 @@ def sip_foc2pix(self, *args):
         """.format(docstrings.TWO_OR_MORE_ARGS('2', 8),
                    docstrings.RETURNS('pixel coordinates', 8))
 
+    def proj_plane_pixel_scales(self):
+        """Same as :func:`astropy.wcs.utils.proj_plane_pixel_scales`
+        but returns a list of `~astropy.units.Quantity`.
+
+        """
```
mcara, line 2455: If this function is going to be the new "workhorse", and also because it is a method of the main class `WCS` from this sub-package, I think it would be better to simply copy the docstring from the `utils` function and maybe make a mention like "For unit-un-aware version - see ..." (something like that)?

mhvk: :+1: - and one can have a `See Also` section to refer the the unit unaware function.

astropy/wcs/wcs.py
```diff
@@ -33,6 +33,7 @@
 import io
 import itertools
 import os
+import operator
```
mhvk, line 36: I think you may not need `operator` any more since you followed my recommendation!

astropy/wcs/wcs.py
mcara: @pllim I did not do any timing tests on this specific PR, but I do not think the answer from https://stackoverflow.com/a/32151422 is accurate. The benchmark does not compare timings of global import vs local import.

I did a quick test on a function that uses numpy's `sin()` to do a simple computation:

#### Local import
```python
In [1]: def f(x):
   ...:     from numpy import sin
   ...:     return sin(x)
   ...:

In [2]: %timeit f(0.5)
1.08 µs ± 10.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: exit
```

#### Global import
```
In [1]: from numpy import sin

In [2]: def f(x):
   ...:     return sin(x)
   ...:

In [3]: %timeit f(0.5)
587 ns ± 7.57 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

pllim: Let's use a more realistic example:

```python
>>> from astropy.io import fits
>>> from astropy.utils.data import get_pkg_data_filename
>>> from astropy.wcs import WCS

>>> path = get_pkg_data_filename("data/dist_lookup.fits.gz", package='astropy.wcs.tests')
>>> with fits.open(path) as hdulist:
...     w = WCS(hdulist[1].header, hdulist)

>>> # Time this first before wcs.utils import below.
>>> # This was done in IPython. I just have a habit to use >>> in markdown.
>>> %timeit w.proj_plane_pixel_scales()
87.6 µs ± 78.8 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

>>> from astropy.wcs.utils import proj_plane_pixel_scales
>>> # Same calculation as the class method
>>> def proj_this(w):
...     values = proj_plane_pixel_scales(w)
...     units = [u.Unit(x) for x in w.wcs.cunit]
...     return [value * unit for (value, unit) in zip(values, units)]

>>> # Time the standalone function
>>> # This was done in IPython. I just have a habit to use >>> in markdown.
>>> %timeit proj_this(w)
87.1 µs ± 149 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Okay, that's 0.5 microseconds improvement in something that takes 87 microseconds to calculate. Is this small improvement worth the extra code complexity in moving the `utils` functions completely out of `utils` just to avoid circular import while being able to do module level import?

mcara: I do not think the code is more complex, it is easier to read (follow), and it is faster. I suspect that is the reason why my older phone is "getting slower" :)

mcara: Could you, please, for completeness, benchmark `utils.proj_plane_pixel_scales(w)`. I am getting like 1/3 of your timing.

pllim: What's the point of that? It doesn't apply units, so it is like comparing orange to apple.

astropy/wcs/wcs.py
```diff
@@ -2449,22 +2448,83 @@ def sip_foc2pix(self, *args):
...
+            A vector of projection plane increments corresponding to each
+            pixel side (axis).
+
+        See Also
+        --------
+        astropy.wcs.utils.proj_plane_pixel_scales
+
+        """  # noqa: E501
+        from astropy.wcs.utils import proj_plane_pixel_scales  # Avoid circular import
+        values = proj_plane_pixel_scales(self.celestial)
```
pllim, line 2480: Re-reading the docstring again, I think this is more correct? That is, by passing in `self.celestial` instead of `self`?

astropy/wcs/wcs.py
pllim: Ops, good catch!

pllim:
I think I have addressed all the comments thus far, except for the suggestion for moving out stuff out of `utils.py`, as I am still not convinced it is worth the effort if the little extra local import is the only thing causing it to happen.

astropy/wcs/wcs.py
mcara: There is nothing conceptually wrong with passing the "entire" `self`. Let's say `self` has 2 celestial axes and one spectral. This function would return something like `CDELT` for each axis. It should be fine. I think you should pass the whole WCS object.

pllim: Ah, okay, I'll change it back then. This snippet from the `utils` docstring confused me.

```rst
    .. note::
        In order to compute the scales corresponding to celestial axes only,
        make sure that the input `~astropy.wcs.WCS` object contains
        celestial axes only, e.g., by passing in the
        `~astropy.wcs.WCS.celestial` WCS object.
```

mcara: That is just to direct the reader what to do if that is what they want (i.e., to get the scale _only_ for celestial axes and exclude non-celestial axes)

mcara: For the "area", it is highly unlikely that anyone would want to compute the area of 1 celestial axis and 1 spectral axis, even though mathematically one can do anything they want. Or, I am not sure most people would find it useful to compute the "volume" of a 2D celestial axes X 1D spectral. Probably, in most cases, user would want to separate these kind of axes but I may be wrong. If there is demand to compute volumes, let me know.

I personally find `proj_plane_pixel_scales` less useful than `proj_plane_pixel_area` because the later is used to approximate "mean pixel scale" (**geometric mean**) which, I personally find more useful than knowing scale along each axes. It also can be used to equalize pixel intensity when displaying an image (like it is done in FLT images).

pllim: I did use `proj_plane_pixel_scales` when I was scratching my head over spacetelescope/stginga#205 😸

mcara (approved):

pllim:
@astrofrog , do you still have objection to this PR?

pllim:
I'll take silence as "no objection".

PR #10873, opened 2020-10-15
Use manylinux1 for linux wheels

astrofrog:
I made this fix already for v4.0.3 but I think we might want to do it for v4.1 - basically manylinux2010 wheels require pip 19 or later which is maybe a little too restrictive, so I think for now and until we make a conscious decision to move to manylinux2010 we should use manylinux1.

Does anyone have a good argument for using manylinux2010?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,6 +26,12 @@ pr:
     include:
     - 'v*'
 
+# Build Linux wheels using manylinux1 for compatibility with old versions
+# of pip and old platforms.
+variables:
+  CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
+  CIBW_MANYLINUX_I686_IMAGE: manylinux1
+
 jobs:
   - template: publish.yml@OpenAstronomy
     parameters:


pllim (approved):
I think it is reasonable.

I am only afraid if it is manylinux2020 . 🔥

pllim:
Not sure what is failing this time...

saimn (approved):
Fine for me. We should consider switching to manylinux10 in the future, but it's probably better to wait a little longer.

Cadair:
I think our minimum supported pip can be driven our minimum Python version, as the "latest stable pip is bundled with CPython". I think therefore we should be able to move to manylinux2010 when we drop 3.6.

eteq (approved):
I agree it's a good idea to do this.

My biggest concern is that this impacts people using docker since the "canonical" way to do this would be to use the appropriate CentOS, which is well past end-of-life now.  But I don't think it's a true blocker, and @Cadair's "last pip at time of oldest python" seems like a good standard that puts this not too far off.

eteq:
Looks like we're all in accord on this being the right thing to do for 4.1 so I will merge and backport now.

PR #10875, opened 2020-10-15
TST: Bump Python for dev jobs

pllim:
### Description

Since 3.9 is out, we should test against Python 3.10 dev. Also use numpy-dev jobs to test against Python 3.9.

- [ ] When python-dev job passes, change it back to cron.


pllim:
Hmm... Python 3.10-dev job failed. Something about extension-helpers -- https://travis-ci.org/github/astropy/astropy/jobs/736184631

pllim:
@saimn says it is too early. Should wait for 3.10b to come out first.

PR #10876, opened 2020-10-15
Handle change in default epoch in matplotlib 3.3 

### LINKED ISSUE #10856: Wrong epoch used in Time.plot_date if matplotlib>=3.3
:beetle: **Problem**

`astropy.Times.plot_date` returns seemingly wrong results when matplotlib>3.3 is installed:

```python
from matplotlib.dates import num2date, date2num
t_astropy = Time('2020-10-14T10:00:00')
num2date(t_astropy.plot_date)
```
Result: ```datetime.datetime(3989, 10, 15, 10, 0, tzinfo=datetime.timezone.utc)```
Expected: ```datetime.datetime(2020, 10, 14, 10, 0, tzinfo=datetime.timezone.utc)``` (1969 years into the future)

This is caused by matplotlib [using a new epoch](https://matplotlib.org/3.3.1/users/whats_new.html#dates-use-a-modern-epoch) starting with 3.3 for dates.

:bulb: **Possible solution**

- Probably astropy should use matplotlib.get_epoch if matlotlib >= 3.3 \[in [time/formats.py#L802](https://github.com/astropy/astropy/blob/e74ae753be1f7d5b0386a38dcf9d98690c669e44/astropy/time/formats.py#L802) \]
  (thanks @maxnoe !)

taldcroft:
### Description

This pull request is to handle the change in default epoch in matplotlib 3.3 for plot dates.

Fixes #10856

### DIFF
--- a/astropy/visualization/tests/test_time.py
+++ b/astropy/visualization/tests/test_time.py
@@ -6,11 +6,17 @@ import pytest
 pytest.importorskip('matplotlib')  # noqa
 
 import matplotlib.pyplot as plt
+import matplotlib.dates
 from erfa import ErfaWarning
 
 from astropy.time import Time
 from astropy.visualization.time import time_support
 from astropy.utils.compat.context import nullcontext
+
+# Matplotlib 3.3 added a settable epoch for plot dates and changed the default
+# from 0000-12-31 to 1970-01-01. This can be checked by the existence of
+# get_epoch() in matplotlib.dates.
+MPL_EPOCH_1970 = hasattr(matplotlib.dates, 'get_epoch')
 
 # Since some of the examples below use times/dates in the future, we use the
 # TAI time scale to avoid ERFA warnings about dubious years.
@@ -123,7 +129,8 @@ FORMAT_CASES = [
   ('jyear', ['2020', '2040', '2060']),
   ('jyear_str', ['J2020.000', 'J2040.000', 'J2060.000']),
   ('mjd', ['60000', '66000', '72000', '78000']),
-  ('plot_date', ['738000', '744000', '750000', '756000']),
+  ('plot_date', (['18000', '24000', '30000', '36000'] if MPL_EPOCH_1970 else
+                 ['738000', '744000', '750000', '756000'])),
   ('unix', ['1500000000', '2000000000', '2500000000', '3000000000']),
   ('yday', ['2020:001:00:00:00.000', '2040:001:00:00:00.000', '2060:001:00:00:00.000']),
 ]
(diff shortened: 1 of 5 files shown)


astropy/visualization/tests/test_time.py
```diff
@@ -6,12 +6,18 @@
...
 from erfa import ErfaWarning
 
 from astropy.time import Time
 from astropy.visualization.time import time_support
 from astropy.utils.compat.context import nullcontext
 
+# Matplotlib 3.3 added a settable epoch for plot dates and changed the default
+# from 0000-12-31 to 1970-01-01. This can be checked by the existence of
+# get_epoch() in matplotlib.dates.
+MPL_EPOCH_1970 = hasattr(matplotlib.dates, 'get_epoch')
```
mhvk, line 19: This assumes the user running the tests has not overridden this... I'm happy to go by that assumption but perhaps mention it in the comment above?

mhvk (approved):
This looks good.  I'm approving despite the one comment, since I cannot quite convince myself it is necessary (and it affects only a test, not the actual code).  But up to you!

astropy/visualization/tests/test_time.py
taldcroft: Is that a configuration setting? Otherwise I don't know how a user could set that prior to running this test. Maybe I'm missing something obvious.

mhvk: I think it is indeed configurable - see https://matplotlib.org/3.3.1/users/whats_new.html#dates-use-a-modern-epoch, in particular
```
Access to the epoch is provided by get_epoch, and there is a new rcParams["date.epoch"] (default: '1970-01-01T00:00:00') rcParam.
```
I guess the alternative would be to actually call `get_epoch` as you do in the code proper...

taldcroft: Because the test is about the string tick values, I'm not sure it is going to be possible or easy to get this right for an arbitrary user epoch. In any case there are many other battles to fight that are probably more worth the time, so I'll accept your offer to just merge now. :smile:

astrofrog:
@taldcroft - this was backported to v4.0.x but is causing failures there because for some reason the tests look like the old epoch is still used even with recent versions of Matplotlib. Could you take a quick look at the failures with ``python setup.py test -P visualization`` in v4.0.x and see if you can figure it out?

taldcroft:
@astrofrog - it turns out this PR needs #10406, which changes the handling of the `epoch` attribute. I tried back-porting that PR into my 4.0.x and it passed the failing visualization tests and time tests. I don't see any problem with backporting #10406.

Issue #10738, opened 2020-09-16
Config woes meta-ticket - add ASTROPY_CONFIG_DIR and ASTROPY_CACHE_DIR


embray:
### Description

This ticket does a few things:

* Attempts to synthesize various open tickets related to configuration system woes
* Proposes adding support for `ASTROPY_CONFIG_DIR` and `ASTROPY_CACHE_DIR` environment variables, allowing total customization of these paths by users if desired
* De-emphasizes use of `XDG_` environment variables for configuring where Astropy-related files go, as it is a somewhat system-biased solution and could cause problems for other apps when used by unwary users (while still maintaining support for the XDG directory specification and documenting it in more "intermediate" config documentation)
* Ditto implementation of #8245 to vendor the [appdirs](https://github.com/ActiveState/appdirs) module rather than maintaining our own config path code (or at least a more minimal amount).

I believe this proposal would, in one stroke, resolve:

* #6511
* #6803
* #10244

and maybe some others.  Also related is #10090 which I don't think is in conflict with this proposal--it would still just suppress loading any config file.  Originally I thought of incorporating #8245 into this, but upon further reflection I think it would be counter-productive.

### Detailed proposal

Much of this proposal is already in line with the status quo, but I wanted to write out the logic clearly:

1. By default all Astropy-related directories remain under the path `~/.astropy/`.  This includes `~/.astropy/config/` and `~/.astropy/cache`; this is the current status quo, and is applicable to all platforms, where its platform-specific expansion depends on how [`os.path.expanduser`](https://docs.python.org/3.6/library/os.path.html#os.path.expanduser) works on that platform (see also #10705)

2. Astropy will honor the XDG base directory specification if and only if `XDG_CONFIG_HOME` and/or `XDG_CACHE_HOME` are explicitly set in the environment.  Otherwise it falls back the the default behavior in 1.  If either one of the `XDG_*_HOME` variables is set we commit to the XDG specification.  For example if `XDG_CONFIG_HOME` is set, then the astropy config dir is `$XDG_CONFIG_HOME/astropy`, and the astropy cache dir is `~/.cache/astropy` and not `~/.astropy/cache`.

   * Use of `XDG_*` environment variables will be de-emphasized in the docs, though this logic will be made clear in an admonition box for Linux/BSD users.

[...]

PR #10877, opened 2020-10-16
Make the configuration file optional

### LINKED ISSUE #5899: Suppress configuration file loading

### LINKED ISSUE #6511: XDG_CONFIG_HOME failing silently caused confusion

### LINKED ISSUE #7290: Concurrent import of astropy occasionally fails with likely race condition on config files.

saimn:
Ref discussion in #10738 and #10090 : stop writing/updating the configuration file by default when importing the package, and do nothing when reading if the file is absent.
I also replaced `update_default_config` with a new function `create_config_file` which can be used to create or update the default file. And removed some old stuff for versions < 0.4. And removed the default config file which is present in the package since it can be generated, this will avoid having to update it in the release procedure.

cc @astrofrog  @pllim @embray 

EDIT:
Fix #5899
Close #7290
Close #6511
Close #10090


astropy/config/configuration.py
```diff
@@ -682,148 +665,70 @@ def reload_config(packageormod=None, rootname=None):
...
+def create_config_file(pkg, rootname='astropy', force=False):
     """
-    Checks if the configuration file for the specified package exists,
-    and if not, copy over the default configuration.  If the
-    configuration file looks like it has already been edited, we do
-    not write over it, but instead write a file alongside it named
-    ``pkg.version.cfg`` as a "template" for the user.
+    Create the default configuration file for the specified package.
+    If the file already exists, it is updated only if it has not been
+    modified.  Otherwise the ``force`` flag is needed to overwrite it.
```
astrofrog, line 681: We could make it even simpler and just error if the file already exists unless force is given, without checking content? Would simplify the code further...

astrofrog (commented):
Just a quick comment below and also wanted to say that we should make sure we aren't breaking anything for downstream packages (the change in behaviour is fine, I mean no removal of functions etc that might break downstream)

pllim:
@saimn , since title says WIP, I converted this PR to draft. FYI.

astropy/config/configuration.py
pllim: In my experience though, if user really comes across this exception, we will have a feature request to ask for option to overwrite... 🤷

astropy/config/configuration.py
```diff
@@ -682,148 +665,70 @@ def reload_config(packageormod=None, rootname=None):
...
         The package to be updated.
-    default_cfg_dir_or_fn : str
-        The filename or directory name where the default configuration file is.
-        If a directory name, ``'pkg.cfg'`` will be used in that directory.
-    version : str, optional
-        The current version of the given package.  If not provided, it will
-        be obtained from ``pkg.__version__``.
     rootname : str
         Name of the root configuration directory.
+    force : bool
```
pllim, line 689: More consistent with our existing API over at unified IO?
```suggestion
    overwrite : bool
```

pllim (commented):
I do like the amount of simplification this introduces, but given the behavior change, this should be in 4.3 or later, at least. I have set the milestone to 5.0 for now, but it is open for discussion.

User facing doc also needs to be written to fully explain what to expect with this change. Perhaps even a section for downstream package maintainers.

This patch should be tested with coordinated packages and a few affiliated ones (I would want to test mine at least) before being merged. The consequences of setting config alias, adding new config, removing old config, or changing existing config data type should also be considered.

saimn:
Thanks @pllim, I could not find how to make it a draft PR...

saimn:
> @astrofrog - Just a quick comment below and also wanted to say that we should make sure we aren't breaking anything for downstream packages (the change in behaviour is fine, I mean no removal of functions etc that might break downstream)

Yep, I released after that astroquery may use `update_default_config` (and it does: https://github.com/astropy/astroquery/blob/master/astroquery/_astropy_init.py#L43). astroquery is probably the only package using the configuration system ?
So we could keep `update_default_config` longer, or make it do nothing (keeping it as an empty function would stop updating the config file as for astropy core) ?

pllim:
> astroquery is probably the only package using the configuration system

Actually, a few of my own packages relies on it. For example, `synphot` and `stsynphot`.

> draft PR

Click the small arrow on the giant green button that says "create PR" and there will be a second option to create draft PR. 😉

astropy/config/configuration.py
saimn: Yes, I'm happy to simplify more.

saimn: Thinking more about this, the problem is that the direct string comparison will fail if a configuration item has been added. So `is_unedited_config_file` is a nice a simple way to check if the config file has not been customized, whatever the version was used to generate it.

astropy/config/configuration.py
saimn: Done.

astropy/config/configuration.py
saimn: I restored `update_default_config` since other packages use it, and it relies on `is_unedited_config_file`.

saimn:
> I do like the amount of simplification this introduces, but given the behavior change, this should be in 4.3 or later, at least. I have set the milestone to 5.0 for now, but it is open for discussion.

5.0 seems a bit far way, and I don't think this change would be a big disruption ;)

> User facing doc also needs to be written to fully explain what to expect with this change. Perhaps even a section for downstream package maintainers.

Sure!

> This patch should be tested with coordinated packages and a few affiliated ones (I would want to test mine at least) before being merged. The consequences of setting config alias, adding new config, removing old config, or changing existing config data type should also be considered.

Not sure what you mean with config alias and data type ?
I think the main thing to test is that the `generate_config`/`create_config_file` functions work well for other packages. I'm trying with astroquery, and there are some issues ;) (one of the subpackages is listed 3 times in the generated config file), so I will try to fix that.

astrofrog:
What about milestoning for 4.3 and then merging as soon as possible after the 4.2 branching which will give ample time for testing with downstream packages?

saimn:
Ref #10893 for a few fixes to get `generate_config` works correctly with astroquery.

astropy/config/configuration.py
```diff
@@ -827,3 +777,63 @@ def update_default_config(pkg, default_cfg_dir_or_fn, version=None, rootname='as
...
+    template_content = template_content.read()
+
+    doupdate = True
+
+    # if the file already exists, check that it has not been modified
+    if cfgfn is not None and path.exists(cfgfn):
+        with open(cfgfn, 'rt', encoding='latin-1') as fd:
+            content = fd.read()
+
+        identical = (content == template_content)
```
embray, line 822: Instead of comparing the config files byte-for-byte, why not just parse the existing config file and compare it to the default configuration?

This might require some refactoring of `generate_config` to have it generate the actual config data structure first, then a separate method for writing it to a file; probably how it should have been done in the first place.

Unless we really care about ensuring that it is byte-for-byte identical.

embray (commented):
:+1: aside from this one question.

astropy/config/configuration.py
saimn: Yes, that's the way it was done. But just after `is_unedited_config_file` loads the config file and checks if any keys are defined. So, since now this will not be run at import, maybe we can just keep the second check?

saimn:
I added a test and updated the documentation, so I think this is now ready. Please review again @pllim @embray @astrofrog

CHANGES.rst
```diff
@@ -334,6 +334,9 @@ Other Changes and Additions
 - Refactored the usage of metaclasses in ``astropy.coordinates`` to instead use
   ``__init_subclass__`` where possible. [#11090]
 
+- The configuration file is no more created by default when importing astropy
+  and its existence is no more required. [#10877]
```
embray, line 338: "no more" -> "no longer" x2

astropy/config/configuration.py
```diff
@@ -43,14 +40,6 @@ class InvalidConfigurationItemWarning(AstropyWarning):
     """
 
 
-class ConfigurationMissingWarning(AstropyWarning):
```
embray, line 46: Removing this might break a tiny handful of existing scripts that deliberately import `ConfigurationMissingWarning` in order to suppress it.  I'm not sure if there is a great (existing) way we could create a deprecation warning for it in this case though.

embray: Maybe this could just be noted in the changelog.

astropy/config/configuration.py
```diff
@@ -852,3 +802,59 @@ def update_default_config(pkg, default_cfg_dir_or_fn, version=None, rootname='as
...
+    template_content = io.StringIO()
+    generate_config(pkg, template_content)
+    template_content.seek(0)
+    template_content = template_content.read()
+
+    doupdate = True
+
+    # if the file already exists, check that it has not been modified
+    if cfgfn is not None and path.exists(cfgfn):
+        with open(cfgfn, 'rt', encoding='latin-1') as fd:
```
embray, line 844: Why does it use latin-1?

docs/config/index.rst
```diff
@@ -14,18 +14,26 @@ parameters used in Astropy or affiliated packages without delving into the
 source code to make those changes.
 
 .. note::
-    The configuration system got a major overhaul in ``astropy`` 0.4 as
-    part of APE3. See :ref:`config-0-4-transition` for information
-    about updating code to use the new API.
+    * Before version 4.3 the configuration file was created by default
+      when importing ``astropy``. Its existence was required, which is
+      no more the case.
```
embray, line 19: "no more" -> "no longer"

embray (dismissed):

astropy/config/configuration.py
saimn: Right, we can keep it for now. Btw I should probably add the deprecation warning for `update_default_config` ?

astropy/config/configuration.py
saimn: No idea. I guess this comes from the old times with Python 2 etc. Since the plan is to remove this function in the future, I guess we can keep it as it is for now.

CHANGES.rst
saimn: Done, same below, thanks.

pllim:
The failures make no sense to me and it is unsettling because I see them in `master` too. I'll hold off review until we can fix the CI.

saimn:
@pllim - See #11305.

pllim:
@saimn , please rebase. 😉

saimn:
I'm not sure if the deprecation warning for `update_default_config` is a good idea... A lot of packages use this function as it's used by the template (https://github.com/search?l=Python&q=update_default_config&type=code) so we should first removed it from the template and give time to projects to adopt those changes.

pllim:
Thanks, @saimn ! @embray , since you requested changes, do you want to re-review?

pllim:
There is a conflict now... From the Coordination Meeting 2021, there is desire to revisit this and get it merged.

taldcroft:
To elaborate on what @pllim said, there was specific feedback that writing the config file is a problem in read-only environments like certain HPC platforms. So 👍 from me on the concept (but I plead ignorance on code details).

saimn:
I removed the deprecation warning, please review again.

pllim (commented):
Some comments/questions below. Also:

* Will https://docs.astropy.org/en/latest/config/astropy_config.html#astropy-config-file still be generated?

After this PR is merged, need to open issue to follow up with deprecation and removal of `update_default_config` and `ConfigurationMissingWarning` after communicating the plan to affected downstream packages

docs/config/index.rst
```diff
@@ -14,18 +14,26 @@ parameters used in Astropy or affiliated packages without delving into the
 source code to make those changes.
 
 .. note::
-    The configuration system got a major overhaul in ``astropy`` 0.4 as
-    part of APE3. See :ref:`config-0-4-transition` for information
-    about updating code to use the new API.
+    * Before version 4.3 the configuration file was created by default
+      when importing ``astropy``. Its existence was required, which is
+      no longer the case.
+    * The configuration system got a major overhaul in ``astropy`` 0.4 as
```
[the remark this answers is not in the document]
saimn: Probably not, maybe for historical purpose ?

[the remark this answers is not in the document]
pllim: I vote that we just remove mention of 0.4. Pretty sure no one uses it now.

docs/config/index.rst
```diff
@@ -216,17 +224,15 @@ Upgrading ``astropy``
 ---------------------
 
 Each time you upgrade to a new major version of ``astropy``, the
-configuration parameters may have changed.
+configuration parameters may have changed. If you want to create a new
```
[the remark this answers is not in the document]
saimn: Nothing bad happens because the default values will be used for configuration items that are not in the file. Same as now since config items are commented by default. So the only thing that will happen is that users may have an old configuration file from an old version, and thus they may miss some configuration items that will be added in or after 4.3. But we don't add config item so often, and the documentation is there to explain how to update this file if needed.

saimn:
image
:astonished:

saimn:
Oh, that's the master→default rename ...

docs/config/index.rst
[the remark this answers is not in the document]
pllim: Well, I guess it is not clear to me whether the doc is meant for end-user or downstream package maintainers, but maybe it is just me.

docs/config/index.rst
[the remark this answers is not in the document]
pllim: I need to merge this pronto to fix CI, so I guess this note entry survives another day.

pllim (approved):
Let's do it! ✔️

docs/config/index.rst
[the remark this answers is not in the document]
saimn: Right, but for the configuration system I'm not sure if downstream packages have other documentation, I guess astropy documentation is the reference one.

Issue #9864, opened 2020-01-16
test_c_reader tests extremely slow on MacOS X with Python 3.8


astrofrog:
The ASCII C reader tests are extremely slow (>5 minutes for just those tests) on my Mac (10.15) with Python 3.8 - I suspect maybe due to the change in default multiprocessing method to 'spawn' (this is a Python change):

```
% pytest astropy/io/ascii/tests/test_c_reader.py --durations=50 -v
Internet access disabled
================================================================= test session starts =================================================================
platform darwin -- Python 3.8.0, pytest-5.3.3.dev61+g4a265ba38.d20200114, py-1.8.0, pluggy-0.13.0 -- /Users/tom/python/dev/bin/python3.8
cachedir: .pytest_cache
Matplotlib: 3.1.1+2995.gbdcd9fcc3
Freetype: 2.6.1
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/tom/tmp/astropy/.hypothesis/examples')

Running tests with Astropy version 4.1.dev27325.
Running tests in astropy/io/ascii/tests/test_c_reader.py.

Date: 2020-01-16T22:07:43

Platform: macOS-10.15.2-x86_64-i386-64bit

Executable: /Users/tom/python/dev/bin/python3.8

Full Python Version: 
3.8.0 (v3.8.0:fa919fdf25, Oct 14 2019, 10:23:27) 
[Clang 6.0 (clang-600.0.57)]

encodings: sys: utf-8, locale: UTF-8, filesystem: utf-8
byteorder: little
float info: dig: 15, mant_dig: 15

Package versions: 
Numpy: 1.18.0
Scipy: 1.3.2
Matplotlib: 3.1.1+2995.gbdcd9fcc3
h5py: 2.10.0
Pandas: 0.25.3
Cython: 0.29.14
Scikit-image: 0.16.2
asdf: 2.4.2
astropy-helpers: 4.0.1

Using Astropy options: remote_data: none.

rootdir: /Users/tom/tmp/astropy, inifile: setup.cfg
plugins: mpl-0.11, repeat-0.8.0, arraydiff-0.3, remotedata-0.3.2, hypothesis-5.1.2, asdf-2.4.2, cov-2.8.1, astropy-header-0.1.2, build-0.1.dev0, doctestplus-0.5.1.dev5+gd3d2207.d20191229, filter-subpackage-0.1.1, openfiles-0.5.0.dev0
[...]

Issue #8858, opened 2019-06-16
Assess performance of parallel io.ascii reader with 'spawn' multiprocessing method


astrofrog:
Based on my understanding of the fast C reader after the work I did in #8853, currently when using the multiprocessing 'spawn' method, it might be that the parallel reader is not efficient because it looks like it will pickle the whole contents of the table (via the ``__reduce__`` method). This would mean memory mapping isn't used. So this issue is a reminder that we should check if this is indeed the case. If so, we might want to consider passing the filename to sub-processes rather than the memory mapped bytes string.

pllim:
And from the other thread, if I understood correctly, performance might also be OS-dependent?

astrofrog:
By default yes, but on Linux or Mac one can easily test both multiprocessing start methods (see #8851 for some example code)

astrofrog:
I finally had a chance to crunch some numbers with a simple CSV table with 10^4 to 10^7 lines - this is on MacOS X with Python 3.8 so with the spawn method:

```python
In [1]: with open('large_ascii.csv', 'w') as f:
   ...:     for line in range(10000):
   ...:         f.write('1,3.2,banana,3.441\n')
   ...: 

In [2]: from astropy.io import ascii
   ...: 

In [3]: %timeit ascii.read('large_ascii.csv', format='csv', delimiter=',', guess=False, fast_reader={'parallel': False, 'use_fast_converter': True})
   ...: 
5.82 ms ± 636 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit ascii.read('large_ascii.csv', format='csv', delimiter=',', guess=False, fast_reader={'parallel': True, 'use_fast_converter': True})
   ...: 
933 ms ± 70.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: with open('large_ascii.csv', 'w') as f:
   ...:     for line in range(100000):
   ...:         f.write('1,3.2,banana,3.441\n')
   ...: 

In [6]: %timeit ascii.read('large_ascii.csv', format='csv', delimiter=',', guess=False, fast_reader={'parallel': False, 'use_fast_converter': True})
   ...: 
57.1 ms ± 6.09 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit ascii.read('large_ascii.csv', format='csv', delimiter=',', guess=False, fast_reader={'parallel': True, 'use_fast_converter': True})
   ...: 
[...]

PR #10880, opened 2020-10-16
Ignore parallel option in fast C reader

astrofrog:
As discussed in https://github.com/astropy/astropy/issues/8858, this ignores for now the parallel option in the fast C reader in io.ascii, since serial reading has better performance. The warning is deliberately not a deprecation warning as this may be fixed in future.

This does mean that we now have some code that isn't tested and I think it's important that we simply remove this code if the bug isn't fixed in a timely way, so I'll open an issue to remind us to remove this code, and I'll milestone it for 4.2 initially though we could discuss what a good time frame is. I think it makes sense to be reasonably aggressive or we'll forget about it. In any case we should keep this present PR simple and we can backport it to LTS and 4.1.

Running the test suite before this PR on MacOS X with Python 3.8:

```
% pytest astropy/io/ascii                            
...
======= 547 passed, 27 skipped, 6 xfailed in 293.08s (0:04:53) =======
```

and with this PR: 

```
======= 547 passed, 27 skipped, 6 xfailed in 18.73s =======
```

Fixes https://github.com/astropy/astropy/issues/9864

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -953,6 +953,11 @@ astropy.extern
 astropy.io.ascii
 ^^^^^^^^^^^^^^^^
 
+- Partially fixed a performance issue when reading in parallel mode. Parallel
+  reading does not work correctly and has worse performance than serial
+  reading, so we now ignore the parallel option and fall back to serial reading.
+  [#10880]
+
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -954,7 +954,7 @@ astropy.io.ascii
 ^^^^^^^^^^^^^^^^
 
 - Partially fixed a performance issue when reading in parallel mode. Parallel
-  reading does not work correctly and has worse performance than serial
+  reading currently has substantially worse performance than the default serial
   reading, so we now ignore the parallel option and fall back to serial reading.
   [#10880]
 
--- a/astropy/io/ascii/cparser.pyx
+++ b/astropy/io/ascii/cparser.pyx
@@ -224,7 +224,8 @@ cdef class CParser:
         # warning. We keep the parallel code below so that it can be fixed in
         # future, but if it cannot be fixed we should remove the parallel code
         # and deprecate the option itself. For now the warning is not a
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -953,6 +953,11 @@ astropy.extern
 astropy.io.ascii
 ^^^^^^^^^^^^^^^^
 
+- Partially fixed a performance issue when reading in parallel mode. Parallel
+  reading does not work correctly and has worse performance than serial
```
taldcroft, line 957: When you say "does not work correctly" that could imply it is returning the wrong answer. I did not see any statements this the answer is wrong, just that it is slow. If so, I would reword this here and elsewhere saying that "Parallel reading currently has substantially worse performance than the default serial reading, so we now ignore ...".

astropy/io/ascii/cparser.pyx
```diff
@@ -219,6 +219,17 @@ cdef class CParser:
...
 
+        # FIXME: for now the parallel mode does not work correctly and is worse
+        # than non-parallel mode so we disable parallel mode if set and emit a
+        # warning. We keep the parallel code below so that it can be fixed in
+        # future, but if it cannot be fixed we should remove the parallel code
+        # and deprecate the option itself. For now the warning is not a
+        # deprecation warning since we may still fix it in future.
+        if parallel:
+            warnings.warn('parallel reading does not currently work, '
+                          'so falling back to serial reading', AstropyWarning)
```
taldcroft, line 230: I think it might help anybody that hits this to have a reference to issue #8858 in the warning text.

taldcroft:
About the milestone for removing the code, since the 4.2 feature freeze is 7 days away that seems entirely unrealistic. I'd just start off with the milestone of 4.3 for either fixing or removing the parallel code.

CHANGES.rst
astrofrog: Done

astropy/io/ascii/cparser.pyx
astrofrog: Done

astrofrog:
@taldcroft - I think your comments are addressed now

taldcroft (approved):
LGTM

PR #10881, opened 2020-10-16
fix division by zero warnings for values near sun

### LINKED ISSUE #10878: Improved light-deflection code can have issues for the center of the Sun
#10666 improved the handling of light deflection for coordinate transformation, but it apparently can have issues when transforming the center of the Sun.  I've seen it happen once on the SunPy CI, where the `RuntimeWarning` is treated as an error.  However, I haven't succeeded yet with reproducing it locally, presumably because it depends on a very precise selection of time (sounds like a job for `hypothesis`...).  Here's an example traceback ([failed build](https://dev.azure.com/sunpy/sunpy/_build/results?buildId=10241&view=logs&j=7f33e5bd-7764-5d8a-ba2e-506e078b9c3f&t=c7676921-0379-54a2-a294-d1797afb29c4&l=4412)), where:
```python
sun = SkyCoord(0*u.deg, 0*u.deg, 0*u.AU, frame='hcrs', obstime=time)
time.utc = "2020-10-15 17:21:50.208105"
```
Traceback:
```python
../../.tox/py38-devdeps/lib/python3.8/site-packages/sunpy/coordinates/sun.py:199: in apparent_longitude
    coord = sun.transform_to(GeocentricMeanEcliptic(equinox=time))
../../.tox/py38-devdeps/lib/python3.8/site-packages/astropy/coordinates/sky_coordinate.py:619: in transform_to
    new_coord = trans(self.frame, generic_frame)
../../.tox/py38-devdeps/lib/python3.8/site-packages/astropy/coordinates/transformations.py:1392: in __call__
../../.tox/py38-devdeps/lib/python3.8/site-packages/astropy/coordinates/transformations.py:995: in __call__
    return supcall(fromcoord, toframe)
../../.tox/py38-devdeps/lib/python3.8/site-packages/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py:135: in icrs_to_gcrs
    gcrs_ra, gcrs_dec = atciqz(srepr.without_differentials(), astrom)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

srepr = <SphericalRepresentation (lon, lat, distance) in (rad, rad, AU)
    (3.50527501, -0.15299223, 0.99699593)>
astrom = (20.78911513, [0.91479605, 0.35636384, 0.15457437], [0.92367677, 0.35156317, 0.15239608], 0.99699593, [-3.97414141e-05...5, 1.41592477e+190, 2.19341577e-152, 1.10639145e+200, 2.34783525e+251, 1.46921998e+195, 3.4430453e-85, 2.65186294e-258)

    def atciqz(srepr, astrom):
        """
        A slightly modified version of the ERFA function ``eraAtciqz``.
    
        ``eraAtciqz`` performs the transformations between two coordinate systems,
        with the details of the transformation being encoded into the ``astrom`` array.
    
        There are two issues with the version of atciqz in ERFA. Both are associated
        with the handling of light deflection.
    
        The companion function ``eraAticq`` is meant to be its inverse. However, this
        is not true for directions close to the Solar centre, since the light deflection
        calculations are numerically unstable and therefore not reversible.
    
        This version sidesteps that problem by artificially reducing the light deflection
        for directions which are within 90 arcseconds of the Sun's position. This is the
        same approach used by the ERFA functions above, except that they use a threshold of
        9 arcseconds.
    
        In addition, ERFA's atciqz assumes a distant source, so there is no difference between
        the object-Sun vector and the observer-Sun vector. This can lead to errors of up to a
        few arcseconds in the worst case (e.g a Venus transit).
    
        Parameters
        ----------
        srepr : `~astropy.coordinates.SphericalRepresentation`
            Astrometric ICRS position of object from observer
        astrom : eraASTROM array
            ERFA astrometry context, as produced by, e.g. ``eraApci13`` or ``eraApcs13``
    
        Returns
        --------
        ri : float or `~numpy.ndarray`
            Right Ascension in radians
        di : float or `~numpy.ndarray`
            Declination in radians
        """
        # ignore parallax effects if no distance, or far away
        srepr_distance = srepr.distance
        ignore_distance = srepr_distance.unit == u.one
    
        # BCRS coordinate direction (unit vector).
        pco = erfa.s2c(srepr.lon.radian, srepr.lat.radian)
    
        # Find BCRS direction of Sun to object
        if ignore_distance:
            # No distance to object, assume a long way away
            q = pco
        else:
            # Find BCRS direction of Sun to object.
            # astrom['eh'] and astrom['em'] contain Sun to observer unit vector,
            # and distance, respectively.
            eh = astrom['em'][..., np.newaxis] * astrom['eh']
            # unit vector from Sun to object
            q = eh + srepr_distance[..., np.newaxis].to_value(u.au) * pco
            sundist = np.linalg.norm(q, axis=-1, keepdims=True)
>           q /= sundist
E           RuntimeWarning: invalid value encountered in true_divide

../../.tox/py38-devdeps/lib/python3.8/site-packages/astropy/coordinates/builtin_frames/utils.py:308: RuntimeWarning
```
Even for Sun center, `sundist` will usually calculate to some very small, but nonzero, distance.  I presume that in this freak circumstance, `sundist` calculated to exactly zero, although I don't understand why NumPy is raising an "invalid value" warning instead of a "divide by zero" warning, so maybe something different has gone wrong?

The obvious workaround is to wrap the problematic division statement with `with np.errstate(divide='ignore', invalid='ignore'):`.

ping @StuartLittlefair

StuartLittlefair:
The changes to light deflection in #10666 caused some problems for the actual centre of the Sun. In that case the code raises division by zero warnings which can break continuous integration tools. These warnings are also ugly and unnecessary since any affected vectors are replaced with the original directions.

As suggested by @ayshih we wrap these lines in ```np.errstate``` to prevent the warnings.
 
#10666 is currently milestoned for 4.1, so this has been given the same milestone.

Fixes #10878

### DIFF
--- a/astropy/coordinates/builtin_frames/utils.py
+++ b/astropy/coordinates/builtin_frames/utils.py
@@ -233,12 +233,8 @@ def aticq(srepr, astrom):
             eh = astrom['em'][..., np.newaxis] * astrom['eh']
             # unit vector from Sun to object
             q = eh + srepr_distance[..., np.newaxis].to_value(u.au) * before
-            sundist = np.linalg.norm(q, axis=-1, keepdims=True)
-            # division by small sun distances will raise warnings/errors that
-            # we can ignore, since any affected values are removed in lines
-            # that follow.
-            with np.errstate(divide='ignore', invalid='ignore'):
-                q /= sundist
+            sundist, q = erfa.pn(q)
+            sundist = sundist[..., np.newaxis]
             # calculation above is extremely unstable very close to the sun
             # in these situations, default back to ldsun-style behaviour,
             # since this is reversible and drops to zero within stellar limb
@@ -308,12 +304,8 @@ def atciqz(srepr, astrom):
         eh = astrom['em'][..., np.newaxis] * astrom['eh']
         # unit vector from Sun to object
         q = eh + srepr_distance[..., np.newaxis].to_value(u.au) * pco
-        sundist = np.linalg.norm(q, axis=-1, keepdims=True)
-        # division by small sun distances will raise warnings/errors that
-        # we can ignore, since any affected values are removed in lines
-        # that follow.
-        with np.errstate(divide='ignore', invalid='ignore'):
-            q /= sundist
+        sundist, q = erfa.pn(q)
+        sundist = sundist[..., np.newaxis]
         # calculation above is extremely unstable very close to the sun
         # in these situations, default back to ldsun-style behaviour,
         # since this is reversible and drops to zero within stellar limb
--- a/astropy/coordinates/builtin_frames/utils.py
+++ b/astropy/coordinates/builtin_frames/utils.py
@@ -234,7 +234,11 @@ def aticq(srepr, astrom):
             # unit vector from Sun to object
             q = eh + srepr_distance[..., np.newaxis].to_value(u.au) * before
             sundist = np.linalg.norm(q, axis=-1, keepdims=True)
-            q /= sundist
+            # division by small sun distances will raise warnings/errors that
+            # we can ignore, since any affected values are removed in lines
+            # that follow.
+            with np.errstate(divide='ignore', invalid='ignore'):
+                q /= sundist
             # calculation above is extremely unstable very close to the sun
             # in these situations, default back to ldsun-style behaviour,
             # since this is reversible and drops to zero within stellar limb
@@ -305,7 +309,11 @@ def atciqz(srepr, astrom):
         # unit vector from Sun to object
         q = eh + srepr_distance[..., np.newaxis].to_value(u.au) * pco
         sundist = np.linalg.norm(q, axis=-1, keepdims=True)
-        q /= sundist
+        # division by small sun distances will raise warnings/errors that
+        # we can ignore, since any affected values are removed in lines
+        # that follow.
+        with np.errstate(divide='ignore', invalid='ignore'):
+            q /= sundist
         # calculation above is extremely unstable very close to the sun
         # in these situations, default back to ldsun-style behaviour,
         # since this is reversible and drops to zero within stellar limb


mhvk (approved):
This looks good. Tests are failing because of an unrelated issue (see #10874).

@eteq - this will need to join #10666 for backporting to 4.0.3

mkbrewer:
Use of `erfa.pn()` would take care of this also.

mhvk:
indeed, [`erfa.pn`](https://pyerfa.readthedocs.io/en/latest/api/erfa.pn.html#erfa.pn) is a logical choice here - @StuartLittlefair - obviously not a big deal, but maybe an idea to change it? It will do the right thing, with xyz as the last dimension.

StuartLittlefair:
```erfa.pn``` is a nice solution, but it doesn't handle shapes perfectly and I lack @mhvk's numpy wizardry to be sure of doing the right thing. e.g suppose the ```q``` array is (N, 3), the way I do it gives ```sundist``` a shape of (N, 1) so I can use it in ```np.where``` later. 

If I use ```erfa.pn``` then ```sundist``` has shape (N,) - so I could use ```sundist[:, np.newaxis]``` but then that doesn't work for scalars? Does the solution below work?

```
sundist, q = erfa.pn(q)
sundist = np.atleast_1d(sundist)[:, np.newaxis]
q = np.where(sundist > 1.0e-10, q, pco)
```

**Comment by mhvk** [comment]
@StuartLittlefair - it would be `sundist[..., np.newaxis]` to append at the end (works for scalars too):
```
erfa.pn([1.,2.,3.])[0][..., np.newaxis]
Out[15]: array([3.74165739])
erfa.pn(np.ones((2,2,3)))[0][..., np.newaxis]
Out[16]: 
array([[[1.73205081],
        [1.73205081]],

       [[1.73205081],
        [1.73205081]]])
```

StuartLittlefair:
God bless you @mhvk and your numpy-Fu. And thanks @mkbrewer for the erfa reminder...

mhvk (approved):
Nicer, indeed!

eteq (approved):
CI failure is unrelated, so I think this is all good and can be backported to 4.1 once the issues still in the #10666 backport  are ironed out.

PR #10883, opened 2020-10-16
Remove extension auto-building machinery

astrofrog:
This might be controversial but I'm on a mission to try and simplify things wherever possible and remove magic.

I personally get frustrated when trying to import astropy in e.g. a notebook and astropy is installed in develop mode and it then just starts trying to build the extension modules on the fly, a slow process. I'd rather just have an error telling me the extension modules aren't built and how to fix it. I'm also not very comfortable with the rolling back of the import which modifies ``sys.modules``.

Furthermore, ``setup.py build_ext --inplace`` may not always work if build-time dependencies (extension-helpers, Cython, etc) are missing, so I'm putting the ``pip`` suggestion first and mentioning the build dependency issue.

But maybe it's just me and others find this feature useful? We should wait for a number of core devs to weigh in before merging this - approve if you agree, leave a comment if you don't.

### DIFF
--- a/astropy/__init__.py
+++ b/astropy/__init__.py
@@ -51,21 +51,6 @@ def _is_astropy_source(path=None):
     return os.path.exists(os.path.join(source_dir, '.astropy-root'))
 
 
-def _is_astropy_setup():
-    """
-    Returns whether we are currently being imported in the context of running
-    Astropy's setup.py.
-    """
-
-    main_mod = sys.modules.get('__main__')
-    if not main_mod:
-        return False
-
-    return (getattr(main_mod, '__file__', False) and
-            os.path.basename(main_mod.__file__).rstrip('co') == 'setup.py' and
-            _is_astropy_source(main_mod.__file__))
-
-
 # The location of the online documentation for astropy
 # This location will normally point to the current released version of astropy
 if 'dev' in __version__:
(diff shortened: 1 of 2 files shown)


Cadair:
Yes +:100: from me

astropy/__init__.py
```diff
@@ -205,48 +190,24 @@ class astronomical_constants(base_constants_version):
...
-                    raise
-
+            raise ImportError('You appear to be trying to import astropy from '
+                              'within a source checkout or from an editable '
+                              'installation without building the extension '
+                              'modules first. Either run:\n\n'
+                              '  pip install -e .\n\nor\n\n'
+                              '  python setup.py build_ext --inplace\n\n'
+                              'to make sure the extension modules are built '
+                              '(note that if you use the latter you may need '
```
saimn, line 207:
```suggestion
                              '(note that if you use the latter you need '
```
It's more than a supposition right?

saimn (approved):
Totally agree! Especially since this will break if ones rely on pip to install build dependencies.

mhvk (approved):
Yes, whenever this happens I tend to ^C because I realize I made a mistake - an error message would be much better -- In the face of ambiguity, refuse the temptation to guess.

mhvk:
p.s. Travis seems to be running twice - maybe related to the switch to travis-ci.com?

pllim:
CircleCI 32-bit/parallel failure should have been fixed. Try rebase?

> related to the switch to travis-ci.com?

Probably. Perhaps delete the webhook to `.org`?

astrofrog:
I thought the org side would turn off by default. I'll investigate.

Cadair (approved):

taldcroft:
:+1: on this.

pllim (approved):
*Just do it*

astropy/__init__.py
```diff
@@ -205,48 +190,24 @@ class astronomical_constants(base_constants_version):
...
+                              'to install build dependencies manually)')
```
eteq, line 208: honestly I'm not sure this warning is necessary. It seems like it comes with the territory.  We're better off documenting this fact in detail or just not bothering to say it at all.  E.g.,
```suggestion
                              'to make sure the extension modules are built ')
```
or
```suggestion
                              'to make sure the extension modules are built '
                              '(see https://docs.astropy.org/en/stable/development/building.html)')
```

eteq (approved):
Even *I* am in favor of this! Clearly this is a thing worth :fire:-ing down.

(I had just one minor doc suggestion)

pllim:
Will this fix the conda failure?

astropy/__init__.py
astrofrog: I went with @eteq's suggestion which was even more drastic

astrofrog:
@pllim which conda failure? I don't think this should have any impact on conda?

pllim:
@astrofrog https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=229626&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=66e4a587-a723-53e9-873c-6c9a2c5971ea&l=4344

astrofrog:
I doubt it will fix that but I'm not 100% sure

embray:
This was a cool feature at the time, but I think it's been mostly obviated by the ease of running `pip install -e .` which should be done anyways.

PR #10884, opened 2020-10-16
Remove config and file related to deprecated bdist_wininst

astrofrog:
The ``bdist_wininst`` setup.py command is [deprecated as of Python 3.8](https://docs.python.org/3/distutils/builtdist.html) and we don't document it or use it anyway (we provide wheels), so I think it makes sense to remove.

Don't think we need a changelog entry since we never really used or advertised this?


pllim (approved):
I don't know this existed. 🔥

pllim:
@astrofrog , feel free to merge once you set the milestone.

eteq (approved):
LFTM too (failures are  #10874 so red herrings)

PR #10885, opened 2020-10-16
Backport light deflection

mhvk:
This reverts the earlier attempt to backport #10666, and then adds a single commit that includes corrected versions of #10666 and #10881. This single commit is likely easy to backport to 4.0.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -685,10 +685,6 @@ astropy.convolution
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
 
-- Fix a bug where for light deflection by the Sun it was always assumed that the
-  source was at infinite distance, which in the (rare and) absolute worst-case
-  scenario could lead to errors up to 3 arcsec. [#10666]
-
 astropy.cosmology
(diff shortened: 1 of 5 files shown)


mhvk:
@eteq - the failures seem unrelated.

mhvk:
And, yes, good to keep this as is, including the history.

mhvk:
Also: I just checked: the commit backports cleanly to 4.0.x (except the changelog entry, which goes in 4.0.3 for some reason). Shall I push that to 4.0.x directly?

eteq:
Thanks for the fixed backport, @mhvk! I see now how intertwined these are but this seems to do it.

Re: v4.0.x, Not yet @mhvk - we should leave that to the next 4.0.x releaser (which I think will be @astrofrog).  Also, sorry, but this raises one more question I want to bring back to #10666 (just trying to put the discussion there instead of here so it's all in one place).

To be clear, though, if we do agree to do the 4.0.x backport there, whoever does the backport should backport #10666 and then this, and that should do the trick.

PR #10889, opened 2020-10-17
Do not copy info when getting scalar Time item

### LINKED ISSUE #10688: Excessive delay and memory usage when calling unsupported Numpy functions on TimeSeries.time
### Description
Trying to call unsupported Numpy ufuncs or other functions like `np.isnan` or `np.mean` on a `Time` objects raises an exception, offering an (implicit) reminder to directly use the `np.array` via `t.value`.
The same operations on an object created as `TimeSeries.time` for moderately large arrays (length a couple 1000) results in Python running for several minutes and allocating up to several GB (~ len(t) * 200 kB) memory.

### Expected behavior

Unsupported functions should raise an exception without undue delay or memory usage of thousands times the original data size, just as for regular `Time`.

### Actual behavior
For a basic `TimeSeries` like
```
t = Time(np.arange(10000), format='mjd', scale='tdb')
ts = TimeSeries(time=t)
np.isnan(ts.time)
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

np.abs(ts.time)
TypeError: bad operand type for abs(): 'Time'
```
the last two lines each take 30-40 seconds, ramping memory usage multiple times up to > 2 GB as shown in Activity Monitor or `top`.

### Steps to Reproduce
Running the attached script as (`pip install memory_profiler`) a function is trying to run a function several times first on the original `Time` object `t`, then the same as element `ts.time` of a `TimeSeries`, a `deepcopy(ts.time)` of the latter, and again the original `t`.
```
> mprof run python3.8 timeseries_leak_prof.py isnan                                                      
mprof: Sampling memory every 0.1s
1.19.1
4.1rc2
ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11     71.9 MiB     71.9 MiB   @profile
    12                             def np_func(x, func):
    13     71.9 MiB      0.0 MiB       try:
    14     73.2 MiB      1.3 MiB           r = func(x)
    15     73.2 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16     73.2 MiB      0.0 MiB           print(err)
    17     73.2 MiB      0.0 MiB       return


ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11     74.1 MiB     74.1 MiB   @profile
    12                             def np_func(x, func):
    13     74.1 MiB      0.0 MiB       try:
    14    167.2 MiB     93.1 MiB           r = func(x)
    15    167.2 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16    167.2 MiB      0.0 MiB           print(err)
    17    167.2 MiB      0.0 MiB       return


ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11    167.2 MiB    167.2 MiB   @profile
    12                             def np_func(x, func):
    13    167.2 MiB      0.0 MiB       try:
    14    185.8 MiB     18.6 MiB           r = func(x)
    15    185.8 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16    185.8 MiB      0.0 MiB           print(err)
    17    185.8 MiB      0.0 MiB       return


ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11    185.8 MiB    185.8 MiB   @profile
    12                             def np_func(x, func):
    13    185.8 MiB      0.0 MiB       try:
    14    176.4 MiB      0.0 MiB           r = func(x)
    15    176.4 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16    176.4 MiB      0.0 MiB           print(err)
    17    176.4 MiB      0.0 MiB       return

> mprof run python3.8 timeseries_leak_prof.py mean                                                       
mprof: Sampling memory every 0.1s
1.19.1
4.1rc2
Unsupported operand type(s) for +: 'Time' and 'Time'
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11     71.5 MiB     71.5 MiB   @profile
    12                             def np_func(x, func):
    13     71.5 MiB      0.0 MiB       try:
    14     77.8 MiB      6.3 MiB           r = func(x)
    15     77.8 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16     73.1 MiB      0.0 MiB           print(err)
    17     73.1 MiB      0.0 MiB       return


Unsupported operand type(s) for +: 'Time' and 'Time'
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11     74.5 MiB     74.5 MiB   @profile
    12                             def np_func(x, func):
    13     74.5 MiB      0.0 MiB       try:
    14   2429.7 MiB   2355.2 MiB           r = func(x)
    15   2429.7 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16    176.6 MiB      0.0 MiB           print(err)
    17    176.6 MiB      0.0 MiB       return


Unsupported operand type(s) for +: 'Time' and 'Time'
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11    176.6 MiB    176.6 MiB   @profile
    12                             def np_func(x, func):
    13    176.6 MiB      0.0 MiB       try:
    14   2430.2 MiB   2253.7 MiB           r = func(x)
    15   2430.2 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16    185.6 MiB      0.0 MiB           print(err)
    17    185.6 MiB      0.0 MiB       return


Unsupported operand type(s) for +: 'Time' and 'Time'
Filename: timeseries_leak_prof.py

Line #    Mem usage    Increment   Line Contents
================================================
    11    185.6 MiB    185.6 MiB   @profile
    12                             def np_func(x, func):
    13    185.6 MiB      0.0 MiB       try:
    14    176.9 MiB      0.0 MiB           r = func(x)
    15    176.9 MiB      0.0 MiB       except (TypeError, SystemError) as err:
    16    176.7 MiB      0.0 MiB           print(err)
    17    176.7 MiB      0.0 MiB       return
```
Note that
1. `memory_profiler` seems to add an additional 5-6 x timing overhead.

2. The `mprof` output does show the memory usage for `np.mean`, but not for ufuncs (because it's all allocated inside compiled C extensions?). However `top` shows similar usage for the latter, as do the attached plots from `mprof plot`.

3. Both for ufuncs and other functions, memory usage failed to trigger the debugger interface running e.g. `python -m memory_profiler --pdb-mmem=200`.

4. Each of `func(ts.time)` and `func(copy.deepcopy(ts.time))` is responsible for 3 successive peaks in memory usage.

5. In a Python session memory generally is not immediately released after running functions like `np.mean` or `np.median`, but after a `np.ufunc`; with older Python or Numpy versions also interpreter segfaults have been observed.

Tested to the same effect with 4.0.1 and current master (including with #10656 merged).

### System Details

See above for other Astropy versions tested, also with 4.1rc2 on Pythons 3.6.12, 3.7.9, 3.9.0rc1.
```
macOS-10.12.6-x86_64-i386-64bit
Python 3.8.5 (default, Jul 22 2020, 22:38:08) 
Numpy 1.19.1
astropy 4.2.dev560+g679a74a46
Scipy 1.5.2
Matplotlib 3.3.0
```

timeseries_leak_isnan
timeseries_leak_mean
[timeseries_leak_prof.py.gz](https://github.com/astropy/astropy/files/5139485/timeseries_leak_prof.py.gz)

taldcroft:
### Description

This pull request is to address a memory and speed performance problem when iterating over a `Time` column that is indexed (e.g. in a `TimeSeries` object). See #10688 for details.

The simple fix is to not copy the `info` attribute when getting a scalar `Time` from an array-valued `Time` object. This is consistent with the behavior of `Column`.

Being very conservative I have milestoned this for 4.2 instead of as a pure performance fix. It seems quite unlikely, but maybe possible that some code is using the `info` data that were previously being copied for a scalar getitem.

Fixes #10688

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -1082,10 +1082,11 @@ class TimeBase(ShapedLikeNDArray):
 
             setattr(tm, attr, val)
 
-        # Copy other 'info' attr only if it has actually been defined.
+        # Copy other 'info' attr only if it has actually been defined and the
+        # time object is not a scalar (issue #10688).
         # See PR #3898 for further explanation and justification, along
         # with Quantity.__array_finalize__
-        if 'info' in self.__dict__:
+        if 'info' in self.__dict__ and tm.shape:
             tm.info = self.info
 
         # Make the new internal _time object corresponding to the format
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -1086,7 +1086,7 @@ class TimeBase(ShapedLikeNDArray):
         # time object is not a scalar (issue #10688).
         # See PR #3898 for further explanation and justification, along
         # with Quantity.__array_finalize__
-        if 'info' in self.__dict__ and tm.shape:
+        if 'info' in self.__dict__:
             tm.info = self.info
 
         # Make the new internal _time object corresponding to the format
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -106,11 +106,6 @@ astropy.time
 - Add new ``isclose()`` method to ``Time`` and ``TimeDelta`` classes to allow
   comparison of time objects to within a specified tolerance. [#10646]
 
-- Change behavior so that when getting a single item out of an array-valued
-  ``Time`` or ``TimeDelta`` object, the ``info`` attribute is no longer copied.
-  This improves performance, especially when the object is an indexed column in
-  a ``Table``. [#10889]
-
 - Improve initialization time by a factor of four when creating a scalar ``Time``
   object in a format like ``unix`` or ``cxcsec`` (time delta from a reference
   epoch time). [#10406]
@@ -236,6 +231,11 @@ astropy.table
   With this, the behaviour becomes the same as that for a regular ``Column``.
   (Note that this does not affect slicing of a table; sliced columns in those
   will continue to carry a sliced version of any indices). [#10890]
+
+- Change behavior so that when getting a single item out of a mixin column such
+  as ``Time``, ``TimeDelta``, ``SkyCoord`` or ``Quantity``, the ``info``
+  attribute is no longer copied. This improves performance, especially when the
+  object is an indexed column in a ``Table``. [#10889]
 
 astropy.tests
 ^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -106,6 +106,11 @@ astropy.time
 - Add new ``isclose()`` method to ``Time`` and ``TimeDelta`` classes to allow
   comparison of time objects to within a specified tolerance. [#10646]
 
+- Change behavior so that when getting a single item out of an array-valued
+  ``Time`` or ``TimeDelta`` object, the ``info`` attribute is no longer copied.
+  This improves performance, especially when the object is an indexed column in
+  a ``Table``. [#10889]
+
 - Improve initialization time by a factor of four when creating a scalar ``Time``
   object in a format like ``unix`` or ``cxcsec`` (time delta from a reference
   epoch time). [#10406]
@@ -115,6 +120,10 @@ astropy.time
 
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
+
+- Improve memory and speed performance when iterating over the entire time
+  column of a ``TimeSeries`` object. Previously this involved O(N^2) operations
+  and memory. [#10889]
 
 astropy.uncertainty
 ^^^^^^^^^^^^^^^^^^^
--- a/astropy/table/tests/test_mixin.py
+++ b/astropy/table/tests/test_mixin.py
@@ -477,9 +477,9 @@ def test_add_column(mixin_cols):
                 assert getattr(t['m'].info, attr) == getattr(t[name].info, attr)
     # Also check that one can set using a scalar.
     s = m[0]
-    if type(s) is type(m):
+    if type(s) is type(m) and 'info' in s.__dict__:
         # We're not going to worry about testing classes for which scalars
-        # are a different class than the real array (and thus loose info, etc.)
+        # are a different class than the real array, or where info is not copied.
         t['s'] = m[0]
         assert_table_name_col_equal(t, 's', m[0])
         for attr in attrs:
@@ -488,7 +488,7 @@ def test_add_column(mixin_cols):
 
     # While we're add it, also check a length-1 table.
     t = QTable([m[1:2]], names=['m'])
-    if type(s) is type(m):
+    if type(s) is type(m) and 'info' in s.__dict__:
         t['s'] = m[0]
         assert_table_name_col_equal(t, 's', m[0])
         for attr in attrs:
--- a/astropy/time/tests/test_basic.py
+++ b/astropy/time/tests/test_basic.py
@@ -2220,21 +2220,6 @@ def test_broadcasting_writeable():
     t[2] = Time(58000, format="mjd")
 
 
-def test_info_no_copy():
-    """Test that getting a single item from a Time object does not copy info.
-    See #10889.
-    """
-    tm = Time([1, 2], format='cxcsec')
-    t = Table([tm], names=['tm'])
-    t.add_index('tm')
-    val = t['tm'][0]
-    assert 'info' not in val.__dict__
-    assert val.info.indices is None
-    val = t['tm'][:]
-    assert 'info' in val.__dict__
-    assert val.info.indices is not None
-
-
 def test_format_subformat_compatibility():
     """Test that changing format with out_subfmt defined is not a problem.
     See #9812, #9810."""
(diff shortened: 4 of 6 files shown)


mhvk:
The PR looks OK, but I'm still wondering whether it is the right approach: the following behaviour seems weird to me:
```
import numpy as np
from astropy.time import Time
t = Table([Time(np.arange(50000, 50010), format='mjd')], names=['t'])
t.add_index('t')
cx = t['t'][3:4]
cx.info.indices
[   t    rows
 ------- ----
 50000.0    0
 50001.0    1
 50002.0    2
 50003.0    3
 50004.0    4
 50005.0    5
 50006.0    6
 50007.0    7
 50008.0    8
 50009.0    9]
# Yet
cx.info.parent_table
# gives None
```
Why would a sliced time still have the full index associated with it?

In contrast, I can see the logic of
```
tx = t[3:4]
tx['t'].info.indices
[Index slice (3, 4, 1) of
    t    rows
 ------- ----
 50000.0    0
 50001.0    1
 50002.0    2
 50003.0    3
 50004.0    4
 50005.0    5
 50006.0    6
 50007.0    7
 50008.0    8
 50009.0    9]
tx['t'].info.parent_table
<Table length=1>
   t   
 object
-------
50003.0
```

I guess I wonder if `indices` should be omitted if the column is sliced; it looks like this would work fine, since the table already takes care of updating it if it is the table rather than the column that is sliced.

The advantage of solving it at the table level would be that the same solution would work for all other possible mixin columns too.

mhvk:
I think I've got an alternative, I think - see #10890

mhvk:
Moving discussion here about whether this is still needed given #10890. I can see the argument in https://github.com/astropy/astropy/pull/10890#issuecomment-711150218 that
> for a scalar getitem I really think we don't want to copy `info` at all. For instance, the info attribute can have meta, and that might be a 5-megabyte structure. I just confirmed that a scalar getitem makes a deepcopy of that meta (in master and in this PR), so doing `np.array(tm)` (where `tm` is a large array-valued Time object with a big info.meta) will explode memory again.

Since this would hold also for `SkyCoord`, `Quantity`, etc., I think we'd need to do the same things for those, which suggests involving `info` itself in the decision about what would need to be copied. E.g., in the mechanism for `DataInfo.__set__` one could add a check on its parents shape and do nothing if that shape was `()`: 
https://github.com/astropy/astropy/blob/079fcb73aedb9396d183454b98811b39f77232dc/astropy/utils/data_info.py#L347-L350

Note that we struggled a bit with this before - see #5047. I think the conclusion there was a good one: we should have clear, predictable behaviour...

p.s. Another possibility for implementation is to use that for most classes any subsetting, etc., goes through `_apply()` - we could change that to call a new, to-be-defined `DataInfo._apply` - that would allow including information about what is being done. But probably overkill.

taldcroft:
> Since this would hold also for SkyCoord, Quantity, etc., I think we'd need to do the same things for those, which suggests involving info itself in the decision

Good plan. The question is whether this should only apply to the `BaseColumnInfo` class or `DataInfo` itself. I'm hesitant about extending this behavior change outside of tables.  For `BaseColumnInfo` I just tried this out and it passes the new test in this PR (after reverting the change in time/core.py):
```
    def __set__(self, instance, value):
        # For Table columns do not set `info` when the instance is a scalar.
        try:
            if not instance.shape:
                return
        except AttributeError:
            pass

        super().__set__(instance, value)
```

mhvk:
Nice! I think I agree that it makes most sense to do it to `BaseColumnInfo`, since then a class that uses it has more or less explicitly signed up for making itself work with `Table` - and for `Table`'s `Column`, scalars also have no connection to the table any more.

taldcroft:
@mhvk - this should be ready for review.

astropy/time/core.py
```diff
@@ -1082,7 +1082,8 @@ def _apply(self, method, *args, format=None, cls=None, **kwargs):
 
             setattr(tm, attr, val)
 
-        # Copy other 'info' attr only if it has actually been defined.
+        # Copy other 'info' attr only if it has actually been defined and the
```
mhvk, line 1085: This can now be removed/reverted? Or link to this PR perhaps instead of the issue.

mhvk (approved):
Looks good modulo the left-over comment in `Time._apply`, which I think can either be removed or should be clarified. Approving since it is so minor.

mhvk:
p.s. Really like the detailed and thorough tests!

PR #10890, opened 2020-10-17
When slicing a table mixin column, do not copy indices by default.

mhvk:
But rather only do it if the table is sliced.

Alternative to #10889 ~for now, just to discuss and check that all tests still pass (have to run...)~

EDIT: the advantage here is that it solves the problem at the table level.

EDIT 2: this actually affects regular columns as well...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -216,6 +216,12 @@ astropy.table
   Instead, use the ``SCEngine`` indexing engine, which is similar in
   performance and relies on the ``sortedcontainers`` package. [#10622]
 
+- When slicing a mixin column in a table that had indices, the indices are no
+  longer copied since they generally are not useful, having the wrong shape.
+  With this, the behaviour becomes the same as that for a regular ``Column``.
+  (Note that this does not affect slicing of a table; sliced columns in those
+  will continue to carry a sliced version of any indices). [#10890]
+
 astropy.tests
 ^^^^^^^^^^^^^
 
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -1200,6 +1200,8 @@ class Table:
                 # still pass if this line is deleted.  (Each col.info attribute access
                 # is expensive).
                 col.info._copy_indices = True
+            else:
+                newcol.info.indices = []
 
             newcols.append(newcol)
 
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3651,6 +3651,7 @@ class QTable(Table):
             q_cls = getattr(col.unit, '_quantity_class', Quantity)
             qcol = q_cls(col.data, col.unit, copy=False)
             qcol.info = col.info
+            qcol.info.indices = col.info.indices
             col = qcol
         else:
             col = super()._convert_col_for_table(col)
--- a/astropy/table/tests/test_index.py
+++ b/astropy/table/tests/test_index.py
@@ -9,7 +9,7 @@ from .test_table import SetupData
 from astropy.table.bst import BST
 from astropy.table.sorted_array import SortedArray
 from astropy.table.soco import SCEngine, HAS_SOCO
-from astropy.table.table import QTable, Row, Table
+from astropy.table import QTable, Row, Table, Column, hstack
 from astropy import units as u
 from astropy.time import Time
 from astropy.table.column import BaseColumn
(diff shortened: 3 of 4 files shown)


mhvk:
(Failure somewhat obscure -- I'll look into it)

mhvk:
@taldcroft - the small bug uncovered by the doctest (ouch!) was because in converting a Column to a Quantity for QTable, the path was not quite the same and the indices were not initialized quite right any more. I now got all tests to pass and included new ones (including one that covers the case identified by the doctest).

For now, I milestoned for 4.2, partially out of worry this breaks something unexpected. But I could go with 4.0.4 too since this does seem a bug fix.

Note: I briefly tried to set the default for `info.indices` to `[]` instead of `None`, but that gave errors I didn't quite understand. Since that is the default for `Column`, I can try to look into that if you think it is better. But perhaps it is best done separately regardless.

taldcroft:
@mhvk - I think this is fine, but I don't think that it is a replacement for #10889, which is explicitly about behavior for getting a scalar item `tm[5]` vs. a slice. The key point is that for a scalar getitem I really think we don't want to copy `info` at all. 

For instance, the info attribute can have `meta`, and that might be a 5-megabyte structure. I just confirmed that a scalar getitem makes a deepcopy of that meta (in master and in this PR), so doing `np.array(tm)` (where `tm` is a large array-valued Time object with a big `info.meta`) will explode memory again.

CHANGES.rst
```diff
@@ -312,6 +312,9 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
+- Fix a bug where slices of columns still carried indices even if they lost
```
taldcroft, line 315: I think this should be called an API change. It was not wrong or unexpected behavior before given the original design in `BaseColumnInfo` (that indices were copied), but we are deciding now that it is more logical / better performance to change the design. Also, although it would be surprising, someone might be expecting the old behavior.

taldcroft (changes requested):
Basically good with the change to the change log. I think that text should just focus on the change which is that if you slice a column (Column or mixin column) that has an index (info.indices), the index is no longer copied into the sliced output.

CHANGES.rst
mhvk: OK to mention it as an API change. The reason I thought it might be more like a bug is that `parent_table` is not copied, which seems to make the indices rather useless. Furthermore, if the slice is taken on the table, the indices are not just copied, but carry information about the slice, so the behaviour is not quite consistent (see below). If it is a bug, in principle, we could backport this to 4.0.

```
In [63]: ts_a
Out[63]: 
<TimeSeries length=5>
          time            flux 
                          mJy  
         object         float64
----------------------- -------
2016-03-22T12:30:31.000     1.0
2016-03-22T12:30:34.000     4.0
2016-03-22T12:30:37.000     5.0
2016-03-22T12:30:40.000     3.0
2016-03-22T12:30:43.000     2.0

In [64]: ts_a['time'].info.indices
Out[64]: 
[          time          rows
 ----------------------- ----
 2016-03-22T12:30:31.000    0
 2016-03-22T12:30:34.000    1
 2016-03-22T12:30:37.000    2
 2016-03-22T12:30:40.000    3
 2016-03-22T12:30:43.000    4]

In [65]: ts_a[:3]['time'].info.indices
Out[65]: 
[Index slice (0, 3, 1) of
           time          rows
 ----------------------- ----
 2016-03-22T12:30:31.000    0
 2016-03-22T12:30:34.000    1
 2016-03-22T12:30:37.000    2
 2016-03-22T12:30:40.000    3
 2016-03-22T12:30:43.000    4]
```

taldcroft:
>  the advantage here is that it solves the problem at the table level.

I would disagree with this characterization. The actual change is in `DataInfo` which is working at the column `info` level. The changes in table were only in reaction to that in order to make Table work again.

mhvk:
Well, it is in `BaseColumnInfo`, which exists to make `info` work with tables... But we're getting at the nitpicking stage here!! Will definitely try to make a clearer changelog entry under API changes.

But what version should we aim at? I.e., should we consider dropping `parent_table` but not `indices` a bug and thus 4.0.4? I'd argue the full indices to a sliced column gives the user at best misleading information.

taldcroft:
I guess I tend to be pretty conservative about what to call a bug if it is going in a bugfix release. Basically the question is whether it is credible that some user out there is taking advantage of this existing behavior and their code will break after installing a bugfix release. I take that contract very seriously. It seems very unlikely that anyone is using that, but it is tough to prove otherwise.

Usually to me for an API change like this to get put into a bug fix, I would say it is OK if the original behavior was so wrong that nobody could be relying on that behavior. Here that is not unambiguously the case, but I agree it is in the grey zone.

mhvk:
@taldcroft - I created an entry in API changes instead of bug fixes. While writing it, I realized again that with this change, mixin columns now behave more similarly to regular columns - indeed, the test I added checks that explicitly. That perhaps suggests a bit more the previous behaviour was not intended. That might argue for a bug-fix in LTS after all, though on the other hand it is not something that will bother many people, so also fine to just go with just 4.2 (as I left it now).

taldcroft (approved):
Looks good, thanks!

PR #10893, opened 2020-10-18
Fix a few issues with generate_config

saimn:
When using it with other packages (tried this with astroquery), this fixes several issues:

- Astropy modules where imported and then added to the config file.
- One module (`astroquery.mpc`) was added 3 times because it is imported in other modules.
- Astroquery instantiates its classes at import which causes a lot of warnings/logging messages, so I choose to silence all outputs.

### DIFF
--- a/astropy/config/configuration.py
+++ b/astropy/config/configuration.py
@@ -610,7 +610,7 @@ def generate_config(pkgname='astropy', filename=None):
     """
     package = importlib.import_module(pkgname)
     with warnings.catch_warnings():
-        warnings.simplefilter('ignore', AstropyDeprecationWarning)
+        warnings.simplefilter('ignore')
         for mod in pkgutil.walk_packages(path=package.__path__,
                                          prefix=package.__name__ + '.'):
 
--- a/astropy/config/configuration.py
+++ b/astropy/config/configuration.py
@@ -652,7 +652,6 @@ def generate_config(pkgname='astropy', filename=None):
             # Check that modules are not processed twice, which can happen
             # when they are imported in another module.
             if mod in processed:
-                print(mod, 'already processed')
                 continue
             else:
                 processed.add(mod)
--- a/astropy/config/configuration.py
+++ b/astropy/config/configuration.py
@@ -609,10 +609,16 @@ def generate_config(pkgname='astropy', filename=None, verbose=False):
         If None, the default configuration path is taken from `get_config`.
 
     """
-    verbosity = nullcontext if verbose else silence
+    if verbose:
+        verbosity = nullcontext
+        filter_warnings = AstropyDeprecationWarning
+    else:
+        verbosity = silence
+        filter_warnings = Warning
+
     package = importlib.import_module(pkgname)
     with verbosity(), warnings.catch_warnings():
-        warnings.simplefilter('ignore')
+        warnings.simplefilter('ignore', category=filter_warnings)
         for mod in pkgutil.walk_packages(path=package.__path__,
                                          prefix=package.__name__ + '.'):
 
--- a/astropy/config/configuration.py
+++ b/astropy/config/configuration.py
@@ -20,12 +20,13 @@ import contextlib
 from os import path
 from textwrap import TextWrapper
 from warnings import warn
-from contextlib import contextmanager, nullcontext
+from contextlib import contextmanager
 
 from astropy.extern.configobj import configobj, validate
 from astropy.utils import find_current_module, silence
 from astropy.utils.exceptions import AstropyDeprecationWarning, AstropyWarning
 from astropy.utils.introspection import resolve_name
+from astropy.utils.compat.context import nullcontext
 
 from .paths import get_config_dir
 
--- a/astropy/config/configuration.py
+++ b/astropy/config/configuration.py
@@ -639,7 +639,24 @@ def generate_config(pkgname='astropy', filename=None):
 
         # Parse the subclasses, ordered by their module name
         subclasses = ConfigNamespace.__subclasses__()
+        processed = set()
+
         for conf in sorted(subclasses, key=lambda x: x.__module__):
+            mod = conf.__module__
+
+            # Skip modules for other packages, e.g. astropy modules that
+            # would be imported when running the function for astroquery.
+            if mod.split('.')[0] != pkgname:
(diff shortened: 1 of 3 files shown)


bsipocz:
> One module (astroquery.mpc) was added 3 times because it is imported in other modules.

what do you mean exactly? It's imported into the pseudo module `solarsystem`, is that causing troubles? (what I don't understand that the same is done for the jpl modules, do those behave differently?)

> Astroquery instantiates its classes at import which causes a lot of warnings/logging messages, so I choose to silence all outputs.

if there are obsessive logging/warnings, we can certainly look into that and do a cull. Could you provide examples?

saimn:
> what do you mean exactly? It's imported into the pseudo module solarsystem, is that causing troubles? (what I don't understand that the same is done for the jpl modules, do those behave differently?)

Yes, it happens when importing `solarsystem.mpc` for example, which creates a new `mpc.Conf` class. I think it happens because importing `mpc` fails because of `astropy._erfa.core`, and so the module import is not cached. And I just realized that I had updated astroquery but I had an old version installed in my site-packages, so with astroquery master the error is gone and it works correctly, without the multiple conf classes. Anyway I think it's still reasonable to have this change to be safe.

> if there are obsessive logging/warnings, we can certainly look into that and do a cull. Could you provide examples?

Sure, this is only when filtering only deprecation warnings: https://gist.github.com/saimn/a18203a6ca86e6c573af0b53fb9b0d37

bsipocz:
Glad to hear that master is being fixed, I recall a few back and forth of how to deal with those erfa things.

> Sure, this is only when filtering only deprecation warnings: https://gist.github.com/saimn/a18203a6ca86e6c573af0b53fb9b0d37

Ahh, thanks. We should raise errors for deprecations, so if any of those are around that's a real issue (with the exception of deprecations coming from pyregion or html5lib). Could you report them back in an astroquery issue?

Most of those are legit warnings, and the rest is already on the radar to be cleaned up.

astropy/config/configuration.py
```diff
@@ -639,7 +639,24 @@ def generate_config(pkgname='astropy', filename=None):
...
+
+            # Skip modules for other packages, e.g. astropy modules that
+            # would be imported when running the function for astroquery.
+            if mod.split('.')[0] != pkgname:
+                continue
+
+            # Check that modules are not processed twice, which can happen
+            # when they are imported in another module.
+            if mod in processed:
+                print(mod, 'already processed')
```
embray, line 655: Was this intended to be left here?

embray (commented):
While making changes in this function anyways, how does my idea in this comment sound https://github.com/astropy/astropy/pull/10877#discussion_r507919535 ?

astropy/config/configuration.py
pllim, line 655: Leftover debug statement?
```suggestion
```

saimn: No indeed !

astropy/config/configuration.py
```diff
@@ -602,15 +602,15 @@ def generate_config(pkgname='astropy', filename=None):
...
+    pkgname : str or None
         The package for which to retrieve the configuration object.
     filename : str or file object or None
         If None, the default configuration path is taken from `get_config`.
 
     """
     package = importlib.import_module(pkgname)
-    with warnings.catch_warnings():
-        warnings.simplefilter('ignore', AstropyDeprecationWarning)
+    with silence(), warnings.catch_warnings():
```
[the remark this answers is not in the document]
saimn: I think it's just annoying to get random warnings/messages unrelated to the purpose of this function. E.g. with astropy you also get the panda :wink:
I could maybe add a verbose flag ?

astropy/config/configuration.py
```diff
@@ -628,7 +628,7 @@ def generate_config(pkgname='astropy', filename=None):
                           width=78)
 
     if filename is None:
-        filename = get_config(package).filename
+        filename = get_config(pkgname).filename
```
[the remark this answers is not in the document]
saimn: Yep. I guess this one was not easy to test since it will generate a file in the default config location. I will see if I can add a test.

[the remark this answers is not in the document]
saimn: Done.

astropy/config/configuration.py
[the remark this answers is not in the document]
saimn: Added a verbose option.

astropy/config/configuration.py
saimn: Thanks, I committed that with other changes to avoid running the CI twice.

astropy/config/configuration.py
```diff
@@ -594,23 +594,24 @@ def get_config(packageormod=None, reload=False, rootname=None):
...
     filename : str or file object or None
         If None, the default configuration path is taken from `get_config`.
 
     """
+    verbosity = nullcontext if verbose else silence
     package = importlib.import_module(pkgname)
-    with warnings.catch_warnings():
-        warnings.simplefilter('ignore', AstropyDeprecationWarning)
+    with verbosity(), warnings.catch_warnings():
+        warnings.simplefilter('ignore')
```
pllim, line 614: Can we also only ignore `AstropyDeprecationWarning` (as before) if `verbose=True`?

astropy/config/configuration.py
```diff
@@ -20,10 +20,10 @@
 from os import path
 from textwrap import TextWrapper
 from warnings import warn
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
```
pllim, line 23: You need to import from `utils.compat` because it doesn't exist in Python 3.6.
```
ERROR ../.. - ImportError: cannot import name 'nullcontext'
```

astropy/config/configuration.py
saimn: Sure, done.

pllim (approved):
LGTM. Thanks!

pllim:
@saimn , perhaps you should answer @embray 's question about https://github.com/astropy/astropy/pull/10877#discussion_r507919535 before merge?

saimn:
@pllim - I think I did so ?

pllim:
Okay then, let's continue discussion on the other issue. Thanks, everyone!

PR #10896, opened 2020-10-18
use f-strings

ikamensh:
This PR changes most simple occurrences of older formatting styles to f-strings. It's derived by running a tool, `flynt astropy`.

I was not sure if this is within scope of CHANGES.rst.

Few stats:
```
Files modified:                            160
Character count reduction:                 10272 (0.10%)
F-string expressions created:              514
```

### DIFF
--- a/astropy/__init__.py
+++ b/astropy/__init__.py
@@ -155,8 +155,7 @@ class base_constants_version(ScienceState):
     @classmethod
     def validate(cls, value):
         if value not in cls._versions:
-            raise ValueError('Must be one of {}'
-                             .format(list(cls._versions.keys())))
+            raise ValueError(f'Must be one of {list(cls._versions.keys())}')
         return cls._versions[value]
 
     @classmethod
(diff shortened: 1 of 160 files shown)


taldcroft:
Thanks @ikamensh! This looks generally good, though we need to get at least one brave soul to scroll through all the changes and make sure they look reasonable.

I milestoned this for 4.3 so that it could be merged after the 4.2 feature freeze / branching. That will prevent merge conflict problems with PRs for 4.2.

pllim:
I thought we already went through this in #8958 . I wonder why these weren't done back then.

taldcroft:
I think the original update code used a very conservative strategy that misses these ones.

ikamensh:
Many changes are multiline format statements - pyupgrade doesn't  change them.

pllim (changes requested):
Need to undo the changes in `extern`. Those files are bundled.

ikamensh:
reverted all changes to `astropy/extern/`

pllim (commented):
Missed some in `samp/hub.py` (beyond "code suggestion" range) but given astropy/pyvo#239 , don't worry about it. `utils/timer.py` will be removed soon too, but since the changes are already made, might as well accept them.

I'll apply my own review and wrap this up. Thanks for your contribution!

pllim (approved):

PR #10897, opened 2020-10-19
Correct the formula for CRPIX computation in WCS.slice()

### LINKED ISSUE #10864: WCS.slice incorrectly computes CRPIX of the new  WCS
### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

```python
# Put your Python code snippet here.
```

mcara:
Fixes #10864

### DIFF
--- a/astropy/wcs/wcs.py
+++ b/astropy/wcs/wcs.py
@@ -3093,11 +3093,7 @@ reduce these to 2 dimensions using the naxis kwarg.
                 if iview.step not in (None, 1):
                     crpix = self.wcs.crpix[wcs_index]
                     cdelt = self.wcs.cdelt[wcs_index]
-                    # equivalently (keep this comment so you can compare eqns):
-                    # wcs_new.wcs.crpix[wcs_index] =
-                    # (crpix - iview.start)*iview.step + 0.5 - iview.step/2.
-                    crp = ((crpix - iview.start - 1.)/iview.step
-                           + 0.5 + 1./iview.step/2.)
+                    crp = (crpix - 1 - iview.start) / iview.step + 1
                     wcs_new.wcs.crpix[wcs_index] = crp
                     if wcs_new.sip is not None:
                         sip_crpix[wcs_index] = crp
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -985,6 +985,7 @@ astropy.visualization
 astropy.wcs
 ^^^^^^^^^^^
 
+- Fixed incorrect value computed for ``CRPIX`` in ``WCS.slice()``. [#10897]
 
 Other Changes and Additions
 ---------------------------
--- a/astropy/wcs/tests/test_utils.py
+++ b/astropy/wcs/tests/test_utils.py
@@ -110,12 +110,12 @@ def test_slice():
     mywcs = WCS(naxis=2)
     mywcs.wcs.crval = [1, 1]
     mywcs.wcs.cdelt = [0.1, 0.1]
-    mywcs.wcs.crpix = [1, 1]
+    mywcs.wcs.crpix = [5, 6]
     mywcs._naxis = [1000, 500]
     pscale = 0.1  # from cdelt
 
     slice_wcs = mywcs.slice([slice(1, None), slice(0, None)])
-    assert np.all(slice_wcs.wcs.crpix == np.array([1, 0]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([5, 5]))
     assert slice_wcs._naxis == [1000, 499]
 
     # test that CRPIX maps to CRVAL:
@@ -125,7 +125,7 @@ def test_slice():
     )
 
     slice_wcs = mywcs.slice([slice(1, None, 2), slice(0, None, 4)])
-    assert np.all(slice_wcs.wcs.crpix == np.array([0.625, 0.25]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([2, 3]))
     assert np.all(slice_wcs.wcs.cdelt == np.array([0.4, 0.2]))
     assert slice_wcs._naxis == [250, 250]
 
@@ -189,20 +189,20 @@ def test_slice_getitem():
     mywcs = WCS(naxis=2)
     mywcs.wcs.crval = [1, 1]
     mywcs.wcs.cdelt = [0.1, 0.1]
-    mywcs.wcs.crpix = [1, 1]
+    mywcs.wcs.crpix = [5, 6]
 
     slice_wcs = mywcs[1::2, 0::4]
-    assert np.all(slice_wcs.wcs.crpix == np.array([0.625, 0.25]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([2, 3]))
     assert np.all(slice_wcs.wcs.cdelt == np.array([0.4, 0.2]))
 
     mywcs.wcs.crpix = [2, 2]
     slice_wcs = mywcs[1::2, 0::4]
-    assert np.all(slice_wcs.wcs.crpix == np.array([0.875, 0.75]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([1.25, 1]))
     assert np.all(slice_wcs.wcs.cdelt == np.array([0.4, 0.2]))
 
     # Default: numpy order
     slice_wcs = mywcs[1::2]
-    assert np.all(slice_wcs.wcs.crpix == np.array([2, 0.75]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([2, 1]))
     assert np.all(slice_wcs.wcs.cdelt == np.array([0.1, 0.2]))
 
 
@@ -210,17 +210,17 @@ def test_slice_fitsorder():
     mywcs = WCS(naxis=2)
     mywcs.wcs.crval = [1, 1]
     mywcs.wcs.cdelt = [0.1, 0.1]
-    mywcs.wcs.crpix = [1, 1]
+    mywcs.wcs.crpix = [3, 3]
 
     slice_wcs = mywcs.slice([slice(1, None), slice(0, None)], numpy_order=False)
-    assert np.all(slice_wcs.wcs.crpix == np.array([0, 1]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([2, 3]))
 
     slice_wcs = mywcs.slice([slice(1, None, 2), slice(0, None, 4)], numpy_order=False)
-    assert np.all(slice_wcs.wcs.crpix == np.array([0.25, 0.625]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([1.5, 1.5]))
     assert np.all(slice_wcs.wcs.cdelt == np.array([0.2, 0.4]))
 
     slice_wcs = mywcs.slice([slice(1, None, 2)], numpy_order=False)
-    assert np.all(slice_wcs.wcs.crpix == np.array([0.25, 1]))
+    assert np.all(slice_wcs.wcs.crpix == np.array([1.5, 3]))
     assert np.all(slice_wcs.wcs.cdelt == np.array([0.2, 0.1]))
 
 


mcara:
I wonder why are unit tests in `test_utils.py` instead of `test_wcs.py`...

astropy/wcs/wcs.py
```diff
@@ -3014,8 +3014,9 @@ def slice(self, view, numpy_order=True):
                     # equivalently (keep this comment so you can compare eqns):
                     # wcs_new.wcs.crpix[wcs_index] =
                     # (crpix - iview.start)*iview.step + 0.5 - iview.step/2.
-                    crp = ((crpix - iview.start - 1.)/iview.step
-                           + 0.5 + 1./iview.step/2.)
+                    # crp = ((crpix - iview.start - 1.)/iview.step
+                    #        + 0.5 + 1./iview.step/2.)
```
pllim, line 3018: Thanks! No point keeping dead code around.
```suggestion
```

mcara: Just above it, even an older comment was kept. Hence I thought it might have some value.

mcara: see:
>equivalently (keep this comment so you can compare eqns):

mcara: @pllim Removed the entire large comment above in the last commit.

pllim (commented):
I am just a user in this context. I am not qualified to review and will have to defer to `wcs` maintainers.

astropy/wcs/wcs.py
```diff
@@ -3093,11 +3093,7 @@ def slice(self, view, numpy_order=True):
                 if iview.step not in (None, 1):
                     crpix = self.wcs.crpix[wcs_index]
                     cdelt = self.wcs.cdelt[wcs_index]
-                    # equivalently (keep this comment so you can compare eqns):
-                    # wcs_new.wcs.crpix[wcs_index] =
-                    # (crpix - iview.start)*iview.step + 0.5 - iview.step/2.
-                    crp = ((crpix - iview.start - 1.)/iview.step
-                           + 0.5 + 1./iview.step/2.)
+                    crp = (crpix - 1 - iview.start) / iview.step + 1
```
dhomeier, line 3096: This implements the first version you discussed in #10864, right? This should be noted somewhere.

dhomeier (changes requested):
The implementation wrt the original issue is clear enough, however I think there is a 3rd possible interpretation, which boils down to the difference between interpreting the pixel values as _sampling_ vs. _binning_ the coordinate values. In the latter view, CRPIX 4 from your example would correspond to a "true coordinate" 3.5 at the centre of pixel 4. Now if `step > 1` implies rebinning to pixels `step` x the original size, that reference point would come to lie at 0.5 to the right of the new pixel 2, binned from original pixels 3+4 and centred at 3.0, making the reference pixel value `2 + 0.5 / step = 2.25` (or `2 + 1./iview.step/2.` in the code).
So in that interpretation the existing formula would be correct!
Seems this was introduced along the way with converting `wcs_slice` from a function to a method in 56d1505; I could not identify an associated PR, but maybe @keflavich can confirm if this was indeed his line of thought.
If so, the code might use some restructuring and a comment to better explain the logic. Independent of which implementation we end up with, the docstring should be updated to delete/replace
https://github.com/astropy/astropy/blob/ac313b3f17c19c734d8cd7e329470ef467ef1a71/astropy/wcs/wcs.py#L3042-L3043
which was apparently overlooked when the `NotImplemented` was removed. :)

astropy/wcs/wcs.py
mcara: I was thinking about re-binning vs sampling as well but I _thought/assumed_ that a `slice` with a parameter like `step` would indicate (or make `me` expect) _sampling_ as this is what essentially happens when you apply a slice to a `numpy` array.

With regard to the comment, I think it should talk about `stop` instead of `step`. I believe `stop` is currently ignored.

mcara:
@dhomeier Actually, re-writing the original equation (currently used in the code) as:
```python
(crpix - 0.5 - start) / step + 0.5
```
vs my formula (proposed in this PR):
```python
(crpix - 1 - start) / step + 1
```
it is clear that the old formula essentially considers pixels corners to be placed at pixel coordinates instead of pixel centers.

astropy/wcs/wcs.py
dhomeier: `stop` seems to be accounted for here https://github.com/astropy/astropy/blob/ac313b3f17c19c734d8cd7e329470ef467ef1a71/astropy/wcs/wcs.py#L3090
But the note on `step` was directly taken from the docstring of the old function, which did not implement it.

It depends indeed on how you interpret the slicing operation, and how it is used _in praxi_. I merely guess the original code was intended to implement binning (which probably is more common when downscaling the number of pixels in an image).

dhomeier:
> Actually, re-writing the original equation (currently used in the code) as:
> 
> ```python
> (crpix - 0.5 - start) / step + 0.5
> ```

Agreed; that makes it much clearer (I was still struggling with where the 0.5-pixel offsets are applied) and reproduces what I would expect for other factors like `step` = 3 or 4.

mcara:
> > Actually, re-writing the original equation (currently used in the code) as:
> > ```python
> > (crpix - 0.5 - start) / step + 0.5
> > ```
> 
> Agreed; that makes it much clearer (I was still struggling with where the 0.5-pixel offsets are applied) and reproduces what I would expect for other factors like `step` = 3 or 4.

Yeah, it kind of answers my question (_"Where do `0.5` and `1./iview.step/2.` come from?"_) from #10864

keflavich:
I definitely have *only* used wcs-downsampling in the context of rebinning, never sampling.  So, @dhomeier, your interpretation is correct.  As you said, in the rebinning case, the center of the first pixel moves.  But in the resampling case - which I had never previously considered - you can reasonably start the sampling anywhere, and it makes sense to treat the original first-pixel center as the start point by default.

We should support both cases, but we need to do it more explicitly - which may mean abandoning the slice syntax in favor of a function.

mcara:
Coming from the `astrodrizzle` side, I would not have thought that `slice()` would be related to binning.

Also, looking at `NDData` and considering one such object (let's say `nd`), what one would expect the `wcs` of a "sliced" object, i.e., `nd[::2, ::2]` be? Computed using resample or rebin approach? (NOTE: After a quick look at the `NDData` - see https://docs.astropy.org/en/stable/nddata/mixins/ndslicing.html#slicing-nddata-with-world-coordinates it is obvious that `step` is not supported when `NDData` has WCS set, at least for now). However, slicing a `NDData` object (without WCS) uses resampling as expected.

So, if everyone (CC: @pllim ) is using `WCS.slice()` assuming it creates a WCS as if data are binned, then no change is necessary.

Also, since the result of slicing depends on the operation on image data (binning, resampling), maybe this WCS re-computation should have been tied to the function that modifies image data and is aware of how data are "sliced".

dhomeier:
It seemed quite logical to me, as the whole `CRPIX`/`CDELT` formalism comes from the FITS image standard, where the interpretation of a pixel as binning the signal over its area is rather natural. But the concept of slicing maybe does not fit that well with it. Offering both interpretation still sounds like a legitimate solution; `timeseries` supports both binned and sampled model as well.
As the method already has a `numpy_order` keyword, handling one for the slicing mode should pose no problems.

Would the interpretation of such "binned slices" at the WCS level create any problems?

OTOH, are there convenient alternatives to `slice` for rebinning image data that could be encouraged instead?

keflavich:
The workflow I've used WCS slicing with is something like this:
 * smooth the data with a 2-pixel kernel (Gaussian or boxcar)
 * downsample the smoothed data with `[::2]` slicing
 * downsample the WCS with `[::2]` slicing.

In this example workflow, both forms of downsampling would be *valid*, but they have different net effects and interpretations.  In the "keep the corner the same" approach (the binning version), the area covered by the downsampled WCS remains the same (at least, in the lower-left corner).

@mcara, you're right, it makes sense to couple this more closely in documentation with the workflow.  I don't believe there are any functions within astropy that use this and therefore should have been more closely coupled (but I may be forgetting).

I'm ambivalent about the API for changing WCS shape/binning as long as the workflow above is supported.

keflavich:
Also, even if no changes are made, we should include your simplification and some comments about this discussion telling future users (future us) why the equations look as they do.

mcara:
Honestly, I am not an user of this function and I was not even aware of it until https://github.com/spacetelescope/stginga/issues/205 and so my opinion should have very low weight. It is really up to those who use it most to say what they expect.

I see that `nddata` has `block_reduce` function - see https://github.com/astropy/astropy/blob/a7fef9422eedaaa16a52d3ed71752813672d7ec5/astropy/nddata/blocks.py#L97 which is also used by CCDProc but I do not see in any place how WCS is handled there.

mcara:
> The workflow I've used WCS slicing with is something like this:
> 
> * smooth the data with a 2-pixel kernel (Gaussian or boxcar)
> * downsample the smoothed data with `[::2]` slicing
> * downsample the WCS with `[::2]` slicing.

I think this workflow does not result in image shift and the formula proposed in this PR should be used to describe a sliced WCS.

Here is a test case: let's simulate a feature - a parabola - centered at some arbitrary pixel and then apply various approaches:

```python
import numpy as np
from photutils import centroid
from astropy import nddata
from astropy.convolution import convolve, Box2DKernel, Gaussian2DKernel

# create an un-resampled (original) image:
cx = 18.3
cy = 17.75
y, x = np.indices((40, 40))
p = -(x - cx)**2 - (y - cy)**2
print(centroid..centroid_quadratic(p))  # to verify simulation

# re-bin this image:
r = nddata.block_reduce(p, 2)
print(centroid.centroid_quadratic(r))  # center is shifted compared to resampling case

# resampling approach ("true slice" IMO):
print(centroid.centroid_quadratic(p[::2, ::2]))

# @keflavich workflow (if I understood it correctly):
bk = Box2DKernel(2, mode='integrate')
gk = Gaussian2DKernel(2)
# both kernels give the same result as resample/array slicing approach and not binning appraoch:
centroids.centroid_quadratic(convolve(p, bk)[::2, ::2])
centroids.centroid_quadratic(convolve(p, gk)[::2, ::2])

**Comment by mcara** [comment]
> The workflow I've used WCS slicing with is something like this:
> 
> * smooth the data with a 2-pixel kernel (Gaussian or boxcar)
> * downsample the smoothed data with `[::2]` slicing
> * downsample the WCS with `[::2]` slicing.

I think this workflow does not result in image shift and the formula proposed in this PR should be used to describe a sliced WCS.

Here is a test case: let's simulate a feature - a parabola - centered at some arbitrary pixel and then apply various approaches:

```python
import numpy as np
from photutils import centroids
from astropy import nddata
from astropy.convolution import convolve, Box2DKernel, Gaussian2DKernel

# create an un-resampled (original) image:
cx = 18.3
cy = 17.75
y, x = np.indices((40, 40))
p = -(x - cx)**2 - (y - cy)**2
print(centroids.centroid_quadratic(p))  # to verify simulation

# re-bin this image:
r = nddata.block_reduce(p, 2)
print(centroids.centroid_quadratic(r))  # center is shifted compared to resampling case

# resampling approach ("true slice" IMO):
print(centroids.centroid_quadratic(p[::2, ::2]))

# @keflavich workflow (if I understood it correctly):
bk = Box2DKernel(2, mode='integrate')
gk = Gaussian2DKernel(2)
# both kernels give the same result as resample/array slicing approach and not binning appraoch:
print(centroids.centroid_quadratic(convolve(p, bk)[::2, ::2]))
print(centroids.centroid_quadratic(convolve(p, gk)[::2, ::2]))
```

keflavich:
ok, so my actual use case made it into spectral-cube:
https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/wcs_utils.py#L205

https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/wcs_utils.py#L306-L322

with tests added here:
https://github.com/radio-astro-tools/spectral-cube/pull/525/files
https://github.com/keflavich/spectral-cube/blob/b16c034c6d56bb5045d08bb088f3ed55894abbe6/spectral_cube/tests/test_wcs_utils.py#L198-L199

though I never had an independent review.

pllim:
For my use case, the main goal is to have the images appear more or less where they are supposed to be in FOV "on the sky." They are not used for science, but rather for quick look. As it is (in the released version), it is already working well enough. It is just that I got very confused when I tried to write a test and couldn't get the original center and the downsampled center to align for the WCSes. At the very least, clearer doc with examples would be very nice indeed. Sorry for the trouble!

dhomeier:
> Center in simulated image:         [18.3  17.75]
> Binning:                           [8.9   8.625]
> Simple resample (odd cols & rows): [9.15  8.875]
> keflavich (Box kernel):            [9.15  8.875]
> keflavich (Gaussian kernel):       [9.15  8.875]

OTOH the binned array `r` is flux conserving (integrated simply by `r.sum() * step**2`), `p[::2, ::2]` is not, so those are slightly different use cases. One could in fact argue that by construction `centroids` is fitting to a flux density distribution _sampled_ at the locations represented by the array indices; otherwise it would have to fit something like the pixel-integrated function values. But that aside, rebinning in fact _cannot_ conserve spatial moments, as the flux from each original pixel will always be allocated to the new pixel position, which is different from at least some of the originals. Whereas undersampling will just discard a fraction of the values, so there should exist some more exotic cases where position will not be preserved either (just consider a high peak at a single pixel that is left out in the resampled array).

Maybe this just confirms the saying that you should almost never re-bin (or undersample) in production code (as opposed to visualisation etc.). But the question remains how a `WCS` should behave in this context.
I tend to go with the solution of offering both formulae via keyword.

mcara:
> just consider a high peak at a single pixel that is left out in the resampled array

Of course! But no one can say where is the position of the source based just on a single pixel. If, on the other hand, you have a reasonably wide (well sampled) source - then you can.

Maybe we should support both but ideally this function should not belong to this class because WCS is not aware (and should not be aware) of what image processing steps are applied to an image. For example, IMO, `nddata.block_reduce` should be taking care of the "sliced" WCS.

keflavich:
I think it belongs in `wcs.utils`, but not as part of the base `WCS` class.  We shouldn't need to re-implement WCS slicing in different tools, but you're right that the `WCS` object isn't "aware" of what it's being used to represent.

dhomeier:
> Of course! But no one can say where is the position of the source based just on a single pixel. If, on the other hand, you have a reasonably wide (well sampled) source - then you can.
> 
No, but the single pixel will contribute to the signal and its centre-of-weight, and removing it will of course change both. But I agree the example is an extreme, possibly pathological case – a single pixel would probably rather be marked as cosmic, and even otherwise it would obviously not be well represented by the distributions assumed in `centroids`.

> Maybe we should support both but ideally this function should not belong to this class because WCS is not aware (and should not be aware) of what image processing steps are applied to an image. For example, IMO, `nddata.block_reduce` should be taking care of the "sliced" WCS.

Yes; I did not know the base `NDData` class was already aware of a WCS, but since it has that attribute, that looks like a better place for this functionality. Having it directly in `wcs` might be convenient, but treacherous.

mcara:
> I think it belongs in `wcs.utils`, but not as part of the base `WCS` class. We shouldn't need to re-implement WCS slicing in different tools, but you're right that the `WCS` object isn't "aware" of what it's being used to represent.

Yes to this too. In this case it would be easier for different tools (i.e., `nddata.block_reduce`) to just call `wcs.utils.slice()` to compute "sliced" WCS according to the image processing algorithm.

PR #10899, opened 2020-10-19
TST: Skip doctest from PR 10848 due to Numpy 1.16 behavior

pllim:
### Description

See https://github.com/astropy/astropy/pull/10848#issuecomment-712476006

@eteq , please adjust milestone as needed.

### DIFF
--- a/docs/known_issues.rst
+++ b/docs/known_issues.rst
@@ -127,7 +127,7 @@ on NumPy before version 1.20 and ignore the unit in later versions:
 .. doctest-requires:: numpy<1.20
 
     >>> my_quantity = u.Quantity(1, u.m)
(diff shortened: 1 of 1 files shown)


docs/known_issues.rst
```diff
@@ -127,7 +127,7 @@ on NumPy before version 1.20 and ignore the unit in later versions:
 .. doctest-requires:: numpy<1.20
```
mhvk, line 127: Maybe change the sentence 2 lines about to
```
on NumPy versions 1.17--1.19 and ignore the unit in other versions:
```

mhvk (approved):
Fine with just skipping the doctest on older astropy - suggestion for a chance in the text inline (which would only be relevant for 4.1/4.0). Approving since I'm also happy to just get it in...

eteq (approved):
LGTM (with @mhvk's change, which I added directly)

PR #3906, opened 2015-06-30
WIP: Use wcslib for astropy.modeling's projections

mdboom:
This adds a templated C extension to use wcslib to perform the evaluation functions of the various WCS-related projections in `astropy.modeling`.  This should be faster than the Python implementations, and it seemed a shame not to use the C implementations since they are sitting right there.  Also, and more importantly for my end goal, this should make implementing the rest of the standard FITS WCS projection models absolutely trivial.

TODO:

- [ ] Support differently-sized yet broadcastable input arrays
- [ ] Refactor some copy-and-pasted code from `astropy._erfa` that probably should live in `astropy_helpers.setup_helpers`.

PR #10903, opened 2020-10-21
changed default in docstring for cylindrical equal area

cshanahan1:
Just a single character change in a docstring to use the correct default value.


pllim (approved):
Wow, looks like the default was added way back in #2276 and the docstring added in #3906 but was already wrong. Good catch!

PR #10904, opened 2020-10-21
Add note to isclose/allclose about atol

### LINKED ISSUE #10892: `quantity_allclose` difference to numpy `allclose`
### Description

For certain values, `astropy.units.allclose` does not match `numpy.allclose`

### Expected behavior

The I/O to be the same, excepting unit considerations.

### Actual behavior

When comparing two numbers where the first array has elements very close to 0 and the second array is all zeros, numpy allclose works as expected, but `units.allclose` returns False, regardless of "rtol"

### Steps to Reproduce

```python
>>> import numpy as np
>>> np.allclose([1e-17, 0, 0], [0, 0, 0], rtol=1e-5)
True
```

```python
>>> import astropy.units as u
>>> u.allclose([1e-17, 0, 0] * u.m, [0, 0, 0] * u.m, rtol=1e-5)
False
```

### System Details

macOS-10.15.7-x86_64-i386-64bit
Python 3.8.2 | packaged by conda-forge | (default, Apr 24 2020, 07:56:27) 
[Clang 9.0.1 ]
Numpy 1.18.4
astropy 4.2.dev877+g4eef52a1c
Scipy 1.4.1
Matplotlib 3.2.1

adrn:
This fixes #10892.

This adds a note to the docstring of both `astropy.units.isclose()` and `astropy.units.allclose()` that describes differences with numpy's equivalent functions that may arise because numpy's functions have default values for `atol`.

cc @nstarman

### DIFF
--- a/astropy/units/quantity.py
+++ b/astropy/units/quantity.py
@@ -1747,6 +1747,17 @@ def isclose(a, b, rtol=1.e-5, atol=None, equal_nan=False, **kwargs):
     Return a boolean array where two arrays are element-wise equal
     within a tolerance.
 
+    .. note::
+
+        `astropy.units.isclose()` behaves slightly different than
+        `numpy.isclose` because of the default absolute tolerance in
+        `numpy.isclose`. In ``numpy``, the default absolute tolerance,
+        ``atol``, is set to ``atol=1e-8`` because there can be a proper
+        absolute comparison between the input without having to scale them by
+        unit conversions. In this function, there is no natural way to set
+        ``atol`` by default because it will depend on the units of the input.
+
+
     Parameters
     ----------
     a, b : array_like or :class:`~astropy.units.Quantity`
--- a/astropy/units/quantity.py
+++ b/astropy/units/quantity.py
@@ -1747,17 +1747,6 @@ def isclose(a, b, rtol=1.e-5, atol=None, equal_nan=False, **kwargs):
     Return a boolean array where two arrays are element-wise equal
     within a tolerance.
 
-    .. note::
-
-        `astropy.units.isclose()` behaves slightly different than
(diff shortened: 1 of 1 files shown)


adrn:
Your call -- it's a minor enhancement, so I'm happy either way!

pllim:
Let's try 4.1.x. Given #9849, backporting to 4.0.x might be painful.

adrn:
OK sounds good, thanks!

astropy/units/quantity.py
```diff
@@ -1747,6 +1747,17 @@ def isclose(a, b, rtol=1.e-5, atol=None, equal_nan=False, **kwargs):
     Return a boolean array where two arrays are element-wise equal
     within a tolerance.
 
+    .. note::
+
+        `astropy.units.isclose()` behaves slightly different than
```
mhvk, line 1752: :+1: to the idea, but can we add it to `Notes` below (in both cases)? And maybe make it shorter, e.g., append "It differs from the numpy function in that the default for the absolute tolerance is not ``atol=1e-8``, but zero, as there is no natural way to set a default given two inputs that may have differently scaled units."

pllim:
CI failure unrelated. In fact, this PR does not need any CI.

astropy/units/quantity.py
adrn: > can we add it to Notes below (in both cases)?

I wanted to make it more prominent, but fine by me as long as @nstarman is happy with moving this description down.

> maybe make it shorter

Sure

mhvk (approved):
Looks all good now!

nstarman:
I agree. Looks great.

adrn:
OK great, thanks for looking!

PR #10909, opened 2020-10-21
Added missing loopback coordinate transformations (plus internal functionality)

### LINKED ISSUE #8835: Ecliptic frames do not have loopback transformations
Is there an intentional reason that the ecliptic frames (and some others) do not have loopback transformations?  When I want to change equinox or observation time and I neglect to explicitly transform through ICRS or some other frame, I get momentarily confused by the no-op.

ayshih:
Edit: Another round of under-the-hood changes.

This PR adds all of the missing loopback transformations (closes #8835).  This PR adds internal functionality to make it easy to create loopback transformations, including auto-detecting the best type of transform to use.  Accordingly, nearly all existing loopback code can be removed (aside from FK4/FK5 loopbacks).  For example, the HCRS loopback is now just one line:
```python
frame_transform_graph._add_merged_transform(HCRS, ICRS, HCRS)  # will register an AffineTransform
```

This PR:
- Adds all of the missing loopback transformations (closes #8835), although not all of them can be immediately written as affine transformations.
- Adds a private method to `TransformGraph` to create and register a single-step transform from a multi-step transformation path, using the transforms that already exist in the graph
  - If the starting frame and the final frame classes are the same, this will be a loopback transformation
  - If all of the transforms are affine, the merged transform will be affine (either `AffineTransform` or `DynamicMatrixTransform`)
  - Otherwise, the merged transform will be `FunctionTransformWithFiniteDifference`
- Adds a private method to `CompositeTransform` to return a single transform that encapsulates all of the constituent transforms.
  - If all of the transforms are affine, the single transform will be affine (either `AffineTransform` or `DynamicMatrixTransform`)
  - Otherwise, the single transform will be `FunctionTransformWithFiniteDifference`
- Adds a private function to combine two sets of affine parameters into one set.
- Modifies the subclassing of `BaseAffineTransform` so that the different types of affine transformations have the same (private) method to obtain the affine parameters.

---
Context:

Currently, a coordinate transformation that is a composite of separately-defined affine transformations is very difficult to implement except as a `FunctionTransform` or `FunctionTransformWithFiniteDifference`.  For example, here's the HCRS loopback transformation:
```python
@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, HCRS, HCRS)
def hcrs_to_hcrs(from_coo, to_frame):
     ...
        return from_coo.transform_to(ICRS()).transform_to(to_frame)
```
Even though the HCRS->ICRS and ICRS->HCRS transformations are both `AffineTransform`, that fact is not exposed outside of the `.transform_to()` machinery.  This approach is problematic because `FunctionTransformWithFiniteDifference` entrains unnecessary computations and inaccuracies when working with coordinates with velocities due to the finite differencing.

### DIFF
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -1188,8 +1188,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray):
         -------
         transframe : coordinate-like
             A new object with the coordinate data represented in the
-            ``newframe`` system (meaning a `~astropy.coordinates.SkyCoord` will
-            remain a SkyCoord).
+            ``newframe`` system.
 
         Raises
         ------
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -24,7 +24,6 @@ from .icrs import ICRS
 from .gcrs import GCRS
 from .cirs import CIRS
 from .hcrs import HCRS
-from .equatorial import TETE
 from .utils import aticq, atciqz, get_offset_sun_from_barycenter
 
 from ..erfa_astrom import erfa_astrom
@@ -94,21 +93,6 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
     return icrs_frame.realize_frame(newrep)
 
 
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, CIRS)
-def cirs_to_cirs(from_coo, to_frame):
-    if (np.all(from_coo.location == to_frame.location) and
-            np.all(from_coo.obstime == to_frame.obstime)):
-        return to_frame.realize_frame(from_coo.data)
-    else:
-        # the CIRS<-> CIRS transform actually goes through ICRS.  This has a
-        # subtle implication that a point in CIRS is uniquely determined
-        # by the corresponding astrometric ICRS coordinate *at its
-        # current time*.  This has some subtle implications in terms of GR, but
-        # is sort of glossed over in the current scheme because we are dropping
-        # distances anyway.
-        return from_coo.transform_to(ICRS()).transform_to(to_frame)
-
-
 # Now the GCRS-related transforms to/from ICRS
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, ICRS, GCRS)
@@ -175,16 +159,6 @@ def gcrs_to_icrs(gcrs_coo, icrs_frame):
         newrep = intermedrep + astrom_eb
 
     return icrs_frame.realize_frame(newrep)
-
-
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GCRS, GCRS)
-def gcrs_to_gcrs(from_coo, to_frame):
-    if (np.all(from_coo.obstime == to_frame.obstime) and
-            np.all(from_coo.obsgeoloc == to_frame.obsgeoloc)):
-        return to_frame.realize_frame(from_coo.data)
-    else:
-        # like CIRS, we do this self-transform via ICRS
-        return from_coo.transform_to(ICRS()).transform_to(to_frame)
 
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GCRS, HCRS)
@@ -260,22 +234,13 @@ def icrs_to_hcrs(icrs_coo, hcrs_frame):
                                                 include_velocity=bool(icrs_coo.data.differentials))
 
 
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, HCRS, HCRS)
-def hcrs_to_hcrs(from_coo, to_frame):
-    if np.all(from_coo.obstime == to_frame.obstime):
-        return to_frame.realize_frame(from_coo.data)
-    else:
-        # like CIRS, we do this self-transform via ICRS
-        return from_coo.transform_to(ICRS()).transform_to(to_frame)
-
-
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, TETE, TETE)
-def tete_to_tete(from_coo, to_frame):
-    if (np.all(from_coo.location == to_frame.location)
-            and np.all(from_coo.obstime == to_frame.obstime)):
-        return to_frame.realize_frame(from_coo.data)
-    else:
-        # We perform the self-transform through ICRS, since any change in obstime
-        # will require handling the change in pos/vel of the Earth and it's impact
-        # on aberration.
-        return from_coo.transform_to(ICRS()).transform_to(to_frame)
+# Create loopback transformations
+frame_transform_graph._add_merged_transform(CIRS, ICRS, CIRS)
+# The CIRS<-> CIRS transform going through ICRS has a
+# subtle implication that a point in CIRS is uniquely determined
+# by the corresponding astrometric ICRS coordinate *at its
+# current time*.  This has some subtle implications in terms of GR, but
+# is sort of glossed over in the current scheme because we are dropping
+# distances anyway.
+frame_transform_graph._add_merged_transform(GCRS, ICRS, GCRS)
+frame_transform_graph._add_merged_transform(HCRS, ICRS, HCRS)
--- a/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
+++ b/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
@@ -275,5 +275,6 @@ def itrs_to_teme(itrs_coo, teme_frame):
 
 # Create loopback transformations
 frame_transform_graph._add_merged_transform(ITRS, CIRS, ITRS)
+frame_transform_graph._add_merged_transform(PrecessedGeocentric, GCRS, PrecessedGeocentric)
 frame_transform_graph._add_merged_transform(TEME, ITRS, TEME)
 frame_transform_graph._add_merged_transform(TETE, ICRS, TETE)
--- a/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
+++ b/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
@@ -13,7 +13,7 @@ from astropy.coordinates.baseframe import frame_transform_graph
 from astropy.coordinates.transformations import FunctionTransformWithFiniteDifference
 from astropy.coordinates.matrix_utilities import matrix_transpose
 
-
+from .icrs import ICRS
 from .gcrs import GCRS, PrecessedGeocentric
 from .cirs import CIRS
 from .itrs import ITRS
@@ -218,11 +218,10 @@ def itrs_to_cirs(itrs_coo, cirs_frame):
     return cirs.transform_to(cirs_frame)
 
 
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, ITRS, ITRS)
-def itrs_to_itrs(from_coo, to_frame):
-    # this self-transform goes through CIRS right now, which implicitly also
-    # goes back to ICRS
-    return from_coo.transform_to(CIRS()).transform_to(to_frame)
+# TODO: implement GCRS<->CIRS if there's call for it.  The thing that's awkward
+# is that they both have obstimes, so an extra set of transformations are necessary.
+# so unless there's a specific need for that, better to just have it go through the above
+# two steps anyway
 
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GCRS, PrecessedGeocentric)
@@ -274,11 +273,7 @@ def itrs_to_teme(itrs_coo, teme_frame):
     return teme_frame.realize_frame(newrepr)
 
 
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, TEME, TEME)
-def teme_to_teme(from_coo, to_frame):
-    if np.all(from_coo.obstime == to_frame.obstime):
-        return to_frame.realize_frame(from_coo.data)
-    else:
-        # this self-transform goes through ITRS right now, which implicitly also
-        # goes back to ICRS
-        return from_coo.transform_to(ITRS(obstime=from_coo.obstime)).transform_to(to_frame)
+# Create loopback transformations
+frame_transform_graph._add_merged_transform(ITRS, CIRS, ITRS)
+frame_transform_graph._add_merged_transform(TEME, ITRS, TEME)
+frame_transform_graph._add_merged_transform(TETE, ICRS, TETE)
--- a/astropy/coordinates/tests/test_celestial_transformations.py
+++ b/astropy/coordinates/tests/test_celestial_transformations.py
@@ -9,10 +9,10 @@ from astropy.coordinates import galactocentric_frame_defaults
 from astropy.coordinates.distances import Distance
 from astropy.coordinates.builtin_frames import (
     ICRS, FK5, FK4, FK4NoETerms, Galactic, CIRS,
-    Supergalactic, Galactocentric, HCRS, GCRS, LSR)
+    Supergalactic, Galactocentric, HCRS, GCRS, LSR, GalacticLSR)
 from astropy.coordinates import SkyCoord
 from astropy.tests.helper import assert_quantity_allclose as assert_allclose
-from astropy.coordinates import EarthLocation, CartesianRepresentation
+from astropy.coordinates import EarthLocation, CartesianRepresentation, CartesianDifferential
 from astropy.time import Time
 from astropy.units import allclose
 
@@ -340,3 +340,46 @@ def test_cirs_icrs():
     # now check ICRS transform gives a decent distance from Barycentre
     moon_icrs = moon_geo.transform_to(ICRS())
     assert_allclose(moon_icrs.distance - 1*u.au, 0.0*u.R_sun, atol=3*u.R_sun)
+
+
+@pytest.mark.parametrize('frame', [LSR, GalacticLSR])
+def test_lsr_loopback(frame):
+    xyz = CartesianRepresentation(1, 2, 3)*u.AU
+    xyz = xyz.with_differentials(CartesianDifferential(4, 5, 6)*u.km/u.s)
+
+    v_bary = CartesianDifferential(5, 10, 15)*u.km/u.s
+
+    # Test that the loopback properly handles a change in v_bary
+    from_coo = frame(xyz)  # default v_bary
+    to_frame = frame(v_bary=v_bary)
+
+    explicit_coo = from_coo.transform_to(ICRS()).transform_to(to_frame)
+    implicit_coo = from_coo.transform_to(to_frame)
+
+    # Confirm that the explicit transformation changes the velocity but not the position
+    assert allclose(explicit_coo.cartesian.xyz, from_coo.cartesian.xyz, rtol=1e-10)
+    assert not allclose(explicit_coo.velocity.d_xyz, from_coo.velocity.d_xyz, rtol=1e-10)
+
+    # Confirm that the loopback matches the explicit transformation
+    assert allclose(explicit_coo.cartesian.xyz, implicit_coo.cartesian.xyz, rtol=1e-10)
+    assert allclose(explicit_coo.velocity.d_xyz, implicit_coo.velocity.d_xyz, rtol=1e-10)
+
+
+@pytest.mark.parametrize('to_frame',
+                         [Galactocentric(galcen_coord=ICRS(300*u.deg, -30*u.deg)),
+                          Galactocentric(galcen_distance=10*u.kpc),
+                          Galactocentric(z_sun=10*u.pc),
+                          Galactocentric(roll=1*u.deg)])
+def test_galactocentric_loopback(to_frame):
+    xyz = CartesianRepresentation(1, 2, 3)*u.pc
+
+    from_coo = Galactocentric(xyz)
+
+    explicit_coo = from_coo.transform_to(ICRS()).transform_to(to_frame)
+    implicit_coo = from_coo.transform_to(to_frame)
+
+    # Confirm that the explicit transformation changes the position
+    assert not allclose(explicit_coo.cartesian.xyz, from_coo.cartesian.xyz, rtol=1e-10)
+
+    # Confirm that the loopback matches the explicit transformation
+    assert allclose(explicit_coo.cartesian.xyz, implicit_coo.cartesian.xyz, rtol=1e-10)
--- a/astropy/coordinates/transformations.py
+++ b/astropy/coordinates/transformations.py
@@ -1408,7 +1408,7 @@ class CompositeTransform(CoordinateTransform):
         curr_coord = fromcoord
         for t in self.transforms:
             # build an intermediate frame with attributes taken from either
-            # `fromframe`, or if not there, `toframe`, or if not there, use
+            # `toframe`, or if not there, `fromcoord`, or if not there, use
             # the defaults
             # TODO: caching this information when creating the transform may
             # speed things up a lot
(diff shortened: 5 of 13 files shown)


ayshih:
I have now added all of the missing loopback transformations in the graph, so this now closes #8835.  There are a few frames remaining that don't need to have loopback transformations defined because there are no changeable frame attributes for those frames.

ayshih:
@StuartLittlefair The ITRS<->TEME transformations from #10149 are bugged when the `obstime` is different between the two frames.  (They both implicitly assume that a TEME loopback existed.)  I have fixed the transformation logic in 43aa4d6 and added a loopback.

ayshih:
Here's a comparison of the transformation graphs before/after this PR.  Note the added loopbacks and the change in transformation type for the HCRS loopback.

## Before
before

## After
after

StuartLittlefair:
> @StuartLittlefair The ITRS<->TEME transformations from #10149 are bugged when the `obstime` is different between the two frames. (They both implicitly assume that a TEME loopback exists.) I have fixed the transformation logic in [43aa4d6](https://github.com/astropy/astropy/commit/43aa4d64c5880f9d7509bafc45712433990fdc84) and added a loopback.

Thanks for adding the TEME loopback - the code was indeed broken without it!

However, given my comment [here](https://github.com/astropy/astropy/pull/10867#discussion_r509989735) I think it is better (more explicit and faster) if the TEME loopback goes through ICRS rather than ITRS.

I'll try and give this a detailed review later since I think it's a lovely idea. My first thought upon skimming it is that the use of ```_get_equivalent_affine_params_to``` requires the developer to know that every step in the eventual transform chain is affine. Surely this can be somehow baked into the frame_transform_graph logic? So that instead of always returning a CompositeTransform, the transform returned is Affine if appropriate? This would mean everyone gets the benefit without the cognitive work of determining which transform is ultimately applied, which as seen in my comment, is not always easy..

ayshih:
> However, given my comment [here](https://github.com/astropy/astropy/pull/10867#discussion_r509989735) I think it is better (more explicit and faster) if the TEME loopback goes through ICRS rather than ITRS.

See my response to your comment for TETE.  For TEME, unlike the case with TETE, there's only one transformation path to ICRS and it goes through ITRS.  Thus, one could swap `ICRS()` for `ITRS()` in my implementation, and the actual transformations used would be the same (i.e., it's not "faster").

My main argument for using the ITRS loopback rather than explicitly looping through ICRS is to naturally support future enhancements.  I can envision someone in the future someone writing an efficient ITRS loopback that doesn't need to transform through ICRS (i.e., a "real" self-transform).  If that were to manifest, then this way of writing the TEME transformations would *automatically* take advantage of it rather than unintentionally circumventing it.

> My first thought upon skimming it is that the use of `_get_equivalent_affine_params_to` requires the developer to know that every step in the eventual transform chain is affine.

Yup, it might be confusing for the developer to determine at the outset, but presumably the `ConvertError` will provide clarity.  I see this as a tool to help developers write transformations in the graph (i.e., the frames on both sides are known), and not something that would be used for frames with unknown types.

> Surely this can be somehow baked into the frame_transform_graph logic? So that instead of always returning a CompositeTransform, the transform returned is Affine if appropriate?

I actually initially started in this direction, but it seemed to add extra complication with minimal benefit, and I think you'd still build it off of the functionality in this PR.  Importantly, you'd need to have something separate from the `.transform_to()` approach because, for example, there'd be no way to interpret the following as a single `AffineTransform`:
```
from_coo.transform_to(ICRS()).transform_to(to_frame)
```
But, I now have some ideas bouncing around my head, so let me mull them over.

StuartLittlefair:
> > However, given my comment [here](https://github.com/astropy/astropy/pull/10867#discussion_r509989735) I think it is better (more explicit and faster) if the TEME loopback goes through ICRS rather than ITRS.
> 
> See my response to your comment for TETE. For TEME, unlike the case with TETE, there's only one transformation path to ICRS and it goes through ITRS. Thus, one could swap `ICRS()` for `ITRS()` in my implementation, and the actual transformations used would be the same (i.e., it's not "faster").
> 

OK, you're right and that is basically fine. I still prefer ICRS since I think it's clearer if the transform obviously goes through ICRS, rather than "hiding" this fact. Note that TEME is a bit of a special case since it's not a "proper" frame - i.e one recognised by the IAU. It doesn't even really have a formal definition, but the informal definition is the Vallardo paper, which defines it w.r.t ITRS which is why I've done it this way. It does make the transform graph weird though, since TEME is an equatorial frame which should really sit alongside CIRS and apparent (TETE) frames. C'est la vie I guess.

ayshih:
I've edited the title/description of this PR to emphasize the user-facing benefit, namely fixing the missing loopback transformations.

@StuartLittlefair: If this PR won't make it into 4.2, I recommend that 43aa4d6 be cherry-picked as a quick PR to fix the bugged transformations.

ayshih:
I've added a new method for `TransformGraph` that makes it simple to add loopback transformations, with auto-detection of the best type of transform to use.  I'll edit the original post.

ayshih:
Here's how this PR now creates all of the ecliptic loopback transformations (which will be a mix of `AffineTransform`, `DynamicMatrixTransform`, and `FunctionTransformWithFiniteDifference`):

https://github.com/astropy/astropy/blob/725a54e7d53c1b2f428c6ccb0452783dfd227637/astropy/coordinates/builtin_frames/ecliptic_transforms.py#L256-L264

And here are some examples where ICRS isn't (explicitly) the intermediate frame:

https://github.com/astropy/astropy/blob/725a54e7d53c1b2f428c6ccb0452783dfd227637/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py#L249-L253

ayshih:
When adding tests, I uncovered a bug in the existing PrecessedGeocentric->GCRS transformation (see a6532e8).

ayshih:
I've added tests for all of the new loopbacks, so I'm calling this PR ready for review.

astropy/coordinates/transformations.py
```diff
@@ -674,6 +674,100 @@ def deco(func):
...
+        first_path = self.get_transform(fromsys, viasys)
+        second_path = self.get_transform(viasys, tosys)
+        if first_path is None or second_path is None:
+            raise ValueError(f"{transform_string} is not possible")
+
+        trans_class_set = set([type(t) for t in first_path.transforms] +
+                              [type(t) for t in second_path.transforms])
+
+        # Shared frame attributes are used for instantiating the intermediate frame
+        # For any attributes that are shared by the "from"/"to" frames, the "to" frame takes precedence
```
StuartLittlefair, line 728: Is it possible to add a note explaining why? I'm sure there's a good reason, but I don't see it immediately, and if there's a good reason it has to be this way, we should make sure the comment is clear so no-one changes it later..

StuartLittlefair (commented):
This is looking really nice @ayshih!

I'm not sure I like the name `_create_direct_transform` - it doesn't explain that the transform is registered, and it's a little confusing because it is not a direct transform, but one via another system. Not sure I have a better suggestion right now.

StuartLittlefair:
I also have another suggestion, which is more radical, but is there a reason we *shouldn't* remove all the lines calling `_create_direct_transform` and instead make it the default (via ICRS) if a transform cannot be found? 

I know this has been discussed in an issue elsewhere, but couldn't find it, and can't think of an obvious reason why this would be a bad idea.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -221,32 +216,38 @@ def precessedgeo_to_gcrs(from_coo, to_frame):
...
-    newrepr = itrs_coo.cartesian.transform(matrix_transpose(pmat))
-    teme = TEME(newrepr, obstime=itrs_coo.obstime)
+    pmat = teme_to_itrs_mat(teme_frame.obstime)
+    newrepr = itrs_coo2.cartesian.transform(matrix_transpose(pmat))
+    return teme_frame.realize_frame(newrepr)
 
-    # now do any needed offsets (no-op if same obstime)
-    return teme.transform_to(teme_frame)
+
+# Create loopback transformations
```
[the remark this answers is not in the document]
ayshih: The short answer is that I don't know.  In all instances, I simply followed the same transformation path as the existing loopback transformations, and didn't question it.

The TETE loopback through ICRS is to pick a transformation path that avoids ITRS (see https://github.com/astropy/astropy/pull/10867#discussion_r510005879).

astropy/coordinates/transformations.py
```diff
@@ -674,6 +674,100 @@ def deco(func):
             return func
         return deco
 
+    def _create_direct_transform(self, fromsys, viasys, tosys, priority=1):
```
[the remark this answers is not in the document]
ayshih: I agree that the name is not clear.  It really should be `_create_and_register_direct_transform()`, but I thought that would be too long...

astropy/coordinates/transformations.py
ayshih: In other words, the reasoning needs to be documented in both places.

ayshih:
The issue is #8996.  I'll point out that an approach like that is not inherently general, if the transformation framework is intended to support graphs where there may not exist a transformation between a frame and ICRS.  A frame transform graph would likely want to declare what its default "common" frame is.

StuartLittlefair:
> The issue is #8996. I'll point out that an approach like that is not inherently general, if the transformation framework is intended to support graphs where there may not exist a transformation between a frame and ICRS. A frame transform graph would likely want to declare what its default "common" frame is.

A very good point!

astropy/coordinates/transformations.py
StuartLittlefair: I think so... Isn't the reasoning just that it *has* to be this way for the last step in the chain, and we use the same code for every step?

astropy/coordinates/transformations.py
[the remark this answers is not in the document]
StuartLittlefair: I'm not sure about the `direct` part either. I'd vote for the more explicit function name. Since this is a method on the frame transfer graph I think `_register...` works, but how about `_register_twostep_transform`?

adrn:
Hey @ayshih - we just discussed this on our astropy.coordinates call. Sorry for dropping the ball on communicating about this for so long, but we do think this is a great idea and would love to get this in before v5.0! Do you have time to rebase this so we can take another look?

ayshih:
Sure, I'll rebase this and make sure it still works as intended

astropy/coordinates/transformations.py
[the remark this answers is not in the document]
ayshih: After mulling it over with fresh eyes, I decided to change the name to ``_add_merged_transform``.

astropy/coordinates/transformations.py
```diff
@@ -674,6 +674,108 @@ def deco(func):
...
+            via_attrs.update(zip(second_overlap, [getattr(to_frame, n) for n in second_overlap]))
+            return via_attrs
+
+        # If all of the transforms are affine, we can create an affine transform
+        if all([issubclass(tc, BaseAffineTransform) for tc in trans_class_set]):
+
+            # If all of the transforms do not return offsets, we can use DynamicMatrixTransform
+            if all([issubclass(tc, (StaticMatrixTransform, DynamicMatrixTransform))
+                    for tc in trans_class_set]):
+                def direct_DMT(from_coo, to_frame):
```
nstarman, line 751: This and the below functions don't need to be defined inside of this method. They might be best as private funcs defined outside. At least for readability, when trying to understand this method, the internals of ``direct_DMT``, etc are not crucial.

astropy/coordinates/transformations.py
```diff
@@ -674,6 +674,108 @@ def deco(func):
             return func
         return deco
 
+    def _add_merged_transform(self, fromsys, viasys, tosys, priority=1):
```
nstarman, line 677: I really like this function, but having just one intermediate frame feels an unnecessary limitation. Perhaps for followup, but 

```suggestion
    def _add_merged_transform(self, fromsys, viasys, *tosys, priority=1):
```

And you use the same recursion trick as ``_get_equivalent_affine_params_to``

nstarman (commented):
This is looking really great @ayshih! Just a few comments about code clarity. Also, some tests on ``_combine_affine_params`` would be excellent. Thanks again for tackling this and for you patience. v5.0 for sure.

adrn (commented):
Very nice work here @ayshih -- thanks for the PR! I think this looks mostly good to go, but would be great to address the few remaining small issues raised by @StuartLittlefair and @nstarman.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -244,32 +243,38 @@ def precessedgeo_to_gcrs(from_coo, to_frame):
     pmat = gcrs_precession_mat(from_coo.equinox)
     crepr = from_coo.cartesian.transform(matrix_transpose(pmat))
     gcrs_coo = GCRS(crepr,
-                    obstime=to_frame.obstime,
-                    obsgeoloc=to_frame.obsgeoloc,
-                    obsgeovel=to_frame.obsgeovel)
+                    obstime=from_coo.obstime,
```
mhvk, line 246: This change suggests we had a bug (your code looks correct, but checking). Is it worth splitting this out as a separate bugfix? Is this something we can write a test case for? Would be good to have a changelog entry as well...

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -244,32 +243,38 @@ def precessedgeo_to_gcrs(from_coo, to_frame):
...
     # then move to the GCRS that's actually desired
     return gcrs_coo.transform_to(to_frame)
 
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, TEME, ITRS)
 def teme_to_itrs(teme_coo, itrs_frame):
-    # first get us to TEME at the target obstime
-    # TODO: self transform?
-    teme_coo2 = teme_coo.transform_to(TEME(obstime=itrs_frame.obstime))
+    # use the pmatrix to transform to ITRS in the source obstime
```
mhvk, line 256: This change looks like it should not affect any results, right?

astropy/coordinates/tests/test_celestial_transformations.py
```diff
@@ -340,3 +340,47 @@ def test_cirs_icrs():
     # now check ICRS transform gives a decent distance from Barycentre
     moon_icrs = moon_geo.transform_to(ICRS())
     assert_allclose(moon_icrs.distance - 1*u.au, 0.0*u.R_sun, atol=3*u.R_sun)
+
+
+@pytest.mark.parametrize('frame',
```
mhvk, line 345: Total nitpick: put on one line?

astropy/coordinates/tests/test_frames.py
```diff
@@ -1530,7 +1530,7 @@ def test_coordinateattribute_transformation():
     class FrameWithCoordinateAttribute(BaseCoordinateFrame):
         coord_attr = CoordinateAttribute(HCRS)
 
-    hcrs = HCRS(1*u.deg, 2*u.deg, 3*u.AU, obstime='2001-02-03')
+    hcrs = HCRS(1*u.AU, 2*u.AU, 3*u.AU, representation_type='cartesian', obstime='2001-02-03')
```
mhvk, line 1533: Why was this change necessary?

astropy/coordinates/transformations.py
```diff
@@ -1426,6 +1530,45 @@ def __call__(self, fromcoord, toframe):
...
+
+    if vec is not None:
+        # Transform the first displacement vector by the second transformation matrix
+        if next_M is not None:
+            vec = vec.transform(next_M)
+
+        # Calculate the new displacement vector
+        if next_vec is not None:
+            if 's' in vec.differentials and 's' in next_vec.differentials:
+                # Adding vectors with velocities takes more steps
```
mhvk, line 1558: Hmm, we really should just support that in `representations`....

astropy/coordinates/transformations.py
mhvk: Another suggestion would be to have these as `staticmethod` on `DynamicMatrixTransform` and `AffineTransform`. But definitely do not let the perfect be the enemy of good enough!

mhvk (commented):
This is really nice, but also from me some comments and suggestions.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
ayshih: Now extracted as #12152, along with the bugfix for the `TEME` transformations;  I've also added a test there.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
ayshih: This change affects results.  In the current code, where no `TEME` loopback is actually defined, that transformation line doesn't actually do anything, even though it looks like it should.  Thus, the current results are incorrect for a change in `obstime`.

I've extracted this fix as #12152, along with the `PrecessedGeocentric`->`GCRS` bugfix.

astropy/coordinates/tests/test_frames.py
ayshih: Ah, yes, that's a little quirk that I will fix.  The reason here is that the automatically generated loopback for `HCRS` will always call `BaseAffineTransform._apply_transform()`, but will simply pass it `None`/`None` for the affine parameters if there is no change in `obstime`.  However, there is no special detection in `_apply_transform()`, so that function will always convert the internal data to Cartesian even if it does nothing further to the data.  So, my change here was a workaround by starting with Cartesian input so that the later equality check would pass.

I'll add special detection to `_apply_transform()` and remove this change.

ayshih:
When self-reviewing the code, I realized that I was implementing this code in the "wrong" place.  I have significantly refactored the internals, and now most of the code is in a private method of `CompositeTransform`.  This vastly simplifies the modifications that are needed to `BaseCoordinateFrame` and `TransformGraph`, and the code is more sensible and easier to maintain.

I still need to address some remaining comments

astropy/coordinates/baseframe.py
```diff
@@ -1227,14 +1209,102 @@ def transform_to(self, new_frame):
...
+                    elif hasattr(self, inter_frame_attr_nm):
+                        attr = getattr(self, inter_frame_attr_nm)
+                        frattrs[inter_frame_attr_nm] = attr
+
+                next_frame = t.tosys(**frattrs)
+
+                next_affine_params = t._affine_params(curr_frame, next_frame)
+                affine_params = _combine_affine_params(affine_params, next_affine_params)
+
+                # The frame needs to have data for the machinery, but that data should not be
```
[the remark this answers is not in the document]
ayshih: It's actually a little trickier than I thought because some existing affine transformations do look at the data instance because they check if the data has differentials before bothering to calculate and return the velocity shift.

astropy/coordinates/baseframe.py
```diff
@@ -1227,14 +1209,102 @@ def transform_to(self, new_frame):
...
+            for t in trans.transforms:
+                if not isinstance(t, BaseAffineTransform):
+                    raise ConvertError(f"The transformation "
+                                       f"from {t.fromsys.__name__} "
+                                       f"to {t.tosys.__name__} "
+                                       f"is a {t.__class__.__name__} "
+                                       "rather than an affine transformation.")
+
+                # build an intermediate frame with attributes taken from either
+                # self, or if not there, `new_frame`, or if not there, use
```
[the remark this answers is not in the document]
ayshih: Moot because removed in refactor

astropy/coordinates/baseframe.py
```diff
@@ -1227,14 +1209,102 @@ def transform_to(self, new_frame):
...
+        new_frame, trans = self._conform_and_get_transformation(new_frame)
+
+        affine_params = (None, None)
+
+        if trans is not None:
+            curr_frame = self
+
+            for t in trans.transforms:
+                if not isinstance(t, BaseAffineTransform):
+                    raise ConvertError(f"The transformation "
```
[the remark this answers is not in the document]
ayshih: Moot because removed in refactor

astropy/coordinates/baseframe.py
```diff
@@ -1174,33 +1174,15 @@ def represent_as(self, base, s='base', in_frame_units=False):
...
-        transframe : coordinate-like
-            A new object with the coordinate data represented in the
-            ``newframe`` system (meaning a `~astropy.coordinates.SkyCoord` will
-            remain a SkyCoord).
-
-        Raises
-        ------
-        ValueError
-            If there is no possible transformation route.
+        Return a tuple with:
```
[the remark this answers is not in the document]
ayshih: Moot because removed in refactor

astropy/coordinates/transformations.py
ayshih: Suggested functionality now implemented (although not via recursion)

astropy/coordinates/transformations.py
ayshih: I think these review comments missed a subtlety.  These transformation functions are not generic functions, but rather are transformation-path-specific functions that access transformation-path-specific variables that are not passed in through their call signatures.  It's not possible to simply define them outside as is.  In principal, one could define generic functions and then dynamically create the necessary transformation-path-specific functions using `partial()`, but that would become significantly less readable.

Please take a look at current version post-refactoring

astropy/coordinates/transformations.py
```diff
@@ -674,6 +674,100 @@ def deco(func):
...
+            The priority if this transform when finding the shortest
+            coordinate transform path - large numbers are lower priorities.
+
+        Notes
+        -----
+        Even though the created transform is a single step in the graph, it
+        will still internally call the constituent transforms.  Thus, there is
+        no performance benefit for using this created transform.
+
+        For Astropy's built-in frames, the most common choice for ``viasys`` is
```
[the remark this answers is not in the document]
ayshih: Comment added

astropy/coordinates/transformations.py
ayshih: I believe it's simply a choice, with either way being implementable.  I have added a comment explicitly noting that I've implemented the same choice as `CompositeTransform.__call__()`

astropy/coordinates/transformations.py
```diff
@@ -1426,6 +1522,45 @@ def __call__(self, fromcoord, toframe):
         return curr_coord
 
 
+def _combine_affine_params(params, next_params):
```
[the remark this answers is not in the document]
ayshih: Test added

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -260,22 +234,7 @@ def icrs_to_hcrs(icrs_coo, hcrs_frame):
...
-    if (np.all(from_coo.location == to_frame.location)
-            and np.all(from_coo.obstime == to_frame.obstime)):
-        return to_frame.realize_frame(from_coo.data)
-    else:
-        # We perform the self-transform through ICRS, since any change in obstime
-        # will require handling the change in pos/vel of the Earth and it's impact
-        # on aberration.
-        return from_coo.transform_to(ICRS()).transform_to(to_frame)
+# Create loopback transformations
+frame_transform_graph._add_merged_transform(CIRS, ICRS, CIRS)
```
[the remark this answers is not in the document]
ayshih: Comment restored

astropy/coordinates/baseframe.py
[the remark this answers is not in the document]
ayshih: Added to the comment

astropy/coordinates/tests/test_celestial_transformations.py
ayshih: Done

ayshih:
CI all green.  Please re-review!

astropy/coordinates/baseframe.py
```diff
@@ -716,6 +716,11 @@ def get_frame_attr_names(cls):
         return {name: getattr(cls, name)
                 for name in cls.frame_attributes}
 
+    @property
+    def _frame_attr_dict(self):
```
mhvk, line 720: My sense would be not to add this new property, given that it is only used in one place, and really baseframe has too many of these already...

mhvk (approved):
@ayshih - this looks great! It looks like just the right solution.  One in-line comment, but also a question: should the new `_add_merged_transform` method be public? I guess in case of doubt, maybe it is best to keep it private, but in principle it seems nicer to use only public methods outside of the module something is defined in.  

I'll approve, though obviously it would be good for @StuartLittlefair and maybe @nstarman and @adrn to have a look as well.

p.s.  Squashing some commits may be good - I think this really has converged!

nstarman (commented):
Agreed with @mhvk that this looks great! https://astropy.slack.com/archives/C0LAGR5TP/p1632750090005900 made me wonder... I have one question about a potential slowdown for some of the transformations.

ayshih:
> should the new `_add_merged_transform` method be public? I guess in case of doubt, maybe it is best to keep it private, but in principle it seems nicer to use only public methods outside of the module something is defined in.

My thinking is that it's useful for developers, but likely too arcane for nearly all users.  Once it's out there, more use cases may become apparent, and then it can be switched to public.

> Squashing some commits may be good

I cleaned up the commits significantly

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -260,22 +234,13 @@ def icrs_to_hcrs(icrs_coo, hcrs_frame):
...
-    else:
-        # like CIRS, we do this self-transform via ICRS
-        return from_coo.transform_to(ICRS()).transform_to(to_frame)
-
-
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, TETE, TETE)
-def tete_to_tete(from_coo, to_frame):
-    if (np.all(from_coo.location == to_frame.location)
-            and np.all(from_coo.obstime == to_frame.obstime)):
-        return to_frame.realize_frame(from_coo.data)
```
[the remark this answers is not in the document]
nstarman: SGTM!

nstarman (approved):

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
```diff
@@ -90,9 +90,6 @@ def observed_to_cirs(observed_coo, cirs_frame):
...
 
 
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, AltAz, AltAz)
-@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, HADec, HADec)
-def observed_to_observed(from_coo, to_frame):
-    # for now we just implement this through CIRS to make sure we get everything
-    # covered
-    return from_coo.transform_to(CIRS(obstime=from_coo.obstime)).transform_to(to_frame)
+# Create loopback transformations
+frame_transform_graph._add_merged_transform(AltAz, CIRS, AltAz)
```
StuartLittlefair, line 94: following #12203, can we route this via ICRS, and add a comment so it is not regressed?

StuartLittlefair (approved):

StuartLittlefair:
This looks great. One optional improvement is suggested, to do with the `AltAz` self transform, but really that is a symptom of a larger issue (https://github.com/astropy/astropy/issues/10997)

adrn (approved):
Thanks @ayshih! This LGTM now, so let's get this in. It looks like one more rebase is needed, and do address @StuartLittlefair's last point, but otherwise I am happy with this.

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
ayshih: Done!

adrn:
Thanks for the patience on this @ayshih! 🎉

PR #10377, opened 2020-05-20
WIP: updates to the release documentation

astrofrog:
This is meant to be a running PR to keep track of updates to the release procedure during the v4.1 release. We can merge it once the release is done.

PR #10910, opened 2020-10-21
add .github_cache to gitignore

eteq:
This is a minor convenience tweak to ``.gitignore`` - ``.github_cache`` is generated by the generate_releaserst script (see #10377) and therefore might show up for people who are doing releases.


pllim (approved):
Seems uncontroversial enough.

PR #10911, opened 2020-10-21
Split the Azure wheel jobs up

Cadair:
This will mean we are always using our allowance of parallel builds and if one fails it will reduce the time taken on a restart.

This thought occurred to me while watching the 4.1 release build.

I tired to come up with a set of jobs which got us closest to 10 (ended up with 11) and which would give us the biggest benefits. Added comments so the next person understands.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,8 +49,16 @@ jobs:
 
       targets:
       - sdist
+      # The linux builds are the fastest so run all three in one job
       - wheels_cp3[678]*linux_i686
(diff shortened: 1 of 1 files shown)


astrofrog:
Can you temporarily make it so it runs in this PR so we can check it works?

astrofrog:
Great thanks! You may remove the commit now :)

Cadair:
It appears to have saved 10 mins on the build vs the 4.1 tag.

astrofrog (approved):
Looks good thanks!

PR #10928, opened 2020-10-23
TST: Skip flaky float cmp doctest

pllim:
### Description

This pull request is to address remote data job failure caused by 10398

- [x] Remove draft PR status when we can confirm remote data job is green because I am not sure if more doctest on that page downstream is going to break or not.


pllim:
Seems like this is the only offending doctest. I thought more would fail but doesn't look like it from https://travis-ci.com/github/astropy/astropy/jobs/405315999:
```
../../docs/coordinates/spectralcoord.rst .                               [ 99%]
```

I cancelled the other jobs because they seem unnecessary and CircleCI is complaining about limits and paid plans.

StuartLittlefair (approved):

astrofrog (approved):

mhvk (approved):
Great that it is only this one!

PR #10281, opened 2020-05-03
Remove deprecated utils.timer and utils.misc.InheritDocstrings

sashank27:
### Description

This pull request is to remove the `utils.timer` and `utils.misc. InheritDocstrings` module as they were expected to be removed after the passing of their deprecation period (i.e. after 4.0)

Fixes #9024
Fixes #9050
EDIT: Close #10847

PR #10932, opened 2020-10-23
MNT: Replace OrderedDict with dict in utils

pllim:
**Blocked by:** https://github.com/astropy/astropy/pull/10932#issuecomment-772468495

### Description

This pull request is to address part of #10888 . I left `timer.py` alone because it is deprecated and removed in #10281 .

**TODO**

- ~Get the rest of `OrderedDict`.~ Too much work. Just smaller focus here.
- [x] Make sure tests pass.
- [x] Get approval from subpackage maintainers.

**Resolved**

I see failures like this and not sure what to make of it. Maybe @taldcroft or @mhvk can advise.

```
__ TestMetaMaskedColumn.test_mapping_init[meta2] __

self = <astropy.table.tests.test_column.TestMetaMaskedColumn object at 0x7f0e559df7f0>, meta = {'a': 1}

    @pytest.mark.parametrize(('meta'), ([dict([('a', 1)]),
                                         dict([('a', 1)]),
                                         OrderedDictSubclass([('a', 1)])]))
    def test_mapping_init(self, meta):
        d = self.test_class(*self.args, meta=meta)
>       assert type(d.meta) == type(meta)
E       AssertionError: assert <class 'dict'> == <class 'astropy.utils.tests.test_metadata.OrderedDictSubclass'>
E        +  where <class 'dict'> = type({'a': 1})
E        +    where {'a': 1} = <MaskedColumn dtype='float64' length=0>\n.meta
E        +  and   <class 'astropy.utils.tests.test_metadata.OrderedDictSubclass'> = type({'a': 1})

astropy/utils/tests/test_metadata.py:30: AssertionError
__ TestMetaMaskedColumn.test_mapping_set[meta2] __
self = <astropy.table.tests.test_column.TestMetaMaskedColumn object at 0x7f0e5598c640>, meta = {'a': 1}

    @pytest.mark.parametrize(('meta'), ([dict([('a', 1)]),
                                         dict([('a', 1)]),
                                         OrderedDictSubclass([('a', 1)])]))
    def test_mapping_set(self, meta):
        d = self.test_class(*self.args, meta=meta)
>       assert type(d.meta) == type(meta)
E       AssertionError: assert <class 'dict'> == <class 'astropy.utils.tests.test_metadata.OrderedDictSubclass'>
E        +  where <class 'dict'> = type({'a': 1})
E        +    where {'a': 1} = <MaskedColumn dtype='float64' length=0>\n.meta
E        +  and   <class 'astropy.utils.tests.test_metadata.OrderedDictSubclass'> = type({'a': 1})

astropy/utils/tests/test_metadata.py:43: AssertionError

FAILED astropy/table/tests/test_column.py::TestMetaMaskedColumn::test_mapping_init[meta2] - AssertionError: assert <class 'dict'> == <class 'astropy.utils.tests.test_metadata.OrderedDictSubclass'>
FAILED astropy/table/tests/test_column.py::TestMetaMaskedColumn::test_mapping_set[meta2] - AssertionError: assert <class 'dict'> == <class 'astropy.utils.tests.test_metadata.OrderedDictSubclass'>
```

https://github.com/astropy/astropy/blob/ad37af00665febfc1b9647ae84f398a6f7a88e73/astropy/utils/tests/test_metadata.py#L32

### DIFF
--- a/astropy/io/ascii/ecsv.py
+++ b/astropy/io/ascii/ecsv.py
@@ -223,11 +223,7 @@ class EcsvData(basic.BasicData):
         # not even be any table meta, so punt in those cases.
 
         try:
-            if isinstance(self.header.table_meta, (list, tuple)):
-                tmeta = dict(self.header.table_meta)
-            else:
-                tmeta = self.header.table_meta
-            scs = tmeta['__serialized_columns__']
+            scs = self.header.table_meta['__serialized_columns__']
         except (AttributeError, KeyError):
             return
 
--- a/astropy/io/ascii/ecsv.py
+++ b/astropy/io/ascii/ecsv.py
@@ -221,8 +221,13 @@ class EcsvData(basic.BasicData):
 
         # Get the serialized columns spec.  It might not exist and there might
         # not even be any table meta, so punt in those cases.
+
         try:
-            scs = self.header.table_meta['__serialized_columns__']
+            if isinstance(self.header.table_meta, (list, tuple)):
+                tmeta = dict(self.header.table_meta)
+            else:
+                tmeta = self.header.table_meta
+            scs = tmeta['__serialized_columns__']
         except (AttributeError, KeyError):
             return
 
--- a/astropy/io/ascii/ecsv.py
+++ b/astropy/io/ascii/ecsv.py
@@ -221,7 +221,6 @@ class EcsvData(basic.BasicData):
 
         # Get the serialized columns spec.  It might not exist and there might
         # not even be any table meta, so punt in those cases.
-
         try:
             scs = self.header.table_meta['__serialized_columns__']
         except (AttributeError, KeyError):
--- a/astropy/table/meta.py
+++ b/astropy/table/meta.py
@@ -1,7 +1,5 @@
 import textwrap
 import copy
-from collections import OrderedDict
-
 
 __all__ = ['get_header_from_yaml', 'get_yaml_from_header', 'get_yaml_from_table']
 
@@ -45,105 +43,6 @@ class ColumnDict(dict):
         way for column attributes.
         """
         return ColumnOrderList(super().items())
-
-
-def _construct_odict(load, node):
-    """
-    Construct OrderedDict from !!omap in yaml safe load.
-
-    Source: https://gist.github.com/weaver/317164
-    License: Unspecified
-
-    This is the same as SafeConstructor.construct_yaml_omap(),
-    except the data type is changed to OrderedDict() and setitem is
-    used instead of append in the loop
-
-    Examples
-    --------
-    ::
-
-      >>> yaml.load('''  # doctest: +SKIP
-      ... !!omap
-      ... - foo: bar
-      ... - mumble: quux
-      ... - baz: gorp
-      ... ''')
-      OrderedDict([('foo', 'bar'), ('mumble', 'quux'), ('baz', 'gorp')])
-
-      >>> yaml.load('''!!omap [ foo: bar, mumble: quux, baz : gorp ]''')  # doctest: +SKIP
-      OrderedDict([('foo', 'bar'), ('mumble', 'quux'), ('baz', 'gorp')])
-    """
-    import yaml
-
-    omap = OrderedDict()
-    yield omap
-    if not isinstance(node, yaml.SequenceNode):
-        raise yaml.constructor.ConstructorError(
-            "while constructing an ordered map", node.start_mark,
-            f"expected a sequence, but found {node.id}", node.start_mark)
-
-    for subnode in node.value:
-        if not isinstance(subnode, yaml.MappingNode):
-            raise yaml.constructor.ConstructorError(
-                "while constructing an ordered map", node.start_mark,
-                f"expected a mapping of length 1, but found {subnode.id}",
-                subnode.start_mark)
-
-        if len(subnode.value) != 1:
-            raise yaml.constructor.ConstructorError(
-                "while constructing an ordered map", node.start_mark,
-                f"expected a single mapping item, but found {len(subnode.value)} items",
-                subnode.start_mark)
-
-        key_node, value_node = subnode.value[0]
-        key = load.construct_object(key_node)
-        value = load.construct_object(value_node)
-        omap[key] = value
-
-
-def _repr_pairs(dump, tag, sequence, flow_style=None):
-    """
-    This is the same code as BaseRepresenter.represent_sequence(),
-    but the value passed to dump.represent_data() in the loop is a
-    dictionary instead of a tuple.
-
-    Source: https://gist.github.com/weaver/317164
-    License: Unspecified
-    """
-    import yaml
-
-    value = []
-    node = yaml.SequenceNode(tag, value, flow_style=flow_style)
-    if dump.alias_key is not None:
-        dump.represented_objects[dump.alias_key] = node
-    best_style = True
-    for (key, val) in sequence:
-        item = dump.represent_data({key: val})
-        if not (isinstance(item, yaml.ScalarNode) and not item.style):
-            best_style = False
-        value.append(item)
-    if flow_style is None:
-        if dump.default_flow_style is not None:
-            node.flow_style = dump.default_flow_style
-        else:
-            node.flow_style = best_style
-    return node
-
-
-def _repr_odict(dumper, data):
-    """
-    Represent OrderedDict in yaml dump.
-
-    Source: https://gist.github.com/weaver/317164
-    License: Unspecified
-
-    >>> data = OrderedDict([('foo', 'bar'), ('mumble', 'quux'), ('baz', 'gorp')])
-    >>> yaml.dump(data, default_flow_style=False)  # doctest: +SKIP
-    '!!omap\\n- foo: bar\\n- mumble: quux\\n- baz: gorp\\n'
-    >>> yaml.dump(data, default_flow_style=True)  # doctest: +SKIP
-    '!!omap [foo: bar, mumble: quux, baz: gorp]\\n'
-    """
-    return _repr_pairs(dumper, 'tag:yaml.org,2002:omap', data.items())
 
 
 def _repr_column_dict(dumper, data):
@@ -238,7 +137,7 @@ def get_yaml_from_header(header):
 
     class TableDumper(AstropyDumper):
         """
-        Custom Dumper that represents OrderedDict as an !!omap object.
+        Custom Dumper that represents dict (insertion-ordered) as an !!omap object.
         """
 
         def represent_mapping(self, tag, mapping, flow_style=None):
@@ -280,7 +179,6 @@ def get_yaml_from_header(header):
                     node.flow_style = best_style
             return node
 
-    TableDumper.add_representer(OrderedDict, _repr_odict)
     TableDumper.add_representer(ColumnDict, _repr_column_dict)
 
     header = copy.copy(header)  # Don't overwrite original
@@ -326,18 +224,10 @@ def get_header_from_yaml(lines):
 
     from astropy.io.misc.yaml import AstropyLoader
 
-    class TableLoader(AstropyLoader):
-        """
-        Custom Loader that constructs OrderedDict from an !!omap object.
-        This does nothing but provide a namespace for adding the
-        custom odict constructor.
-        """
-
-    TableLoader.add_constructor('tag:yaml.org,2002:omap', _construct_odict)
     # Now actually load the YAML data structure into `meta`
     header_yaml = textwrap.dedent('\n'.join(lines))
     try:
-        header = yaml.load(header_yaml, Loader=TableLoader)
+        header = yaml.load(header_yaml, Loader=AstropyLoader)
     except Exception as err:
         raise YamlParseError(str(err))
 
--- a/astropy/table/meta.py
+++ b/astropy/table/meta.py
@@ -1,59 +1,6 @@
 import textwrap
-import copy
 
 __all__ = ['get_header_from_yaml', 'get_yaml_from_header', 'get_yaml_from_table']
-
-
-class ColumnOrderList(list):
-    """
-    List of tuples that sorts in a specific order that makes sense for
-    astropy table column attributes.
-    """
-
-    def sort(self, *args, **kwargs):
-        super().sort()
-
-        column_keys = ['name', 'unit', 'datatype', 'format', 'description', 'meta']
-        in_dict = dict(self)
-        out_list = []
-
-        for key in column_keys:
-            if key in in_dict:
-                out_list.append((key, in_dict[key]))
-        for key, val in self:
-            if key not in column_keys:
-                out_list.append((key, val))
-
-        # Clear list in-place
-        del self[:]
-
-        self.extend(out_list)
-
-
-class ColumnDict(dict):
-    """
-    Specialized dict subclass to represent attributes of a Column
-    and return items() in a preferred order.  This is only for use
-    in generating a YAML map representation that has a fixed order.
-    """
-
-    def items(self):
-        """
-        Return items as a ColumnOrderList, which sorts in the preferred
-        way for column attributes.
-        """
-        return ColumnOrderList(super().items())
-
-
-def _repr_column_dict(dumper, data):
-    """
-    Represent ColumnDict in yaml dump.
-
-    This is the same as an ordinary mapping except that the keys
-    are written in a fixed order that makes sense for astropy table
-    columns.
-    """
-    return dumper.represent_mapping('tag:yaml.org,2002:map', data)
 
 
 def _get_col_attributes(col):
@@ -61,23 +8,29 @@ def _get_col_attributes(col):
     Extract information from a column (apart from the values) that is required
     to fully serialize the column.
     """
-    attrs = ColumnDict()
-    attrs['name'] = col.info.name
+    attrs = {}
 
-    type_name = col.info.dtype.type.__name__
-    if type_name.startswith(('bytes', 'str')):
-        type_name = 'string'
-    if type_name.endswith('_'):
-        type_name = type_name[:-1]  # string_ and bool_ lose the final _ for ECSV
-    attrs['datatype'] = type_name
+    def get_datatype(dtype):
+        datatype = dtype.type.__name__
+        if datatype.startswith(('bytes', 'str')):
+            datatype = 'string'
+        if datatype.endswith('_'):
+            datatype = datatype[:-1]  # string_ and bool_ lose the final _ for ECSV
+        return datatype
 
     # Set the output attributes
-    for attr, nontrivial, xform in (('unit', lambda x: x is not None, str),
+    for attr, nontrivial, xform in (('name', lambda x: True, None),
+                                    ('unit', lambda x: x is not None, str),
+                                    ('dtype', lambda x: True, get_datatype),
                                     ('format', lambda x: x is not None, None),
                                     ('description', lambda x: x is not None, None),
                                     ('meta', lambda x: x, None)):
         col_attr = getattr(col.info, attr)
         if nontrivial(col_attr):
+            if attr == 'dtype':
+                # ECSV uses `datatype` not `dtype` for similarity with ASDF and
+                # numpy agnosticism.
+                attr = 'datatype'
             attrs[attr] = xform(col_attr) if xform else col_attr
 
     return attrs
@@ -135,58 +88,12 @@ def get_yaml_from_header(header):
 
     from astropy.io.misc.yaml import AstropyDumper
 
-    class TableDumper(AstropyDumper):
-        """
-        Custom Dumper that represents dict (insertion-ordered) as an !!omap object.
-        """
-
-        def represent_mapping(self, tag, mapping, flow_style=None):
-            """
-            This is a combination of the Python 2 and 3 versions of this method
-            in the PyYAML library to allow the required key ordering via the
-            ColumnOrderList object.  The Python 3 version insists on turning the
-            items() mapping into a list object and sorting, which results in
-            alphabetical order for the column keys.
-            """
-            value = []
-            node = yaml.MappingNode(tag, value, flow_style=flow_style)
-            if self.alias_key is not None:
-                self.represented_objects[self.alias_key] = node
-            best_style = True
-            if hasattr(mapping, 'items'):
-                mapping = mapping.items()
-                if hasattr(mapping, 'sort'):
-                    mapping.sort()
-                else:
-                    mapping = list(mapping)
-                    try:
-                        mapping = sorted(mapping)
-                    except TypeError:
-                        pass
-
-            for item_key, item_value in mapping:
-                node_key = self.represent_data(item_key)
-                node_value = self.represent_data(item_value)
-                if not (isinstance(node_key, yaml.ScalarNode) and not node_key.style):
-                    best_style = False
-                if not (isinstance(node_value, yaml.ScalarNode) and not node_value.style):
-                    best_style = False
-                value.append((node_key, node_value))
-            if flow_style is None:
-                if self.default_flow_style is not None:
-                    node.flow_style = self.default_flow_style
-                else:
-                    node.flow_style = best_style
-            return node
-
-    TableDumper.add_representer(ColumnDict, _repr_column_dict)
-
-    header = copy.copy(header)  # Don't overwrite original
     header['datatype'] = [_get_col_attributes(col) for col in header['cols']]
+    header = {key: header[key] for key in sorted(header)}
     del header['cols']
 
-    lines = yaml.dump(header, default_flow_style=None,
-                      Dumper=TableDumper, width=130).splitlines()
+    lines = yaml.dump(header, default_flow_style=None, sort_keys=False,
+                      Dumper=AstropyDumper, width=130).splitlines()
     return lines
 
 
--- a/astropy/table/meta.py
+++ b/astropy/table/meta.py
@@ -3,6 +3,45 @@ import textwrap
 __all__ = ['get_header_from_yaml', 'get_yaml_from_header', 'get_yaml_from_table']
 
 
+def _construct_dict_from_omap(load, node):
+    """
+    Construct dict from !!omap in yaml safe load.
+
+    Source: https://gist.github.com/weaver/317164
+    License: Unspecified
+
+    This is the same as SafeConstructor.construct_yaml_omap(),
+    except the data type is changed to dict() and setitem is
+    used instead of append in the loop.
+    """
+    import yaml
+
+    omap = {}
+    yield omap
+    if not isinstance(node, yaml.SequenceNode):
+        raise yaml.constructor.ConstructorError(
+            "while constructing an ordered map", node.start_mark,
+            f"expected a sequence, but found {node.id}", node.start_mark)
+
+    for subnode in node.value:
+        if not isinstance(subnode, yaml.MappingNode):
+            raise yaml.constructor.ConstructorError(
+                "while constructing an ordered map", node.start_mark,
+                f"expected a mapping of length 1, but found {subnode.id}",
+                subnode.start_mark)
+
+        if len(subnode.value) != 1:
+            raise yaml.constructor.ConstructorError(
+                "while constructing an ordered map", node.start_mark,
+                f"expected a single mapping item, but found {len(subnode.value)} items",
+                subnode.start_mark)
+
+        key_node, value_node = subnode.value[0]
+        key = load.construct_object(key_node)
+        value = load.construct_object(value_node)
+        omap[key] = value
+
+
 def _get_col_attributes(col):
     """
     Extract information from a column (apart from the values) that is required
@@ -131,10 +170,19 @@ def get_header_from_yaml(lines):
 
     from astropy.io.misc.yaml import AstropyLoader
 
+    class TableLoader(AstropyLoader):
+        """
+        Custom Loader that constructs OrderedDict from an !!omap object.
+        This does nothing but provide a namespace for adding the
+        custom odict constructor.
+        """
+
+    TableLoader.add_constructor('tag:yaml.org,2002:omap', _construct_dict_from_omap)
+
     # Now actually load the YAML data structure into `meta`
     header_yaml = textwrap.dedent('\n'.join(lines))
     try:
-        header = yaml.load(header_yaml, Loader=AstropyLoader)
+        header = yaml.load(header_yaml, Loader=TableLoader)
     except Exception as err:
         raise YamlParseError(str(err))
 
--- a/astropy/utils/metadata.py
+++ b/astropy/utils/metadata.py
@@ -7,7 +7,6 @@ from functools import wraps
 
 import warnings
 
-from collections import OrderedDict
 from collections.abc import Mapping
 from copy import deepcopy
 
@@ -398,12 +397,12 @@ class MetaData:
         if instance is None:
             return self
         if not hasattr(instance, '_meta'):
-            instance._meta = OrderedDict()
+            instance._meta = dict()
         return instance._meta
 
     def __set__(self, instance, value):
         if value is None:
-            instance._meta = OrderedDict()
+            instance._meta = dict()
         else:
             if isinstance(value, Mapping):
                 if self.copy:
--- a/astropy/utils/metadata.py
+++ b/astropy/utils/metadata.py
@@ -405,9 +405,6 @@ class MetaData:
             instance._meta = dict()
             return
 
-        if isinstance(value, (list, tuple)):
-            value = dict(value)
-
         if isinstance(value, Mapping):
             if self.copy:
                 instance._meta = deepcopy(value)
--- a/astropy/utils/metadata.py
+++ b/astropy/utils/metadata.py
@@ -403,14 +403,18 @@ class MetaData:
     def __set__(self, instance, value):
         if value is None:
             instance._meta = dict()
-        else:
-            if isinstance(value, Mapping):
-                if self.copy:
-                    instance._meta = deepcopy(value)
-                else:
-                    instance._meta = value
+            return
+
+        if isinstance(value, (list, tuple)):
+            value = dict(value)
+
+        if isinstance(value, Mapping):
+            if self.copy:
+                instance._meta = deepcopy(value)
             else:
-                raise TypeError("meta attribute must be dict-like")
+                instance._meta = value
+        else:
+            raise TypeError(f"meta attribute must be dict-like, but got {value}")
 
 
 class MetaAttribute:
--- a/astropy/utils/tests/test_metadata.py
+++ b/astropy/utils/tests/test_metadata.py
@@ -48,7 +48,7 @@ class MetaBaseTest:
     @pytest.mark.parametrize(('meta'), (["ceci n'est pas un meta", 1.2, [1, 2, 3]]))
     def test_non_mapping_set(self, meta):
         with pytest.raises(TypeError):
-            d = self.test_class(*self.args, meta=meta)
+            self.test_class(*self.args, meta=meta)
 
     def test_meta_fits_header(self):
 
--- a/astropy/utils/tests/test_metadata.py
+++ b/astropy/utils/tests/test_metadata.py
@@ -21,7 +21,7 @@ class MetaBaseTest:
 
     def test_none(self):
         d = self.test_class(*self.args)
-        assert isinstance(d.meta, OrderedDict)
+        assert isinstance(d.meta, dict)
         assert len(d.meta) == 0
 
     @pytest.mark.parametrize(('meta'), ([dict([('a', 1)]),
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -324,6 +324,8 @@ Other Changes and Additions
 
 - Officially declared minversion of ``ipython`` to be 4.2. [#10942]
 
+- Minimum version of required PyYAML is now 5.1. [#10932]
+
 - Refactor conversions between ``GCRS`` and ``CIRS,TETE`` for better accuracy
   and substantially improved speed. [#11069]
 
--- a/astropy/io/ascii/tests/test_ecsv.py
+++ b/astropy/io/ascii/tests/test_ecsv.py
@@ -88,6 +88,8 @@ def test_write_full():
     """
     Write a full-featured table with common types and explicitly checkout output
     """
+    from astropy.utils.diff import report_diff_values
+
     t = T_DTYPES['bool', 'int64', 'float64', 'str']
     lines = ['# %ECSV 0.9',
              '# ---',
@@ -122,7 +124,7 @@ def test_write_full():
 
     out = StringIO()
     t.write(out, format='ascii.ecsv')
-    assert out.getvalue().splitlines() == lines
+    assert report_diff_values(os.linesep.join(lines), out.getvalue(), fileobj=sys.stderr)
 
 
 @pytest.mark.skipif('not HAS_YAML')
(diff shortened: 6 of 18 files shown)


astropy/utils/tests/test_metadata.py
```diff
@@ -11,7 +9,7 @@
 from astropy.io import fits
 
 
-class OrderedDictSubclass(OrderedDict):
+class OrderedDictSubclass(dict):
```
mhvk, line 12: This one we should keep the same - it specifically tests what happens if someone subclasses `OrderedDict` (which should still be allowed)

astropy/utils/tests/test_metadata.py
```diff
@@ -21,11 +19,11 @@ class MetaBaseTest:
 
     def test_none(self):
         d = self.test_class(*self.args)
-        assert isinstance(d.meta, OrderedDict)
+        assert isinstance(d.meta, dict)
         assert len(d.meta) == 0
 
     @pytest.mark.parametrize(('meta'), ([dict([('a', 1)]),
-                                         OrderedDict([('a', 1)]),
```
mhvk, line 28: This should also still be tested.

astropy/utils/tests/test_metadata.py
```diff
@@ -38,7 +36,7 @@ def test_non_mapping_init(self, meta):
             self.test_class(*self.args, meta=meta)
 
     @pytest.mark.parametrize(('meta'), ([dict([('a', 1)]),
-                                         OrderedDict([('a', 1)]),
```
mhvk, line 41: Also keep this one.

mhvk (commented):
@pllim - looks mostly good, but in `test_metadata`, the `OrderedDict` stuff should be kept - it is specifically to test that the class (and a subclass) work, and that should remain the case. This should also solve the failure (I think...)

astropy/utils/tests/test_metadata.py
pllim: Makes sense. Thanks!

pllim:
I don't have enough Table metadata parsing with YAML expertise to fix this error:
```
     # %ECSV 0.9
     # ---
     # datatype:
     # - name: bool
     #   unit: m / s
     #   datatype: bool
     #   description: descr_bool
     #   meta: {meta bool: 1}
     # - name: int64
     #   unit: m / s
     #   datatype: int64
     #   description: descr_int64
     #   meta: {meta int64: 1}
     # - name: float64
     #   unit: m / s
     #   datatype: float64
     #   description: descr_float64
     #   meta: {meta float64: 1}
     # - name: str
     #   unit: m / s
     #   datatype: string
     #   description: descr_str
     #   meta: {meta str: 1}
  a> # meta: !!omap
  b> # meta:
  a> # - comments: [comment1, comment2]
   ?   ^
  b> #   comments: [comment1, comment2]
   ?   ^
     # schema: astropy-2.0
     bool int64 float64 str
     False 0 0.0 "ab 0"
     True 1 1.0 "ab, 1"
     False 2 2.0 ab2
```

mhvk:
I think that change is expected - the `yaml` file effectively stored the fact that it was an ordered dict. But cc @taldcroft just in case...

astropy/io/ascii/ecsv.py
```diff
@@ -221,8 +221,13 @@ def _set_fill_values(self, cols):
 
         # Get the serialized columns spec.  It might not exist and there might
         # not even be any table meta, so punt in those cases.
+
         try:
-            scs = self.header.table_meta['__serialized_columns__']
+            if isinstance(self.header.table_meta, (list, tuple)):
```
pllim, line 226: I dunno... I have to make some weird patching to get the tests to pass...

astropy/utils/metadata.py
```diff
@@ -398,20 +397,24 @@ def __get__(self, instance, owner):
...
-        else:
-            if isinstance(value, Mapping):
-                if self.copy:
-                    instance._meta = deepcopy(value)
-                else:
-                    instance._meta = value
+            instance._meta = dict()
+            return
+
+        if isinstance(value, (list, tuple)):
```
pllim, line 408: Another weird patch here.

pllim:
Update: Got the tests to pass, though I see many more occurrences of `OrderedDict` across different subpackages. This might take a while... Should we break the work up into a series of PRs instead? Should we be touching FITS?

**coordinates**

```
./astropy/coordinates/tests/test_representation.py:5:from collections import OrderedDict
./astropy/coordinates/tests/test_representation.py:1092:        attr_classes = OrderedDict([('lon', Longitude180),
./astropy/coordinates/tests/test_representation.py:1113:        attr_classes = OrderedDict([('lon', Longitude),
./astropy/coordinates/tests/test_representation.py:1161:            attr_classes = OrderedDict([('lon', Longitude),
./astropy/coordinates/tests/test_representation.py:1172:            attr_classes = OrderedDict([('lon', Longitude),
./astropy/coordinates/tests/test_representation.py:1527:        attr_classes = OrderedDict([('phi', Angle),
./astropy/coordinates/tests/test_representation.py:1556:            return OrderedDict(
./astropy/coordinates/tests/test_representation.py:1565:            return OrderedDict((('phi', sintheta),
./astropy/coordinates/tests/test_unit_representation.py:5:from collections import OrderedDict
./astropy/coordinates/tests/test_unit_representation.py:48:        attr_classes = OrderedDict([('lon', Longitude180),
./astropy/coordinates/tests/test_unit_representation.py:52:        attr_classes = OrderedDict([('lon', Longitude180),
./astropy/coordinates/transformations.py:24:from collections import defaultdict, OrderedDict
./astropy/coordinates/transformations.py:1430:trans_to_color = OrderedDict()
./astropy/coordinates/baseframe.py:13:from collections import namedtuple, OrderedDict, defaultdict
./astropy/coordinates/baseframe.py:412:    frame_attributes = OrderedDict()
./astropy/coordinates/baseframe.py:714:        return OrderedDict((name, getattr(cls, name))
./astropy/coordinates/baseframe.py:849:        out = OrderedDict()
./astropy/coordinates/baseframe.py:860:        out = OrderedDict()
./astropy/coordinates/baseframe.py:1896:        self.frame_attributes = OrderedDict()
./astropy/coordinates/representation.py:10:from collections import OrderedDict
./astropy/coordinates/representation.py:682:    and define an ``attr_classes`` attribute, an `~collections.OrderedDict`
./astropy/coordinates/representation.py:1256:    attr_classes = OrderedDict([('x', u.Quantity),
./astropy/coordinates/representation.py:1316:        return OrderedDict(
./astropy/coordinates/representation.py:1323:        return OrderedDict((('x', l), ('y', l), ('z', l)))
./astropy/coordinates/representation.py:1549:    attr_classes = OrderedDict([('lon', Longitude),
./astropy/coordinates/representation.py:1584:        return OrderedDict(
./astropy/coordinates/representation.py:1592:        return OrderedDict((('lon', sf_lon),
./astropy/coordinates/representation.py:1749:    attr_classes = OrderedDict([('distance', u.Quantity)])
./astropy/coordinates/representation.py:1768:        return OrderedDict((('distance', l),))
./astropy/coordinates/representation.py:1835:    attr_classes = OrderedDict([('lon', Longitude),
./astropy/coordinates/representation.py:1886:        return OrderedDict(
./astropy/coordinates/representation.py:1897:        return OrderedDict((('lon', sf_lon),
./astropy/coordinates/representation.py:1999:    attr_classes = OrderedDict([('phi', Angle),
./astropy/coordinates/representation.py:2045:        return OrderedDict(
./astropy/coordinates/representation.py:2057:        return OrderedDict((('phi', r * sintheta),
./astropy/coordinates/representation.py:2155:    attr_classes = OrderedDict([('rho', u.Quantity),
./astropy/coordinates/representation.py:2189:        return OrderedDict(
./astropy/coordinates/representation.py:2197:        return OrderedDict((('rho', l),
./astropy/coordinates/representation.py:2247:            cls.attr_classes = OrderedDict([('d_' + c, u.Quantity)
./astropy/coordinates/representation.py:2887:    attr_classes = OrderedDict([('d_lon_coslat', u.Quantity),
./astropy/coordinates/representation.py:2950:    attr_classes = OrderedDict([('d_lon_coslat', u.Quantity),
./astropy/coordinates/solar_system.py:8:from collections import OrderedDict
./astropy/coordinates/solar_system.py:33:BODY_NAME_TO_KERNEL_SPEC = OrderedDict([
./astropy/coordinates/solar_system.py:49:PLAN94_BODY_NAME_TO_PLANET_INDEX = OrderedDict(
./docs/coordinates/frames.rst:52:    OrderedDict([('equinox', <Time object: scale='tt' format='jyear_str' value=J2000.000>)])
./docs/coordinates/frames.rst:112:    OrderedDict([('ra', 'lon'), ('dec', 'lat'), ('distance', 'distance')])
./docs/coordinates/representations.rst:598:* ``attr_classes`` class attribute (``OrderedDict``):
./docs/coordinates/representations.rst:662:        attr_classes = OrderedDict([('comp1', ComponentClass1),
./docs/coordinates/inplace.rst:34:    OrderedDict([('ra', 'lon'), ('dec', 'lat'), ('distance', 'distance')])
./docs/coordinates/skycoord.rst:535:  OrderedDict([('l', 'lon'), ('b', 'lat'), ('distance', 'distance')])
./docs/coordinates/skycoord.rst:538:  OrderedDict([('l', Unit("deg")), ('b', Unit("deg"))])
./docs/coordinates/skycoord.rst:554:  OrderedDict([('equinox', <Time object: scale='tt' format='byear_str' value=B1950.000>), ('obstime', None)])
./docs/coordinates/skycoord.rst:840:    OrderedDict([('ra', 'lon'), ('dec', 'lat'), ('distance', 'distance')])
```

**io.ascii**
```
./astropy/io/ascii/tests/test_read.py:6:from collections import OrderedDict
./astropy/io/ascii/tests/test_read.py:570:    keywords = OrderedDict((('intval', 1),
./astropy/io/ascii/fastbasic.py:5:from collections import OrderedDict
./astropy/io/ascii/fastbasic.py:138:        meta = OrderedDict()
./astropy/io/ascii/fastbasic.py:266:        meta = OrderedDict()
./astropy/io/ascii/ecsv.py:8:from collections import OrderedDict
./astropy/io/ascii/ecsv.py:152:        header_cols = OrderedDict((x['name'], x) for x in header['datatype'])
./astropy/io/ascii/core.py:22:from collections import OrderedDict
./astropy/io/ascii/core.py:556:        the table ``lines`` and update the OrderedDict ``meta`` in place.  This base
./astropy/io/ascii/core.py:1214:        self.meta = OrderedDict(table=OrderedDict(),
./astropy/io/ascii/core.py:1215:                                cols=OrderedDict())
./astropy/io/ascii/daophot.py:15:from collections import defaultdict, OrderedDict
./astropy/io/ascii/daophot.py:120:                keywords = OrderedDict(map(self.extract_keyword_line, grouped_lines_dict['#K']))
./astropy/io/ascii/ipac.py:13:from collections import defaultdict, OrderedDict
./astropy/io/ascii/ipac.py:119:        table_meta['keywords'] = OrderedDict()
```

**io.fits**
```
./astropy/io/fits/tests/test_connect.py:614:        # Mixin info.meta can None instead of empty OrderedDict(), #6720 would
./astropy/io/fits/tests/test_header.py:61:    dict1 = collections.OrderedDict(list1)
./astropy/io/fits/_utils.pyx:2:from collections import OrderedDict
./astropy/io/fits/_utils.pyx:23:    cards = OrderedDict()
./astropy/io/fits/column.py:12:from collections import OrderedDict
./astropy/io/fits/column.py:146:KEYWORD_TO_ATTRIBUTE = OrderedDict(zip(KEYWORD_NAMES, KEYWORD_ATTRIBUTES))
./astropy/io/fits/column.py:148:ATTRIBUTE_TO_KEYWORD = OrderedDict(zip(KEYWORD_ATTRIBUTES, KEYWORD_NAMES))
./astropy/io/fits/fitstime.py:5:from collections import defaultdict, OrderedDict
./astropy/io/fits/fitstime.py:435:    time_columns = defaultdict(OrderedDict)
./astropy/io/fits/fitstime.py:523:    newtable.meta['__coordinate_columns__'] = defaultdict(OrderedDict)
./docs/io/fits/appendix/header_transition.rst:66:(or :class:`~collections.OrderedDict` in Python) for ease of use by new users
./docs/io/fits/appendix/header_transition.rst:103:similar to an :class:`~collections.OrderedDict` in that the order of insertions
```

**io.misc**

```
./astropy/io/misc/tests/test_hdf5.py:652:        # Mixin info.meta can None instead of empty OrderedDict(), #6720 would
./astropy/io/misc/asdf/tags/time/tests/test_time.py:5:from collections import OrderedDict
./astropy/io/misc/asdf/tags/time/tests/test_time.py:21:    newschema = OrderedDict()
./astropy/io/misc/asdf/tags/time/tests/test_time.py:34:                cursor = cursor.setdefault('items', OrderedDict())
./astropy/io/misc/asdf/tags/time/tests/test_time.py:36:                cursor = cursor.setdefault('properties', OrderedDict())
./astropy/io/misc/asdf/tags/time/tests/test_time.py:40:                    cursor = cursor.setdefault(part, OrderedDict())
```

**io.registry**
```
./astropy/io/registry.py:10:from collections import OrderedDict
./astropy/io/registry.py:23:_readers = OrderedDict()
./astropy/io/registry.py:24:_writers = OrderedDict()
./astropy/io/registry.py:25:_identifiers = OrderedDict()
```

**modeling**

```
./astropy/modeling/tests/test_models_quantities.py:3:from collections import OrderedDict
./astropy/modeling/tests/test_models_quantities.py:220:            kwargs = OrderedDict(zip(('x', 'y'), args))
./astropy/modeling/tests/test_models_quantities.py:222:            kwargs = OrderedDict(zip(('x', 'y', 'z'), args))
./astropy/modeling/utils.py:368:    storage, for example an `OrderedDict`.  However, even in this case
./astropy/modeling/core.py:25:from collections import defaultdict, OrderedDict, deque
./astropy/modeling/core.py:74:        return OrderedDict()
./astropy/modeling/core.py:137:        pdict = OrderedDict()
./astropy/modeling/core.py:3124:        self._parameters_ = OrderedDict()
./astropy/modeling/core.py:3712:    params = OrderedDict((param.name, Parameter(param.name,
./astropy/modeling/core.py:3721:    members = OrderedDict([
```

**nddata**

```
./astropy/nddata/tests/test_decorators.py:343:    # Meta is always set (OrderedDict) so it has a special case that it's
./astropy/nddata/tests/test_nddata.py:6:from collections import OrderedDict
./astropy/nddata/nddata.py:60:        an empty `collections.OrderedDict` is created.
./astropy/nddata/bitmask.py:8:from collections import OrderedDict
./astropy/nddata/bitmask.py:88:            cache = OrderedDict()
./astropy/nddata/flag_collection.py:4:from collections import OrderedDict
./astropy/nddata/flag_collection.py:13:class FlagCollection(OrderedDict):
./astropy/nddata/flag_collection.py:22:    `~collections.OrderedDict`, but with the addition of a ``shape=``
./astropy/nddata/flag_collection.py:37:        OrderedDict.__init__(self, *args, **kwargs)
./astropy/nddata/flag_collection.py:43:                OrderedDict.__setitem__(self, item, value, **kwargs)
```

**table**

```
./astropy/table/table.py:6:from collections import OrderedDict, defaultdict
./astropy/table/table.py:196:# be sure to change the OrderedDict ref(s) in Row and in __len__().
./astropy/table/table.py:198:class TableColumns(OrderedDict):
./astropy/table/table.py:199:    """OrderedDict subclass for a set of columns.
./astropy/table/table.py:239:            return OrderedDict.__getitem__(self, item)
./astropy/table/table.py:667:        columns = OrderedDict((key, col if isinstance(col, BaseColumn) else col_copy(col))
./astropy/table/table.py:1801:            return len(OrderedDict.__getitem__(self.columns, self._first_colname))
./astropy/table/table.py:3417:        out = OrderedDict()
./astropy/table/table.py:3507:        out = OrderedDict()
./astropy/table/tests/test_info.py:7:from collections import OrderedDict
./astropy/table/tests/test_info.py:135:        assert cinfo == OrderedDict([('name', 'name'),
./astropy/table/tests/test_info.py:162:        assert cinfo == OrderedDict([('name', 'name'),
./astropy/table/tests/test_info.py:180:        assert cinfo == OrderedDict([('dtype', 'int64'),
./astropy/table/tests/test_operations.py:3:from collections import OrderedDict
./astropy/table/tests/test_operations.py:52:        self.t1.meta.update(OrderedDict([('b', [1, 2]), ('c', {'a': 1}), ('d', 1)]))
./astropy/table/tests/test_operations.py:53:        self.t2.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:54:        self.t3.meta.update(OrderedDict([('b', 3), ('c', [1, 2]), ('d', 2), ('a', 1)]))
./astropy/table/tests/test_operations.py:56:        self.meta_merge = OrderedDict([('b', [1, 2, 3, 4]),
./astropy/table/tests/test_operations.py:398:        meta1 = OrderedDict([('b', [1, 2]), ('c', {'a': 1}), ('d', 1)])
./astropy/table/tests/test_operations.py:399:        meta2 = OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)])
./astropy/table/tests/test_operations.py:807:        self.t1.meta.update(OrderedDict([('b', [1, 2]), ('c', {'a': 1}), ('d', 1)]))
./astropy/table/tests/test_operations.py:808:        self.t2.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:809:        self.t4.meta.update(OrderedDict([('b', [5, 6]), ('c', {'c': 1}), ('e', 1)]))
./astropy/table/tests/test_operations.py:810:        self.t5.meta.update(OrderedDict([('b', 3), ('c', 'k'), ('d', 1)]))
./astropy/table/tests/test_operations.py:811:        self.meta_merge = OrderedDict([('b', [1, 2, 3, 4, 5, 6]),
./astropy/table/tests/test_operations.py:1002:        t1['a'].info.meta.update(OrderedDict([('b', [1, 2]), ('c', {'a': 1}), ('d', 1)]))
./astropy/table/tests/test_operations.py:1003:        t2['a'].info.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:1004:        t4['a'].info.meta.update(OrderedDict([('b', [5, 6]), ('c', {'c': 1}), ('e', 1)]))
./astropy/table/tests/test_operations.py:1007:        t2['b'].info.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:1049:        assert out['b'].info.meta == OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)])
./astropy/table/tests/test_operations.py:1076:        t1['a'].info.meta.update(OrderedDict([('b', [1, 2]), ('c', {'a': 1}), ('d', 1)]))
./astropy/table/tests/test_operations.py:1077:        t2['a'].info.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:1078:        t4['a'].info.meta.update(OrderedDict([('b', [5, 6]), ('c', {'c': 1}), ('e', 1)]))
./astropy/table/tests/test_operations.py:1081:        t2['b'].info.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:1101:        assert out['b'].info.meta == OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)])
./astropy/table/tests/test_operations.py:1334:        self.t1.meta.update(OrderedDict([('b', [1, 2]), ('c', {'a': 1}), ('d', 1)]))
./astropy/table/tests/test_operations.py:1335:        self.t2.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:1336:        self.t4.meta.update(OrderedDict([('b', [5, 6]), ('c', {'c': 1}), ('e', 1)]))
./astropy/table/tests/test_operations.py:1337:        self.t5.meta.update(OrderedDict([('b', 3), ('c', 'k'), ('d', 1)]))
./astropy/table/tests/test_operations.py:1338:        self.meta_merge = OrderedDict([('b', [1, 2, 3, 4, 5, 6]),
./astropy/table/tests/test_operations.py:1497:        meta1 = OrderedDict([('b', [1, 2]), ('c', {'a': 1}), ('d', 1)])
./astropy/table/tests/test_operations.py:1502:        t3['d'].info.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/test_operations.py:1503:        t4['g'].info.meta.update(OrderedDict([('b', [5, 6]), ('c', {'c': 1}), ('e', 1)]))
./astropy/table/tests/test_operations.py:1504:        t3['e'].info.meta.update(OrderedDict([('b', [3, 4]), ('c', {'b': 1}), ('a', 1)]))
./astropy/table/tests/conftest.py:13:from collections import OrderedDict
./astropy/table/tests/conftest.py:164:    cols = OrderedDict()
./astropy/table/tests/test_table.py:9:from collections import OrderedDict
./astropy/table/tests/test_table.py:2124:            self.meta = OrderedDict([('a', 1), ('b', 2)])
./astropy/table/tests/test_table.py:2165:        meta = OrderedDict([('c', 3)])
./astropy/table/tests/test_table.py:2197:    assert t2.meta is not t.meta  # NOT the same OrderedDict object but equal
./astropy/table/tests/test_table.py:2203:    assert t2.meta is not t.meta  # NOT the same OrderedDict object but equal
./astropy/table/tests/test_table.py:2209:    assert t2.meta is not t.meta  # NOT the same OrderedDict object but equal
./astropy/table/tests/test_table.py:2816:    assert isinstance(t.values(), type(OrderedDict().values()))
./astropy/table/tests/test_table.py:2817:    assert isinstance(t.columns.values(), type(OrderedDict().values()))
./astropy/table/tests/test_table.py:2818:    assert isinstance(t.columns.keys(), type(OrderedDict().keys()))
./astropy/table/tests/test_table.py:2829:    assert isinstance(t.items(), type(OrderedDict({}).items()))
./astropy/table/tests/test_init_table.py:3:from collections import OrderedDict, UserDict
./astropy/table/tests/test_init_table.py:300:class TestInitFromOrderedDict(BaseInitFromDictLike):
./astropy/table/tests/test_init_table.py:303:        self.data = OrderedDict([('a', Column(name='x', data=[1, 3])),
./astropy/table/tests/test_init_table.py:540:def test_init_from_row_OrderedDict(table_type):
./astropy/table/tests/test_init_table.py:541:    row1 = OrderedDict([('b', 1), ('a', 0)])
./astropy/table/np_utils.py:9:from collections import OrderedDict, Counter
./astropy/table/np_utils.py:72:    col_name_map = OrderedDict((name, col_name_map[name]) for name in col_name_list)
./astropy/table/row.py:4:from collections import OrderedDict
./astropy/table/row.py:51:            out = OrderedDict.__getitem__(self._table.columns, item)[self._index]
./astropy/table/serialize.py:4:from collections import OrderedDict
./astropy/table/serialize.py:254:class _TableLite(OrderedDict):
./astropy/table/info.py:29:    column as its single argument.  The function must return an OrderedDict
./astropy/table/soco.py:8:from collections import OrderedDict
./astropy/table/soco.py:132:        result = OrderedDict()
./astropy/table/operations.py:15:from collections import OrderedDict, Counter
./astropy/table/operations.py:382:    col_name_map = OrderedDict()
./astropy/table/operations.py:635:    col_name_map = OrderedDict()
./astropy/table/operations.py:706:    col_name_map = OrderedDict()
./astropy/table/operations.py:898:    col_name_map = OrderedDict((name, col_name_map[name]) for name in col_name_list)
./astropy/table/operations.py:1268:        col_name_map = OrderedDict((name, in_names) for name, in_names in col_name_map.items()
```

**tests.runner**

```
./astropy/tests/runner.py:12:from collections import OrderedDict
./astropy/tests/runner.py:108:        cls.keywords = OrderedDict()
```

**time**

```
./astropy/time/formats.py:10:from collections import OrderedDict, defaultdict
./astropy/time/formats.py:38:# Use an OrderedDict to fix the order in which formats are tried.
./astropy/time/formats.py:40:TIME_FORMATS = OrderedDict()
./astropy/time/formats.py:41:TIME_DELTA_FORMATS = OrderedDict()
```

**utils**

These are ignored on purpose:
```
./astropy/utils/tests/test_timer.py:52:    p.time_func(2.2)  # Test OrderedDict
./astropy/utils/tests/test_metadata.py:3:from collections import OrderedDict
./astropy/utils/tests/test_metadata.py:14:class OrderedDictSubclass(OrderedDict):
./astropy/utils/tests/test_metadata.py:28:                                         OrderedDict([('a', 1)]),
./astropy/utils/tests/test_metadata.py:29:                                         OrderedDictSubclass([('a', 1)])]))
./astropy/utils/tests/test_metadata.py:41:                                         OrderedDict([('a', 1)]),
./astropy/utils/tests/test_metadata.py:42:                                         OrderedDictSubclass([('a', 1)])]))
./astropy/utils/timer.py:7:from collections import OrderedDict
./astropy/utils/timer.py:156:        self._cache_good = OrderedDict()
./astropy/utils/timer.py:158:        self._cache_est = OrderedDict()
./astropy/utils/timer.py:159:        self._cache_out = OrderedDict()
./astropy/utils/timer.py:255:        self._cache_est = OrderedDict()
```

**visualization.wcsaxes**

```
./astropy/visualization/wcsaxes/coordinates_map.py:4:from collections import OrderedDict
./astropy/visualization/wcsaxes/coordinates_map.py:174:            row = OrderedDict([('index', icoord), ('aliases', ' '.join(aliases)),
./astropy/visualization/wcsaxes/frame.py:5:from collections import OrderedDict
./astropy/visualization/wcsaxes/frame.py:149:class BaseFrame(OrderedDict, metaclass=abc.ABCMeta):
./astropy/visualization/wcsaxes/frame.py:215:        spines = OrderedDict()
```

**wcs.wcsapi**
```
./astropy/wcs/wcsapi/high_level_api.py:2:from collections import defaultdict, OrderedDict
./astropy/wcs/wcsapi/high_level_api.py:138:        classes = OrderedDict()
```

astropy/table/meta.py
```diff
@@ -47,105 +45,6 @@ def items(self):
         return ColumnOrderList(super().items())
 
 
-def _construct_odict(load, node):
```
mhvk, line 50: Is this OK? If someone has an old yaml file, will it still load?

mhvk:
I think spreading it over multiple PRs seems wise. That way test failures are easier to check/understand, and we also ensure everything gets properly looked at. Plus, we can share the load!

astropy/table/meta.py
pllim: I think so? There was a failing test trying to load something in `coordinates` but it was fixed with the weird patches I pointed out above.

pllim:
@mhvk , in that case, do you think the current state of the PR is atomic enough? Or should I include some other stuff so that the changes in `utils` make sense?

mhvk:
@pllim - I think this PR is quite nicely atomic, just focussing on `meta` data, so let's keep it as is. As you note, though, some of the changes seem a bit odd. @taldcroft - `meta` is a bit more up your alley, would you be able to have a look?

pllim:
Thanks, @mhvk ! I'm limiting the scope of this PR to `utils` and the things that directly affected and mark this as ready for review. Also rebased.

pllim:
RTD timeout is unrelated.

mhvk:
@taldcroft - could you review this PR? I am not familiar enough with the test cases to be confident that we wouldn't accidentally break loading of old yaml files, etc.

taldcroft:
@pllim @mhvk - there appear to be some issues with this PR and legacy files.  For instance:
```
from collections import OrderedDict
t = simple_table()
t.meta['a'] = OrderedDict()
t.meta['a']['b'] = 1
t.meta['a']['a'] = 2
qt = QTable(t)
qt.write('qtable.ecsv')
```
Running that in astropy LTS generates this ECSV file:
```
# %ECSV 0.9
# ---
# datatype:
# - {name: a, datatype: int64}
# - {name: b, unit: m, datatype: float64}
# - {name: c, datatype: string}
# meta: !!omap
# - a: !!omap
#   - {b: 1}
#   - {a: 2}
# - __serialized_columns__:
#     b:
#       __class__: astropy.units.quantity.Quantity
#       unit: !astropy.units.Unit {unit: m}
#       value: !astropy.table.SerializedColumn {name: b}
# schema: astropy-2.0
a b c
1 1.0 c
2 2.0 d
3 3.0 e
```
However, reading that back using this PR gives the wrong value for `qt.meta['a']`:
```
In [20]: qt = QTable.read('qtable.ecsv')

In [21]: qt.meta['a']                                                                                                                                                                 
Out[21]: [('b', 1), ('a', 2)]

In [22]: type(qt.meta['a'])                                                                                                                                                           
Out[22]: list
```
In LTS this works as expected:
```
In [14]: qt = QTable.read('qtable.ecsv')                                                                                                                                              

In [15]: qt.meta['a']                                                                                                                                                                 
Out[15]: OrderedDict([('b', 1), ('a', 2)])
```

taldcroft:
Another problem seems to be more fundamental, namely that the ordering of `meta` seems to be lost:
```
In [13]: t.meta                                                                                                                                                                       
Out[13]: {'b': [1, 2], 'a': 0}

In [14]: t.write('table.ecsv', overwrite=True)                                                                                                                                        

In [15]: cat table.ecsv                                                                                                                                                               
# %ECSV 0.9
# ---
# datatype:
# - {name: a, datatype: int64}
# - {name: b, datatype: float64}
# - {name: c, datatype: string}
# meta:
#   a: 0
#   b: [1, 2]
# schema: astropy-2.0
a b c
1 1.0 c
2 2.0 d
3 3.0 e
```
But this is actually about pyyaml, I think we need to get a `sort_keys=False` in the loop somewhere (e.g. https://stackoverflow.com/a/55858112/323631).

taldcroft:
We also need to be concerned about someone generating an ECSV file with new astropy and then someone else trying to read it with LTS.  Which is to say that this is all a little bit tricky and will either (a) require some more careful thought, or (b) require leaving `omap` and maybe the OrderedDict.  Not sure!

taldcroft:
I don't have time today to look at this more, but I'm wondering if it would be possible to represent a plain `dict` as `!!omap` in ECSV, and leave in all the handling for OrderedDict. That might solve most of the back/forward compatibility issues. An ECSV generated with astropy 4.3 would turn dict into omap. Astropy 4.3 would read that as dict, but astropy < 4.3 would read it as OrderedDict.

What I'm not sure about are whether it is possible in YAML to override how `dict` is represented. I'm guessing it should be fine, but would need to test.

mhvk:
Yes, some trickery to make sure that the `yaml` output does not actually change would seem best, otherwise we have to move to a new ecsv version number (which I think would be overkill just to get rid of a class that still exists!). 

@pllim - For now, I converted the PR to a draft since I fear this needs real thought (so good that we split it off from the easier parts!)

taldcroft:
> I'm wondering if it would be possible to represent a plain dict as !!omap

First try at that failed with a max recursion error. That might just not be possible.

taldcroft:
I'm a little stumped, and having thought more, generically turning dict into `!!omap` is probably a bad idea anyway.

One end-run to use the native `dict` without changing much of the machinery is to just change the definition of `OrderedDict` from the version in collections to `class OrderedDict(dict): pass`. In practice all this is doing is tagging certain objects with a class so that they get `!!omap` in YAML.

Another option is just to go to plain `dict` and bump the ECSV version. LTS 4.0.x requires Python 3.6+, which for CPython 3.6 does actually have an ordered `dict` class, even if it is not in the language spec. So an LTS user that gets an ECSV file from a user running astropy 4.3 won't actually have a problem if they are running CPython 3.6+. I don't know about PyPy, but astropy does not formally support that anyway.

I guess I like the 2nd option but I'm not adamant.

taldcroft:
@pllim @mhvk - looking at the code more I realized another big chunk was just dealing with the old unordered dict, so I cut that out, yay.

I was not able to locally reproduce the test failures that led to the two "weird patches", so I reverted those. If they come up in CI then I can investigate.

taldcroft:
Here is a regression test to check that writing a complex table with astropy 4.3 and reading it back in astropy 3.0 gives the same answer. For some reason the `unit` is not being explicitly written out, presumably because `deg` is the default and isn't provided in some kwargs along the way.

Anyway, I *think* this is good enough to say that files written in astropy 4.3 will read back correctly in old astropy with Python 3.6.

### Astropy 4.3 dev + Py 3.7
```
In [3]: t = QTable.read('astropy/coordinates/tests/accuracy/data/rv.ecsv')
In [4]: t.write('rv43.ecsv')   
```
### Astropy 3.0.3 + Py 3.6
```
In [19]: pwd
Out[19]: '/Users/aldcroft'

In [20]: t43 = QTable.read('/Users/aldcroft/git/astropy/rv43.ecsv')

In [21]: t43.info
Out[21]: 
<QTable length=50>
   name    dtype    unit   class  
--------- ------- ------- --------
   target  object deg,deg SkyCoord
  obstime  object             Time
   obslon float64     deg    Angle
   obslat float64     deg    Angle
  geocent float64           Column
heliocent float64           Column
     lsrk float64           Column
     lsrd float64           Column
 galactoc float64           Column
 localgrp float64           Column

In [23]: t43.write('git/astropy/rv43_30.ecsv')
In [24]: !opendiff git/astropy/rv43_30.ecsv git/astropy/astropy/coordinates/tests/accuracy/data/rv.ecsv
```
image

astropy/utils/metadata.py
```diff
@@ -398,20 +397,21 @@ def __get__(self, instance, owner):
...
-        else:
-            if isinstance(value, Mapping):
-                if self.copy:
-                    instance._meta = deepcopy(value)
-                else:
-                    instance._meta = value
+            instance._meta = dict()
+            return
+
+        if isinstance(value, Mapping):
```
taldcroft, line 408: This code is the same as the original except OrderedDict => dict and doing a `return` instead of an `else` block.

taldcroft (approved):
This looks good to me now, but I'm not quite an independent reviewer any more. Need review from @mhv as well.

pllim:
Error seems real with oldest deps. Does this mean we need to bump oldest supported PyYAML version too?
```
TypeError: dump_all() got an unexpected keyword argument 'sort_keys'
```

taldcroft:
> Does this mean we need to bump oldest supported PyYAML version too?

Yeah, I'd say so.  From https://github.com/yaml/pyyaml/blob/master/CHANGES it looks like we need version 5.1, which is from 2019-03-13.

mhvk (commented):
This looks great! Code-wise, two nitpicky comments, and one small query about sorting keys.

Also, I do think/fear we should be bumping the ECSV version number: even if astropy can just read the files back, other code may have their expectations about the presence of `!!omap` (and the corresponding `-` at the start of the next line). And this expectation would be reasonable, as `!!omap` is explicitly mentioned in APE 6.

Less sure whether this would be 0.9.1 or 1.0. Since it is not really a breaking change, perhaps 0.9.1 makes most sense? In the end, what changes from the file format perspective is that we stress that the *expectation* is that for a mapping, order will be preserved (in principle, `meta` could always have been anything). Still, that new expectation deserves a version change, I think...

pllim:
@taldcroft and @mhvk , I'll let you two finalize things here. While this is technically my PR, the implementation details are out of my depths now. If you don't get around to bumping minversion of PyYAML, I'll do it, but please ping me to coordinate, so we don't clobber each other's work. Thank you so much!

astropy/table/meta.py
```diff
@@ -236,59 +127,12 @@ def get_yaml_from_header(header):
...
-                else:
-                    node.flow_style = best_style
-            return node
-
-    TableDumper.add_representer(OrderedDict, _repr_odict)
-    TableDumper.add_representer(ColumnDict, _repr_column_dict)
-
-    header = copy.copy(header)  # Don't overwrite original
     header['datatype'] = [_get_col_attributes(col) for col in header['cols']]
+    header = {key: header[key] for key in sorted(header)}
```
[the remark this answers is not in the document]
taldcroft: It doesn't *need* to be sorted, but it does keep the ECSV header consistent with past behavior and also tidier going forward. I.e. if we change the code so that keys get added in a different order then the output will be the same. I noticed this of course because of a regression fail without the `sorted`.

taldcroft:
Looking back into the YAML spec, I guess we need to ask the hard question about whether it is really OK to use a plain mapping which we are taking to be ordered because the Python (PyYAML) implementation supports it. The YAML spec says explicitly that mapping keys are not ordered and if you need them to be ordered then use something like `!!omap`: https://yaml.org/spec/1.2/spec.html#id2765608.

astropy/table/meta.py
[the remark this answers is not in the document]
mhvk: OK, I mostly wondered if it was already done automatically, but you had already checked. Agree there is no need to change the behaviour.

mhvk:
Ouch!  I'd tend to agree with you that we should follow that specification, especially as it really leaves no doubt as to the intent.

taldcroft:
@mhvk - agreed (grumpily).  

This points to my original plan A of making a local version of ordered dict with `class OrderedDict(dict): pass` and using that for the default container for Meta. At least this will save the messiness of changing the ECSV version and/or schema. Using `dict` as the base class will make the meta object behave and look much more like a regular dict, so there is at least some benefit from all this.

taldcroft:
@pllim - we do still need to bump the min version of PyYAML. Where to do that?

pllim:
@taldcroft , I identified the following files that needs updating from a global `grep`. There are multiple occurrences, so I recommend a careful search and replace in those files. Might have missed something but hopefully the tests will smoke them out. Will also need a change log entry for this. 

```
./astropy/io/misc/yaml.py:.. note:: This module requires PyYaml version 3.13 or later.
./astropy/io/misc/yaml.py:    raise ImportError('The PyYAML package is required for astropy.io.misc.yaml and must be 3.13 or later')
./astropy/io/misc/yaml.py:        raise ImportError('The PyYAML package is required for astropy.io.misc.yaml and must be 3.13 or later')
./astropy/__init__.py:# PyYAML is an optional dependency, but this is the minimum version that is
./setup.cfg:    PyYAML>=3.13
./tox.ini:    oldestdeps: pyyaml==3.13
```

pllim:
Hmm, there are conflicts now. I forgot where we were with this PR... Are there still things I need to follow up on, or it would be ready for merge once I rebase?

astropy/utils/metadata.py
pllim: Is this just a note or something I still have to do? I don't quite remember. 😬

pllim:
Oh no.... I think I clobbered Tom's edit. Good thing there is a hash in the discussions here. Lemme see if I can undo my own damage.

pllim:
@taldcroft and @mhvk -- I removed conflicts, bumped PyYAML minversion, and addressed the remaining comments. Please review carefully one more time. And if it looks good, let's get this in before I get more conflicts. Thanks!

taldcroft:
@pllim - sorry, but there is a real problem with the current implementation, see https://github.com/astropy/astropy/pull/10932#issuecomment-729316348 and https://github.com/astropy/astropy/pull/10932#issuecomment-729587756. It's been on my list of things to do but I haven't quite been looking forward to it. :smile:

pllim:
Re: https://github.com/astropy/astropy/pull/10932#issuecomment-772468495 -- Ah, sorry; I thought by bumping PyYAML minversion, that would be enough but I had not been following the convo very closely. I'll turn this PR back to draft then. Thanks for the update!

pllim:

pllim:
Oh, wait this is my own PR. Hmm too many conflicts now. I am just going to abandon it.

PR #10931, opened 2020-10-23
Time fast reader bigendian fix

mhvk:
Fixes #10930 

This should work on big-endian. Marking it as a draft only because once it succeeds, the `s390` run should be taken out of the regular CI runs again (or should it? but let's leave that for another issue...)

Note that following https://github.com/astropy/astropy/issues/10930#issuecomment-715578126, I've replaced the ascii C check with a faster numpy one.

PR #10933, opened 2020-10-24
Upgrade doctestplus on travis using pip

mhvk:
The version in focal is too old and the deprecation warning is not
consistently caught.  So, now we can also remove its exception from
setup.cfg.

Follow-up of a comment of @pllim in #10908. Not sure why earlier I thought this didn't work.

Note: not running the tests again, since it all passed in https://travis-ci.com/github/astropy/astropy/builds/192443455
This PR does the changes to travis/setup only, since it can be backported, while #10931 will get only the changes to the time fast reader, which are for 4.2 only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -272,6 +272,7 @@ install:
       fi
 
 script:
+    # for apt, doctestplus on focal is rather out of date, so get new version.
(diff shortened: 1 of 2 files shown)


pllim (approved):
LGTM. Thanks!

PR #10934, opened 2020-10-24
More ufuncs work with quantities + test #6390

### LINKED ISSUE #6390: multiple functions from scipy.special do not work with quantities
I just tried to use SciPy's error function, and Astropy politely requested that I raise this issue here:
```Python
>>> from astropy import units
>>> from scipy.special import erf, erfc
>>> r1 = 5*u.m
>>> r2 = 6*u.m
>>> erf(r1/r2)
TypeError: Unknown ufunc erf.  Please raise issue on https://github.com/astropy/astropy
>>> erfc(r1/r2)
TypeError: Unknown ufunc erfc.  Please raise issue on https://github.com/astropy/astropy
>>> erf(r1.value/r2.value)  # workaround
0.76140717068356445
```
In scipy.special, ufuncs that I would prioritize include:
- [ ] [Error function, etc](https://docs.scipy.org/doc/scipy/reference/special.html#error-function-and-fresnel-integrals)
- [ ] [Bessel functions](https://docs.scipy.org/doc/scipy/reference/special.html#bessel-functions)
- [ ] [Gamma and related functions](https://docs.scipy.org/doc/scipy/reference/special.html#gamma-and-related-functions)
- [ ] [Legendre functions](https://docs.scipy.org/doc/scipy/reference/special.html#legendre-functions)
- [ ] `zeta` and `zetac` - Riemann zeta functions

Thanks!

DiegoAsterio:
### Description

This pull request is to address multiple functions from scipy.special do not work with quantities

EDIT: reworded to ensure we don't close the issue: ~Fixes~ addresses #6390 partially

### DIFF
--- a/astropy/units/quantity_helper/scipy_special.py
+++ b/astropy/units/quantity_helper/scipy_special.py
@@ -5,6 +5,7 @@
 Available ufuncs in this module are at
 https://docs.scipy.org/doc/scipy/reference/special.html
 """
+import numpy as np
 
 from astropy.units.core import UnitsError, UnitTypeError, dimensionless_unscaled
 from . import UFUNC_HELPERS
@@ -16,12 +17,12 @@ from .helpers import (get_converter,
 
 # ufuncs that require dimensionless input and give dimensionless output.
 dimensionless_to_dimensionless_sps_ufuncs = (
-    'erf','erfc', 'erfcx', 'erfi', 'erfinv', 'erfcinv', 'gamma', 'gammaln',
-    'loggamma', 'gammasgn', 'psi', 'rgamma', 'digamma',
+    'erf', 'erfc', 'erfcx', 'erfi', 'erfinv', 'erfcinv',
+    'gamma', 'gammaln', 'loggamma', 'gammasgn', 'psi', 'rgamma', 'digamma',
     'wofz', 'dawsn', 'entr', 'exprel', 'expm1', 'log1p', 'exp2', 'exp10',
     'j0', 'j1', 'y0', 'y1', 'i0', 'i0e', 'i1', 'i1e',
-    'k0', 'k0e', 'k1', 'k1e', 'itj0y0', 'it2j0y0', 'iti0k0', 'it2i0k0'
-    ,'ndtr', 'ndtri')
+    'k0', 'k0e', 'k1', 'k1e', 'itj0y0', 'it2j0y0', 'iti0k0', 'it2i0k0',
+    'ndtr', 'ndtri')
 
 
 scipy_special_ufuncs = dimensionless_to_dimensionless_sps_ufuncs
@@ -66,8 +67,9 @@ def get_scipy_special_helpers():
     import scipy.special as sps
     SCIPY_HELPERS = {}
     for name in dimensionless_to_dimensionless_sps_ufuncs:
+        # In SCIPY_LT_1_5, erfinv and erfcinv are not ufuncs.
         ufunc = getattr(sps, name, None)
-        if ufunc:
+        if isinstance(ufunc, np.ufunc):
             SCIPY_HELPERS[ufunc] = helper_dimensionless_to_dimensionless
 
     for ufunc in degree_to_dimensionless_sps_ufuncs:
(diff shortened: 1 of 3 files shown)


astropy/units/quantity_helper/scipy_special.py
```diff
@@ -16,11 +16,13 @@
 
 # ufuncs that require dimensionless input and give dimensionless output.
 dimensionless_to_dimensionless_sps_ufuncs = (
-    'erf', 'gamma', 'gammasgn', 'psi', 'rgamma', 'erfc', 'erfcx', 'erfi',
+    'erf','erfc', 'erfcx', 'erfi', 'erfinv', 'erfcinv', 'gamma', 'gammaln',
+    'loggamma', 'gammasgn', 'psi', 'rgamma', 'digamma',
```
mhvk, line 20: I'm not quite sure about `gammaln` and `loggamma` - arguably, since they return the natural logarithm, perhaps they should return a functional log-based unit? (Would be a new one, since we have only `dex` for log-10 based...). But then, again `np.log(dimensionless)` gives back dimensionless, so probably best to let them go in...

`rgamma` is fine, as are `digamma`

mhvk (changes requested):
This looks good - there is an inline comment, but mostly there as I was wondering about it. In the end, OK as is.

The one thing to do would be to add an entry in `CHANGES.rst`. As the deadline for 4.2 just passed, this would go in the 4.3 release, but the entries for that have not yet been added, so maybe best to wait a bit until that is the case, then rebase and add the entry.

DiegoAsterio:
Changed CHANGES.srt

DiegoAsterio:
I am running the test on Travis CI and I get two errors:
1. The converters_and_units test function when applied to erfinv displays the following error:
```
astropy.units.core.UnitConversionError: Can only apply 'add' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)
```
2. The test_exp_invalid_units function when applied to erfcinv 
```
AssertionError: assert 'Can only app...ss quantities' == 'Can only app...ss quantities'

E         - Can only apply 'erfcinv' function to dimensionless quantities

E         ?                 ^ -- --

E         + Can only apply 'ndtri' function to dimensionless quantities

E         ?                 ^^^
```
I do not fully understand the first error because it is not covered in the docstring.
The second error seems a bit more comprehensible. As ndtri is called inside erfcinv the error refers to ndtri hence the two strings are not the same.

mhvk:
@DiegoAsterio - it looks like both `erfinv` and `erfcinv` are not pure C based ufuncs, but rather some python version. I'm quite surprised about that, since there clearly is C code at https://github.com/scipy/scipy/blob/4e8635247518d9ca14340cbc33d1338b98ed8066/scipy/special/cephes/erfinv.c#L46. I will investigate... (only have a bit of time now, so if it isn't something obvious, this may take a while)

mhvk:
@DiegoAsterio - the PR does seem to work on recent scipy, but with the oldestdeps I cannot even seem to install locally... But we just updated our minimum requirements, so I've gone ahead and rebased on the current master, force-pushing to your branch. If it was something related to scipy moving from python to C code, that should solve it... :crossed_fingers:

astrojuanlu:
I rebased this again on top of current `master`, since the infrastructure has changed quite a bit.

mhvk:
`erfinv` and `erfcinv` still fail on our oldest scipy. I think we need a version check for their inclusion -- using something like
```
from astropy.utils import minversion

SCIPY_LT_X.X = minversion('scipy', x.x)

if not SCIPY_LT_X.X:
    dimensionless_to_dimensionless_ufuncs += ('erfinv', 'erfcinv')
```
where X.X is the scipy version which introduced a proper ufunc for those.

mhvk:
p.s. Of course, we should also check if we could not just increase the minimum version...

mhvk:
Rebased again. Probably should just increase minimum scipy version... (if still needed)

eteq:

mhvk:
Hmm, the milestone move reminded me of this one... I rebased and added the version check that was missing here. I'll approve though obviously the tests still have to pass.

mhvk (approved):
All OK now.

mhvk:
p.s. While listed under "bugfix", I think it is enough of just an enhancement for recent scipy that I don't think it needs back-porting.

mhvk:
Thanks, @DiegoAsterio - even if it took quite a while!

PR #10946, opened 2020-10-27
Skip setup_package files when running generate_config

saimn:
Follow-up to the recent issue with  abspath in setup package (#10927).

### DIFF
--- a/astropy/config/tests/test_configs.py
+++ b/astropy/config/tests/test_configs.py
@@ -147,13 +147,7 @@ def test_config_file():
 def test_generate_config(tmp_path):
     from astropy.config.configuration import generate_config
     out = io.StringIO()
-    try:
-        generate_config('astropy', out)
-    except ValueError as err:
-        if sys.platform.startswith('win') and 'path is on mount' in str(err):
-            pytest.xfail("See Github issue 10747")
-        else:
-            raise
+    generate_config('astropy', out)
     conf = out.getvalue()
 
     outfile = tmp_path / 'astropy.cfg'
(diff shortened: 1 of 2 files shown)


astropy/config/tests/test_configs.py
```diff
@@ -147,13 +147,7 @@ def test_config_file():
...
     from astropy.config.configuration import generate_config
     out = io.StringIO()
-    try:
-        generate_config('astropy', out)
-    except ValueError as err:
-        if sys.platform.startswith('win') and 'path is on mount' in str(err):
-            pytest.xfail("See Github issue 10747")
-        else:
-            raise
+    generate_config('astropy', out)
```
pllim, line 150: Need to also remove the try/except from `test_generate_config2` below.

pllim (commented):
This is fine (after you address the comment about test), as there are no `Conf` defined in `setup_package`. However, in the future, if somehow someone is compelled to define a configuration object inside `setup_package.py`, it will be ignored by `generate_config`. I am wondering if there is a way we can tell people not to do that.

astropy/config/tests/test_configs.py
saimn: I missed that one, thanks.

saimn:
There is really no reason to add a `Conf` in `setup_package`, since `setup_package` files are executed only during setup.

pllim (approved):
LGTM. Thanks!

pllim:
I changed the labels a bit because this is still technically a bug fix but as a direct follow-up of #10927.

PR #10947, opened 2020-10-27
DOC: Explain why bounding box test only for Matplotlib 3.3

pllim:
### Description

This pull request is to address https://github.com/astropy/astropy/commit/361f723ec00ea7dfdc08f4d5bcf48c8e63f8d0dc#r43603162 from #10945 , which was a spin-off from #10900 .


dhomeier:
Would probably do if the "just to be safe" provides enough information.
Though frankly, since the test has always been passing for all mpl3* versions, and still is per #10952, I still do not see the point in rationalising about the wrong label and further restricting the test rather then correcting the label. ;-)

dhomeier:
So if @Cadair agrees I suggest to use #10952 instead.

pllim:
Okay, it is fine by me! Closing in favor of  #10952

Issue #1759, opened 2013-11-09
Support operator= for scalar Quantities


aconley:
Right now things like

```
>>> import astropy.units as u
>>> ex1 = u.Quantity([1, 2, 3], u.m)
>>> ex1 *= 0.1
```

work, but

```
>>> ex2 = u.Quantity(1, u.m)
>>> ex2 *= 0.01
```

does not.  It would be nice to allow both.

Note that these work fine if you change them to 1.0 instead of 1.  But for some reason (unlike numpy scalar arrays), trying this on u.Quantity integer objects throws a TypeError.

PR #10950, opened 2020-10-27
BUG: fix QTable Column to Quantity dtype loss

### LINKED ISSUE #10943: QTable changes dtype when converting column to Quantity
### Description
When `QTable` converts a column to `Quantity`, it often changes the `dtype`, which would be unexpected.

Version affected: observed in `4.1rc1`. I believe it has been like that.

### Expected behavior
The `dtype` should be preserved after the `Quantity` conversion

### Actual behavior
`dtype` becomes `float64`, which I believe is the default for `Quantity`

### Steps to Reproduce

Example code:
```python
ts = astropy.table.QTable(dict(time=[1, 2, 3]))

ts['cadenceno'] = [11, 12, 13]
print(ts['cadenceno'].dtype)  # int32
ts['cadenceno'].unit = u.dimensionless_unscaled
print(ts['cadenceno'].dtype)  # float64
```

The conversion from `int32` to `float64` happens irrespective of the specific unit being assigned, e.g., `u.pixel`.

### System Details

```
Windows-10-10.0.19041-SP0
Python 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 01:53:57) [MSC v.1916 64 bit (AMD64)]
Numpy 1.19.1
astropy 4.1rc1
Scipy 1.5.2
Matplotlib 3.1.0
```

orionlee:
### Description

This pull request is to address the bug that for `QTable`, when a column is converted from `Column` to `Quantity`, the column 's `dtype` is changed in some cases, e.g., columns of `int` dtype is incorrectly changed to `float64`.

Fixes #10943

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -403,6 +403,9 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
+- Fixed a bug when a column in ``QTable`` is converted to ``Quantity``,
+  datatype is sometimes not preserved. [#10950]
+
 astropy.tests
 ^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -403,9 +403,6 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
-- Fixed a bug when a column in ``QTable`` is converted to ``Quantity``,
-  datatype is sometimes not preserved. [#10950]
-
 astropy.tests
 ^^^^^^^^^^^^^
 
@@ -988,6 +985,9 @@ astropy.stats
 
 astropy.table
 ^^^^^^^^^^^^^
+
+- Fixed a bug when an integer-type ``Column`` in ``QTable`` is converted to
+  ``Quantity`` by setting the unit, the datatype is changed to float. [#10950]
 
 astropy.tests
 ^^^^^^^^^^^^^
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3648,7 +3648,7 @@ class QTable(Table):
             # We need to turn the column into a quantity, or a subclass
             # identified in the unit (such as u.mag()).
             q_cls = getattr(col.unit, '_quantity_class', Quantity)
-            qcol = q_cls(col.data, col.unit, copy=False)
+            qcol = q_cls(col.data, col.unit, dtype=col.dtype, copy=False)
             qcol.info = col.info
             qcol.info.indices = col.info.indices
             col = qcol
--- a/astropy/table/tests/test_table.py
+++ b/astropy/table/tests/test_table.py
@@ -2278,6 +2278,15 @@ def test_replace_column_qtable():
     assert t['a'].info.format is None
 
 
+@pytest.mark.parametrize("col_vals", [[1.1, 2.2], [3, 4]])
+def test_column_quantity_conversion_qtable(col_vals):
+    """Ensure dtype is preserved in ``QTable`` ``Column`` to ``Quantity`` conversion.  #10943"""
+    t = table.QTable([col_vals], names=['a'])
+    exp_dtype = t['a'].dtype
+    t['a'].unit = u.m
+    assert t['a'].dtype is exp_dtype
+
+
 def test_replace_update_column_via_setitem():
     """
     Test table update like ``t['a'] = value``.  This leverages off the


CHANGES.rst
```diff
@@ -403,6 +403,9 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
+- Fixed a bug when a column in ``QTable`` is converted to ``Quantity``,
```
taldcroft, line 406: This can be backported to the LTS 4.0.4 version, so it should go in the 4.0.4 section.

CHANGES.rst
```diff
@@ -403,6 +403,9 @@ astropy.stats
...
+  datatype is sometimes not preserved. [#10950]
```
taldcroft, line 407: I think you can be more specific here. At least from my thinking about it, the only case of dtype not being preserved is when the original column is an integer type. So you could say:
```
Fixed a bug when an integer-type ``Column`` in ``QTable`` is converted to ``Quantity``
by setting the unit, the datatype was changed to float. [#10950]
```

taldcroft (changes requested):
@orionlee - thanks! This looks great, the only comment is just adjustments to the change log entry. When you make the commit you can add `[skip ci]` at the end of the commit message so we don't burn CI time needlessly on testing.

CHANGES.rst
orionlee: Feedback incorporated.

BTW, do you happen to know the reason(s) why `QTable` performs `Quantity` conversion only for columns with units, rather than all numeric columns? As it is, users would need to deal with 2 different APIs (`Quantity` and `Column`) when using `QTable` to handle with a mix of numeric data with and without unit, e.g., flux, cadence, quality flags.

mhvk:
@orionlee - thanks for the fix. Before just going for it, it is probably good to think through the consequences: if a `Quantity` is `int`, any in-place multiplication, division or unit change (e.g., 'm' to 'km') will lead to truncation. Indeed, this is the reason that generally we always convert to float - unless one works at the quantum level (say, electron spin), integers with units just do not make all that much physical sense. Hence, I worry that this change is going to cause more problems than it solves - it seems quite likely someone's code out there is relying on the existing behaviour...

The answer to your more general question is related: I think one should only use `Quantity` for physically meaningful items where a unit (possibly dimensionless) makes sense. So it seems to me there is a clear place for `Column` to hold items like flags or strings, etc., where quantity methods like `.to()` or numpy overrides like for `np.sin` are meaningless (but which provides a way to hold metadata, etc.).

orionlee:
@mhvk I think I can understand your concern about integer truncation for in-place operations. 

However, I am not aware of any in-place operations for `Quantity`. E.g., `some_quantity.to()` , `np.sin(some_quantity)`, `some_quantity / 2` all produce new `Quantity` objects, and the results are correctly in float. What am I missing here?

In the context of [Lightkurve](https://github.com/KeplerGO/lightkurve), our use case is that when users handle the numeric columns, e.g., physical quantity `flux`, dimensionless metadata such as `cadenceno`, `quality`, we'd like to reduce users' (and our internal codes') burden in dealing with 2 different APIs (the most glaring one is accessing raw data via `.value` vs `.data`).

If automatic float conversion is to be kept, it'd be desirable to a) document it and b) emit a warning about the conversion. Otherwise, the codes such as the following would unexpectedly fail, and baffle the users.
```python
    ts['quality'] = read_quality_column()  # e.g., [1, 0, 3, 8, ...]
    ts['quality'].unit = u.dimensionless_unscaled  # prefer to work with `Quantity` API
    quality_mask = ts['quality'].value & quality_bitmask  # unexpectedly failed, as ts['quality'].dtype is float64
```

mhvk:
@orionlee - I meant things like the following, but actually trying them I see that they already lead to `TypeError`, since numpy does not allow the in-place modification:
```
qt = QTable([u.Quantity([1033, 26], u.m, dtype=int)], names=['s'])
qt['s'] /= 5
# TypeError: No loop matching the specified signature and casting was found for ufunc true_divide
qt['s'] <<= u.km
# UFuncTypeError: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```
So, at least my worry that this PR would silently corrupt data analysis that previously worked is misplaced. 

Though of course this PR will still break existing code that does in-place division or multiplication and uses the existing Column to Quantity conversion. And expecting float by default may be quite baked in, since the present conversion is done implicitly in, e.g., reading FITS tables:
```
In [48]: Table([Column(np.arange(10), unit=u.m, name='s')]).write('t.fits')

In [49]: QTable.read('t.fits')
Out[49]: 
<QTable length=10>
   s   
   m   
float64
-------
    0.0
    1.0
    2.0
    3.0
    4.0
    5.0
    6.0
    7.0
    8.0
    9.0
```

mhvk:
Your example of assigning a unit and having the `dtype` change can indeed fairly be called suprising...

And it is indeed not great that the API differs - it's basically a historical artifact. The only thing that works on both is pass the column through `np.asarray`... (or `.view(np.ndarray)`). Though ideally of course one doesn't need to get the value out - the point of `Quantity` is not losing sight of the unit! (And I would still recommend not using Quantity for flags, etc....)

orionlee:
Having seen the various examples, I suppose it could be true that expecting float by default may be quite baked in.

1. So what's the reason for `Quantity` to support int types to begin with? From the discussion so far, it seems that using int with Quantity just leads to complication. For performance reason?
2. If we keep the existing behavior (converting to float during Column to Quantity conversion), what can be done in reducing surprises? Emitting a warning in `_convert_col_for_table()` would be be undesirable, as it would generate excessive warning when implicit conversion is done, say, during reading FITS file

mhvk:
@orionlee - supporting `int` comes basically by default, by being an `ndarray` subclass. At some point (#1759), it became clear this caused problems and in #1776 we changed the initializer to got for `float` unless an explicit dtype was given, but perhaps foolishly we allowed people to explicitly set an integer dtype if they really wanted to.

As for avoiding surprises, it is indeed tricky. Perhaps at least we could warn for the case where one sets a unit on an integer `Column`. Obviously, better documentation would also help, but it is difficult to think exactly where to put this so that it will be found by those who need it, but doesn't interfere for those who don't.

orionlee:
> Though ideally of course one doesn't need to get the value out - the point of Quantity is not losing sight of the unit!

In `Lightkurve`'s case,  we're forced to use raw values from time to time, primarily in plotting via `bokeh`.

mhvk:
@orionlee - OK, so indeed other packages. Interoperation between them is a hard problem!!

Aside: have you tried our relatively recent `Time` interface for matplotlib? See https://docs.astropy.org/en/latest/visualization/matplotlib_integration.html

orionlee:
1. Based on the conversation, it might be a good idea to deprecate use of integer `dtype` in `Quantity`.

2. I'm closing the PR and has filed a new issue #10964 to address how to make the current behavior less surprising to users.

PR #10953, opened 2020-10-27
Test refactored version of pytest-mpl

Cadair:
This is to test that https://github.com/matplotlib/pytest-mpl/pull/97 works IRL


Cadair:
well that seems to work :tada:

PR #10912, opened 2020-10-21
Fix biweight stats when MAD=0 and axis is input

larrybradley:
As demonstrated in #10820, biweight stat functions currently give inconsistent results when `median_absolute_deviation=0` and using an ``axis`` or not.  This PR fixes the issue for all cases when `MAD=0`, which occurs when the data are constant or mostly-constant.  When `MAD=0`:

* `biweight_location` returns the median
* `biweight_scale`, `biweight_midvariance`, and `biweight_midcovariance` return 0.0

Fixes #10820.

PR #10959, opened 2020-10-28
Fix biweight stats to return scalar values if axis is None

larrybradley:
PR #10912 introduced a regression where scalar values were no longer returned if `axis=None`.  This PR fixes the regression and adds several regression tests.


pllim (approved):
LGTM. Thanks!

PR #10962, opened 2020-10-29
Add `Column.value` as an alias for `Column.data` for consistency with `Quantity` and `Time`

### LINKED ISSUE #10859: Consider adding `__array__` method to `TimeBase` to return `value`?
### Description

From [this slack question](https://app.slack.com/client/T067V5AR2/C0LAMJCRE/thread/C0LAMJCRE-1602705646.012700), I was wondering about a protocol for `Time` that `np.array(tm)` or `np.asarray(tm)` would return `tm.value`. In theory this could apply to `SkyCoord`. Currently in both cases doing `np.array(..)` returns an array of `Time` or `SkyCoord` objects, which in practice is rarely what you wanted.

I guess one snag is if the Time value has missing values and `tm.value` returns a MaskedArray.

There is also the original suggestion of adding a `.value` attribute to `Column` to make it more like `Time` and `Quantity`. Again this could in theory apply to `SkyCoord`, where it would return e.g. an `N x 3` np.array.

This is definitely a question for @mhvk!

barentsen:
(Fixes #10859.)

### Description

This PR proposes to make the `Table` API more user-friendly by adding `Column.value` as an alias for `Column.data`, so that the underlying array can be accessed in a way that is consistent with the `Time` and `Quantity` classes, which provide `Time.value` and `Quantity.value` properties.

### Example

This change would particularly benefit the user experience when dealing with tables which contain a mix of `Time`, `Quantity`, `Column` and/or `MaskedColumn` fields. This will be common in `TimeSeries` objects.  For example, I frequently make the following mistake at present:

```python
In [1]: from astropy.timeseries import TimeSeries
   ...: from astropy.time import Time
   ...: from astropy.units import Quantity
   ...: from astropy.table import Column
   ...:
   ...: ts = TimeSeries({'time': Time([50000, 50001, 50002], format='mjd'),
   ...:                  'flux': Quantity([1., 2., 3.], unit='electron/s'),
   ...:                  'flag': Column(['clear', 'clear', 'cloudy'])})

In [2]: ts['time'].value
Out[2]: array([50000., 50001., 50002.])

In [3]: ts['flux'].value
Out[3]: array([1., 2., 3.])

In [4]: ts['flag'].value
<ipython-input-4-7973ebc06f9d> in <module>
----> 1 ts['flag'].value

AttributeError: 'Column' object has no attribute 'value'
```

If we merge this PR, the new behavior will be:
```python
In [4]: ts['flag'].value
Out[4]: array(['clear', 'clear', 'cloudy'], dtype='<U6')
```

### Do we really need this? 

I am aware that, in an ideal world, users should not have to worry about accessing `.value`.  Indeed it looks like `Column` objects work seamlessly with ~all numpy/astropy functions these days (thank you whoever made this happen!!).  The situation is different for `Time` however, e.g. `np.median(Time)` does not work.  And, while `Quantity` objects are seamless, I believe its `.value` property will always be a popular (if dangerous) shortcut for silencing conversion errors.

/cc @taldcroft @mhvk @pllim @orionlee

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -127,7 +127,7 @@ astropy.table
 
 - Added ``Column.value`` as an alias for the existing ``Column.data`` attribute.
   This makes accessing a column's underlying data array consistent with the
-  ``.value`` attribute available for ``Time`` and ``Quantity`` objects. [#10859]
+  ``.value`` attribute available for ``Time`` and ``Quantity`` objects. [#10962]
 
 astropy.tests
 ^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -125,6 +125,10 @@ astropy.stats
 astropy.table
 ^^^^^^^^^^^^^
 
+- Added ``Column.value`` as an alias for the existing ``Column.data`` attribute.
+  This makes accessing a column's underlying data array consistent with the
+  ``.value`` attribute available for ``Time`` and ``Quantity`` objects. [#10859]
+
 astropy.tests
 ^^^^^^^^^^^^^
 
--- a/astropy/table/column.py
+++ b/astropy/table/column.py
@@ -405,6 +405,10 @@ class BaseColumn(_ColumnGetitemShim, np.ndarray):
         return self.view(np.ndarray)
 
     @property
+    def value(self):
+        return self.data
+
+    @property
     def parent_table(self):
         # Note: It seems there are some cases where _parent_table is not set,
         # such after restoring from a pickled Column.  Perhaps that should be
--- a/astropy/table/tests/test_column.py
+++ b/astropy/table/tests/test_column.py
@@ -10,6 +10,7 @@ from numpy.testing import assert_array_equal
 
 from astropy.tests.helper import assert_follows_unicode_guidelines
 from astropy import table
+from astropy import time
 from astropy import units as u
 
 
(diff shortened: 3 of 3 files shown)


mhvk:
@barentsen - I like this, and as I mentioned, think @taldcroft and I actually discussed it. One thing that needs a slight bit of thought it what should happen for `MaskedColumn` - should `.value` return a `MaskedArray`?

Also, added your (implicit) request for `np.median(time)` to #8610 - we know how to do it, but just have to get around to it!

barentsen:
> what should happen for MaskedColumn - should .value return a MaskedArray?

This is the current behavior in this PR.

I spend 15 minutes trying to identify an example application where `np.array` works and `np.MaskedArray` breaks, e.g. using various scipy functions, and failed to find an example.  Admittedly, it's equally difficult to find examples where passing a `Column` instead of an `np.array` matters these days.

mhvk:
Ah, yes, since you just make it an alias for `.data`, it of course works the same way. I think this is a good approach. Something discussed with @taldcroft would be whether eventually we'd want to deprecate `.data` for `MaskedColumn`, so it could become what it is for `MaskedArray`, i.e., return a non-masked item (`Column` in this case). But even if so I think that should be a separate issue/PR - this one is clean and simple as is.

So, I'm in favour and think the PR looks good! But @taldcroft definitely should have a look too.

taldcroft (approved):
LGTM @barentsen, thanks!

PR #10966, opened 2020-10-29
TST: Undo PR 10863 now fix is upstream in numpy

### LINKED ISSUE #10815: numpy-dev using a dtype with a subarray field is deprecated

pllim:
### Description

This pull request is to un-ignore the warning because it should be gone now with new changes in numpy dev.

Close #10815 and close #10818

**TODO**

- [ ] Make sure numpy dev job passes. It is allowed to fail.
  - Still failed as of 2020-10-29, see https://travis-ci.com/github/astropy/astropy/jobs/418922167

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -131,7 +131,6 @@ filterwarnings =
     ignore:Unknown pytest.mark.mpl_image_compare:pytest.PytestUnknownMarkWarning
     ignore:Unknown config option:pytest.PytestConfigWarning
     ignore:matplotlibrc text.usetex:UserWarning:matplotlib
-    ignore:using a dtype with a subarray field is deprecated:DeprecationWarning
     # Triggered by ProgressBar > ipykernel.iostream
     ignore:the imp module is deprecated:DeprecationWarning
     # toolz internal deprecation warning https://github.com/pytoolz/toolz/issues/500


mhvk:
Now passed (dev has an URL error only)

mhvk (approved):
Test now passed, so merging. Thanks, @pllim!

PR #10969, opened 2020-10-30
Improved norm handling for Differential classes

ayshih:
Calling `.norm()` on non-Cartesian `Differential` classes without providing the `base` argument gives an unhelpful error/traceback:
```python
>>> from astropy.coordinates import SphericalDifferential
>>> import astropy.units as u
>>> s = SphericalDifferential(1*u.deg/u.s, 2*u.deg/u.s, 3*u.km/u.s)
>>> s.norm()
  File "c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\representation.py", line 2510, in norm
    return self.to_cartesian(base).norm()
  File "c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\representation.py", line 2369, in to_cartesian
    base_e, base_sf = self._get_base_vectors(base)
  File "c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\representation.py", line 2353, in _get_base_vectors
    cls._check_base(base)
  File "c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\representation.py", line 2295, in _check_base
    if cls not in base._compatible_differentials:
AttributeError: 'NoneType' object has no attribute '_compatible_differentials'
```
A similarly unhelpful error occurs for `RadialDifferential` because it tries to go through Cartesian, but the norm can be returned without doing so.

This PR:
* Adds a check for `.norm()` when the user fails to provide `base` for non-Cartesian differentials.  With this PR:
```python
>>> s.norm()
  File "c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\representation.py", line 2512, in norm
    raise ValueError("`base` must be provided to calculate the norm of a"
ValueError: `base` must be provided to calculate the norm of a SphericalDifferential
```
* Overrides `.norm()` for `RadialDifferential` so it doesn't go through Cartesian.

### DIFF
--- a/astropy/coordinates/representation.py
+++ b/astropy/coordinates/representation.py
@@ -2402,7 +2402,7 @@ class BaseDifferential(BaseRepresentationOrDifferential,
         ----------
         other_class : `~astropy.coordinates.BaseRepresentation` subclass
             The type of representation to turn the coordinates into.
-        base : instance of ``self.base_representation``, optional
+        base : instance of ``self.base_representation``
             Base relative to which the differentials are defined.  If the other
             class is a differential representation, the base will be converted
             to its ``base_representation``.
@@ -2500,13 +2500,17 @@ class BaseDifferential(BaseRepresentationOrDifferential,
         base : instance of ``self.base_representation``
             Base relative to which the differentials are defined. This is
             required to calculate the physical size of the differential for
-            all but cartesian differentials.
+            all but Cartesian differentials or radial differentials.
 
         Returns
         -------
         norm : `astropy.units.Quantity`
             Vector norm, with the same shape as the representation.
         """
+        # RadialDifferential overrides this function, so there is no handling here
+        if not isinstance(self, CartesianDifferential) and base is None:
+            raise ValueError("`base` must be provided to calculate the norm of a"
+                             f" {type(self).__name__}")
         return self.to_cartesian(base).norm()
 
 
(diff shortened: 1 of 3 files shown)


astropy/coordinates/representation.py
```diff
@@ -3006,6 +3010,9 @@ def to_cartesian(self, base):
         return self.d_distance * base.represent_as(
             UnitSphericalRepresentation).to_cartesian()
 
+    def norm(self, base=None):
```
pllim, line 3013: Need a docstring? Also, looks like this also needs a change log for 4.3. Thanks!

ayshih: It will inherit the docstring from `BaseDifferential.norm()`.  I followed the same approach as taken for the other overridden methods in `RadialDifferential`.

mhvk (approved):
Looks great, thanks! I agree that we might as well backport this.

mhvk:
The readthedocs failure looks unrelated.

Obviously, if we backport, the changelog entry should be moved...

StuartLittlefair (approved):

adrn (approved):

ayshih:
How far back will this be backported (i.e., to where should I move the changelog entry)?

mhvk:
It would be for the LTS version, i.e., 4.0.4. Thanks!

ayshih:
Okay, moved

mhvk:
Assuming that the readthedocs failure is unrelated (which it looked like when I checked before), ready to go in if travis passes.

adrn:
Yes, I think the RTD failure is unrelated (many other builds are timing out)

adrn:
Thanks @ayshih!

PR #10970, opened 2020-10-30
Fixed missing induced velocity for helioecliptic transforms

ayshih:
The affine transformations between ICRS and any of the helioecliptic frames (`HeliocentricMeanEcliptic`, `HeliocentricTrueEcliptic`, and `HeliocentricEclipticIAU76`) are missing the induced-velocity offset if the to-be-transformed coordinate has differentials.  I move the code from the ICRS<->HCRS transformations to a separate function in `astropy.coordinates.utils` and used it for all of the transformations.

### DIFF
--- a/astropy/coordinates/builtin_frames/ecliptic_transforms.py
+++ b/astropy/coordinates/builtin_frames/ecliptic_transforms.py
@@ -20,7 +20,7 @@ from .gcrs import GCRS
 from .ecliptic import (GeocentricMeanEcliptic, BarycentricMeanEcliptic, HeliocentricMeanEcliptic,
                        GeocentricTrueEcliptic, BarycentricTrueEcliptic, HeliocentricTrueEcliptic,
                        HeliocentricEclipticIAU76, CustomBarycentricEcliptic)
-from .utils import get_jd12, EQUINOX_J2000
+from .utils import get_jd12, get_offset_sun_from_barycenter, EQUINOX_J2000
 from astropy.coordinates.errors import UnitsError
 
 
@@ -105,15 +105,14 @@ def icrs_to_helioecliptic(from_coo, to_frame):
     if not u.m.is_equivalent(from_coo.cartesian.x.unit):
         raise UnitsError(_NEED_ORIGIN_HINT.format(from_coo.__class__.__name__))
 
-    # get barycentric sun coordinate
-    # this goes here to avoid circular import errors
-    from astropy.coordinates.solar_system import get_body_barycentric
-    bary_sun_pos = get_body_barycentric('sun', to_frame.obstime)
+    # get the offset of the barycenter from the Sun
+    ssb_from_sun = get_offset_sun_from_barycenter(to_frame.obstime, reverse=True,
+                                                  include_velocity=bool(from_coo.data.differentials))
 
     # now compute the matrix to precess to the right orientation
     rmat = _mean_ecliptic_rotation_matrix(to_frame.equinox)
 
-    return rmat, (-bary_sun_pos).transform(rmat)
+    return rmat, ssb_from_sun.transform(rmat)
 
 
 @frame_transform_graph.transform(AffineTransform,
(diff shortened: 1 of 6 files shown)


pllim:
Hello. Is this a bug fix or enhancement? Thanks!

ayshih:
Bug fix: the velocity transformations between ICRS and any of the helioecliptic frames is currently not inaccurate (by a shift of the velocity of the Sun: ~15 m/s)

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -376,3 +376,40 @@ def prepare_earth_position_vel(time):
         )
 
     return earth_pv, earth_heliocentric
+
+
+def offset_sun_from_barycenter(time, include_velocity=False, reverse=False):
```
mhvk, line 381: It is very annoying that representations cannot be simply negated, as that would avoid `reverse` here. In principle, that is not difficult to solve, but would be a new feature, so not so helpful here (but see #10987). Looking at `solar_system`, I also think that, really, `get_body_barycentric` itself should have the `include_velocity` option. We could consider calling it `get_sun_barycentric(time, include_velocity=False, reverse=False)` here (and then later deprecate reverse).

astropy/coordinates/builtin_frames/ecliptic_transforms.py
```diff
@@ -105,15 +105,14 @@ def icrs_to_helioecliptic(from_coo, to_frame):
     if not u.m.is_equivalent(from_coo.cartesian.x.unit):
         raise UnitsError(_NEED_ORIGIN_HINT.format(from_coo.__class__.__name__))
 
-    # get barycentric sun coordinate
-    # this goes here to avoid circular import errors
-    from astropy.coordinates.solar_system import get_body_barycentric
-    bary_sun_pos = get_body_barycentric('sun', to_frame.obstime)
+    # get the offset of the barycenter from the Sun
+    sun_bary_pos = offset_sun_from_barycenter(to_frame.obstime, reverse=True,
```
mhvk, line 109: Here and below, I find the name rather confusion. Maybe `sun_to_bary_offset`? Alternatively, maybe it is an idea to just put the function call in the return statement?

mhvk (commented):
This looks very good! Since it is a bug fix, I think we should backport to 4.0.4. Could you add a changelog entry?

I don't really have any substantial comments, except that I'm not very keen on the names! But admittedly I don't really have good alternative suggestions either.

astropy/coordinates/builtin_frames/ecliptic_transforms.py
ayshih: - For the ICRS<->HCRS transformations, I put the function call in the return statement.
- For the ecliptic transformations, I changed the names `sun_from_ssb` and `ssb_from_sun`.  Putting it in the return statement would get too unreadable since half of the transformations require a rotation of the offset.

ayshih:
- Changelog entry added
- Prepended "get_" to the name of the utility function (since "offset" can be read as a verb)
- Changed some variable names (https://github.com/astropy/astropy/pull/10970#discussion_r517091668)

astropy/coordinates/builtin_frames/ecliptic_transforms.py
mhvk: Nice!

astropy/coordinates/builtin_frames/utils.py
```diff
@@ -376,3 +376,40 @@ def prepare_earth_position_vel(time):
...
+    Returns the offset of the Sun center from the solar-system barycenter (SSB).
+
+    Parameters
+    ----------
+    time : `~astropy.time.Time`
+        Time at which to calculate the offset
+    include_velocity : `bool`
+        If ``True``, attach the velocity as a differential.  Defaults to ``False``.
+    reverse : `bool`
+        If ``True``, return the offset of the barycenter from the SSB.  Defaults to ``False``.
```
mhvk, line 392: Typo: SSB -> Sun

mhvk (commented):
Looks good modulo the typo noted - can just be a `[skip ci]` commit.

ayshih:
Whoops!  Typo fixed

astropy/coordinates/builtin_frames/utils.py
ayshih: Fixed

mhvk (approved):
OK, looks all good now! Merging...

Issue #10639, opened 2020-08-10
TST: Unpin matplotlib in pyinstaller cron job


pllim:
When pyinstaller supports `matplotlib>=3.3`, pinning done in #10637 should be undone. cc @astrofrog 

https://github.com/astropy/astropy/blob/d39544138f8afb833ecb3923d49c9c1297b05036/tox.ini#L165-L166

Also see pyinstaller/pyinstaller#5004

pllim:
If #10779 changes the pinning spec, update the issue.

pllim:
Looks like it is already unpinned in `main`.

PR #10976, opened 2020-10-31
Un-pin matplotlib in pyinstaller CI

dstansby:
Fixes https://github.com/astropy/astropy/issues/10639 - the upstream issue on the pyinstaller side seems to be fixed now, so I think this can be removed.


pllim:
Need a temporary commit to run pyinstaller on this PR as a sanity check. This commit is to be removed and squashed out when things are proven to work. Thanks.

PR #10984, opened 2020-11-03
Remove temporary warning in modeling

nden:
Removed a Warning when evaluating math models:
`Models in math_functions are experimental.`


pllim (approved):
LGTM. RTD timeout seems unrelated.

Issue #3670, opened 2015-04-07
Non-linear fitters should support model sets


embray:
Currently, when a [model set](http://docs.astropy.org/en/stable/modeling/models.html#model-sets) is passed to a non-linear fitter like `LevMarLSQFitter`, an exception is raised that non-linear fitters do no support multiple data sets.

This is a mostly superficial restriction though; there's no reason this can't be implemented--the logic just isn't there yet.

I put a TODO note in the code about this (in the process of some refactoring I'm working on), but I thought this was important enough to make sure there was an issue for it.

embray:
Note, the refactoring I'm doing I think should make this easier, but I haven't tried yet.

sirpercival:
This would be an incredibly useful feature for me to have (even more so than the other issues I've opened in the last couple days). Would you be willing to share the work you've done already and I can see if I can make progress on it?

embray:
@sirpercval: Sorry I didn't get back on this sooner.  I'm going to do what I can to post a prototype for what I'm talking about soon.  Just can't make any guarantees right now.

jehturner:
Should I be worried that @embray refers to a "prototype" here? That sounds more involved than the changes to `LevMarLSQFitter` in #7243, so I wonder whether there might be any complications that we're overlooking. I suppose the proof of the pudding is in the eating...

Cadair:
To make sure this issue contains all the info on this, @nden has a commit on this branch: https://github.com/nden/astropy/tree/fit-msets-nonlin which is related to this I believe.

Cadair:
@astrofrog Do you think we should close this as addressed by the work that closed #14610

astrofrog:
Yes I think so

PR #10985, opened 2020-11-03
Experiments with model sets and non-linear fitters

Cadair:
This is a very draft set of hacks to improve the workflow around, and hopefully parallelise non-linear fitters over model sets.

At the moment this PR enables you to update parameters in place in a model, and also adds a flag to the `LevMarLSQFitter` to not copy the model before fitting it. Using this allows the following code to work:

```python
gen_curve = m.Gaussian1D(amplitude=12,
                         stddev=6,
                         mean=0)

x = np.arange(-20, 21)

n_models = 12

data = gen_curve(x)
data = np.repeat(data[None,:], n_models, axis=0)
data += np.random.normal(0., data*0.1, size=data.shape)

curve = m.Gaussian1D(amplitude=np.zeros((n_models,)) + 12,
                     stddev=np.zeros((n_models,)) + 6,
                     mean=np.zeros((n_models,)) + 1,
                     model_set_axis=-1)

fit = fitting.LevMarLSQFitter(inplace=True)

for sub, sub_data in zip(curve.iter_models, data):
    fit(sub, x, sub_data)

print(curve)
```
```
Model: Gaussian1D
Inputs: ('x',)
Outputs: ('y',)
Model set size: 12
Parameters:
        amplitude               mean               stddev     
    ------------------ --------------------- -----------------
     11.95422133571381  -0.07266095118471086  5.99545327963244
    12.156119072050434  -0.07862513063747723 5.896005724939767
    12.661097809541358  -0.12937266599014197 5.914236165562634
    12.145417357362323  0.061064016793466167 5.902466367263756
    11.908985742644665   0.06705609240645635   6.1386506413625
     11.32213839228508    0.0728347476209994 6.110827573154081
    12.824222438558532  -0.02555400772144983 5.972700110188168
    11.564129843035015    0.0529231958392127 6.197163202260534
     11.51237244367314  -0.08100330081726208  6.08075381821545
    12.069120479400457 -0.042293890464727577 5.985331749660404
    10.994796721724828   0.16992100295058366 6.270956201788193
     11.27045189430589   -0.0313423860725926 6.124522140219055

```

This is already a lot cleaner than having to manually reconstruct the model set from the output of 12 calls to the fitter. The potential next steps for this are:

* [x] Provide a method or helper function which returns an iterable of sub models for each element in a model set
* [x] Provide an example of using something like subprocess.map to parallelise fitting across the model sets
* [ ] Use dask to distribute the fitting.

I am opening this as draft now to get initial feedback from people who are more familiar with the fitting machinery, so they can wave me off if this is the wrong tree to bark up.

@astrofrog @perrygreenfield @nden

related to https://github.com/astropy/astropy/issues/3670

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -1958,6 +1958,22 @@ class Model(metaclass=_ModelMeta):
         """
         return 1
 
+    @property
+    def iter_models(self):
+        """
+        An iterable of model instances with parameters shared with this model set.
+        """
+        if len(self) == 1:
+            return self
+
+        for i in range(len(self)):
+            params = {}
+            for pname in self.param_names:
+                pslice = [slice(None)] * len(self._param_metrics[pname]['shape'])
+                pslice[self._model_set_axis] = slice(i, i+1)
+                params[pname] = getattr(self, pname)[tuple(pslice)]
+            yield type(self)(**params)
+
     def _initialize_constraints(self, kwargs):
         """
         Pop parameter constraint values off the keyword arguments passed to
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -1974,6 +1974,46 @@ class Model(metaclass=_ModelMeta):
                 params[pname] = getattr(self, pname)[tuple(pslice)]
             yield type(self)(**params)
 
+    @classmethod
+    def from_models(cls, models, model_set_axis=None):
+        """
+        Generate a model set from an iterable of models.
+
+        Parameters
+        ----------
+        models : iterable of `.Model`
+            A list of models to convert into a model set.
+
+        model_set_axis: int, optional
+            This argument changes how parameter values are combined.  Normally the
+            first axis of each input parameter array (properly the 0th axis) is
+            taken as the axis corresponding to the model sets.  However, any axis
+            of an input array may be taken as this "model set axis".  This accepts
+            negative integers as well--for example use ``model_set_axis=-1`` if the
+            last (most rapidly changing) axis should be associated with the model
+            sets.
+
+        Returns
+        -------
+        model: `.Model`
+            An instance of the model with length equal to the length of the input model list.
+        """
+        if not isiterable(models) and all(isinstance(m, Model) for m in models):
+            raise TypeError("models must be an iterable of model instances.")
+
+        models = list(models)
+        if not all(isinstance(m, cls) for m in models):
+            raise TypeError("All models in the input must be of the same type as this model.")
+
+        if not all((m._parameters.shape == models[0]._parameters.shape for m in models)):
+            raise ValueError("All models in the input must have the same shape parameters.")
+
+        params = {pname: np.stack([getattr(m, pname).value for m in models],
+                                  axis=model_set_axis or 0) for pname in cls.param_names}
+        output_model = cls(n_models=len(models), model_set_axis=model_set_axis, **params)
+
+        return output_model
+
     def _initialize_constraints(self, kwargs):
         """
         Pop parameter constraint values off the keyword arguments passed to
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -2155,12 +2155,12 @@ class Model(metaclass=_ModelMeta):
                 _val = param._setter(value)
             if isinstance(_val, Quantity):
                 param.internal_unit = _val.unit
-                param._internal_value = np.array(_val.value)
+                param._internal_value = np.array(_val.value, copy=False)
             else:
                 param.internal_unit = None
-                param._internal_value = np.array(_val)
+                param._internal_value = np.array(_val, copy=False)
         else:
-            param._value = np.array(value)
+            param._value = np.array(value, copy=False)
 
     def _initialize_slices(self):
 
--- a/astropy/modeling/fitting.py
+++ b/astropy/modeling/fitting.py
@@ -1050,7 +1050,7 @@ class LevMarLSQFitter(metaclass=_FitterMeta):
     The constraint types supported by this fitter type.
     """
 
-    def __init__(self, calc_uncertainties=False):
+    def __init__(self, calc_uncertainties=False, inplace=False):
         self.fit_info = {'nfev': None,
                          'fvec': None,
                          'fjac': None,
@@ -1060,7 +1060,8 @@ class LevMarLSQFitter(metaclass=_FitterMeta):
                          'ierr': None,
                          'param_jac': None,
                          'param_cov': None}
-        self._calc_uncertainties=calc_uncertainties
+        self._calc_uncertainties = calc_uncertainties
+        self._inplace = inplace
         super().__init__()
 
     def objective_function(self, fps, *args):
@@ -1146,7 +1147,7 @@ class LevMarLSQFitter(metaclass=_FitterMeta):
 
         from scipy import optimize
 
-        model_copy = _validate_model(model, self.supported_constraints)
+        model_copy = _validate_model(model, self.supported_constraints, copy=not self._inplace)
         farg = (model_copy, weights, ) + _convert_input(x, y, z)
         if model_copy.fit_deriv is None or estimate_jacobian:
             dfunc = None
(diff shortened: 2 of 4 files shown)


Cadair:
One thing I have just noticed with this is that it leads to the model's `._parameters` getting out of sync with the arrays in the parameter objects. I have no idea what the `._parameters` array is for or why this would be an issue.

Cadair:
I have now pushed a third commit which adds an iterator to make this a little cleaner:

```python
fit = fitting.LevMarLSQFitter(inplace=True)

for sub, sub_data in zip(curve.iter_models, data):
    fit(sub, x, sub_data)

print(curve)
```

Cadair:
Finally, just to prove it's possible I ran this with multiprocessing. All be it without the shared memory back to the original mode instance.

```python
import multiprocessing

pool = multiprocessing.Pool(2)

fit = fitting.LevMarLSQFitter()

pool.starmap(fit, [(sub, x, sub_data) for sub, sub_data in zip(curve.iter_models, data)])
```
```
[<Gaussian1D(amplitude=12.13056208, mean=0.08627905, stddev=5.89127721)>,
 <Gaussian1D(amplitude=11.43620798, mean=-0.11744416, stddev=6.08924581)>,
 <Gaussian1D(amplitude=12.12824768, mean=-0.13536278, stddev=5.89532783)>,
 <Gaussian1D(amplitude=12.42841051, mean=0.07618924, stddev=5.8025821)>,
 <Gaussian1D(amplitude=12.46726706, mean=0.06087045, stddev=5.86021525)>,
 <Gaussian1D(amplitude=11.96075239, mean=-0.03319208, stddev=6.06174972)>,
 <Gaussian1D(amplitude=12.34658916, mean=0.10149613, stddev=6.00840042)>,
 <Gaussian1D(amplitude=11.85058529, mean=-0.07094919, stddev=6.12856275)>,
 <Gaussian1D(amplitude=12.54250288, mean=0.10285538, stddev=5.7743818)>,
 <Gaussian1D(amplitude=12.19364611, mean=-0.0611381, stddev=5.81404733)>,
 <Gaussian1D(amplitude=12.25321791, mean=0.00247555, stddev=5.92130551)>,
 <Gaussian1D(amplitude=12.04788155, mean=-0.01036336, stddev=6.02245627)>]

It is however nice and easy to parallelise now.

perrygreenfield:
@Cadair are you more concerned about the interface or the use of memory?

With regard to self._parameters, it is the interface to the fitters so that fitters not need concern themselves with the names of parameters. They just see a set of number that they change until the fit criteria are met. This presents a problem keeping the values in parameter objects and the model in sync. When parameter change the values that leads to an update in the parameter array, but it doesn't work the other way since changing an element of an array is hard to intercept without massive wrapping of numpy. The current approach is either you replace the whole parameter array (in which case setattr knows to update the parameters) and if you only update values in the array, it is up to you to force a sync (by calling the `_array_to_parameters()` method

Cadair:
> are you more concerned about the interface or the use of memory?

Ideally both. At this stage I am thinking more about interface, but I try and apply this to a 4k x 4k grid of spectral points then memory is going to matter as well.

With regards to `._parameters` I am setting the values in the parameter objects (I think) but the values in the arrays aren't updating in the above examples. It looks like they are still zeros at the end (or maybe empty). Which means I don't understand how the fitter is (or isn't) using them. 

Does it matter if the `._parameters` array is actually a numpy array? Could it be a light weight proxy object to the actual arrays in the Parameter objects? (Not suggesting making it so just trying to work out what's going on)

perrygreenfield:
When you say that the values in the arrays aren't updating, how are you showing that? print(curve.amplitude)?

Cadair:
I was running `print(curve._parameters)`.

> are you more concerned about the interface or the use of memory?

Thinking more about this overnight, I think that the memory is probably more important. At the end of all this I hope to have some kind of `ParrallelFitter` class so that could paper over the worst of the interface complexity.

perrygreenfield:
I'm guessing (not completely certain since it takes me time to follow all the interactions to be sure) that in the example above you are creating a new model for each pixel (or whatever). And what happens is that for each new model a new _parameter array is created (small, since it only represents one case). That is filled by the fitter, and then those values are back populated into the parameter values, which happen to be views in the original model parameter arrays. So in the end, those parameter  objects value are fully populated, but the original model's _parameter array is not updated (unless you call curve._parameters_to_array() (which is called automatically in certain contexts, but not in what you are doing). Am I 100% sure, no, but that is what I think is happening.

The question becomes whether there is a way of avoiding creating a new object for each pixel. Ultimately there should be.

Lastly, regarding the memory issue. I always wonder about strategies that save a factor of two (or three) in memory usage. Sure, it helps when you are close to the limit, but in the full range of memory issues, it is a small range of concern (in a log sense). There are probably more robust solutions that deal with the broader range of memory (e.g., iterating over partitions of the original problem that will fit into memory).

Finally (did I say lastly?) I'm not sure what your intend use of iter_models is since you didn't use it in your example.

Cadair:
> The question becomes whether there is a way of avoiding creating a new object for each pixel. Ultimately there should be.

I am going to start here as I think it's the most fundamental issue.

This is where I started. I want to pass a model set to the non linear fitters and iterate over the elements in the model set. I spent some time staring at the `__call__` methods of some fitters and I came to the conclusion that given the constructor to a model seems reasonably lightweight, it would be a lot easier to write something up at this level here to make the model sets work with existing fitters rather than having to bend all the fitters to the model sets.

> Lastly, regarding the memory issue. I always wonder about strategies that save a factor of two (or three) in memory usage. Sure, it helps when you are close to the limit, but in the full range of memory issues, it is a small range of concern (in a log sense).

With regards to memory, it's not my current focus, the copy based changes presented here were mostly aimed at improving the API and maybe taking down some of the performance overhead and memory overhead. Also, they were me poking at this thing trying to understand it. They seem useful to me, but maybe they aren't, or maybe a whole different approach is needed (see below).

> There are probably more robust solutions that deal with the broader range of memory (e.g., iterating over partitions of the original problem that will fit into memory).

I am currently pondering if this is the next step I want to take. The main motivations here for me are to both make larger than memory fitting work at all, and also to be able to make use of parallelisation. Both the DKIST and the spectral cube use cases driving this are using dask, so I was considering chasing down a full dask array based solution here.

I think this would be a lot more work, I think to really do that effectively (with Dasks distributed scheduler) we would have to allow the arrays in Parameter / Model to be Dask array not just numpy array, which given the fact I don't understand any of the subtlety here is rather daunting.

> I'm not sure what your intended use of iter_models was since you didn't use it in your example.

See https://github.com/astropy/astropy/pull/10985#issuecomment-721278295 for the use I was thinking about. Maybe this is just a stop-gap on being able to chunk up the problem with Dask, or maybe this is actually a useful addition to the API for model sets?

perrygreenfield:
Thinking about this a bit more, I think that what you did above wouldn't have used any more memory in the original case either  since the fitting is done on one at a time, and any copying of the parameters would only have been done on array lengths of 1. If the loop instead used the same instance of a single model (i.e., no model sets) case, and you instead assigned the parameters to the same model instance, it probably would avoid much of a new model creation (which may or may not be negligible anyway), and just assign the result to parameter arrays. True, you need multiple lines to do that rather than putting them in a class initializer.

If I recall correctly, the thought we had was doing all this within the nonlinear fitters, but then the logic of multiprocessing would have to go there too, or splitting up the model sets into N model set classes to run on multiple cores or processors. So far we haven't done much to take advantage of multiprocessing, partly because it involves may possible situations and the need to understand how best to split up the work.

Cadair:
> If the loop instead used the same instance of a single model (i.e., no model sets) case, and you instead assigned the parameters to the same model instance, it probably would avoid much of a new model creation (which may or may not be negligible anyway), and just assign the result to parameter arrays. True, you need multiple lines to do that rather than putting them in a class initializer.

This could probably work as well. What I implemented here was more a few helper functions and stuff to make this easier. I don't know if the inplace stuff is truly required to make all this work, but for a shared-memory type use (i.e. thread parallel or shared memory with multiprocessing) it might be useful.

There is a lot of demand for these features for both the solar and radio communities, so I should probably try and de-stall this. Would removing the shared memory stuff from this make the benefits less muddy? I think the `from_models` and `iter_models` helpers would certainly make the implementation of a `MultiprocessingFitter` easier.

pllim:
Since this was only experimental, I am closing.

PR #10990, opened 2020-11-04
Remove OrderedDict from astropy.coordinates, small BaseCoordinateFrame.__init__ refactor

astrojuanlu:
### Description

* Remove `OrderedDict` from `astropy.coordinates`, see gh-10888
* Small refactor of `BaseCoordinateFrame.__init__` to clarify some comments and move part of the code to two separate internal methods. It was not justified in itself, but used this opportunity to include it. The removal of `OrderedDict` is in a separate commit that applies cleanly without the rest.

### DIFF
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -423,8 +423,8 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
         _normalize_representation_type(kwargs)
         representation_type = kwargs.pop('representation_type', representation_type)
 
-        self._representation = self._infer_representation_differential_type(representation_type, differential_type)
-        self._data = self._infer_representation_differential_data(args, copy, kwargs)  # possibly None.
+        self._representation = self._infer_representation(representation_type, differential_type)
+        self._data = self._infer_data(args, copy, kwargs)  # possibly None.
 
         # Set frame attributes, if any
 
@@ -491,7 +491,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
             # Set up representation cache.
             self.cache['representation'][key] = self._data
 
-    def _infer_representation_differential_type(self, representation_type, differential_type):
+    def _infer_representation(self, representation_type, differential_type):
         if representation_type is None and differential_type is None:
             return {'base': self.default_representation, 's': self.default_differential}
 
@@ -518,7 +518,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
 
         return _get_repr_classes(representation_type, **differential_type)
 
-    def _infer_representation_differential_data(self, args, copy, kwargs):
+    def _infer_data(self, args, copy, kwargs):
         # if not set below, this is a frame with no data
         representation_data = None
         differential_data = None
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -10,7 +10,7 @@ classes.
 import abc
 import copy
 import inspect
-from collections import namedtuple, OrderedDict, defaultdict
+from collections import namedtuple, defaultdict
 import warnings
 
 # Dependencies
@@ -409,7 +409,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
 
     _inherit_descriptors_ = (Attribute,)
 
-    frame_attributes = OrderedDict()
+    frame_attributes = {}
     # Default empty frame_attributes dict
 
     def __init__(self, *args, copy=True, representation_type=None,
@@ -711,8 +711,8 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
 
     @classmethod
     def get_frame_attr_names(cls):
-        return OrderedDict((name, getattr(cls, name))
-                           for name in cls.frame_attributes)
+        return {name: getattr(cls, name)
+                for name in cls.frame_attributes}
 
     def get_representation_cls(self, which='base'):
         """The class used for part of this frame's data.
@@ -846,7 +846,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
         return self._get_representation_info()
 
     def get_representation_component_names(self, which='base'):
-        out = OrderedDict()
+        out = {}
         repr_or_diff_cls = self.get_representation_cls(which)
         if repr_or_diff_cls is None:
             return out
@@ -857,7 +857,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
         return out
 
     def get_representation_component_units(self, which='base'):
-        out = OrderedDict()
+        out = {}
         repr_or_diff_cls = self.get_representation_cls(which)
         if repr_or_diff_cls is None:
             return out
@@ -1893,7 +1893,7 @@ class GenericFrame(BaseCoordinateFrame):
     name = None  # it's not a "real" frame so it doesn't have a name
 
     def __init__(self, frame_attrs):
-        self.frame_attributes = OrderedDict()
+        self.frame_attributes = {}
         for name, default in frame_attrs.items():
             self.frame_attributes[name] = Attribute(default)
             setattr(self, '_' + name, default)
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -427,6 +427,9 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
             representation_type, differential_type = self._infer_representation_differential_type(
                 representation_type, differential_type)
             self.set_representation_cls(representation_type, **differential_type)
+        else:
+            self._representation = {'base': self.default_representation,
+                                    's': self.default_differential}
 
         representation_data, differential_data = self._infer_representation_differential_data(
             args, copy, kwargs)
@@ -552,10 +555,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray, metaclass=FrameMeta):
                                      'velocity differential is supported. Got: '
                                      '{}'.format(diffs))
 
-        # This calls self.get_representation_cls()
-        # and sets self._representation to the defaults if it was not set earlier
-        # (specifically if representation_type or differential_type were passed to the initializer)
-        elif self.representation_type:
+        else:
             representation_cls = self.get_representation_cls()
             # Get any representation data passed in to the frame initializer
             # using keyword or positional arguments for the component names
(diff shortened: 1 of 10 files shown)


astropy/coordinates/baseframe.py
```diff
@@ -424,39 +424,121 @@ def __init__(self, *args, copy=True, representation_type=None,
...
 
         if representation_type is not None or differential_type is not None:
+            representation_type, differential_type = self._infer_representation_differential_type(
+                representation_type, differential_type)
+            self.set_representation_cls(representation_type, **differential_type)
+        else:
+            self._representation = {'base': self.default_representation,
+                                    's': self.default_differential}
 
-            if representation_type is None:
```
astrojuanlu, line 428: The diff is a bit messed up, but this line up to the next comment was moved to `_infer_representation_differential_data`

astrojuanlu: (Edit: typo)

astropy/coordinates/baseframe.py
```diff
@@ -424,39 +424,121 @@ def __init__(self, *args, copy=True, representation_type=None,
...
-                representation_type = self.default_representation
+        representation_data, differential_data = self._infer_representation_differential_data(
```
astrojuanlu, line 434: `_infer_representation_differential_data` contains now a significant chunk of logic

astropy/coordinates/baseframe.py
```diff
@@ -424,39 +424,121 @@ def __init__(self, *args, copy=True, representation_type=None,
...
+            # Read-only frame attributes are defined as FrameAttribute
+            # descriptors which are not settable, so set 'real' attributes as
+            # the name prefaced with an underscore.
+
+            if fnm in kwargs:
+                value = kwargs.pop(fnm)
+                setattr(self, '_' + fnm, value)
+                # Validate attribute by getting it. If the instance has data,
+                # this also checks its shape is OK. If not, we do it below.
+                values[fnm] = getattr(self, fnm)
```
astrojuanlu, line 456: I wanted to decouple the setting of the frame attributes from the rest, but found that a lot of things are happening in this `getattr(self, fnm)`: it's where the validation of the shapes takes place _if the frame has `_data` already_ (see my extra code comment in `Attribute.__get__`). I wonder if that part could be splitted out to another function, instead of validating on the fly like this?

astropy/coordinates/baseframe.py
```diff
@@ -473,7 +555,7 @@ def __init__(self, *args, copy=True, representation_type=None,
                                      'velocity differential is supported. Got: '
                                      '{}'.format(diffs))
 
-        elif self.representation_type:
+        else:
```
astrojuanlu, line 558: With the changes in `__init__`, `self.representation_type` is always set, or should be

astropy/coordinates/baseframe.py
```diff
@@ -729,10 +748,6 @@ def get_representation_cls(self, which='base'):
         -------
         representation : `~astropy.coordinates.BaseRepresentation` or `~astropy.coordinates.BaseDifferential`.
         """
-        if not hasattr(self, '_representation'):
```
astrojuanlu, line 732: Now we do this in `__init__`

astrojuanlu:
RTD failure seems unrelated.

pllim:
Yeah, seems to happen quite often lately. 🤔

astropy/coordinates/baseframe.py
```diff
@@ -424,39 +424,121 @@ def __init__(self, *args, copy=True, representation_type=None,
         representation_type = kwargs.pop('representation_type', representation_type)
 
         if representation_type is not None or differential_type is not None:
```
mhvk, line 426: Maybe also move this class to the new `_infer_representation_differential_type`?

astropy/coordinates/baseframe.py
```diff
@@ -424,39 +424,121 @@ def __init__(self, *args, copy=True, representation_type=None,
...
+        else:
+            self._representation = {'base': self.default_representation,
+                                    's': self.default_differential}
 
-            if representation_type is None:
-                representation_type = self.default_representation
+        representation_data, differential_data = self._infer_representation_differential_data(
+            args, copy, kwargs)
+
+        if differential_data:
```
mhvk, line 437: I think just
```
if differential_data:
    representation_data = representation_data.with_differentials({'s': differential_data})
self._data = representation_data
```
would be slightly clearer. If so, I think one can also consider moving the `if` statement inside the function call (which could then just be `_infer_representation_data(...)`) - in the end the differentials are supposed to be attached to a representation.

astropy/coordinates/baseframe.py
mhvk: Yes, this is quite ugly (I refactored it to get shape consistency before). I think it may be an idea to treat the data and values as one giant to-be-broadcasted whole instead. But my suggestion would be to leave it for another PR.

astropy/coordinates/baseframe.py
```diff
@@ -424,39 +424,121 @@ def __init__(self, *args, copy=True, representation_type=None,
...
+                    for fnm in shapes:
+                        getattr(self, fnm)
                 else:
-                    differential_type = {'s': 'base'}  # see set_representation_cls()
+                    self._no_data_shape = shapes.popitem()[1]
+
+            else:
+                self._no_data_shape = ()
+
+        # else:
```
mhvk, line 492: I sort-of see your point of not using `else` - though in a way both clauses are "wrap-up" for the two cases. In any case, don't leave it commented out here!

Aside: I do wonder whether this is actually needed - but a good thorough look at the cache is also for another time!

mhvk (commented):
@astrojuanlu - this looks nice! I like the gathering of those unwieldy parts of `__init__` in their own methods. A few small comments about going perhaps just a little further.

p.s. Personally, I'd have slightly preferred to have the change from `OrderedDict` to `dict` in a separate PR, as it now mixes in some way trivial stuff with somewhat more tricky changes - though really what is most tricky is how the difference happens to be type-set.

astrojuanlu:
> p.s. Personally, I'd have slightly preferred to have the change from OrderedDict to dict in a separate PR, as it now mixes in some way trivial stuff with somewhat more tricky changes

What I usually do in these cases is to review only a range of commits:

commit-range

But, if there are other reasons to split it, I'll happily do so :)

Addressing your comments now.

astropy/coordinates/baseframe.py
astrojuanlu: Absolutely, let's leave this for another PR. Just touching base to make sure I was not missing anything.

astropy/coordinates/baseframe.py
astrojuanlu: > In any case, don't leave it commented out here!

Fair enough, removing it!

> Aside: I do wonder whether this is actually needed - but a good thorough look at the cache is also for another time!

Definitely - I still do not understand what that chunk does :sweat_smile:

astropy/coordinates/baseframe.py
astrojuanlu: My initial philosophy was: "let these new `_infer` methods _not_ set any attributes, and do that only in `__init__`". That way, even though I distribute the logic and make `__init__` easier to read, the "mutation" only happens in one place. What do you think?

astropy/coordinates/baseframe.py
astrojuanlu: Nice! I like this solution much more. Changing.

astrojuanlu:
Thanks for your prompt review @mhvk! Just addressed your comments.

astropy/coordinates/baseframe.py
```diff
@@ -424,39 +424,114 @@ def __init__(self, *args, copy=True, representation_type=None,
         representation_type = kwargs.pop('representation_type', representation_type)
 
         if representation_type is not None or differential_type is not None:
+            representation_type, differential_type = self._infer_representation_differential_type(
+                representation_type, differential_type)
+            self.set_representation_cls(representation_type, **differential_type)
+        else:
+            self._representation = {'base': self.default_representation,
+                                    's': self.default_differential}
```
astrojuanlu, line 432: I was not entirely satisfied with this mix between `self.set_representation_cls(...)` and `self._representation = ...`, but this chunk of code:

```python
        if base is None:
            base = self._representation['base']
```

makes `set_representation_cls` a bit obtuse, and I'm still not sure how to untangle it.

astropy/coordinates/baseframe.py
mhvk: Ah, I see I was not clear. I definitely agree that we should set the attributes here, I just thought we could deal with the `None, None` input also in your new method (which would then just return `self.default_representation, self.default_differential`), i.e., here you would simply have
```
        representation_type, differential_type = self._infer_representation_differential_type(
            representation_type, differential_type)
        self.set_representation_cls(representation_type, **differential_type)
```

mhvk: Actually, if your method returns `representation_type` also in the dict as the 'base' entry, you can pass it directly on (with name change):
```
self.set_representation_cls(**self._infer_representation_cls(
            representation_type, differential_type))
```

mhvk: Sorry, now see your next comment: I think I would just do
```
self._representation = self._infer_representation(representation_type, differential_type)
```
and let `_infer_representation` return a properly organized `dict`.

mhvk:
@astrojuanlu - thanks, looks pretty good now. Some more comments on the representation inferrence - I think we are thinking along the same lines there!

p.s. As you note, the ordereddict stuff is all in a separate commit, so let's leave it here.

astropy/coordinates/baseframe.py
astrojuanlu: Excellent suggestion, now the code is much more compact :heart_eyes:

astropy/coordinates/baseframe.py
```diff
@@ -423,40 +423,110 @@ def __init__(self, *args, copy=True, representation_type=None,
         _normalize_representation_type(kwargs)
         representation_type = kwargs.pop('representation_type', representation_type)
 
-        if representation_type is not None or differential_type is not None:
+        self._representation = self._infer_representation_differential_type(representation_type, differential_type)
```
mhvk, line 426: Ah, seeing it together it is even nicer than I thought! Really only a nitpick left - my tendency would be to shorten the method names to `_infer_representation(...)` and `_infer_data(...)` (or latter perhaps `_infer_representation_data` - the differential being part of the representation data).

mhvk:
This looks quite beautiful (modulo the suggestion for a shorter name for the method). @eteq, @adrn, @StuartLittlefair - it might be good if any of you could have a quick look too; does any of you want to? (I'm ready to sign off on it otherwise.) Note that the diff looks annoyingly worse than it is - mostly this is just a refactoring, but one that makes the initialization easier to follow.

astropy/coordinates/baseframe.py
astrojuanlu: Done!

mhvk (approved):
OK, it all looks good from my point of view. Will leave it open a few days just in case.

StuartLittlefair:
I will leave this to your capable hands @mhvk. Delving into the machinery of the coordinates classes makes my head hurt.

adrn:
Planning to take a look at this but very busy week -- will get to it over the weekend!

adrn (approved):
No comments from me! I think this looks great, and I like the restructuring / factorization of the __init__! Thanks @astrojuanlu for this!

PR #10992, opened 2020-11-04
Make QTable Quantity int to float conversion less surprising

### LINKED ISSUE #10964: Make QTable Quantity conversion's implicit float conversion less surprising
### Description

Based on the conversation in PR #10950 , it is clear that it is often desirable (and at times necessary) that for `QTable`, when a `Column` of integer `dtype` is converted to `Quantity`, the `dtype` is converted to float in the resulting `Quantity`

It has been the **existing** behavior. However, the behavior:
1. is not documented,
2. is not explicitly tested in unit test suite,
3. could be surprising to astropy users in some cases. See below for an example [*]

It'd be good to have (1) and (2) addressed. Bonus would be to address (3).

### Additional context

[*] Example where the current behavior could be surprising to users, drawn from issue #10950 where the behavior was perceived to be a bug.
```python
ts = astropy.table.QTable(dict(time=[1, 2, 3]))
ts['cadenceno'] = [11, 12, 13]
print(ts['cadenceno'].dtype)  # int32
ts['cadenceno'].unit = u.dimensionless_unscaled
print(ts['cadenceno'].dtype)  # float64
# ^ dtype is changed after assigning an unit (i.e, converting to Quantity)
```

orionlee:
### Description

This pull request is to address the issue that during `QTable` `Quantity` conversion, it will convert `int` to `float`.
- add test to make it explicit that the behavior is intentional (rather than a bug)
- Add note to `QTable` apidoc.
- Emit warnings when int-to-float conversion happened. The intent is to catch use cases such as `qtab['col'].unit = u.m`. 

Fixes #10964

### DIFF
--- a/astropy/table/connect.py
+++ b/astropy/table/connect.py
@@ -73,9 +73,7 @@ class TableRead(registry.UnifiedReadWrite):
                 raise TypeError('could not convert reader output to {} '
                                 'class.'.format(cls.__name__))
 
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", message=".*dtype is converted to float.*")
-            out._set_column_attribute('unit', units)
+        out._set_column_attribute('unit', units)
         out._set_column_attribute('description', descriptions)
 
         return out
--- a/astropy/table/connect.py
+++ b/astropy/table/connect.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import warnings
 from astropy.io import registry
 
 from .info import serialize_method_as
@@ -72,7 +73,9 @@ class TableRead(registry.UnifiedReadWrite):
                 raise TypeError('could not convert reader output to {} '
                                 'class.'.format(cls.__name__))
 
-        out._set_column_attribute('unit', units)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message=".*dtype is converted to float.*")
+            out._set_column_attribute('unit', units)
         out._set_column_attribute('description', descriptions)
 
         return out
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3665,6 +3665,17 @@ class QTable(Table):
                           "".format(col.info.name))
             self._dtype_before_conversion = None
 
+    def _set_column_attribute(self, attr, values):
+        # some internal codes use ``_set_column_attribute()`` to assign units to a column,
+        # which might trigger the warnings. Suppress them as ,they are internal codes that
+        # should know what they are doing.
+        if attr != 'unit':
+            super()._set_column_attribute(attr, values)
+        else:
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", message=".*dtype is converted to float.*")
+                super()._set_column_attribute(attr, values)
+
 
 class NdarrayMixin(np.ndarray):
     """
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3593,6 +3593,9 @@ class QTable(Table):
     except that columns with an associated ``unit`` attribute are converted to
     `~astropy.units.Quantity` objects.
 
+    Note: when a column of ``int`` ``dtype`` is converted to ``Quantity``, its
+    ``dtype`` is converted to ``float``.
+
     See also:
 
     - https://docs.astropy.org/en/stable/table/
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3624,6 +3624,8 @@ class QTable(Table):
 
     """
 
+    _dtype_before_conversion = None
+
     def _is_mixin_for_table(self, col):
         """
         Determine if ``col`` should be added to the table directly as
@@ -3647,11 +3649,21 @@ class QTable(Table):
             qcol = q_cls(col.data, col.unit, copy=False)
             qcol.info = col.info
             qcol.info.indices = col.info.indices
+            if np.issubdtype(getattr(col, 'dtype', None), np.int):
+                self._dtype_before_conversion = col.dtype
             col = qcol
         else:
             col = super()._convert_col_for_table(col)
 
         return col
+
+    def replace_column(self, name, col, copy=True):
+        super().replace_column(name, col, copy)
+        if self._dtype_before_conversion is not None:
+            warnings.warn("dtype is converted to float for Quantity column '{}'. "
+                          "See https://docs.astropy.org/en/stable/api/astropy.table.QTable.html"
+                          "".format(col.info.name))
+            self._dtype_before_conversion = None
 
 
 class NdarrayMixin(np.ndarray):
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -3593,9 +3593,8 @@ class QTable(Table):
     except that columns with an associated ``unit`` attribute are converted to
     `~astropy.units.Quantity` objects.
 
-    Note: when a column of ``int`` ``dtype`` is converted to ``Quantity``, its
-    ``dtype`` is converted to ``float``.
-
+    .. note:: When a column of ``int`` ``dtype`` is converted to ``Quantity``, its
+              ``dtype`` is converted to ``float``.
     See also:
 
     - https://docs.astropy.org/en/stable/table/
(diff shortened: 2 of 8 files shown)


orionlee:
Regarding the warnings,
- it's emitted in the narrow case that conversion happened during column update, e.g., `qtab['col'].unit = u.m`
- No warning is emitted for the case when a column is added, e.g., adding `Column([1, 2], unit=u.m)`, as it will simply generate many spurious warnings.
- the current implementation (tracking it in an internal state `QTable._dtype_before_conversion`) is a bit hacky. I tried to add the state to the column `qcol` object, its `.info`, and `.meta`, but none of them work (the state was somehow lost).

astropy/table/connect.py
```diff
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import warnings
```
pllim, line 3: Not used?
```suggestion
```

astropy/table/table.py
```diff
@@ -3593,6 +3593,9 @@ class QTable(Table):
     except that columns with an associated ``unit`` attribute are converted to
     `~astropy.units.Quantity` objects.
 
+    Note: when a column of ``int`` ``dtype`` is converted to ``Quantity``, its
+    ``dtype`` is converted to ``float``.
```
pllim, line 3597:
```suggestion
    .. note:: When a column of ``int`` ``dtype`` is converted to ``Quantity``, its
              ``dtype`` is converted to ``float``.
```

astropy/table/table.py
```diff
@@ -3644,12 +3649,33 @@ def _convert_col_for_table(self, col):
...
             col = super()._convert_col_for_table(col)
 
         return col
 
+    def replace_column(self, name, col, copy=True):
+        super().replace_column(name, col, copy)
+        if self._dtype_before_conversion is not None:
+            warnings.warn("dtype is converted to float for Quantity column '{}'. "
+                          "See https://docs.astropy.org/en/stable/api/astropy.table.QTable.html"
+                          "".format(col.info.name))
```
pllim, line 3665:
```suggestion
            warnings.warn(f"dtype is converted to float for Quantity column '{col.info.name}'. "
                          "See https://docs.astropy.org/en/stable/api/astropy.table.QTable.html",
                          AstropyUserWarning)
```

astropy/table/table.py
```diff
@@ -3644,12 +3649,33 @@ def _convert_col_for_table(self, col):
...
+            self._dtype_before_conversion = None
+
+    def _set_column_attribute(self, attr, values):
+        # some internal codes use ``_set_column_attribute()`` to assign units to a column,
+        # which might trigger the warnings. Suppress them as ,they are internal codes that
```
pllim, line 3670:
```suggestion
        # which might trigger the warnings. Suppress them as they are internal codes that
```

astropy/table/tests/test_column.py
```diff
@@ -595,7 +595,8 @@ def test_qtable_column_conversion():
     assert isinstance(qtab['i'], table.column.Column)
     assert isinstance(qtab['f'], table.column.Column)
 
-    qtab['i'].unit = 'km/s'
+    with pytest.warns(UserWarning, match="dtype is converted to float"):
```
pllim, line 598:
```suggestion
    with pytest.warns(AstropyUserWarning, match="dtype is converted to float"):
```

astropy/table/tests/test_table.py
```diff
@@ -2258,6 +2259,46 @@ def test_table_meta_copy_with_meta_arg():
...
 
+def test_qtable_quantity_int_conversion():
+    """Ensure the behavior when converting ``int`` ``Column`` to ``Quantity``, the dtype will be ``float``.
+
+    See https://github.com/astropy/astropy/issues/10964 for the rationale
+    """
+
+    tab = QTable(dict(time=[1, 2, 3]))
+    tab['length'] = [9, 8, 7]
+    with pytest.warns(UserWarning, match="dtype is converted to float"):
```
pllim, line 2270:
```suggestion
    with pytest.warns(AstropyUserWarning, match="dtype is converted to float"):
```

astropy/table/tests/test_table.py
```diff
@@ -2258,6 +2259,46 @@ def test_table_meta_copy_with_meta_arg():
...
+        tab['length'].unit = u.m
+    assert np.issubdtype(tab['length'].dtype, np.float)
+
+    # same for dimensionless unit
+    tab['col2'] = [6, 5, 4]
+    with pytest.warns(UserWarning, match="dtype is converted to float"):
```
pllim, line 2276:
```suggestion
    with pytest.warns(AstropyUserWarning, match="dtype is converted to float"):
```

pllim (commented):
Thanks! I left a general review and a thorough review would be done by subpackage maintainers.

As this is technically a behavior change, it needs a change log under "API change". Thanks!

pllim:
p.s. Failures look like they are related to this PR and they need to be fixed. For example:
```
FutureWarning: Conversion of the second argument of issubdtype from `int` to `np.signedinteger` is deprecated. In future, it will be treated as `np.int64 == np.dtype(int).type`.
```

orionlee:
Fixed wording / formatting, and added changelog entry.

The rest of suggested code changes look reasonable. I'm holding off the changes waiting for more feedback.

Specifically, I myself is not completely sold on the new behavior (emit warning during column update / replacement), and would like to get some feedback.

saimn:
ping @mhvk @taldcroft in case you missed this one.

astropy/table/table.py
```diff
@@ -3644,12 +3648,33 @@ def _convert_col_for_table(self, col):
             qcol = q_cls(col.data, col.unit, copy=False)
             qcol.info = col.info
             qcol.info.indices = col.info.indices
+            if np.issubdtype(getattr(col, 'dtype', None), np.int):
```
mhvk, line 3651: Change `np.int` to `np.integer` (which is the actual type class, while `np.int is int`).

astropy/table/table.py
```diff
@@ -3644,12 +3648,33 @@ def _convert_col_for_table(self, col):
...
+                self._dtype_before_conversion = col.dtype
```
mhvk, line 3652: I think logically the warning should be given here. Is there a particular reason not to do that?

It is risky to use some outside attribute to communicate between methods, so would be good to avoid it!

astropy/table/table.py
mhvk: Remind me: is `AstropyUserWarning` just shown once by default, or every time? I'm not sure we want the every time behaviour here - people will just stop looking at warnings if they happen too frequently...

mhvk (commented):
Definitely can see the benefit of a warning here, but a question about the approach.

astropy/table/table.py
orionlee: I apologize for the long delay.

Using the attribute is indeed hacky, the part that I am most uneasy about in this PR.

The reason that a warning cannot be issued in here in `_convert_col_for_table()` is that it is used in various other scenarios, in addition to the use case conversion in unit assignment `tab['a'].unit = 'km'` this PR tried to address.

E.g., if a warning is issued here, the following codes will also emit warning, even though from an user's perspective, there is no conversion taken place. `t['b']` is consistently a float-based Quantity. 

```python
t = QTable([[1, 2], [3, 4]], names=['a', 'b'])
t['b'] = Column([4, 5], unit='cm')
```

Any idea on how to address this?

pllim:
There are conflicts now. Also, we have switched the way we do change log, so you have to undo changes to `CHANGES.rst` and follow instructions at https://github.com/astropy/astropy/blob/main/docs/changes/README.rst

orionlee:
@pllim I'd like to gauge some feedback before making more commits that would take care of conflicts, etc, as I am not entirely comfortable with the PR. See https://github.com/astropy/astropy/pull/10992#discussion_r622436271

pllim:
In that case, given the impending feature freeze, I'll take the liberty to move milestone. Thanks for your understanding!

orionlee:
@pilim One thing we could safely do is to document the existing behavior (without issuing warnings). I could create another PR for that in the next day or so. Let me know if you'd like me to do so.

pllim:
> document the existing behavior (without issuing warnings)

@orionlee , yes, that would be nice to have until this PR can progress in the future. Thank you!

Issue #10887, opened 2020-10-17
Implement direct ICRS -> Observed transform


StuartLittlefair:
### Description

After years of digging around in the coordinates code I had somehow never realised that we do not implement a direct ICRS -> ALTAZ transform, despite this being perhaps the most commonly used transformation "in the wild". Instead, we apply ICRS -> CIRS -> ALTAZ. This is the way it is done internally inside ```erfa``` of course, but there is overhead in doing it this way from the creation of ```SkyCoord``` objects and the like.

We should implement the direct transformation within ```erfa``` for a speedup of around 6x.

### Additional context

A crude attempt at doing this is the following snippet:

```python
import erfa
import numpy as np
from astropy import units as u
from astropy.coordinates.representation import UnitSphericalRepresentation
from astropy.coordinates.transformations import FunctionTransformWithFiniteDifference
from astropy.coordinates.builtin_frames import ICRS, AltAz
from astropy.coordinates.builtin_frames.utils import (get_jd12,
                                                      get_polar_motion,
                                                      get_cip,
                                                      prepare_earth_position_vel)
from astropy.coordinates.baseframe import frame_transform_graph


@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, ICRS, AltAz)
def icrs_to_altaz(icrs_coo, altaz_frame):

    jd1_tt, jd2_tt = get_jd12(altaz_frame.obstime, 'tt')
    xp, yp = get_polar_motion(altaz_frame.obstime)
    x, y, s = get_cip(jd1_tt, jd2_tt)
    era = erfa.era00(*get_jd12(altaz_frame.obstime, 'ut1'))
    earth_pv, earth_heliocentric = prepare_earth_position_vel(altaz_frame.obstime)

    # refraction constants
    refa, refb = erfa.refco(
        altaz_frame.pressure.to_value(u.hPa),
        altaz_frame.temperature.to_value(u.deg_C),
        # Relative humidity can be a quantity or a number.
        u.Quantity(altaz_frame.relative_humidity).value,
        altaz_frame.obswl.to_value(u.micron)
    )

    astrom = erfa.apco(
        jd1_tt, jd2_tt, earth_pv, earth_heliocentric, x, y, s, era,
        altaz_frame.location.lon.to_value(u.radian),
        altaz_frame.location.lat.to_value(u.radian),
        altaz_frame.location.height.to_value(u.m),
        xp, yp, s, refa, refb
    )

[...]

PR #10994, opened 2020-11-05
Add topocentric CIRS and fix the CIRS to AltAz transformations

### LINKED ISSUE #10356: CIRS is not at the ultimate accuracy

StuartLittlefair:
### Description

As pointed out by @mkbrewer, our lack of topocentric CIRS frames meant that transformations into the AltAz
frames did not properly handle aberration and therefore could be incorrect by several arcseconds for nearby objects like the moon, and up to three degrees for the ridiculous case of an object 10km above the Earth's surface.

This PR adds a topocentric position to the CIRS frame and provides accurate CIRS transformations such that Alt-Az is now comparable to MKB's code to hundredths of a milli-arcsecond, and compares to Skyfield to accuracies of tens of milli-arcseconds, if one disables polar motion.

The cost of this precision is a rather severe slowdown in the transformation code - approximately a factor of 10. This is because the calls to ```erfa.apcs``` and calculation of the celestial-to-intermediate matrix are more expensive that the calls to ```erfa.apci``` we use in the master branch. On the face of it, this trade-off between speed and accuracy would be a challenge for us, but I believe we can sidestep it by implementing the direct ICRS->AltAz transform as requested in #10887, which will make the ICRS->AltAz transform faster than the current master.

An implementation detail that is open for discussion: I chose to specify observer location for CIRS frames using the GCRS position and velocity. This is to make transforms easy, since attributes can be shared between GCRS and CIRS, but also to make it easy to define a frame using `EarthLocation.get_gcrs_posvel`. This is similar to the recently implemented TETE frame.

Pinging @maxnoe here as this touches his Erfa interpolation code.

Fixes #10356 and the ancient #4269.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,8 +16,9 @@ astropy.convolution
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
 
-- Adds the ability to create topocentric ``CIRS`` frames, improving accuracy of
-  ``AltAz`` calculations to of order milli-arcsecond. [#10994]
+- Adds the ability to create topocentric ``CIRS`` frames. Using these,
+  ``AltAz`` calculations are now accurate down to the milli-arcsecond
+  level. [#10994]
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ astropy.convolution
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
 
+- Adds the ability to create topocentric ``CIRS`` frames, improving accuracy of
+  ``AltAz`` calculations to of order milli-arcsecond. [#10994]
+
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 
--- a/astropy/coordinates/builtin_frames/cirs.py
+++ b/astropy/coordinates/builtin_frames/cirs.py
@@ -4,8 +4,9 @@
 from astropy import units as u
 from astropy.utils.decorators import format_doc
 from astropy.coordinates.attributes import (TimeAttribute,
-                                            CartesianRepresentationAttribute)
+                                            EarthLocationAttribute)
 from astropy.coordinates.baseframe import base_doc
+from astropy.coordinates.earth import EarthLocation
 from .baseradec import doc_components, BaseRADecFrame
 from .utils import DEFAULT_OBSTIME
 
@@ -18,18 +19,10 @@ doc_footer = """
     obstime : `~astropy.time.Time`
         The time at which the observation is taken.  Used for determining the
         position of the Earth and its precession
-    obsgeoloc : `~astropy.coordinates.CartesianRepresentation`, `~astropy.units.Quantity`
-        The position of the observer relative to the center-of-mass of the
-        Earth, oriented the same as GCRS. Either [0, 0, 0],
-        `~astropy.coordinates.CartesianRepresentation`, or proper input for one,
-        i.e., a `~astropy.units.Quantity` with shape (3, ...) and length units.
-        Defaults to [0, 0, 0], meaning a geocentric observer.
-    obsgeovel : `~astropy.coordinates.CartesianRepresentation`, `~astropy.units.Quantity`
-        The velocity of the observer relative to the center-of-mass of the
-        Earth, oriented the same as GCRS. Either [0, 0, 0],
-        `~astropy.coordinates.CartesianRepresentation`, or proper input for one,
-        i.e., a `~astropy.units.Quantity` with shape (3, ...) and velocity
-        units.  Defaults to [0, 0, 0], meaning a geocentric observer.
+    location : `~astropy.coordinates.EarthLocation`
+        The location on the Earth.  This can be specified either as an
+        `~astropy.coordinates.EarthLocation` object or as anything that can be
+        transformed to an `~astropy.coordinates.ITRS` frame.
 """
 
 
@@ -42,10 +35,7 @@ class CIRS(BaseRADecFrame):
     """
 
     obstime = TimeAttribute(default=DEFAULT_OBSTIME)
-    obsgeoloc = CartesianRepresentationAttribute(default=[0, 0, 0],
-                                                 unit=u.m)
-    obsgeovel = CartesianRepresentationAttribute(default=[0, 0, 0],
-                                                 unit=u.m/u.s)
+    location = EarthLocationAttribute(default=EarthLocation(0*u.km, 0*u.km, 0*u.km))
 
 # The "self-transform" is defined in icrs_cirs_transformations.py, because in
 # the current implementation it goes through ICRS (like GCRS)
--- a/astropy/coordinates/builtin_frames/cirs.py
+++ b/astropy/coordinates/builtin_frames/cirs.py
@@ -1,14 +1,12 @@
 # -*- coding: utf-8 -*-
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-from astropy import units as u
 from astropy.utils.decorators import format_doc
 from astropy.coordinates.attributes import (TimeAttribute,
                                             EarthLocationAttribute)
 from astropy.coordinates.baseframe import base_doc
-from astropy.coordinates.earth import EarthLocation
 from .baseradec import doc_components, BaseRADecFrame
-from .utils import DEFAULT_OBSTIME
+from .utils import DEFAULT_OBSTIME, EARTH_CENTER
 
 __all__ = ['CIRS']
 
@@ -36,7 +34,7 @@ class CIRS(BaseRADecFrame):
     """
 
     obstime = TimeAttribute(default=DEFAULT_OBSTIME)
-    location = EarthLocationAttribute(default=EarthLocation(0*u.km, 0*u.km, 0*u.km))
+    location = EarthLocationAttribute(default=EARTH_CENTER)
 
 # The "self-transform" is defined in icrs_cirs_transformations.py, because in
 # the current implementation it goes through ICRS (like GCRS)
--- a/astropy/coordinates/builtin_frames/cirs.py
+++ b/astropy/coordinates/builtin_frames/cirs.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+from astropy import units as u
 from astropy.utils.decorators import format_doc
-from astropy.coordinates.attributes import TimeAttribute
+from astropy.coordinates.attributes import (TimeAttribute,
+                                            CartesianRepresentationAttribute)
 from astropy.coordinates.baseframe import base_doc
 from .baseradec import doc_components, BaseRADecFrame
 from .utils import DEFAULT_OBSTIME
@@ -15,7 +17,19 @@ doc_footer = """
     ----------------
     obstime : `~astropy.time.Time`
         The time at which the observation is taken.  Used for determining the
-        position of the Earth and its precession.
+        position of the Earth and its precession
+    obsgeoloc : `~astropy.coordinates.CartesianRepresentation`, `~astropy.units.Quantity`
+        The position of the observer relative to the center-of-mass of the
+        Earth, oriented the same as GCRS. Either [0, 0, 0],
+        `~astropy.coordinates.CartesianRepresentation`, or proper input for one,
+        i.e., a `~astropy.units.Quantity` with shape (3, ...) and length units.
+        Defaults to [0, 0, 0], meaning a geocentric observer.
+    obsgeovel : `~astropy.coordinates.CartesianRepresentation`, `~astropy.units.Quantity`
+        The velocity of the observer relative to the center-of-mass of the
+        Earth, oriented the same as GCRS. Either [0, 0, 0],
+        `~astropy.coordinates.CartesianRepresentation`, or proper input for one,
+        i.e., a `~astropy.units.Quantity` with shape (3, ...) and velocity
+        units.  Defaults to [0, 0, 0], meaning a geocentric observer.
 """
 
 
@@ -28,7 +42,10 @@ class CIRS(BaseRADecFrame):
     """
 
     obstime = TimeAttribute(default=DEFAULT_OBSTIME)
-
+    obsgeoloc = CartesianRepresentationAttribute(default=[0, 0, 0],
+                                                 unit=u.m)
+    obsgeovel = CartesianRepresentationAttribute(default=[0, 0, 0],
+                                                 unit=u.m/u.s)
 
 # The "self-transform" is defined in icrs_cirs_transformations.py, because in
 # the current implementation it goes through ICRS (like GCRS)
--- a/astropy/coordinates/builtin_frames/cirs.py
+++ b/astropy/coordinates/builtin_frames/cirs.py
@@ -18,11 +18,12 @@ doc_footer = """
     ----------------
     obstime : `~astropy.time.Time`
         The time at which the observation is taken.  Used for determining the
-        position of the Earth and its precession
+        position of the Earth and its precession.
     location : `~astropy.coordinates.EarthLocation`
         The location on the Earth.  This can be specified either as an
         `~astropy.coordinates.EarthLocation` object or as anything that can be
-        transformed to an `~astropy.coordinates.ITRS` frame.
+        transformed to an `~astropy.coordinates.ITRS` frame. The default is the
+        centre of the Earth.
 """
 
 
--- a/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
@@ -22,11 +22,13 @@ from ..erfa_astrom import erfa_astrom
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, AltAz)
 def cirs_to_altaz(cirs_coo, altaz_frame):
-    if np.any(cirs_coo.obstime != altaz_frame.obstime):
-        # the only frame attribute for the current CIRS is the obstime, but this
-        # would need to be updated if a future change allowed specifying an
-        # Earth location algorithm or something
-        cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime))
+
+    obsgeoloc, obsgeovel = altaz_frame.location.get_gcrs_posvel(altaz_frame.obstime)
+    if (np.any(cirs_coo.obstime != altaz_frame.obstime) or
+            np.any(cirs_coo.obsgeoloc != obsgeoloc)):
+        cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime,
+                                              obsgeoloc=obsgeoloc,
+                                              obsgeovel=obsgeovel))
 
     # we use the same obstime everywhere now that we know they're the same
     obstime = cirs_coo.obstime
@@ -35,22 +37,15 @@ def cirs_to_altaz(cirs_coo, altaz_frame):
     is_unitspherical = (isinstance(cirs_coo.data, UnitSphericalRepresentation) or
                         cirs_coo.cartesian.x.unit == u.one)
 
-    if is_unitspherical:
-        usrepr = cirs_coo.represent_as(UnitSphericalRepresentation)
-        cirs_ra = usrepr.lon.to_value(u.radian)
-        cirs_dec = usrepr.lat.to_value(u.radian)
-    else:
-        # compute an "astrometric" ra/dec -i.e., the direction of the
-        # displacement vector from the observer to the target in CIRS
-        loccirs = altaz_frame.location.get_itrs(cirs_coo.obstime).transform_to(cirs_coo)
-        diffrepr = (cirs_coo.cartesian - loccirs.cartesian).represent_as(UnitSphericalRepresentation)
-
-        cirs_ra = diffrepr.lon.to_value(u.radian)
-        cirs_dec = diffrepr.lat.to_value(u.radian)
+    # We used to do "astrometric" corrections here, but these are no longer necesssary
+    # CIRS has proper topocentric behaviour
+    usrepr = cirs_coo.represent_as(UnitSphericalRepresentation)
+    cirs_ra = usrepr.lon.to_value(u.radian)
+    cirs_dec = usrepr.lat.to_value(u.radian)
 
     # first set up the astrometry context for CIRS<->AltAz
     astrom = erfa_astrom.get().apio13(altaz_frame)
-
+    astrom['diurab'] = 0
     az, zen, _, _, _ = erfa.atioq(cirs_ra, cirs_dec, astrom)
 
     if is_unitspherical:
@@ -77,23 +72,20 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
 
     # first set up the astrometry context for ICRS<->CIRS at the altaz_coo time
     astrom = erfa_astrom.get().apio13(altaz_coo)
+    astrom['diurab'] = 0
 
     # the 'A' indicates zen/az inputs
     cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
+    obsgeoloc, obsgeovel = altaz_coo.location.get_gcrs_posvel(altaz_coo.obstime)
     if isinstance(altaz_coo.data, UnitSphericalRepresentation) or altaz_coo.cartesian.x.unit == u.one:
-        cirs_at_aa_time = CIRS(ra=cirs_ra, dec=cirs_dec, distance=None,
-                               obstime=altaz_coo.obstime)
+        distance = None
     else:
-        # treat the output of atoiq as an "astrometric" RA/DEC, so to get the
-        # actual RA/Dec from the observers vantage point, we have to reverse
-        # the vector operation of cirs_to_altaz (see there for more detail)
+        distance = altaz_coo.distance
 
-        loccirs = altaz_coo.location.get_itrs(altaz_coo.obstime).transform_to(cirs_frame)
-
-        astrometric_rep = SphericalRepresentation(lon=cirs_ra, lat=cirs_dec,
-                                                  distance=altaz_coo.distance)
-        newrepr = astrometric_rep + loccirs.cartesian
-        cirs_at_aa_time = CIRS(newrepr, obstime=altaz_coo.obstime)
+    cirs_at_aa_time = CIRS(ra=cirs_ra, dec=cirs_dec, distance=distance,
+                           obstime=altaz_coo.obstime,
+                           obsgeoloc=obsgeoloc,
+                           obsgeovel=obsgeovel)
 
     # this final transform may be a no-op if the obstimes are the same
     return cirs_at_aa_time.transform_to(cirs_frame)
--- a/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
@@ -45,6 +45,10 @@ def cirs_to_altaz(cirs_coo, altaz_frame):
 
     # first set up the astrometry context for CIRS<->AltAz
     astrom = erfa_astrom.get().apio13(altaz_frame)
+    # Note: this is inefficient. Really we should write our own apio13 version and
+    # strip out pvobs calls. However, we don't need a diurnal aberration/parallax
+    # term because we have already accounted for that in transforming to topocentric
+    # CIRS.
     astrom['diurab'] = 0
     az, zen, _, _, _ = erfa.atioq(cirs_ra, cirs_dec, astrom)
 
@@ -53,13 +57,10 @@ def cirs_to_altaz(cirs_coo, altaz_frame):
                                           lon=u.Quantity(az, u.radian, copy=False),
                                           copy=False)
     else:
-        # now we get the distance as the cartesian distance from the earth
-        # location to the coordinate location
-        locitrs = altaz_frame.location.get_itrs(obstime)
-        distance = locitrs.separation_3d(cirs_coo)
+        # since we've transformed to CIRS at the observatory location, just use CIRS distance
         rep = SphericalRepresentation(lat=u.Quantity(PIOVER2 - zen, u.radian, copy=False),
                                       lon=u.Quantity(az, u.radian, copy=False),
-                                      distance=distance,
+                                      distance=cirs_coo.distance,
                                       copy=False)
     return altaz_frame.realize_frame(rep)
 
@@ -87,7 +88,7 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
                            obsgeoloc=obsgeoloc,
                            obsgeovel=obsgeovel)
 
-    # this final transform may be a no-op if the obstimes are the same
+    # this final transform may be a no-op if the obstimes and locations are the same
     return cirs_at_aa_time.transform_to(cirs_frame)
 
 
--- a/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
@@ -22,13 +22,10 @@ from ..erfa_astrom import erfa_astrom
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, AltAz)
 def cirs_to_altaz(cirs_coo, altaz_frame):
-
-    obsgeoloc, obsgeovel = altaz_frame.location.get_gcrs_posvel(altaz_frame.obstime)
-    if (np.any(cirs_coo.obstime != altaz_frame.obstime) or
-            np.any(cirs_coo.obsgeoloc != obsgeoloc)):
+    if (altaz_frame.location != cirs_coo.location or
+            np.any(cirs_coo.obstime != altaz_frame.obstime)):
         cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime,
-                                              obsgeoloc=obsgeoloc,
-                                              obsgeovel=obsgeovel))
+                                              location=altaz_frame.location))
 
     # if the data are UnitSphericalRepresentation, we can skip the distance calculations
     is_unitspherical = (isinstance(cirs_coo.data, UnitSphericalRepresentation) or
@@ -68,7 +65,6 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
 
     # the 'A' indicates zen/az inputs
     cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
-    obsgeoloc, obsgeovel = altaz_coo.location.get_gcrs_posvel(altaz_coo.obstime)
     if isinstance(altaz_coo.data, UnitSphericalRepresentation) or altaz_coo.cartesian.x.unit == u.one:
         distance = None
     else:
@@ -76,8 +72,7 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
 
     cirs_at_aa_time = CIRS(ra=cirs_ra, dec=cirs_dec, distance=distance,
                            obstime=altaz_coo.obstime,
-                           obsgeoloc=obsgeoloc,
-                           obsgeovel=obsgeovel)
+                           location=altaz_coo.location)
 
     # this final transform may be a no-op if the obstimes and locations are the same
     return cirs_at_aa_time.transform_to(cirs_frame)
--- a/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
@@ -22,7 +22,7 @@ from ..erfa_astrom import erfa_astrom
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, AltAz)
 def cirs_to_altaz(cirs_coo, altaz_frame):
-    if (altaz_frame.location != cirs_coo.location or
+    if (np.any(altaz_frame.location != cirs_coo.location) or
             np.any(cirs_coo.obstime != altaz_frame.obstime)):
         cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime,
                                               location=altaz_frame.location))
@@ -64,7 +64,7 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
     astrom = erfa_astrom.get().apio(altaz_coo)
 
     # the 'A' indicates zen/az inputs
-    cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
+    cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom) << u.radian
     if isinstance(altaz_coo.data, UnitSphericalRepresentation) or altaz_coo.cartesian.x.unit == u.one:
         distance = None
     else:
--- a/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
@@ -41,7 +41,7 @@ def cirs_to_altaz(cirs_coo, altaz_frame):
     cirs_dec = usrepr.lat.to_value(u.radian)
 
     # first set up the astrometry context for CIRS<->AltAz
-    astrom = erfa_astrom.get().apio13(altaz_frame)
+    astrom = erfa_astrom.get().apio(altaz_frame)
     az, zen, _, _, _ = erfa.atioq(cirs_ra, cirs_dec, astrom)
 
     if is_unitspherical:
@@ -64,7 +64,7 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
     zen = PIOVER2 - usrepr.lat.to_value(u.radian)
 
     # first set up the astrometry context for ICRS<->CIRS at the altaz_coo time
-    astrom = erfa_astrom.get().apio13(altaz_coo)
+    astrom = erfa_astrom.get().apio(altaz_coo)
 
     # the 'A' indicates zen/az inputs
     cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
--- a/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/cirs_observed_transforms.py
@@ -30,9 +30,6 @@ def cirs_to_altaz(cirs_coo, altaz_frame):
                                               obsgeoloc=obsgeoloc,
                                               obsgeovel=obsgeovel))
 
-    # we use the same obstime everywhere now that we know they're the same
-    obstime = cirs_coo.obstime
-
     # if the data are UnitSphericalRepresentation, we can skip the distance calculations
     is_unitspherical = (isinstance(cirs_coo.data, UnitSphericalRepresentation) or
                         cirs_coo.cartesian.x.unit == u.one)
@@ -45,11 +42,6 @@ def cirs_to_altaz(cirs_coo, altaz_frame):
 
     # first set up the astrometry context for CIRS<->AltAz
     astrom = erfa_astrom.get().apio13(altaz_frame)
-    # Note: this is inefficient. Really we should write our own apio13 version and
-    # strip out pvobs calls. However, we don't need a diurnal aberration/parallax
-    # term because we have already accounted for that in transforming to topocentric
-    # CIRS.
-    astrom['diurab'] = 0
     az, zen, _, _, _ = erfa.atioq(cirs_ra, cirs_dec, astrom)
 
     if is_unitspherical:
@@ -73,7 +65,6 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
 
     # first set up the astrometry context for ICRS<->CIRS at the altaz_coo time
     astrom = erfa_astrom.get().apio13(altaz_coo)
-    astrom['diurab'] = 0
 
     # the 'A' indicates zen/az inputs
     cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -34,7 +34,7 @@ from ..erfa_astrom import erfa_astrom
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, ICRS, CIRS)
 def icrs_to_cirs(icrs_coo, cirs_frame):
     # first set up the astrometry context for ICRS<->CIRS
-    astrom = erfa_astrom.get().apci(cirs_frame)
+    astrom = erfa_astrom.get().apco(cirs_frame)
 
     if icrs_coo.data.get_name() == 'unitspherical' or icrs_coo.data.to_cartesian().x.unit == u.one:
         # if no distance, just do the infinite-distance/no parallax calculation
@@ -66,7 +66,7 @@ def icrs_to_cirs(icrs_coo, cirs_frame):
 def cirs_to_icrs(cirs_coo, icrs_frame):
     # set up the astrometry context for ICRS<->cirs and then convert to
     # astrometric coordinate direction
-    astrom = erfa_astrom.get().apci(cirs_coo)
+    astrom = erfa_astrom.get().apco(cirs_coo)
     srepr = cirs_coo.represent_as(SphericalRepresentation)
     i_ra, i_dec = aticq(srepr.without_differentials(), astrom)
 
@@ -96,8 +96,8 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, CIRS)
 def cirs_to_cirs(from_coo, to_frame):
-    if (np.all(from_coo.obstime == to_frame.obstime) and
-            np.all(from_coo.obsgeoloc == to_frame.obsgeoloc)):
+    if (from_coo.location == to_frame.location and
+            np.all(from_coo.obstime == to_frame.obstime)):
         return to_frame.realize_frame(from_coo.data)
     else:
         # the CIRS<-> CIRS transform actually goes through ICRS.  This has a
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -96,16 +96,15 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, CIRS)
 def cirs_to_cirs(from_coo, to_frame):
-    if np.all(from_coo.obstime == to_frame.obstime):
+    if (np.all(from_coo.obstime == to_frame.obstime) and
+            np.all(from_coo.obsgeoloc == to_frame.obsgeoloc)):
         return to_frame.realize_frame(from_coo.data)
     else:
-        # the CIRS<-> CIRS transform actually goes through ICRS.  This has a
-        # subtle implication that a point in CIRS is uniquely determined
-        # by the corresponding astrometric ICRS coordinate *at its
-        # current time*.  This has some subtle implications in terms of GR, but
-        # is sort of glossed over in the current scheme because we are dropping
-        # distances anyway.
-        return from_coo.transform_to(ICRS()).transform_to(to_frame)
+        # the CIRS<-> CIRS transform actually goes through GCRS. Previously
+        # this went through ICRS, but this seems unecessary since the
+        # transform between GCRS and ICRS for the same topocentric position
+        # is a simple matrix transform.
+        return from_coo.transform_to(GCRS()).transform_to(to_frame)
 
 
 # Now the GCRS-related transforms to/from ICRS
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -100,14 +100,16 @@ def cirs_to_cirs(from_coo, to_frame):
             np.all(from_coo.obsgeoloc == to_frame.obsgeoloc)):
         return to_frame.realize_frame(from_coo.data)
     else:
-        # the CIRS<-> CIRS transform actually goes through GCRS. Previously
-        # this went through ICRS, but this seems unecessary since the
-        # transform between GCRS and ICRS for the same topocentric position
-        # is a simple matrix transform.
-        return from_coo.transform_to(GCRS()).transform_to(to_frame)
-
+        # the CIRS<-> CIRS transform actually goes through ICRS.  This has a
+        # subtle implication that a point in CIRS is uniquely determined
+        # by the corresponding astrometric ICRS coordinate *at its
+        # current time*.  This has some subtle implications in terms of GR, but
+        # is sort of glossed over in the current scheme because we are dropping
+        # distances anyway.
+        return from_coo.transform_to(ICRS()).transform_to(to_frame)
 
 # Now the GCRS-related transforms to/from ICRS
+
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, ICRS, GCRS)
 def icrs_to_gcrs(icrs_coo, gcrs_frame):
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -108,8 +108,8 @@ def cirs_to_cirs(from_coo, to_frame):
         # distances anyway.
         return from_coo.transform_to(ICRS()).transform_to(to_frame)
 
-# Now the GCRS-related transforms to/from ICRS
 
+# Now the GCRS-related transforms to/from ICRS
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, ICRS, GCRS)
 def icrs_to_gcrs(icrs_coo, gcrs_frame):
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -96,7 +96,7 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, CIRS)
 def cirs_to_cirs(from_coo, to_frame):
-    if (from_coo.location == to_frame.location and
+    if (np.all(from_coo.location == to_frame.location) and
             np.all(from_coo.obstime == to_frame.obstime)):
         return to_frame.realize_frame(from_coo.data)
     else:
--- a/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
+++ b/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
@@ -13,13 +13,13 @@ import astropy.units as u
 from astropy.coordinates.baseframe import frame_transform_graph
 from astropy.coordinates.transformations import FunctionTransformWithFiniteDifference
 from astropy.coordinates.matrix_utilities import matrix_transpose
-from astropy.coordinates.earth import EarthLocation
+
 
 from .gcrs import GCRS, PrecessedGeocentric
 from .cirs import CIRS
 from .itrs import ITRS
 from .equatorial import TEME, TETE
-from .utils import get_polar_motion, get_jd12
+from .utils import get_polar_motion, get_jd12, EARTH_CENTER
 
 # # first define helper functions
 
@@ -194,7 +194,7 @@ def cirs_to_gcrs(cirs_coo, gcrs_frame):
 def cirs_to_itrs(cirs_coo, itrs_frame):
     # first get us to geocentric CIRS at the target obstime
     cirs_coo2 = cirs_coo.transform_to(CIRS(obstime=itrs_frame.obstime,
-                                           location=EarthLocation(0*u.km, 0*u.km, 0*u.km)))
+                                           location=EARTH_CENTER))
 
     # now get the pmatrix
     pmat = cirs_to_itrs_mat(itrs_frame.obstime)
--- a/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
+++ b/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
@@ -13,6 +13,7 @@ import astropy.units as u
 from astropy.coordinates.baseframe import frame_transform_graph
 from astropy.coordinates.transformations import FunctionTransformWithFiniteDifference
 from astropy.coordinates.matrix_utilities import matrix_transpose
+from astropy.coordinates.earth import EarthLocation
 
 from .gcrs import GCRS, PrecessedGeocentric
 from .cirs import CIRS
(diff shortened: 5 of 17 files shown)


maxnoe:
I just want to comment that while I think this is great, I also think this should not make it into a release without the faster, direct ICRS to AltAz transformation.

A new release with slower transformation between ICRS and AltAz to gain some arcseconds for close-by objects would be catastrophic for us who are mostly interested in fast, not terribly precises ICRS to AltAz transforms for FAR away objects.

StuartLittlefair:
Totally agree @maxnoe

StuartLittlefair:
These test failures look unrelated - a timeout on readthedocs, and the coverage test is failing due to stale code - that I am leaving in place since I intend to use it in a later PR. I'm not sure why the windows build on travis failed to tear down properly.

mhvk:
Ideally, we keep the implementation of ICRS<->AltAz decoupled from this PR (which I have not looked at but suspect will not be super-simple!), and it could be either in parallel or after - since we have quite a while before 4.3, either would seem fine.

StuartLittlefair:
I think it is best to leave the ICRS transform to another PR. There’s enough to review here already, although it’s not that complicated in the end given how long I’ve spent on it.

mkbrewer:
> The cost of this precision is a rather severe slowdown in the transformation code - approximately a factor of 10. 

I am curious about this slowdown. Is this in the ICRS<->CIRS or CIRS<->AltAz transforms? I don't see it being in the former as not much has really changed there. 

> This is because the calls to erfa.apcs and calculation of the celestial-to-intermediate matrix are more expensive that the calls to erfa.apci we use in the master branch

This isn't true. Calculation of the celestial to intermediate matrix has always been required as it is needed by erfa.apci. Then there's the fact that erfa.apci ends up calling erfa.apcs anyway. 

If the slowdown is in the CIRS<->AltAz transformations, this is because of the call to `altaz_coo.location.get_gcrs_posvel()` necessitated by your choice of GCRS position and velocity to specify observer location for CIRS frames. While I completely understand the convenience and consistency of this approach, that call does require the costly calculation of the celestial to intermediate matrix. An alternate approach would be to use CIRS position and velocity for the observer location for CIRS frames. Then an `Earthlocation.get_cirs_posvel()` function could be written for use here that would be much less costly. Using CIRS position and velocity wouldn't be all that inconvenient either. In the GCRS<->CIRS transforms, it is necessary to calculate the celestial to intermediate matrix anyway, so this could be applied to the observer's position and velocity also during these transformations. In the ICRS<->CIRS transforms, the observer's GCRS position and velocity required by the call to erfa.apcs could also be easily obtained by using the celestial to intermediate matrix which is also calculated in `erfa_astrom.get().apci()` anyway.

astropy/coordinates/erfa_astrom.py
```diff
@@ -105,6 +105,12 @@ def apio13(frame_or_coord):
             frame_or_coord.relative_humidity.value,
             frame_or_coord.obswl.value,
         )
+        # Note: this is inefficient. Really we should write our own apio13 version and
+        # strip out pvobs calls. However, we don't need a diurnal aberration/parallax
+        # term because we have already accounted for that in transforming to topocentric
+        # CIRS.
```
mkbrewer, line 111: I would suggest that you go ahead an do this as it only requires changing two lines of code in erfa.apio(). Otherwise, inertia may set in and might never get done.

StuartLittlefair: The main thing stopping me was not being able to work out how to make an empty astrom structure

StuartLittlefair:
@mkbrewer, you are wrong about the cause of the slowdown, but then again so am I. 

The way it is coded is slightly slower but the bulk of the slowdown comes from the fact that CIRS->AltAz contains an ICRS self transform to get topocentric CIRS. 

So what was a fairly direct ICRS->CIRS->AA becomes ICRS->CIRS->ICRS->CIRS’->AA so there is double the number of actual transforms, each of which is slower.

astropy/coordinates/erfa_astrom.py
mhvk: I still have to actually look at the PR, but making an empty structure is easy:
```
astrom = np.empty(shape, dtype=erfa.dt_eraASTROM)
```

StuartLittlefair: perfect - obvious now I see it. This might save some microseconds

StuartLittlefair:
Looking into the slowdown further, there is something *very* odd going on here, which I'll have to dig into

```
import numpy as np
from astropy.time import Time
import astropy.units as u
from astropy.coordinates import EarthLocation, CIRS, AltAz, SkyCoord
from astropy.coordinates.tests.utils import randomly_sample_sphere

# lots of times
times = Time('2020-08-21T00:00') + np.linspace(-3*u.hour, 3*u.hour, 100)
location = EarthLocation.of_site('lapalma')

# aa frame at lots of times
aa_frame = AltAz(obstime=times[:, np.newaxis], location=location)
ra, dec, _ = randomly_sample_sphere(100)
coos = SkyCoord(ra, dec)

aa = coos.transform_to(aa_frame) # 3.2 sec (60 msec on master)
aa = coos.transform_to(CIRS(obstime=times)).transform_to(aa_frame) # 320 msec
```

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
```diff
@@ -22,11 +22,13 @@
...
-    if np.any(cirs_coo.obstime != altaz_frame.obstime):
-        # the only frame attribute for the current CIRS is the obstime, but this
-        # would need to be updated if a future change allowed specifying an
-        # Earth location algorithm or something
-        cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime))
+
+    obsgeoloc, obsgeovel = altaz_frame.location.get_gcrs_posvel(altaz_frame.obstime)
+    if (np.any(cirs_coo.obstime != altaz_frame.obstime) or
+            np.any(cirs_coo.obsgeoloc != obsgeoloc):
+        cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime,
```
mkbrewer, line 29: One way to speed up a straight through CIRS->AltAz transformation would be to allow the AltAz `obstime` and `location` to be `None` which would then just use the CIRS frame attributes.

StuartLittlefair:
There is something about the speed here that I don't understand - if you make a large number of times and ICRS coordinates and broadcast them to make an altaz frame then the second line of code below is about 18x faster than the first... 

```python
coos.transform_to(aa_frame)
coos.transform_to(CIRS()).transform_to(aa_frame)
```

mkbrewer:
> The way it is coded is slightly slower but the bulk of the slowdown comes from the fact that CIRS->AltAz contains an ICRS self transform to get topocentric CIRS.

Is this true in the case of a solar system body where one already has a topocentric GCRS position? Or is this just in the case where one is asking for the AltAz coordinates of an ICRS SkyCoord where the time and location are necessarily derived from the AltAz frame itself. In that case, couldn't you just derive the CIRS `obstime` and `location` from the AltAz frame before leaving the ICRS in the first place? You must already be doing that for`obstime` as an ICRS SkyCoord has no `obstime` attribute.

StuartLittlefair:
@mkbrewer - see my comment above yours. The majority of the slowdown is not in the transforms themselves, but in the astropy coordinate transform machinery...

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
StuartLittlefair: Sorry @mkbrewer - this isn't a very useful suggestion since it requires re-writing all of how astropy works from the ground up. By construction you get an AltAz coordinate at the location and time specified on the AltAz frame.

mkbrewer:
Yeah, I seem to be consistently running one comment behind you. That (N, M) business is way out of my league, but it looks like you're on it.

StuartLittlefair:
Well I know what the problem is, but not how to fix it! You are right about the cause of the slowdown for a single point though

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
mkbrewer: Isn't that because the CIRS frame didn't have a `location` attribute before?

mhvk: In the current framework, we can only define an `AltAz` frame using explicit locations, i.e.,  `AltAz(obsgeoloc=..., obsgeovel=...)`, but there is no reason the code couldn't realize when doing the transform that those are identical to the CIRS ones. 

Maybe more directly similar, I think at least at the `SkyCoord.transform_to` level, we could have at least the string version ('altaz') mean "create an `AltAz` frame with attributes as available on the `SkyCoord`" (at the `BaseFrame.transform_to` level, I think we want to stick to being super explicit). So, I think this could be made to work quite conveniently!

Anyway, would be for a follow-up PR!

StuartLittlefair: It works that way for all transforms. The target frame sets the properties.  What you suggest would change the API, and we don’t do that

mhvk:
@StuartLittlefair - your diagnosis for (N, M) makes sense. Apparently, we broadcast before transforming. That should not be necessary and can probably be solved, but needs some thought... (broadcasting is super powerful but tricky to get exactly right)

Could you raise a separate issue? It does seem quite important generally since presumably this means that single-ICRS-position to AltAz for many times -- surely one of the more common transforms! -- is also much slower than it needs to be.

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
mkbrewer: I see. So it's a matter of philosophy then. I've always wondered why I needed to specify the time and location again when I wanted AltAz from the result of a call to `get_body()`.

mhvk: I think we should distinguish frames from `SkyCoord` here. In frames, everything is explicit, while `SkyCoord` can carry attributes not directly relevant to a given frame (say, `obsgeoloc` for ICRS). I do think it is very possible for the `SkyCoord` output of `get_body()` to pass on the time and location in `.transform_to` - though obviously we actually need to implement it!

mkbrewer: I should think that it would be generally advisable to allow data to flow downstream unimpeded rather than encountering data at the end of a chain that requires the entire chain to be recomputed. I don't really see the application of computing coordinates at one place and time only to change that place and time further downstream, but I must be missing something there.

mhvk: It is really mostly about where data is stored. The `ICRS` class has only coordinates, nothing about an observer, so it cannot transform to, say, `AltAz`, without adding those pieces of information, but it can transform to `AltAz(obstime=..., ...)`.

On the other hand,  `SkyCoord`, can hold extra information.

Of course, really this should all be an implementation detail - `SkyCoord` is the user-facing class. And I think it is actually meant to work on `SkyCoord`, but doesn't - see #10996.

StuartLittlefair:
Thanks @mhvk - I've raised the issue there and will not try to address it as part of this PR.

astropy/coordinates/erfa_astrom.py
StuartLittlefair: done

StuartLittlefair:
I now have a working branch of the direct ICRS->AltAz transform, but it requires many of the same changes to the tests included in this PR, and many of the tests are failing since ICRS->AltAz now gives a different result to ICRS->CIRS->AltAz (because of the innacuracies this PR should fix).

Comments on including that transform as part of this PR?

astropy/coordinates/erfa_astrom.py
```diff
@@ -213,13 +254,20 @@ def apci(self, frame_or_coord):
             For this function, a CIRS frame is expected.
         '''
         obstime = frame_or_coord.obstime
+        # no point in interpolating for a single value
         support = self._get_support_points(obstime)
 
-        cip = self._get_cip(support, obstime)
```
mhvk, line 218: I think this means that the static method `_get_cip` is no longer used...

astropy/coordinates/tests/test_celestial_transformations.py
```diff
@@ -313,3 +313,31 @@ def test_hcrs_icrs_differentials():
...
+    # now check that the distance change is similar to earth radius
+    assert 1000*u.km < np.abs(moon_topo.distance - moon_geo.distance).to(u.au) < 7000*u.km
+
+    # now check that it round-trips
+    moon2 = moon_topo.transform_to(moon_geo)
+    assert_allclose(moon_geo.cartesian.xyz, moon2.cartesian.xyz)
+
+    # now check ICRS transform gives a decent distance from Barycentre
+    moon_icrs = moon_geo.transform_to(ICRS())
+    assert moon_icrs.distance - 1*u.au < 1*u.R_sun
```
mhvk, line 343: Needs absolute value to make it a more secure test. (or use `assert_allclose`)

astropy/coordinates/tests/test_erfa_astrom.py
```diff
@@ -85,8 +85,9 @@ def test_interpolation_nd():
 
         altaz = AltAz(location=fact, obstime=obstime)
```
mhvk, line 86: altaz no longer used?

docs/coordinates/common_errors.rst
```diff
@@ -0,0 +1,93 @@
...
+    >>> from astropy.time import Time
+    >>> from astropy import units as u
+
+    >>> t = Time('J2010')
+    >>> obj = EarthLocation(-1*u.deg, 52*u.deg, height=10.*u.km)
+    >>> home = EarthLocation(-1*u.deg, 52*u.deg, height=0.*u.km)
+    >>> obsloc_gcrs, obsvel_gcrs = home.get_gcrs_posvel(t)
+
+    >>> # An object that appears straight overhead - FOR A GEOCENTRIC OBSERVER.
+    >>> gcrs_geo = obj.get_itrs(t).transform_to(GCRS(obstime=t))
```
mhvk, line 77: Maybe try to keep names a bit more similar, so use `obs` instead of `home`? And then here `obj_gcrs`?

docs/coordinates/common_errors.rst
```diff
@@ -0,0 +1,93 @@
...
+Due to aberration, the actual position of this source will be displaced from its apparent position, by an angle of
+approximately 20.5 arcseconds. Because this source is about one Earth radius away, it's actual position is therefore
+around 600 metres away from where it appears to be. This 600 metre shift, for an object 10 km above the Earth's surface
+is an angular difference of just over three degrees - which is why this object does not appear overhead for a topocentric
+observer - one on the surface of the Earth.
+
+The correct way to construct a |SkyCoord| object for a source that is directly overhead a topocentric observer is
+as follows::
+
+    >>> from astropy.coordinates import EarthLocation, AltAz
```
mhvk, line 67: This one seems very long compared to the "faulty" one - is there really no better way??

mhvk (commented):
This looks very good. A number of smaller comments. And wondering whether the correct way of doing things in the example could be simplified...

astropy/coordinates/erfa_astrom.py
```diff
@@ -30,11 +30,14 @@ class ErfaAstrom:
...
     erfa functions and return the astrom object.
     '''
-
     @staticmethod
     def apci(frame_or_coord):
         '''
-        Wrapper for ``erfa.apci``, used in conversions CIRS <-> ICRS
+        Prepare astrometry context used in conversions CIRS <-> ICRS
+
+        Despite the name this doesn't actually call erfa.apci, since that does not
```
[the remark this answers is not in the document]
StuartLittlefair: Not sure what we would rename to? Although it's not erfa's `apci` any more, it is a version of `apci` - i.e calculates astrometry for celestial-to-intermediate coords. It's not `apcs` certainly...

astropy/coordinates/erfa_astrom.py
```diff
@@ -69,9 +77,12 @@ def apcs(frame_or_coord):
         return erfa.apcs(jd1_tt, jd2_tt, obs_pv, earth_pv, earth_heliocentric)
 
     @staticmethod
-    def apio13(frame_or_coord):
+    def apio(frame_or_coord):
```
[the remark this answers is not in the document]
StuartLittlefair: I changed this name because this really is a version of `apio`, not `apio13`

docs/coordinates/common_errors.rst
StuartLittlefair: Hmm. I can save one line:

```python
itrs_vec = obj.get_itrs(t).cartesian - home.get_itrs(t).cartesian
obsloc_gcrs, obsvel_gcrs = home.get_gcrs_posvel(t)
topocentric_cirs_frame = CIRS(obstime=t, obsgeoloc=obsloc_gcrs, obsgeovel=obsvel_gcrs)
cirs_vec = itrs_vec.transform(matrix_transpose(cirs_to_itrs_mat(t)))
cirs_topo = topocentric_cirs_frame.realize_frame(cirs_vec)
aa = cirs_topo.transform_to(AltAz(obstime=t, location=home))
```

Not sure it's any clearer?

StuartLittlefair: No, I can do 5 clearer lines:

```python
itrs_vec = obj.get_itrs(t).cartesian - home.get_itrs(t).cartesian
gcrs_vec = ITRS().realize_frame(itrs_vec).transform_to(GCRS(obstime=t))
obsloc_gcrs, obsvel_gcrs = home.get_gcrs_posvel(t)
topo_gcrs = GCRS(gcrs_vec, obstime=t, obsgeoloc=obsloc_gcrs, obsgeovel=obsvel_gcrs)
aa = topo_gcrs.transform_to(AltAz(obstime=t, location=home))
```

astropy/coordinates/erfa_astrom.py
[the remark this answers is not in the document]
mhvk: OK, let's leave it as is, then!

astropy/coordinates/erfa_astrom.py
```diff
@@ -213,13 +254,20 @@ def apci(self, frame_or_coord):
             For this function, a CIRS frame is expected.
         '''
         obstime = frame_or_coord.obstime
+        # no point in interpolating for a single value
```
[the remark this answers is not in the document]
maxnoe: probably a merge artifact. It was there before, but removed in #10833

mkbrewer:
Is the method of specifying the observer's location still up for discussion here? If so, after giving this a lot of thought, I truly believe that the cleanest thing to do would be to use an `EarthLocation`. Not only is this the natural attribute to use in CIRS<->AltAz transformations, but it makes things a bit easier for the user also as it doesn't require a call to `EarthLocation.get_gcrs_posvel()` with its unnecessary attendant overhead. It also follows the way SOFA does things more closely. In this scenario, the `ErfaAstrom` method `apci()` would be replaced with `apco()` like this:
```
    @staticmethod
    def apco(frame_or_coord):
        '''
        Prepare astrometry context used in conversions CIRS <-> ICRS

        Arguments
        ---------
        frame_or_coord: ``astropy.coordinates.BaseCoordinateFrame`` or ``astropy.coordinates.SkyCoord``
            Frame or coordinate instance in the corresponding frame
            for which to calculate the calculate the astrom values.
            For this function, a CIRS frame is expected.
        '''
        jd1_tt, jd2_tt = get_jd12(altaz_frame.obstime, 'tt')
        xp, yp = get_polar_motion(altaz_frame.obstime)
        sp = erfa.sp00(jd1_tt, jd2_tt)
        x, y, s = get_cip(jd1_tt, jd2_tt)
        era = erfa.era00(*get_jd12(altaz_frame.obstime, 'ut1'))
        earth_pv, earth_heliocentric = prepare_earth_position_vel(altaz_frame.obstime)

        astrom = erfa.apco(
            jd1_tt, jd2_tt, earth_pv, earth_heliocentric, x, y, s, era,
            frame_or_coord.location.lon.to_value(u.radian),
            frame_or_coord.location.lat.to_value(u.radian),
            frame_or_coord.location.height.to_value(u.m),
            xp, yp, sp, 0.0, 0.0
        )
        return astrom
```
This makes the intermediate rotation transform `gcrs_to_cirs()` very slightly more complicated in that it would be necessary to transform the ITRS coordinates of the CIRS `EarthLocation` into the GCRS by calling `erfa.pvtob()` followed by multiplication by the transpose of the celestial to intermediate matrix which is required for this transform anyway. Frankly, I don't understand the intermediate rotation transform `cirs_to_gcrs()` as it seems to violate the principle of transforms always taking their attributes from the target frame, but the same could be done there as well.

The same could actually apply to TETE as well. Since this is derived from GCRS, the intermediate rotation transforms supply the celestial to intermediate matrix necessary to convert an `EarthLocation` observer's position attribute there also.

mkbrewer:
There are a few refinements that could be made to this. The calls to `pvtob()` which include a geodetic to geocentric transformation could be eliminated (including the one on `apco()`) by inclusion of `get_cirs_posvel()` (and `get_tete_posvel()`) methods in the `EarthLocation` class. `ErfaAstrom.apco()` could be used in a direct ICRS to AltAz transform with the inclusion of arguments for the refraction constants.

StuartLittlefair:
It's certainly up for discussion @mkbrewer, but this is probably not the best place, since it would have to be a separate PR to make these changes. The discussion at #10994 is probably the best place to talk about it, since the discussion is heading that way. 

I've linked to your comment on that issue.

mkbrewer:
I see that you meant issue #10996 there.

mhvk:
I like @mkbrewer's suggestion of using `location` for all earth-bound systems. It does seem directly relevant to this PR, as this is the PR that introduces new arguments for CIRS and thus a good moment to chose either `obsgeoloc` and `obsgeovel` *or* `location`.

That said, as long as we have not made a release, we are still free to switch, and this PR brings us to a state with known-good results and tests that verify those (hurray!!!). So, I can also see the case for reviewing/merging and then discussing further.

Since it seems most important now just to be sure that we make a "final" decision before 4.3, I set the milestone of #10996 to 4.3.

StuartLittlefair:
I thought of a seperate PR because it makes sense to do the same thing for all the Earth bound systems and it seems like “mission creep” to start doing it here. 

Although if we can find a nice solution first perhaps it makes sense to sort out the location issue first and then update this PR to use the new scheme?

mhvk:
@StuartLittlefair - you're right, we should do any new location definition for all at the same time. As noted, I think it is fine to do this PR first.

mhvk:
The above said, even just for CIRS we can either follow AltAz or TETE - maybe it is more logical to just do `location` here? From a quick look, it would seem to make CIRS->AltAz faster to use `EarthLocation`, while it would make CIRS<->GCRS more complicated. But the former would seem more important...

mkbrewer:
Unless I'm missing something, CIRS<->GCRS wouldn't really be more complicated. Here's what I'm thinking:
```
@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GCRS, CIRS)
def gcrs_to_cirs(gcrs_coo, cirs_frame):
    # First get the pmatrix
    pmat = gcrs_to_cirs_mat(cirs_frame.obstime)

    # Now get us to GCRS at the target obstime and pos/vel
    obscirsloc, obscirsvel = cirs_frame.location.get_cirs_posvel()
    obsgeoloc = erfa.trxp(pmat, obscirsloc)
    obsgeovel = erfa.trxp(pmat, obscirsvel)
    gcrs_coo2 = gcrs_coo.transform_to(GCRS(obstime=cirs_frame.obstime,
                                           obsgeoloc=obsgeoloc,
                                           obsgeovel=obsgeovel))

    crepr = gcrs_coo2.cartesian.transform(pmat)
    return cirs_frame.realize_frame(crepr)
```

mhvk:
@StuartLittlefair - I think @mkbrewer's arguments have more or less convinced me that independent of the larger discussion, it might be better to try to use `location` here for CIRS instead of `obsgeo{loc|vel}`. What do you think? If you like it but don't have time to try to make the change, I could give it a go (though likely only towards the end of the week). It does seem that apart from that, this is ready.

StuartLittlefair:
I think I’ll give it a go in this PR. If nothing else it will give us a feel for moving all the other frames in that direction.

StuartLittlefair:
@mhvk @mkbrewer Using `location` for CIRS is rather neat in the end, and has the neat side effect of getting away from the major slowdowns mentioned in #10997, so these transformations are not *achingly* slow. 

I would support a transition to `location` attributes for TETE as well, although that's a bit annoying since we'll have to deprecate the use of `obsgeoloc` and `obsgeovel`, even though they are probably not used in the wild yet.

As a note to reviewers, especially @maxnoe - I've left the `apci` transforms in `erfa_astrom` even though they are not currently used. The reason is that I am not getting the kind of speed ups from interpolation that those transforms provided and I want to leave them in place until I understand where the bottlenecks are.

StuartLittlefair:
@maxnoe I think this is of similar performance now to before this PR. Running the test case you supplied in https://github.com/astropy/astropy/pull/10647#issuecomment-673569823 I see similar performance to astropy master:

master branch:

```
No Interpolation took 9.7546 s
Interpolation with  10.0 s   resolution took 1.0634 s
Interpolation with   5.0 min resolution took 0.2984 s
Interpolation with  30.0 min resolution took 0.2053 s
```

this pr:
```
No Interpolation took 10.8796 s
Interpolation with  10.0 s   resolution took 1.1008 s
Interpolation with   5.0 min resolution took 0.2722 s
Interpolation with  30.0 min resolution took 0.2027 s
```

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
```diff
@@ -22,49 +22,37 @@
...
-        # the only frame attribute for the current CIRS is the obstime, but this
-        # would need to be updated if a future change allowed specifying an
-        # Earth location algorithm or something
-        cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime))
 
-    # we use the same obstime everywhere now that we know they're the same
-    obstime = cirs_coo.obstime
+    obsgeoloc, obsgeovel = altaz_frame.location.get_gcrs_posvel(altaz_frame.obstime)
+    if (np.any(cirs_coo.obstime != altaz_frame.obstime) or
+            np.any(cirs_coo.obsgeoloc != obsgeoloc)):
```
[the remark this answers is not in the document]
StuartLittlefair: not really an issue now we just have `location`

astropy/coordinates/erfa_astrom.py
StuartLittlefair: it has come back again as a result of adding `apco`

astropy/coordinates/tests/test_erfa_astrom.py
StuartLittlefair: I think it is? Maybe this comment is outdated...

maxnoe:
@StuartLittlefair awesome, numbers look good, nothing I would be worried about (even faster?)

Does it still hold that you think implementing a direct Altaz <-> ICRS transform could make this even faster?

astropy/coordinates/tests/test_celestial_transformations.py
StuartLittlefair: done.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -162,16 +164,21 @@ def cirs_to_gcrs(cirs_coo, gcrs_frame):
     # compute the pmatrix, and then multiply by its transpose
     pmat = gcrs_to_cirs_mat(cirs_coo.obstime)
     newrepr = cirs_coo.cartesian.transform(matrix_transpose(pmat))
-    gcrs = GCRS(newrepr, obstime=cirs_coo.obstime)
+    gcrs = GCRS(newrepr,
+                obstime=cirs_coo.obstime,
+                obsgeoloc=cirs_coo.obsgeoloc,
+                obsgeovel=cirs_coo.obsgeovel)
 
     # now do any needed offsets (no-op if same obstime and 0 pos/vel)
```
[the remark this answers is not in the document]
StuartLittlefair: agreed

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -107,9 +108,9 @@ def cirs_to_cirs(from_coo, to_frame):
         # distances anyway.
         return from_coo.transform_to(ICRS()).transform_to(to_frame)
 
-
```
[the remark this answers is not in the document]
StuartLittlefair: changed this now I think - wasn't 100% sure I understood

docs/coordinates/common_errors.rst
StuartLittlefair: I've re-worked the example below.

docs/coordinates/common_errors.rst
StuartLittlefair: Is it simpler now? There's still quite a few steps, but hopefully the comments explain the rationale a bit more.

StuartLittlefair:
> @StuartLittlefair awesome, numbers look good, nothing I would be worried about (even faster?)
> 
> Does it still hold that you think implementing a direct Altaz <-> ICRS transform could make this even faster?
> 

A little bit faster yes - we'd only be computing the astrometry context once, and there's none of the overhead of making the intermediate SkyCoord object.

> The `docs/coordinates/performance` now contains a plot for this. Which could tell more than the example I made.
> 
> I will have a look at the PR later this week, if that's fast enough.

Plenty fast enough. There isn't even a date for feature freeze on 4.3...

astropy/coordinates/erfa_astrom.py
```diff
@@ -238,6 +283,78 @@ def _get_cip(support, obstime):
...
+            for polar_motion_component in polar_motion_support
+        )
+
+    @staticmethod
+    def _get_sp00(support, obstime):
+        # this is already done in _get_cip. Improve efficiency?
+        jd1_tt_support, jd2_tt_support = get_jd12(support, 'tt')
+        sp00_support = erfa.sp00(jd1_tt_support, jd2_tt_support)
+        return np.interp(obstime.mjd, support.mjd, sp00_support)
+
```
mkbrewer, line 277: I don't see the point of interpolating `erfa.era00()` and `erfa.sp00()` here. It would likely be faster not to do so as they are just a few lines of code. In the case of `erfa.sp00()`, only 2. Also the earth rotation angle is the one thing that changes the fastest here, so I really don't think linear interpolation is a good thing to do with that.

mkbrewer:
> @mhvk @mkbrewer Using location for CIRS is rather neat in the end, and has the neat side effect of getting away from the major slowdowns mentioned in #10997, so these transformations are not achingly slow.

That isn't a side effect, it was one of the main reasons that I suggested going this way.

astropy/coordinates/erfa_astrom.py
StuartLittlefair: Very true; it’s faster and more accurate to not interpolate - good catch.

astropy/coordinates/earth.py
```diff
@@ -667,14 +679,24 @@ def get_gcrs(self, obstime):
...
+        return self._in_alternate_frame(GCRS, obstime)
 
-        itrs = self.get_itrs(obstime)
-        # Assume the observatory itself is fixed on the ground.
-        # We do a direct assignment rather than an update to avoid validation
-        # and creation of a new object.
-        zeros = np.broadcast_to(0. * u.km / u.s, (3,) + itrs.shape, subok=True)
-        itrs.data.differentials['s'] = CartesianDifferential(zeros)
-        return itrs.transform_to(GCRS(obstime=obstime))
+    def get_cirs(self, obstime):
```
mhvk, line 684: This works, and I like the refactoring with `_in_alternate_frame`, but I wonder if we shouldn't try to avoid a proliferation of methos...

One option to do so would be to make `EarthLocation` a bit more like a frame and define a `.transform_to(frame)` method which is the same as your `_in_alternate_frame, except that it takes a frame instance rather than a frame class (i.e., the `obstime` should already be done by the user).

Another option would be to give `get_itrs` a new keyword argument that asks to include a zero velocity component.  Then, inside code a user can just to `el.get_itrs(obstime, include_velocity=True).transform_to(...)`

astropy/coordinates/earth.py
```diff
@@ -699,6 +721,29 @@ def get_gcrs_posvel(self, obstime):
         obsgeovel = gcrs_data.differentials['s'].to_cartesian()
         return obsgeopos, obsgeovel
 
+    def get_cirs_posvel(self, obstime):
```
mhvk, line 724: If we do the above, this would be  small function inside the transformation code; a further advantage is that if we move eventually to a new `observer` argument (or gain the ability to store velocity on `location`), we have not introduced a method with relatively little use.

astropy/coordinates/erfa_astrom.py
```diff
@@ -81,22 +92,38 @@ def apio13(frame_or_coord):
...
-            lon.to_value(u.radian),
-            lat.to_value(u.radian),
-            height.to_value(u.m),
-            *get_polar_motion(frame_or_coord.obstime),
-            # all below are already in correct units because they are QuantityFrameAttribues
-            frame_or_coord.pressure.value,
-            frame_or_coord.temperature.value,
-            frame_or_coord.relative_humidity.value,
-            frame_or_coord.obswl.value,
+        astrom = np.zeros(frame_or_coord.obstime.shape, dtype=erfa.dt_eraASTROM)
```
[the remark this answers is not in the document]
mhvk: This is resolved by the added docstring

astropy/coordinates/tests/test_erfa_astrom.py
mhvk: I don't think it is: you do not use `altaz` any more in the `for` loop below...

astropy/coordinates/tests/test_intermediate_transformations.py
```diff
@@ -631,3 +634,79 @@ def test_tete_transforms():
...
+    aa = cirs_topo.transform_to(AltAz(obstime=t, location=home))
+    assert_allclose(aa.alt, 90*u.deg)
+
+
+@pytest.mark.remote_data
+def test_aa_high_precision():
+    """
+    These tests are provided by @mkbrewer - see issue #10356.
+
+    The code that produces them agrees very well (<0.5 mas) with SkyField once Polar motion
```
mhvk, line 673: Wasn't SkyField just updated?

docs/coordinates/common_errors.rst
```diff
@@ -0,0 +1,90 @@
...
+    >>> obj = EarthLocation(-1*u.deg, 52*u.deg, height=10.*u.km)
+    >>> home = EarthLocation(-1*u.deg, 52*u.deg, height=0.*u.km)
+
+    >>> # Now we make a ITRS vector of a straight overhead object
+    >>> itrs_vec = obj.get_itrs(t).cartesian - home.get_itrs(t).cartesian
+
+    >>> # Now we create a topocentric coordinate with this data
+    >>> # Any topocentric frame will work, we use CIRS
+    >>> # Start by transforming the ITRS vector to CIRS
+    >>> cirs_vec = ITRS(obstime=t).realize_frame(itrs_vec).transform_to(CIRS(obstime=t))
```
mhvk, line 81: A small simplification would be `ITRS(itrs_vec, obstime=t).transform_to(CIRS(obstime=t))`

Actually, I'd also add `.cartesian`, so your `_vec` always refers to a representation.

docs/coordinates/common_errors.rst
```diff
@@ -0,0 +1,90 @@
...
+    >>> # Now create a topocentric CIRS frame
+    >>> topocentric_cirs_frame = CIRS(obstime=t, location=home)
+    >>> # Finally, make CIRS frame object with the correct data
+    >>> cirs_topo = topocentric_cirs_frame.realize_frame(cirs_vec.cartesian)
```
mhvk, line 85: Similarly, I think this can just be `cirs_topo = CIRS(cirs_vec.cartesian, obstime=t, location=home)` (or without the `.cartesian` if one adds that above)

mhvk (commented):
@StuartLittlefair - this looks very good indeed! I like how things end up being quite a bit simpler.

My main comment is about the new methods for `EarthLocation` - see in-line comment.
Note that apart from the concrete suggestions in-line, we could also opt for just keeping `get_cirs_posvel` inside of the built-in transformations for now - that way we can decide separately how to adjust `EarthLocation`.

astropy/coordinates/tests/test_intermediate_transformations.py
mkbrewer: The SkyField update is still a work in progress at this time. The basic ITRS<->TIRS transforms are tested and working, but there is a bit of infrastructure that still needs doing, so it'll be a few weeks.

astropy/coordinates/earth.py
mkbrewer: You will also need a `get_tete_posvel()` method if TETE is converted to use an `EarthLocation`. The difference being the use of greenwich apparent sidereal time rather than earth rotation angle.

mhvk: Hmm, with TETE also in play, all the more reason to try to find a more general way to do this...

mkbrewer: My suggestion for implementing these functions in `EarthLocation` was just a matter of convenience for use in the relevant intermediate rotation transforms that require conversion of an `EarthLocation` to GCRS position and velocity. Of course, you can just use the ITRS position from the `EarthLocation` and convert it with the `itrs_to_cirs/tete` matrix, but then you also need to calculate the relevant velocity.

mhvk: From the new `_in_alternate_frame()` private method, it seems clear that all that is really needed is to provide a way to have an ITRS with associated velocities of 0 -- after that, the usual transformation machinery suffices. So, in that sense, I don't think we need a whole lot of extra methods.

astropy/coordinates/tests/test_erfa_astrom.py
StuartLittlefair: You mean the for loop that starts as below?

```python
for frame, tcode in zip([altaz, cirs, gcrs], ['apio', 'apco', 'apcs']):
```

because `altaz` is definitely used in that loop.

astropy/coordinates/builtin_frames/cirs.py
```diff
@@ -15,7 +18,11 @@
     ----------------
     obstime : `~astropy.time.Time`
         The time at which the observation is taken.  Used for determining the
-        position of the Earth and its precession.
+        position of the Earth and its precession
```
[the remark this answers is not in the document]
StuartLittlefair: fixed

astropy/coordinates/builtin_frames/cirs.py
```diff
@@ -15,7 +18,11 @@
...
+    location : `~astropy.coordinates.EarthLocation`
+        The location on the Earth.  This can be specified either as an
+        `~astropy.coordinates.EarthLocation` object or as anything that can be
+        transformed to an `~astropy.coordinates.ITRS` frame.
```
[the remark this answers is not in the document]
StuartLittlefair: done

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
```diff
@@ -22,49 +22,34 @@
...
 def cirs_to_altaz(cirs_coo, altaz_frame):
-    if np.any(cirs_coo.obstime != altaz_frame.obstime):
-        # the only frame attribute for the current CIRS is the obstime, but this
-        # would need to be updated if a future change allowed specifying an
-        # Earth location algorithm or something
-        cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime))
-
-    # we use the same obstime everywhere now that we know they're the same
-    obstime = cirs_coo.obstime
+    if (altaz_frame.location != cirs_coo.location or
```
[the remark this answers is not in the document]
StuartLittlefair: done

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
```diff
@@ -76,26 +61,20 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
     zen = PIOVER2 - usrepr.lat.to_value(u.radian)
 
     # first set up the astrometry context for ICRS<->CIRS at the altaz_coo time
-    astrom = erfa_astrom.get().apio13(altaz_coo)
+    astrom = erfa_astrom.get().apio(altaz_coo)
 
     # the 'A' indicates zen/az inputs
     cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
```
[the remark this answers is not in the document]
StuartLittlefair: nice.

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -96,7 +96,8 @@ def cirs_to_icrs(cirs_coo, icrs_frame):
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, CIRS)
 def cirs_to_cirs(from_coo, to_frame):
-    if np.all(from_coo.obstime == to_frame.obstime):
+    if (from_coo.location == to_frame.location and
```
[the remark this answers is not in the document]
StuartLittlefair: done

docs/coordinates/common_errors.rst
StuartLittlefair: ok

docs/coordinates/common_errors.rst
StuartLittlefair: much neater, thanks

StuartLittlefair:
Thanks @mhvk - I see what you mean about the proliferation of methods on `EarthLocation`. I have gone for a single `get_posvel` routine that allows one to specify the frame. A similar method has to live somewhere, so it may as well be on `EarthLocation` I think. I left `get_gcrs` and `get_gcrs_posvel` intact, although we could deprecate if we wanted.

astropy/coordinates/tests/test_erfa_astrom.py
mhvk: Duh, blinded as I was by seeing `altaz` changed to `cirs`, I didn't notice the other `altaz` - sorry!

mkbrewer:
I do hope that this passes muster with @mhvk as I think it is quite elegant, especially if `get_gcrs_posvel()` is eventually consolidated into it.  `get_posvel()` is very specifically meant to return the position and velocity of an `EarthLocation`, so, in my view, it is rightfully a method of that class.

mkbrewer:
In fact, you could begin that transition by modifying `get_gcrs_posvel()` to just call `get_posvel()`.

astropy/coordinates/earth.py
```diff
@@ -646,12 +651,45 @@ def get_itrs(self, obstime=None):
...
+        location : `~astropy.coordinates.CartesianRepresentation`
+            The position of the object in the specified ``frame``.
+        velocity : `~astropy.coordinates.CartesianRepresentation`
+            The velocity of the object in the specified ``frame``.
+        """
+        if not inspect.isclass(frame):
+            raise TypeError('frame must be a class')
+
+        frame_instance = frame(obstime=obstime)
+        data = self.get_itrs(obstime=obstime, include_velocity=True).transform_to(frame_instance).data
```
mhvk, line 688: here, you'd need `frame.obstime`.  Would one also want to do `.cartesian` instead of `.data`? Then, the differentials get already transformed automatically, and you could just do
```
return data.without_differentials(), data.differentials['s']
```

The brevity to me also somewhat argues against the need for the method...  I'll make a PR to your branch so you can see what I have in mind...

mhvk (commented):
Small suggestions; PR to your PR to come...

astropy/coordinates/erfa_astrom.py
```diff
@@ -81,23 +109,37 @@ def apio13(frame_or_coord):
...
-            *get_polar_motion(frame_or_coord.obstime),
-            # all below are already in correct units because they are QuantityFrameAttribues
-            frame_or_coord.pressure.value,
-            frame_or_coord.temperature.value,
+        # longitude with adjustment for TIO locator s'
+        astrom['along'] = lon.to_value(u.radian) + erfa.sp00(jd1_tt, jd2_tt)
+
+        # Polar motion, rotated onto local meridian
+        xp, yp = get_polar_motion(frame_or_coord.obstime)
+        sl = np.sin(astrom['along'])
```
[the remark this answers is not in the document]
mkbrewer: This is a direct copy from the relevant `erfa` routines and I think it is best to leave it as is for clarity sake.

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
```diff
@@ -22,49 +22,34 @@
...
 def cirs_to_altaz(cirs_coo, altaz_frame):
-    if np.any(cirs_coo.obstime != altaz_frame.obstime):
-        # the only frame attribute for the current CIRS is the obstime, but this
-        # would need to be updated if a future change allowed specifying an
-        # Earth location algorithm or something
-        cirs_coo = cirs_coo.transform_to(CIRS(obstime=altaz_frame.obstime))
-
-    # we use the same obstime everywhere now that we know they're the same
-    obstime = cirs_coo.obstime
+    if (np.any(altaz_frame.location != cirs_coo.location) or
```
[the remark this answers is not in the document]
StuartLittlefair: No it does not, because the default for AltAz is `None`, but the default for CIRS is `EarthLocation(0, 0, 0)`

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -162,16 +170,24 @@ def cirs_to_gcrs(cirs_coo, gcrs_frame):
     # compute the pmatrix, and then multiply by its transpose
     pmat = gcrs_to_cirs_mat(cirs_coo.obstime)
     newrepr = cirs_coo.cartesian.transform(matrix_transpose(pmat))
-    gcrs = GCRS(newrepr, obstime=cirs_coo.obstime)
-
-    # now do any needed offsets (no-op if same obstime and 0 pos/vel)
+    # can't use location.get_gcrs_posvel as that would trigger infinite recursion
+    obscirsloc, obscirsvel = cirs_coo.location.get_posvel(cirs_coo.obstime, CIRS)
```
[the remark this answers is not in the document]
StuartLittlefair: I'm not quite sure what you mean. It doesn't stop the recursion, which arises because if I try to transform this to GCRS the astropy machinery, then `cirs_to_gcrs` is called inside `cirs_to_gcrs` forever

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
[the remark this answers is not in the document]
maxnoe: Ok, but the error message will change now? What is the result of:

```
from astropy.coordinates import SkyCoord, AltAz

SkyCoord(alt='70d', az='0d', frame=AltAz()).transform_to(AltAz())
```

Currently it is `"NoneType" object has no attribute "to_geodetic"`

astropy/coordinates/earth.py
mkbrewer: However this ends up, I do think it is important for it to be fairly intuitive for the user. For instance, I use `get_gcrs_posvel()` myself in order to form an astrometric position and will continue to do so until an astrometric frame is available.

astropy/coordinates/erfa_astrom.py
```diff
@@ -192,6 +234,16 @@ def _prepare_earth_position_vel(support, obstime):
...
         return earth_pv, earth_heliocentric
 
+    @staticmethod
+    def _get_c2i(support, obstime):
+        jd1_tt_support, jd2_tt_support = get_jd12(support, 'tt')
+        c2i_support = erfa.c2i06a(jd1_tt_support, jd2_tt_support)
+        c2i = np.empty(obstime.shape + (3, 3))
+        for dim1 in range(3):
+            for dim2 in range(3):
+                c2i[..., dim1, dim2] = np.interp(obstime.mjd, support.mjd, c2i_support[..., dim1, dim2])
```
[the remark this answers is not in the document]
mhvk: I've also tried to think if one couldn't do something with broadcasting here! Sadly, cannot count on `scipy` as it is an optional dependency...

mhvk:
@StuartLittlefair - see https://github.com/StuartLittlefair/astropy/pull/5 for my promised PR. I think this ended up quite neat and avoids adding extra stuff to `EarthLocation` beyond the for sure handy `include_velocity` keyword argument.

Since everything in `EarthLocation` is user-facing API, I think we should make changes there in a separate PR, where we can really think what is most useful. Personally, I like to move more in the direction of treating these locations as coordinates (as done by sunpy for their `observer`).

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
```diff
@@ -76,26 +61,20 @@ def altaz_to_cirs(altaz_coo, cirs_frame):
...
-        # the vector operation of cirs_to_altaz (see there for more detail)
-
-        loccirs = altaz_coo.location.get_itrs(altaz_coo.obstime).transform_to(cirs_frame)
+        distance = altaz_coo.distance
 
-        astrometric_rep = SphericalRepresentation(lon=cirs_ra, lat=cirs_dec,
-                                                  distance=altaz_coo.distance)
-        newrepr = astrometric_rep + loccirs.cartesian
-        cirs_at_aa_time = CIRS(newrepr, obstime=altaz_coo.obstime)
+    cirs_at_aa_time = CIRS(ra=cirs_ra, dec=cirs_dec, distance=distance,
```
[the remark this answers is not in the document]
mhvk: Tastes differ...  I don't much like closing parentheses on their own line - it breaks the change in indentation as the way to indicate the end of something to me (of course, I may well get used to it as I'm seeing it more and more as people "blacken" their code).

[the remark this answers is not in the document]
maxnoe: I like the property that old lines don't show up in a `git diff` when a new option is added

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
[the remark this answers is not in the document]
maxnoe: I mean doing: 

```
obsgeoloc, obsgeovel = cirs_coo.location.get_posvel(cirs_coo.obstime, CIRS, transform=matrix_transpose(pmat))
```

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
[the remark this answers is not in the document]
mhvk: Yes, no denying that it has advantages too... Though it depends in part who one writes the code for, a reviewer or someone new trying to understand what is there (and thus seeing the whole thing).

Anyway, previous discussions like these (about line length, etc.) led to the conclusion that for existing files one should follow the style already there...

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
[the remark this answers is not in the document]
StuartLittlefair: Ah ok. I’m not a fan of that, as I think it’s a confusing api - one request a position in one frame explicitly, but the transform argument means you get a result in another frame. 
 
Anyway, this is made redundant by @mhvk’s PR to this PR

mkbrewer:
Too bad. I was looking forward to being able to easily retrieve the observer's position and velocity in any frame using `get_posvel()`. Are you planning on generalizing `get_location_gcrs()` for use with TETE, or just creating a second function?

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -162,16 +170,24 @@ def cirs_to_gcrs(cirs_coo, gcrs_frame):
...
     return gcrs.transform_to(gcrs_frame)
 
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, CIRS, ITRS)
 def cirs_to_itrs(cirs_coo, itrs_frame):
-    # first get us to CIRS at the target obstime
-    cirs_coo2 = cirs_coo.transform_to(CIRS(obstime=itrs_frame.obstime))
+    # first get us to geocentric CIRS at the target obstime
+    cirs_coo2 = cirs_coo.transform_to(CIRS(obstime=itrs_frame.obstime,
+                                           location=EarthLocation(0*u.km, 0*u.km, 0*u.km)))
```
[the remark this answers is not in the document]
StuartLittlefair: Is this ever likely to be a problem?

I'm not keen on the idea of static members of `EarthLocation` since that is going to get messy with the site registry - when should something be a static member and when in the site list for example. I have the same thoughts for the `ZERO_SPEED` constant.

mhvk:
@mkbrewer - the fact that you would use it is definitely a good reason to have a method like `get_posvel` - I really just wanted to avoid having discussions about what exactly its argument should be delay merging this very good PR!

It does bring up a different question, of whether it would be useful generally to have a convenience method on any coordinate (or representation?) that returns position and velocity as two separate representations.
(And now I am distracting in this PR! To be moved to a new issue...)

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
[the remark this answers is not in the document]
mhvk: Maybe just define them as constants for `intermediate_rotation_transforms`? Or in `utils.py`, together with `DEFAULT_OBSTIME`? 

Although in this particular case, since the centre of the Earth is already the default, perhaps just omit it?

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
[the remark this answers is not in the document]
mhvk: This PR does not change that...

astropy/coordinates/earth.py
```diff
@@ -646,12 +651,45 @@ def get_itrs(self, obstime=None):
 
         # do this here to prevent a series of complicated circular imports
         from .builtin_frames import ITRS
-        return ITRS(x=self.x, y=self.y, z=self.z, obstime=obstime)
+        itrs_coo = ITRS(x=self.x, y=self.y, z=self.z, obstime=obstime)
```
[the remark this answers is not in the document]
mhvk: I wondered about this too, though perhaps it ends up being better to have something like an `Observer` analogue of a `SkyCoord`. Though since this would change API, perhaps best discussed outside of the PR.

mhvk:
@StuartLittlefair - I think this now is basically OK, and I'd approve if there weren't a few smaller comments from @maxnoe remaining (the module-level constants seem a good idea). If you do those, and are up to it, perhaps squash the various fixup commits (it would seem a good idea to keep the CIRS with obsgeoloc/vel in the git history, just in case we want to look back at the code and reasoning). A really nice improvement, and nice multi-way collaboration getting there too!

astropy/coordinates/earth.py
```diff
@@ -646,12 +651,45 @@ def get_itrs(self, obstime=None):
...
+        if include_velocity:
+            zeros = np.broadcast_to(0. * u.km / u.s, (3,) + itrs_coo.shape, subok=True)
```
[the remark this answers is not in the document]
StuartLittlefair: It's the broadcasting that's slow here - not the creation of the `Quantity`.

```python
In [30]: %timeit np.broadcast_to(ZERO_SPEED, itrs.shape, subok=True)                                                                                        
79.3 µs ± 3.84 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [31]: %timeit np.broadcast_to(0. * (u.km / u.s), (3,) + itrs.shape, subok=True)                                                                          
72.2 µs ± 5.58 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

I've included the brackets here though

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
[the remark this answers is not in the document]
StuartLittlefair: done

astropy/coordinates/erfa_astrom.py
```diff
@@ -192,6 +234,16 @@ def _prepare_earth_position_vel(support, obstime):
 
         return earth_pv, earth_heliocentric
 
+    @staticmethod
+    def _get_c2i(support, obstime):
```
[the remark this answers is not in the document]
StuartLittlefair: I've added this, and ones to your own private routines with no docstrings ;)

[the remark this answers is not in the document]
maxnoe: Ha, thanks and sorry ;)

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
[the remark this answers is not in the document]
StuartLittlefair: Indeed - in fact I'm not sure how one would change that. An AltAz frame without a location is undefined IMO, and the issue is with the fact that you are allowed to create one. 

Rather than fix the error message in the snippet above I would prefer the code died when attempting to create the AltAz frame, but the habit of defining AltAz frames like this, and picking up the location for the AltAz from the `from_coo` is in the wild so we are stuck with it I think

StuartLittlefair:
Fixes #4269 BTW!

StuartLittlefair:
OK @mhvk I think I have addressed all comments. I attempted the squash but backed out in the end - I've never been comfortable with the more advanced git stuff.

mkbrewer:
I will just note here that the calls to `pvtob()` and `trxpv()` in `apco()` could be replaced by the procedure in `intermediate_rotation_transforms` to derive the GCRS position and velocity of the observer. I highly doubt that this would make a noticeable difference though. The only truly redundant operation that this would avoid is the geodetic to geocentric transform which is done only once per observer location.

astropy/coordinates/earth.py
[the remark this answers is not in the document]
mhvk: Surprising!  I played around a bit and fastest is
```
np.broadcast_to(np.zeros(()), (3,)+itrs.shape) << u.km/u.s
```
In the end, though, it would all be negligible compared to the rest of the routine, so let's keep it as you have it now.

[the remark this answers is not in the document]
maxnoe: I disagree. If there are easy gains **that don't harm legibility** we should always chose the most efficient solution. So replacing with your proposal is the way to go I think.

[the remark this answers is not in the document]
maxnoe: Wait, is this just `np.zeros((3, ) + itrs_coo.shape) << u.km / u.s`? Why the broadcast for a single number?

pllim:
@StuartLittlefair , regarding your comment in https://github.com/astropy/astropy/pull/10994#issuecomment-730263401 , if this PR fixes additional issue, please add "fixes ..." on the original post or GitHub will not auto-close the issue when this is merged. Thanks!

Disclaimer: Have not been following the convo.

astropy/coordinates/earth.py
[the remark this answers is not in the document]
mhvk: Save memory...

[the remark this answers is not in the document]
maxnoe: Ha! TIL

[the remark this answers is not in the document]
StuartLittlefair: I find @mhvk's fastest solution not very legible, so I'll stick with it as it is now.

mhvk:
@StuartLittlefair - if you don't mind, I'll do a rebase -- with the excuse to just address the two remaining comments of @maxnoe, so this can go in. 

p.s. I am no git expert either, but I find that what works quite well is `git rebase -i upstream/master`, which allows one to move commits around and possibly squash them together (handiest if one has a `gitk` window open at the same time)

CHANGES.rst
```diff
@@ -16,6 +16,9 @@ astropy.convolution
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
 
+- Adds the ability to create topocentric ``CIRS`` frames, improving accuracy of
+  ``AltAz`` calculations to of order milli-arcsecond. [#10994]
```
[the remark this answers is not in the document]
mhvk: I also find it a bit weird and will change to
```
- Adds the ability to create topocentric ``CIRS`` frames. Using these,
  ``AltAz`` calculations are now accurate down to the milli-arcsecond
  level. [#10994]
```

StuartLittlefair:
I don't mind at all!

astropy/coordinates/builtin_frames/cirs_observed_transforms.py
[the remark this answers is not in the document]
mhvk: Going to call this a separate issue - though one we should fix, if likely at a different level!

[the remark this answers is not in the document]
maxnoe: See https://github.com/astropy/astropy/issues/6777

mhvk (approved):
Some reduction in number of commits, but most were already quite logical steps. Anyway, in the process looked over everything one last time, and I think it is all OK. Thanks so much all!

mhvk:
p.s. Sounds like you did the same thing as I did, but it is very easy to just get one thing out of order and then everything breaks...

StuartLittlefair:
Indeed - thanks to all. Especially @mkbrewer without which this wouldn’t have happened.

mhvk:
For lack of a better place: 4.2 is not yet out, and it is the release in which TETE is introduced - should be have a "bug fix" to make it use `location` instead of `obsgeoloc/vel`? (We won't be loved for it...)

mhvk:
p.s. I'm trying to create a quick PR so we have something to discuss

StuartLittlefair:
I think we should ask politely. There’s a strong case for it, and I can turn it around tomorrow but if it causes too much pain maybe not? 

Who handles the 4.2 release? Is it @pllim ?

pllim:
> Who handles the 4.2 release? 

Maybe @astrofrog

mhvk:
See #11061 for a PR that does it, if only to show that it does not have to cause delay...

astrofrog:
@bsipocz is handling 4.2

bsipocz:
Delay is caused by other factors. I can get to the release over the weekend but not much before, so if the bugfix is merged by then it's all good from my side.

mkbrewer:
Does this mean that the CIRS fix may make it into the 4.2 release also?

ludwigschwardt:
Thanks for this!! This saved us from a recent rabbit-hole chasing 100 mas errors on Solar System bodies :tada: 😅 

> ... transformations into the AltAz frames did not properly handle aberration ...

I would like to understand the fix better. Is it really (diurnal) aberration? I would have thought that aberration would be independent of the object distance. It seems more like parallax, with larger effects for closer objects, but it's also smaller than the full geocentric vs topocentric parallax.

brandon-rhodes:
@ludwigschwardt — The *angle* produced by aberration is indeed completely independent of object distance. Light does not know how far it’s traveled, and has the same relative velocity to the observer as it arrives whether it came from nearby or from very far away.

But applying the same angle to two vectors of different length will require two different offsets: if *v₂* has twice the length of *v₁*, then the correction applied will need to have twice the length to produce the same angle.

Does that perhaps make better sense of what the code is doing? Or does it not quite match the line you’re looking at?

mkbrewer:
@ludwigschwardt The problem was that the component of diurnal aberration produced by diurnal parallax was missing. Aberration is dependent on the dot product of the position vector and the velocity vector and the position vector has a diurnal component for nearby objects . Aberration was being computed at the geocenter and, while a correction for diurnal velocity was added, no correction was made for diurnal parallax.

mkbrewer:
For more on this, please see my comment here: https://github.com/astropy/astropy/issues/10356#issuecomment-678830640

ludwigschwardt:
@brandon-rhodes @mkbrewer Thanks a lot for the feedback, it's much appreciated!

I carefully reread your posts several times until something dawned. Let's see if I got it this time...

diurnal_aberration

Let aberration be a function _a(v, d)_ of the velocity vector _v_ of the observer and the direction vector _d_ towards the object (just before applying aberration), which typically involves the dot product of _v_ and _d_. Then the total aberration for the geocenter (= annual aberration) is

_a( v_<sub>E</sub>, _d_<sub>E</sub> _)_

and the total aberration for the topocentric observer (= annual + diurnal aberration) is

_a( v_<sub>T</sub>, _d_<sub>T</sub> _)_ = _a( v_<sub>E</sub> + &Delta;_v_<sub>T</sub>, _d_<sub>T</sub> _)_.

If I understand you correctly, Astropy instead used to calculate

_a( v_<sub>E</sub> + &Delta;_v_<sub>T</sub>, _d_<sub>E</sub> _)_.

On top of this we also have the diurnal parallax _p_ = _d_<sub>T</sub> - _d_<sub>E</sub> which is hopefully already taken care of (up to 1 degree for the Moon!).

If everything is linear and you wanted to split up the total topocentric aberration into components, one could then write it as

_a( v_<sub>T</sub>, _d_<sub>T</sub> _)_ = _a( v_<sub>E</sub>, _d_<sub>E</sub> _)_ + _a(_ &Delta;_v_<sub>T</sub>, _d_<sub>E</sub> _)_ + _a( v_<sub>E</sub>, _p )_ + _a(_ &Delta;_v_<sub>T</sub>, _p )_
= annual + diurnal + diurnal due to parallax + tiny

For the Moon the maximum error in aberration is on the order of (max parallax) * (annual aberration) = 1/60 * 20" ~ 0.3", which is what I got before, but along a way different route... :joy: At least I now realise that the Moon's own velocity does not affect aberration.

mkbrewer:
That's correct. Aberration was being calculated at the geocenter (your first term), then the position was moved to the `EarthLocaton` adding diurnal parallax, then the aberration correction for diurnal velocity (your second term) was applied, but your third and fourth terms were, unfortunately, neglected.

mkbrewer:
In the current master, aberration is calculated at the `EarthLocation` as it should have been done originally.

PR #10995, opened 2020-11-05
Support masked bool data in ASCII files with blank string

### LINKED ISSUE #10988: Support nulls in ECSV boolean columns?
### Description
The ECSV format is described, and as far as I know defined, in https://github.com/astropy/astropy-APEs/blob/master/APE6.rst. That document does not discuss in any detail the encoding of blank values, but e.g. an empty string value in an `int32`-typed column can be read by the Astropy implementation, and is represented by a masked value.  However an empty string value in a `bool`-typed column causes a read error.  If `file` refers to the following text:

    # %ECSV 0.9
    # ---
    # datatype: [
    #   { name: AA, datatype: int32   },
    #   { name: BB, datatype: bool    },
    #   { name: CC, datatype: float64 },
    # ]
    AA BB CC
    1 True 2.1
    2 False 5.4
    "" "" nan
    9 True -9.9

then `astropy.table.Table.read(file, format='ascii.ecsv')` gives me a `ValueError: Column BB failed to convert: bool input strings must be only False or True`.  This is Astropy 4.0, Python 3.6.

Is this intended behaviour?  There are cases where a boolean column could contain nulls as well as Trues and Falses (I came across this myself when trying to represent certain Gaia data in ECSV - the table in question had boolean columns containing null values, which meant the Astropy reader could not read my ECSV serialisation of the table).  APE6 does say _"Boolean fields are represented as the case-sensitive string False or True"_, but since nulls are not discussed elsewhere (e.g. it doesn't mention nulls in integer columns) it's not clear whether this counts as a declaration that nulls are intentionally outlawed for the `bool` type.

The ECSV I/O handlers I have written for STIL/TOPCAT do work with blanks in `bool`-typed columns.  I request that the Astropy ECSV implementation considers doing the same.

taldcroft:
### Description

This pull request is to address issue #10988 to support nulls in ECSV data.

The implementation brought along another "feature", namely support for "0" and "1" as False and True, respectively. I did not necessarily *want* to make that change, but the way masking is supported in `io.ascii` it was unavoidable without a much larger code change. Masking works by finding empty "" data values and replacing those with the string "0" and setting that entry as masked, so it was necessary to allow "0" as a column value. For symmetry I added "1".  Some people might view this change as a feature.

Once this was done for ECSV it also worked for reading from a regular character-separated data file as long as an appropriate `converters` is provided.

For now I milestoned this for 4.3 because it changes the behavior somewhat. But I can't immediately think of a way this would change existing working code, so maybe it could be a bug-fix for LTS.

Fixes #10988

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1186,6 +1186,13 @@ astropy.io.ascii
   reading, so we now ignore the parallel option and fall back to serial reading.
   [#10880]
 
+- Fixed a bug where "" (blank string) as input data for a boolean type column
+  was causing an exception instead of indicating a masked value. As a
+  consequence of the fix, the values "0" and "1" are now also allowed as valid
+  inputs for boolean type columns. These new allowed values apply for both ECSV
+  and for basic character-delimited data files ('basic' format with appropriate
+  ``converters`` specified). [#10995]
+
 astropy.io.fits
 ^^^^^^^^^^^^^^^
 
--- a/astropy/io/ascii/core.py
+++ b/astropy/io/ascii/core.py
@@ -944,13 +944,18 @@ def convert_numpy(numpy_type):
         # Try a smaller subset first for a long array
         if len(vals) > 10000:
             svals = numpy.asarray(vals[:1000])
-            if not numpy.all((svals == 'False') | (svals == 'True')):
-                raise ValueError('bool input strings must be only False or True')
+            if not numpy.all((svals == 'False')
+                             | (svals == 'True')
+                             | (svals == '0')
+                             | (svals == '1')):
+                raise ValueError('bool input strings must be False, True, 0, 1, or ""')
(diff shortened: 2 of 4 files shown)


hamogu (approved):
I approve this change (obviously, the test that fails needs to be skipped if pyyaml is not installed). While I agree that strictly speaking this could be considered a change in behavior, I would classify this as a bug fix and will re-milestone for LTS.

mhvk (approved):
That is a nice, clean solution! Agreed that interpreting 0 and 1 as False and True is, if anything, an improvement.

The case for backporting is indeed a bit confusing - I cannot really make up my mind!

mhvk:
p.s. The failure is real - the test needs `yaml`.

taldcroft:
OK, 4.0.4 it is!

PR #11001, opened 2020-11-08
Simplify transformation to true ecliptic coordinates

### LINKED ISSUE #11000: ecliptic_transforms._true_ecliptic_rotation_matrix() calculates nutation twice
### Description

I noticed that the function `ecliptic_transforms._true_ecliptic_rotation_matrix()` calculates nutation twice. Once in the call to `erfa.pnm06a()` and then again explicitly right afterwards. Calculating nutation is quite expensive, so it would be better to just explictily call the three functions that `erfa.pnm06a()` calls.

### Additional context

What `erfa.pnm06a()` actually does:
```
/* Fukushima-Williams angles for frame bias and precession. */
   eraPfw06(date1, date2, &gamb, &phib, &psib, &epsa);

/* Nutation components. */
   eraNut06a(date1, date2, &dp, &de);

/* Equinox based nutation x precession x bias matrix. */
   eraFw2m(gamb, phib, psib + dp, epsa + de, rnpb);
```

mhvk:
As noted by @mkbrewer in #11000, if one looks in detail at the erfa functions called, one can simplify the calculation, avoiding twice doing the (expensive) nutation calculation. EDIT: ~In the second commit, I also combined two rotations.~ [Commit removed later]

fixes #11000

### DIFF
--- a/astropy/coordinates/builtin_frames/ecliptic_transforms.py
+++ b/astropy/coordinates/builtin_frames/ecliptic_transforms.py
@@ -37,15 +37,24 @@ def _mean_ecliptic_rotation_matrix(equinox):
 
 
 def _true_ecliptic_rotation_matrix(equinox):
-    # This code calls pnm06a from ERFA, which retrieves the precession
-    # matrix (including frame bias) according to the IAU 2006 model, and
-    # including the nutation. This family of systems is less popular
+    # This code calls the same routines as done in pnm06a from ERFA, which
+    # retrieves the precession matrix (including frame bias) according to
+    # the IAU 2006 model, and including the nutation.
+    # This family of systems is less popular
     # (see https://github.com/astropy/astropy/pull/6508).
     jd1, jd2 = get_jd12(equinox, 'tt')
-    rnpb = erfa.pnm06a(jd1, jd2)
-    _, nut_obl = erfa.nut06a(jd1, jd2)*u.radian
-    obl = erfa.obl06(jd1, jd2)*u.radian + nut_obl  # calculate the true obliquity of the ecliptic
-    return matrix_product(rotation_matrix(obl, 'x'), rnpb)
+    # Here, we call the three routines from erfa.pnm06a separately,
+    # so that we can keep the nutation for calculating the true obliquity
+    # (which is a fairly expensive operation); see gh-11000.
+    # pnm06a: Fukushima-Williams angles for frame bias and precession.
+    gamb, phib, psib, epsa = erfa.pfw06(jd1, jd2)
+    # pnm06a: Nutation components.
+    dpsi, deps = erfa.nut06a(jd1, jd2)
+    # pnm06a: Equinox based nutation x precession x bias matrix.
+    rnpb = erfa.fw2m(gamb, phib, psib+dpsi, epsa+deps)
+    # calculate the true obliquity of the ecliptic
+    obl = erfa.obl06(jd1, jd2)+deps
+    return matrix_product(rotation_matrix(obl << u.radian, 'x'), rnpb)
 
 
 def _obliquity_only_rotation_matrix(obl=erfa.obl80(EQUINOX_J2000.jd1, EQUINOX_J2000.jd2) * u.radian):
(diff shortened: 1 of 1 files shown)


mkbrewer:
Personally, I would not change the arguments for `fw2m()` for clarity. This function is supposed to yield the BPN. The final rotation is not expensive and makes it clear that you are rotating to the ecliptic.

mhvk:
@mkbrewer - OK, you're right, clarity is more important here than making execution fractionally faster. I removed the second commit, i.e., this is now just your original suggestion.

astropy/coordinates/builtin_frames/ecliptic_transforms.py
```diff
@@ -37,15 +37,24 @@ def _mean_ecliptic_rotation_matrix(equinox):
...
     jd1, jd2 = get_jd12(equinox, 'tt')
-    rnpb = erfa.pnm06a(jd1, jd2)
-    _, nut_obl = erfa.nut06a(jd1, jd2)*u.radian
-    obl = erfa.obl06(jd1, jd2)*u.radian + nut_obl  # calculate the true obliquity of the ecliptic
-    return matrix_product(rotation_matrix(obl, 'x'), rnpb)
+    # Here, we call the three routines from erfa.pnm06a separately,
+    # so that we can keep the nutation for calculating the true obliquity
+    # (which is a fairly expensive operation); see gh-11000.
+    # pnm06a: Fukushima-Williams angles for frame bias and precession.
+    gamb, phib, psib, epsa = erfa.pfw06(jd1, jd2)
```
StuartLittlefair, line 50: Is there any harm in expanding these out? I find the short variable names used internally by sofa to hinder easy reading of the code, we could use `gamma_bar`, `phi_bar`, `psi_bar` and `epsilon_A` for example, which are easier to cross-ref with the Fukushima paper.

astropy/coordinates/builtin_frames/ecliptic_transforms.py
```diff
@@ -37,15 +37,24 @@ def _mean_ecliptic_rotation_matrix(equinox):
...
+    # pnm06a: Nutation components.
+    dpsi, deps = erfa.nut06a(jd1, jd2)
```
StuartLittlefair, line 52: Similarly here, these are nutation components in longitude and obliquity right? So are `nut_lon` and `nut_obl` better names?

Or if we wanted to keep the consistency with the F-W angle names we could use `nut_psi` and `nut_eps`

StuartLittlefair (approved):
Looks good @mhvk. 

I've some minor suggestions about variable names but I am ambivalent; there is a case for keeping consistency with the SOFA internal names as well.

astropy/coordinates/builtin_frames/ecliptic_transforms.py
mhvk: This is indeed tricky. In the end, I think using the variable names as in ERFA is most useful, as I'd think that is the first thing people might look at. And in this case, the F-W names are not clearer in themselves, only by reference. So, maybe just a comment cross-linking them? I've done that for now.

astropy/coordinates/builtin_frames/ecliptic_transforms.py
mhvk: Here, I agree more with you, but with once one decides on the one above, this one follows (and is clearer, since `dpsi` matches `psi` and `deps` matches `eps` -- this would be true even if going for the F-W names). So, again I thought it easiest to just expand the comment above to `pnm06a: Nutation components (in longitude and obliquity).`

mhvk:
Ouch, `[ci skip]` does not work any more...

StuartLittlefair (approved):

StuartLittlefair:
Ouch indeed. Nevertheless, this can be merged when the tests are done IMO.

pllim:
Re: skip ci -- Yeah... You could hop over to https://github.com/astropy/astropy/actions to cancel it but that's about it for now. Cannot cancel just one job either; it is all or nothing. 😞

PR #11015, opened 2020-11-09
Updating spinner

### LINKED ISSUE #11004: Outdated documentation for Spinner
### Description

The current docstring in the [Spinner](https://github.com/astropy/astropy/blob/ab1a1d3c776ff636b844be6799c1c9b91b32a21f/astropy/utils/console.py#L855) class demonstrates using `s.next()` to update the spinner.  But this is a Python 2-ism that no longer exists for iterators in Python 3.  Instead the correct documentation would be `next(s)`.

Additionally, I think it's confusing that the API for `Spinner` is so different from `ProgressBar` and `ProgressBarOrSpinner`.  In the latter two cases, their `__enter__` method (when using `with ProgressBar(...) as bar:` returns the `ProgressBar(OrSpinner)` instance itself, which is iterable itself.  It also provides an `update()` method which can be called to update the progress bar.  Whereas `with Spinner(...) as s:` does not return the `Spinner` instance itself, but rather a generator.

It should be easy to update the API for `Spinner` (make it iterable, make `__enter__` return the `Spinner` instance, and add an `update()` method) so that it is more like `ProgressBar`.

I would suggest making the updates to `Spinner` and fixing the documentation at the same time.

Iamsoto:
Attempting to update `astropy.utils.console.spinner ` as discussed in issue #11004

I'm not sure if we want new tests for this class or not. Though writing tests for this might be rough. Let me know though, and I'll do my best. 

Otherwise I'm hoping this is similar to what you were asking for @embray 

Fixes #11004

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -151,6 +151,9 @@ astropy.utils
 
 - Removed deprecated ``utils.misc.InheritDocstrings`` and ``utils.timer``. [#10281]
 
+- Updated ``utils.console.Spinner`` to better resemble the API of ``utils.console.ProgressBar`` 
+  including an ``update()`` method and iterator support. [#11015]
+
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
 
--- a/astropy/utils/console.py
+++ b/astropy/utils/console.py
@@ -860,7 +860,7 @@ class Spinner:
 
         with Spinner("Reticulating splines", "green") as s:
             for item in enumerate(items):
-                s.next()
+                s.update()
     """
     _default_unicode_chars = "◓◑◒◐"
     _default_ascii_chars = "-/|\\"
@@ -887,7 +887,7 @@ class Spinner:
             completely silent.
 
         step : int, optional
-            Only update the spinner every *step* steps
+            Only update the spinner every *step* steps. Must be >= 0.
 
         chars : str, optional
             The character sequence to use for the spinner
@@ -899,52 +899,27 @@ class Spinner:
         self._msg = msg
         self._color = color
         self._file = file
+        if step < 0:
+            raise ValueError("step must be >= 0")
+
         self._step = step
+        self._cur_step = 0
         if chars is None:
             if conf.unicode_output:
                 chars = self._default_unicode_chars
             else:
                 chars = self._default_ascii_chars
         self._chars = chars
-
+        self._cur_char = 0
         self._silent = not isatty(file)
 
-    def _iterator(self):
-        chars = self._chars
-        index = 0
-        file = self._file
-        write = file.write
-        flush = file.flush
-        try_fallback = True
-
-        while True:
-            write('\r')
-            color_print(self._msg, self._color, file=file, end='')
-            write(' ')
-            try:
-                if try_fallback:
-                    write = _write_with_fallback(chars[index], write, file)
-                else:
-                    write(chars[index])
-            except UnicodeError:
-                # If even _write_with_fallback failed for any reason just give
-                # up on trying to use the unicode characters
-                chars = self._default_ascii_chars
-                write(chars[index])
-                try_fallback = False  # No good will come of using this again
-            flush()
-            yield
-
-            for i in range(self._step):
-                yield
-
-            index = (index + 1) % len(chars)
+        if self._silent:
+            self._iter = self._silent_iterator
+        else:
+            self._iter = self._iterator
 
     def __enter__(self):
-        if self._silent:
-            return self._silent_iterator()
-        else:
-            return self._iterator()
+        return self
 
     def __exit__(self, exc_type, exc_value, traceback):
         file = self._file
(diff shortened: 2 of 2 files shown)


astropy/utils/console.py
```diff
@@ -860,7 +860,7 @@ class Spinner:
 
         with Spinner("Reticulating splines", "green") as s:
             for item in enumerate(items):
-                s.next()
+                next(s) # or s.update()
```
embray, line 863: I would almost rather just document it as `s.update()` since I think that's less confusing and surprising to beginners (who might not understand where `next()` came from).

astropy/utils/console.py
```diff
@@ -899,73 +899,80 @@ def __init__(self, msg, color='default', file=None, step=1,
...
+            self._cur_char = (self._cur_char + 1) % len(chars)
+        self._cur_step = self._cur_step + 1
 
     def _silent_iterator(self):
         color_print(self._msg, self._color, file=self._file, end='')
         self._file.flush()
 
-        while True:
-            yield
+    def update(self):
```
embray, line 974: This should have a docstring.

embray: Although not as useful as it is in a progress bar, for symmetry it would be nice if this also took an optional `value` argument.

You could write:

```
def update(self, value=None):
    if value is None:
        value = 1
    for _ in range(value):
        next(self)
```

Then that would allow you to do away entirely with this conditional: https://github.com/astropy/astropy/pull/11015/files#diff-582a94e0a0f7293634d5bb6531b7bdf8a847ecd2a380ec691a65bac8beaff546R1048

astropy/utils/console.py
```diff
@@ -887,7 +887,7 @@ def __init__(self, msg, color='default', file=None, step=1,
             completely silent.
 
         step : int, optional
-            Only update the spinner every *step* steps
+            Only update the spinner every *step* steps. Must be >= 0.
```
embray, line 890: Please make this explicit by raising a `ValueError` if `step < 1`.

astropy/utils/console.py
```diff
@@ -899,73 +899,80 @@ def __init__(self, msg, color='default', file=None, step=1,
         self._msg = msg
         self._color = color
         self._file = file
-        self._step = step
+        self._step = step + 1
```
embray, line 902: Why do add `+ 1` here?  Since the default step is 1 this means by default it will only update every 2 steps.

astropy/utils/console.py
```diff
@@ -899,73 +899,80 @@ def __init__(self, msg, color='default', file=None, step=1,
...
+                    write = _write_with_fallback(chars[self._cur_char], write, file)
                 else:
-                    write(chars[index])
+                    write(chars[self._cur_char])
             except UnicodeError:
                 # If even _write_with_fallback failed for any reason just give
                 # up on trying to use the unicode characters
                 chars = self._default_ascii_chars
-                write(chars[index])
+                write(chars[self._cur_char])
```
embray, line 963: Since this is no longer a generator, you should actually update `self._chars` on line 962 so that it reads `self._chars = self._default_ascii_chars`.  Otherwise it will keep trying to use the unicode chars on every loop.

astropy/utils/console.py
```diff
@@ -899,73 +899,80 @@ def __init__(self, msg, color='default', file=None, step=1,
...
+        """
         chars = self._chars
-        index = 0
         file = self._file
         write = file.write
-        flush = file.flush
         try_fallback = True
 
-        while True:
+        if self._cur_step % self._step == 0:
```
embray, line 950: I don't think changing this to not be a generator was necessary.  My thinking was in `__init__` you can do: `self._iter = self._iterate()`, and in `__next__` write:  `next(self._iter)`.

That said, you already went to the effort of changing it, so as long as it works I think it's a pretty harmless change.

embray (changes requested):
A few minor issues, but looks good otherwise, thanks!

embray:
Will also need a changelog entry.  I'm not sure why the changelog checker didn't run (@pllim?)

pllim:
The bot been flaky. Sometimes it works, sometimes it doesn't. Either Heroku isn't as reliable anymore or we hit limits on its free tier.

astropy/utils/console.py
Iamsoto: Uff. I thought I tried that approach, and at the time thought it didn't work. Though I just tried exactly what you wrote there and it works fine, so now I feel silly

astropy/utils/console.py
Iamsoto: The loop on the iterator does self._cur_iter % self._step, so if step is 1 the method will update every 'other' cycle. It was my attempt at emulating the generator. Though, I don't think I need to add a 1 here, it can just be accounted for in the loop

astropy/utils/console.py
Iamsoto: Good catch

astropy/utils/console.py
Iamsoto: Perhaps its better to just write `def update(self, value=1)` ?

astropy/utils/console.py
Iamsoto: Done

astropy/utils/console.py
Iamsoto: Should be good now

Iamsoto:
I'm not sure why the codecov patch is failing. I did a rebase + push, thought that would solve it

CHANGES.rst
```diff
@@ -147,6 +147,9 @@ astropy.utils
 
 - Removed deprecated ``utils.misc.InheritDocstrings`` and ``utils.timer``. [#10281]
 
+- Updated docstring for console.spinner, console.spinner now implements iterator protocal
```
embray, line 150: Please reword this as

```
Updated ``utils.console.Spinner`` to better resemble the API of ``utils.console.ProgressBar`` including an ``update()`` method and iterator support.
```

astropy/utils/console.py
```diff
@@ -899,52 +899,27 @@ def __init__(self, msg, color='default', file=None, step=1,
         self._msg = msg
         self._color = color
         self._file = file
+        if step < 0:
+            raise ValueError(f"step must be >= 0")
```
embray, line 903: no need to make this an f-string

embray (changes requested):
minor nitpicks

embray:
No worries about codecov.  I still don't understand exactly why, but sometimes it shows totally random coverage changes that have nothing to do with your PR.

CHANGES.rst
Iamsoto: Done!

astropy/utils/console.py
Iamsoto: Done!

Iamsoto:
Hey, 

I made PR #11046 which updates the spinner API using your original recommendations, i.e. keeping the enumerator.
I hope you'll find the option useful

astropy/utils/console.py
embray: I see now.  If you ask me the original semantics of this argument are actually a bug too.  The docs say "only update the spinner every {step} steps".  So if `step=1` that should mean update the spinner every 1 step, i.e., every step.  The current semantics with `step=1` mean it updates every 2 steps.  Unless I'm missing something that doesn't make any sense.

embray: In which case it should be that `step >= 1` and `step=1` should be the default.

embray: In fact I see the default is already `step=1`.  I think this was just a minor bug in the original code.  It should have read:

```python
for _ in range(self._step - 1):
    yield
```

astropy/utils/console.py
embray: Don't worry about it; the way you rewrote it is fine too :)

embray:
Ah ok, since you did that, let's close this PR in favor of your new one which is simpler.

astropy/utils/console.py
pllim: Opened an issue at #11656

PR #11017, opened 2020-11-10
allow metaclass keywords in Model subclasses

nstarman:

### Description

In the Model class construction the metaclass ``_ModelMeta`` does not permit keyword arguments. Adding this option does not change anything in the class creation process, merely allows subclasses of ``_ModelMeta`` to inherit methods without having to wrap them to avoid passing keyword argument.

no-changelog-needed

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -85,7 +85,7 @@ class _ModelMeta(abc.ABCMeta):
     # Default empty dict for _parameters_, which will be empty on model
     # classes that don't have any Parameters
 
-    def __new__(mcls, name, bases, members):
+    def __new__(mcls, name, bases, members, **kwds):
         # See the docstring for _is_dynamic above
         if '_is_dynamic' not in members:
             members['_is_dynamic'] = mcls._is_dynamic
@@ -105,7 +105,7 @@ class _ModelMeta(abc.ABCMeta):
 
         for opermethod, opercall in opermethods:
             members[opermethod] = opercall
-        cls = super().__new__(mcls, name, bases, members)
+        cls = super().__new__(mcls, name, bases, members, **kwds)
 
         param_names = list(members['_parameters_'])
 
@@ -128,8 +128,8 @@ class _ModelMeta(abc.ABCMeta):
 
         return cls
 
-    def __init__(cls, name, bases, members):
-        super(_ModelMeta, cls).__init__(name, bases, members)
+    def __init__(cls, name, bases, members, **kwds):
+        super(_ModelMeta, cls).__init__(name, bases, members, **kwds)
         cls._create_inverse_property(members)
         cls._create_bounding_box_property(members)
         pdict = {}
@@ -694,6 +694,9 @@ class Model(metaclass=_ModelMeta):
     # If cov_matrix is available, then std will set as well
     _cov_matrix = None
     _stds = None
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__()
 
     def __init__(self, *args, meta=None, name=None, **kwargs):
         super().__init__()
--- a/astropy/modeling/tests/test_models.py
+++ b/astropy/modeling/tests/test_models.py
@@ -16,7 +16,7 @@ from astropy import units as u
 from astropy.modeling import fitting, models
 from astropy.modeling.models import Gaussian2D
 from astropy.modeling.bounding_box import ModelBoundingBox
-from astropy.modeling.core import FittableModel
+from astropy.modeling.core import FittableModel, _ModelMeta
 from astropy.modeling.parameters import Parameter
 from astropy.modeling.polynomial import PolynomialBase
 from astropy.modeling.powerlaws import SmoothlyBrokenPowerLaw1D
@@ -1023,3 +1023,35 @@ def test_SmoothlyBrokenPowerLaw1D_fit_deriv():
                                          estimate_jacobian=True)
     assert_allclose(new_model_with_deriv.parameters,
                     new_model_no_deriv.parameters, atol=0.5)
+
+
+class _ExtendedModelMeta(_ModelMeta):
+    @classmethod
+    def __prepare__(mcls, name, bases, **kwds):
+        # this shows the parent class machinery still applies
+        namespace = super().__prepare__(name, bases, **kwds)
+        # the custom bit
+        namespace.update(kwds)
+        return namespace
+
+    model = models.Gaussian1D(1.5, 2.5, 3.5)
+    assert model.amplitude._description == "Amplitude (peak value) of the Gaussian"
+    assert model.mean._description == "Position of peak (Gaussian)"
+
+
+def test_metaclass_kwargs():
+    """Test can pass kwargs to Models"""
+    class ClassModel(FittableModel, flag="flag"):
+        def evaluate(self):
+            pass
+
+    # Nothing further to test, just making the class is good enough.
+
+
+def test_submetaclass_kwargs():
+    """Test can pass kwargs to Model subclasses."""
+    class ClassModel(FittableModel, metaclass=_ExtendedModelMeta, flag="flag"):
+        def evaluate(self):
+            pass
+
+    assert ClassModel.flag == "flag"


pllim:
Hello and thanks! Can you please clarify what use cases that this patch is trying to solve? Also, this is technically an API change, so it needs a change log and perhaps some user documentation update as well.

nstarman:
@pllim, sure thing. The general use case is allowing modifications to the Model class construction process in the metaclass, rather than in __new__, which is not the best place for it. For instance, the Model subclasses have the ``inverse`` property overwritten by the metaclass, changing it to ``_inverse`` and using the ``Model.inverse`` function. While this is fine for Model and most subclasses, there is no good way to control this action. Subclasses of Model which wish to exert such type of control need to be able to accept an argument to do so. This may be implemented as a class attribute, or for a more inviolable option, as an arg/kwarg to the class construction itself.

```python
class ExampleModel(Model, metaclass=_ExtendedModelMeta, some_flag=True):
    pass
```

This PR just makes this possible to implement ``_ExtendedModelMeta `` by users foolhardy enough to wade into the murk of metaclasses.

From my experience, I want to override the signature of the evaluate method for a series of subclasses of Polynomial models. Aspects of this process are best / most efficiently accomplished at class declaration, not instantiation. Having flags allows the classes to be powerfully tailored in a manner which may be changed in the child classes and store the choices with inheritable private attributes.

nstarman:
@pllim squashed with tests, change-log, and docs.

pllim:
Thanks. Now let's wait for review from modeling maintainers. In the meantime, we ask for your continued patience.

nden (approved):

nden:
Looks good

nstarman:
codecov seems to be getting this wrong, none of the coverage changes are related to this PR.

nden (commented):
See [comment](https://github.com/astropy/astropy/pull/11098#discussion_r532220641)

nstarman:
> See [comment](https://github.com/astropy/astropy/pull/11098#discussion_r532220641)

Great. I can rebase as necessary. Or vice versa if this is approved first.

astropy/modeling/core.py
```diff
@@ -70,7 +70,7 @@ class _ModelMeta(abc.ABCMeta):
     Parameter descriptors declared at the class-level of Model subclasses.
     """
     @classmethod
-    def __prepare__(mcls, name, bases):
+    def __prepare__(mcls, name, bases, **kwds):
```
embray, line 73: Per https://github.com/astropy/astropy/pull/11098#discussion_r532220641 this will have to be rebased on #11098 and remove `__prepare__`.  Alternatively, if this PR is merged first I can rebase mine.  Either way.

docs/modeling/new-model.rst
```diff
@@ -277,3 +277,24 @@ input``.
...
+Advanced Notes
+==============
+
+The Model class implements a custom metaclass, ``modeling.core._ModelMeta``,
+which handles a number of modifications to the standard class creation
+process. For instance, user-defined ``inverse`` properties are overwritten,
+changing it to ``_inverse`` so that ``Model.inverse`` is always used (it calls
+``_inverse``).
+
+Subclassing this MetaClass, allows for further customization. This is not encouraged since ``_ModelMeta`` is currently a private class and is free to be changed at whim.
```
embray, line 291: Please wrap this line.

astropy/modeling/core.py
```diff
@@ -695,6 +695,9 @@ class Model(metaclass=_ModelMeta):
     _cov_matrix = None
     _stds = None
 
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__()
+
```
embray, line 700: Just to clarify, if you pass keyword arguments to a class, do they get passed to `__init_subclass__` too?  This is just a use case I' haven't seen before.  I would have assumed that if that's the case then the default `__init_subclass__` also takes kwargs but maybe not (hence the need for this).

embray:
I'm still a little confused about this.  What is the *specific* case motivating this?  I'm not aware of any case where it is necessary to subclass `_ModelMeta`, nor do I think the user documentation should say *anything* about extending `_ModelMeta`.

It does seem like accepting some arguments to Model subclasses could be useful in general though.

You mentioned something about the default `_ModelMeta` wrapping `inverse`.  Is there some reason you need it to *not* do that?  What is that reason?  And if so, maybe there is a more directed solution to that problem that can be resolved internally, rather than exposing such machinery at a user-documented level.

docs/modeling/new-model.rst
embray: Actually, per [this comment](https://github.com/astropy/astropy/pull/11017#issuecomment-736006874) I don't think this should be documented at all, nor should there be a changelog entry.

If there's a case where you want to be able to prevent wrapping `Model.inverse` maybe that can be an option directly to `_ModelMeta`.  I don't think we should be making anything about `_ModelMeta` publicly documented unless we were to make it a public method in the first place and document explicit use cases for subclassing it.

astropy/modeling/core.py
nstarman: Thanks! I've rebased. `__prepare__` is gone.

docs/modeling/new-model.rst
nstarman: > Actually, per this comment I don't think this should be documented at all, nor should there be a changelog entry.

I agree. I've removed the documentation.

> f there's a case where you want to be able to prevent wrapping Model.inverse maybe that can be an option directly to _ModelMeta.

Maybe for a future PR. 🤞

astropy/modeling/core.py
nstarman: yeah, fortunately or unfortunately the default `__init_subclass__` doesn't take a kwarg. This isn't a problem unless the class has a custom meta.

nstarman:
@embray The *specific* case is something in my personal research.

I want to override the signature of the ``evaluate`` method for a series of subclasses of Polynomial models. Aspects of this process are best / most efficiently accomplished at class declaration, not instantiation. However, when there is a custom metaclass, this type of manipulation is only possible if the metaclass allows for kwargs, which they often do because people (myself including) lazily define the `__new__` and `__init__` methods with `(*args, **kwargs)`. `_ModelMeta` was actually well programmed, which is a problem in this case. I just want to add `**kwargs` to the metaclass. These are non-functional on `_ModelMeta`,  but allow custom subclasses a greater degree of flexibility.

> It does seem like accepting some arguments to Model subclasses could be useful in general though

agreed!

embray:
@nstarman What do you mean by "override the signature of the evaluate method for a series of subclasses of Polynomial models" and is there a reason this couldn't be done with a simple class attribute?

nstarman:
> What do you mean by "override the signature of the evaluate method for a series of subclasses of Polynomial models" and is there a reason this couldn't be done with a simple class attribute?

@embray Sorry for the delay! I kind of lost track of this PR.

I've found a way around this in ``__new__``. What I wanted was to set a flag on the class that it should change the signature off ``evaluate`` when instantiated, while for other subclasses, I did not want that flag set to True. An example of this type of use is https://github.com/adrn/gala/blob/f7184e6b09fbc42a349f6b5a2bca6242f1e9936e/gala/potential/scf/bfe_class.pyx#L53, where the flag ``GSL_only`` is used.
This PR should not change any current functionality or create any new functionality, just enable others to add functionality on top of Model metaclasses.

nstarman:
@nden @embray @perrygreenfield, rebased to fix merge conflict.

nstarman:
@nden @embray @perrygreenfield, the above discussion, while relevant to the reason I pushed this PR, I feel is actually unrelated to the content of this PR, which only enables extending the Model metaclass, and doesn't actually do so here. This PR is certainly not critical for v4.3 and could be pushed to v5.0, but...  #ready-for-final-review ?

embray:
I had a look at that `GSL_only` example, and while I see what that's doing what confuses me is still why this syntax of passing arguments in the `class` statement needs to be used at all in this case, as opposed to the more common syntax of just declaring this in a class attribute like:

```
class SCFPotential(CPotentialBase):
    GSL_only = True
    ...
```

There's certainly nothing *wrong* with the other approach.  It's just less common and seems like overkill to me.  But maybe this is purely an aesthetic/subjective question.  It got me thinking, and I really don't know what a good use case is for custom arguments to `class` so I asked about this here: https://discuss.python.org/t/motivation-for-custom-keyword-arguments-to-class-statement/8402

But maybe there's something I'm completely missing as to why this is necessary.  So I'm definitely not opposed to this PR, and and mostly just struggling to understand the use case.

nstarman:
> But maybe there's something I'm completely missing as to why this is necessary. So I'm definitely not opposed to this PR, and and mostly just struggling to understand the use case.

It's funny because I worked around my use case as well :p ... As I understand it, the limitations of class arguments are twofold:
1. They actually need to be defined *on* the class and can be mutated like any other class method. An argument *to* the class itself need never be stored. It can set a behavior and then be done.
2. clarity: arguments *on* the class can be anything, e.g. ``n_inputs``, while arguments *to* the class are very explicitly arguments to the metaclass constructor.

In the context of  https://discuss.python.org/t/motivation-for-custom-keyword-arguments-to-class-statement/8402, that example is just a flag. But one idea for something that could be done that is more complicated is: wrapping methods for numpy ufuncs versus numba.


```
class ScalarClassWithMethods:
    def __init_subclass__(cls, wrap_method):
        wrapper = numba_wrapper if wrap_method == "numba" else numpy_wrapper
        for name, method in _get_my_methods_:
            setattr(cls, name, wrapper(method))

class NumpyWrapped(ScalarClassWithMethods, wrap_method="numpy"):
    pass

class NumbaWrapped(ScalarClassWithMethods, wrap_method="numba"):
    pass
```

pllim:
Seems like discussions are continuing... I am going to re-milestone to 5.0 so we don't make rash decisions due to lack of time.

astrofrog:
I am re-milestoning to v5.1 as we've now branched v5.0.x

pllim:
Hmm, this was approved by @nden back in Dec 2020 but has been languishing since and yet has a "ready for final review" label. Should the label be removed? What is the current status? Thanks!

nstarman:
Personally, I think it's ready to merge. I might actually remove some of the tests. I'm not sure if the change is large enough to merit testing.

WilliamJamieson (approved):
These changes seem reasonable to me.

pllim:
Does @nden want to re-review or is an approval from @WilliamJamieson enough?

nden:
William is a maintainer.
Time to merge this.

PR #11020, opened 2020-11-10
TST: Relax warning check in remote coordinates test with IERS download

pllim:
### Description

This pull request is to address the transient remote data failure caused by a warning we cannot control. This warning affects unrelated PRs, causing their remote data jobs to fail.

```
E           astropy.utils.exceptions.AstropyWarning: failed to download
ftp://anonymous:mail%40astropy.org@gdc.cddis.eosdis.nasa.gov/pub/products/iers/finals2000A.all
and https://datacenter.iers.org/data/9/finals2000A.all, using local IERS-B:
<urlopen error Unable to open any source!
Exceptions were {'ftp://anonymous:mail%40astropy.org@gdc.cddis.eosdis.nasa.gov/pub/products/iers/finals2000A.all':
URLError("ftp error: error_temp('425 Security: Bad IP connecting.')"),
'https://datacenter.iers.org/data/9/finals2000A.all':
URLError(gaierror(-5, 'No address associated with hostname'))}>
```

Example log: https://travis-ci.com/github/astropy/astropy/jobs/432522565

### DIFF
--- a/astropy/coordinates/tests/test_intermediate_transformations.py
+++ b/astropy/coordinates/tests/test_intermediate_transformations.py
@@ -2,6 +2,7 @@
 """Accuracy tests for GCRS coordinate transformations, primarily to/from AltAz.
 
 """
+import os
 
 import pytest
 import numpy as np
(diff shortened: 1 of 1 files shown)


mhvk (approved):
Pity that now we may not always know whether something else breaks, but I think this is the only reasonable solution to avoid getting so many false positives that we just stop looking. So, let's go for it!

pllim:
Thanks, @mhvk ! I left the check during local runs still strict, hoping we can still catch stuff if devs are diligent in running tests locally when they submit PR and during RC testings. 🤞

PR #11022, opened 2020-11-10
TST: Relax tolerances in test_bbox_size on Freetype version mismatch

### LINKED ISSUE #11007: [4.2rc1] test_bbox_size failure with Python 3.8/3.9
### Description

When I run the test suite during the build of the Debian package, `test_bbox_size` fails on Python 3.8 and 3.9.

### Expected behavior

The tests should succeed on all supported Python versions.

### Actual behavior

I get the following failure:
```
________________________________ test_bbox_size ________________________________

    def test_bbox_size():
        # Test for the size of a WCSAxes bbox
        x0 = 11.38888888888889
        y0 = 3.5
        # Upper bounding box limits (only have Matplotlib >= 3.0 now)
        x1 = 576.0
        y1 = 432.0
    
        fig = plt.figure()
        ax = WCSAxes(fig, [0.1, 0.1, 0.8, 0.8])
        fig.add_axes(ax)
        fig.canvas.draw()
        renderer = fig.canvas.renderer
        ax_bbox = ax.get_tightbbox(renderer)
>       assert np.allclose(ax_bbox.x0, x0)
E       assert False
E        +  where False = <function allclose at 0x7fb26d2d40d0>(11.26388888888889, 11.38888888888889)
E        +    where <function allclose at 0x7fb26d2d40d0> = np.allclose
E        +    and   11.26388888888889 = Bbox([[11.26388888888889, 3.5], [576.0, 432.0]]).x0

astropy/visualization/wcsaxes/tests/test_misc.py:521: AssertionError
= 1 failed, 15010 passed, 294 skipped, 60 xfailed, 12 xpassed in 717.27 seconds =
```

### Steps to Reproduce

Build (and test) the Debian experimental package (soon [in salsa](https://salsa.debian.org/debian-astro-team/astropy/-/tree/experimental)).

### System Details

 * Python 3.8.6, 3.9.0
 * Numpy 1.19.4
 * Astropy 4.2rc1
 * Scipy 1.5.3
 * Matplotlib 3.3.2

dhomeier:
### Description
Checks for Freetype library Matplotlib is linked against; if this is not the reference version 2.6.1 which Matplotlib mandates for its own tests to reproduce exact comparison images, allow `test_bbox_size` to check the bounding box extent with relaxed tolerance.

Fixes #11007

This combines several possible approaches to the problem: assuming the failure on Debian is related to its Freetype 2.10, run the test once with wider tolerance and allow it to fail on the stricter comparison. We may only need one of these – assuming the test as introduced by @dstansby in #10797 should assure that the plot fits comfortably within the `tightbbox`, `atol` of 0.1 – 0.2 pt would certainly be always acceptable.

### DIFF
--- a/astropy/visualization/wcsaxes/tests/test_misc.py
+++ b/astropy/visualization/wcsaxes/tests/test_misc.py
@@ -24,7 +24,9 @@ from astropy.visualization.wcsaxes.utils import get_coord_meta
 from astropy.visualization.wcsaxes.transforms import CurvedTransform
 
 mpl_version = Version(matplotlib.__version__)
+ft_version = Version(matplotlib.ft2font.__freetype_version__)
 MATPLOTLIB_LT_31 = mpl_version < Version('3.1')
+FREETYPE_261 = ft_version == Version("2.6.1")
 TEX_UNAVAILABLE = not matplotlib.checkdep_usetex(True)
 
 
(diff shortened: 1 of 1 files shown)


pllim (commented):
I don't think I know enough about the intent of this test to review, so I'll have to defer to others.

dstansby (approved):
Seems like a reasonable approach to take

pllim (approved):
Thanks, @dstansby ! I'll approve to remove merge block but will leave it open in case others want to review. @dhomeier , feel free to merge if no one objects after a few days?

dhomeier:
OK, thanks for reviewing – ideally @olebole would test the patch with the Debian build to confirm it actually fixes the failure, but at least it should not break anything!

astrofrog (approved):

olebole (approved):
I can confirm that this patch removes the error on Debian. It now builds nicely, thank you!

Cadair (approved):

PR #11028, opened 2020-11-11
EXP: Add duration measurements for RTD

pllim:
### Description

This pull request is to prove that `sphinx.ext.duration` does not work on RTD.

For readthedocs/readthedocs.org#7659

Also see #10807


pllim:
Wrong rabbit hole.

PR #11037, opened 2020-11-12
TST: Add back --open-files to one job

pllim:
### Description

This pull request is to investigate if we can still use `--open-files`.

This is a follow up of #10388

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -28,7 +28,7 @@ jobs:
             python: 3.8
             toxenv: py38-test-alldeps
             toxargs: -v --develop
-            toxposargs: --durations=50
+            toxposargs: --open-files
 
           # NOTE: In the build below we also check that tests do not open and
           # leave open any files. This has a performance impact on running the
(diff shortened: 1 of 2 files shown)


pllim:
Seems to work but I just don't know if adding `--open-files` to the job that also measures duration will mess up the latter. What do you think, @astropy/devops-maintainers @astrofrog @saimn et al?

astrofrog:
Yeah it definitely doesn't make sense to have the one that mesures durations be the same one.

pllim:
@astrofrog , any idea where is the next best place to put it without creating a new job just for it? We don't want it to miss too many tests and we know it doesn't work for `oldest-deps` and probably unwise to put it under Windows (not sure about OSX).

astrofrog:
What about adding open files to the 'Python 3.8 with all optional dependencies' on Linux and measuring durations on the MacOS X and Windows equivalent ones?

pllim:
Like this, @astrofrog ?

astrofrog (approved):
Yes that's fine by me :)

pllim:
Gah... It uses `openfiles-0.5.0` and took 1.5 hours. I thought I pinned it but apparently not. Too many actions.

Issue #11031, opened 2020-11-12
Are keys on a io.fits.Header object case insensitive?


dstansby:
The documentation for `astropy.io.fits.Header` does not mention anything about whether keys are case sensitive or insensitive, however the code has the following comment which would suggest the keys are case insensitive
https://github.com/astropy/astropy/blob/3627a7473e7c73f2e46e74157eaa4f078ccb8574/astropy/io/fits/header.py#L1634-L1635

It would be great to work out the sensitiveness of keys, and documented this somewhere in the docstring.

pllim:
By FITS standards, the keys are case insensitive, I think?

saimn:
FITS keywords are uppercase:

> The keyword name shall be a left justified, eight-character,
space-filled, ASCII string with no embedded spaces. All digits
0 through 9 (decimal ASCII codes 48 to 57, or hexadecimal 30
to 39) and upper case Latin alphabetic characters ’A’ through
’Z’ (decimal 65 to 90 or hexadecimal 41 to 5A) are permitted;
lower-case characters shall not be used.

PR #11039, opened 2020-11-13
Make sure access to Header items is case insensitive

dstansby:
Fixes https://github.com/astropy/astropy/issues/11031. This adds a test to check that access is case insensitive, and adds a note to the `Header` docstring.

### DIFF
--- a/astropy/io/fits/header.py
+++ b/astropy/io/fits/header.py
@@ -73,6 +73,11 @@ class Header:
         ...
 
     See the Astropy documentation for more details on working with headers.
+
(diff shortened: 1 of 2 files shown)


Cadair (approved):

saimn:
Looks good, thanks. Please add the changelog entry.

saimn:
I realized afterwards that a changelog entry is not really necessary for a docstring change, sorry.
So, looks good to me :+1:

saimn (approved):

PR #11040, opened 2020-11-13
upload any codecov in GitHub CI

nstarman:

### Description

codecov report is uploaded for any test with the "-cov" flag.

@pllim , you've been testing the GitHub CI, so I'm tagging you here.

no-changelog-needed ?

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -80,7 +80,7 @@ jobs:
       run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
     # TODO: Do we need --gcov-glob "*cextern*" ?
     - name: Upload coverage to codecov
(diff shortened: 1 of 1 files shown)


pllim:
@nstarman , why this change? There should only be one coverage calculation per build.

nstarman:
For the simple case that the test name can change. I've been setting up my GitHub CI and ran into this. Also, some features that rely on optional dependencies and have conditionals to test for an import. Codecov auto merges coverage reports, so I've taken to uploading 2: one for minimal dependencies, and another for all optional dependencies.

pllim:
Maybe this is more for the template than for core library. cc @astropy/devops-maintainers and @astrofrog

astrofrog (approved):
I personally think this change makes sense, it protects us against the case where we eg update the python version in the tox env name and forget to update the codecov action. It's also more concise and covers case where we might add one or more additional builds with coverage.

saimn (approved):
I agree it looks better :+1:

Issue #11038, opened 2020-11-13
TST: Remaining GitHub Actions actions


pllim:
### Description

Remaining things to port from `.travis.yml`:

- [x] ~Caching~ with `ccache` and `.hypothesis`. (`.hypothesis` caching PR opened at #11161, @saimn is interested to look at the rest when he gets the chance). **UPDATE: Haven't seen a real need for caching so far.**
  - Formal documentation at https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows (there is a 5 GB hard limit)
  - There is a [blog about using ccache in Actions](https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/) that maybe we can try.
  - ~Do we want to start caching pip dependencies too?~ See https://github.com/actions/cache/blob/main/examples.md#python---pip (The answer is "no" according to https://github.com/astropy/astropy/issues/11038#issuecomment-786770402 .)
- [x] `pytest` conditions in test modules using `TRAVIS` env var. See #11128
  - Example: https://github.com/astropy/astropy/blob/f9ba0f8e68abde11e5af05979213e3909c3f6fc7/conftest.py#L29
- [x] ~Slack notification~ (both on success and failure) for cron job. (@saimn tried something but it didn't work, @Cadair has a cunning plan at https://github.com/astropy/astropy/issues/11038#issuecomment-743229649.) **UPDATE: Nothing is happening on this front and no one has asked for it lately.**
- Cron:
  - [x] Big-endian s390x architecture with apt (`arch: s390x`, `ASTROPY_USE_SYSTEM_ALL=1`) (@astrofrog opened #11045 )
  - [x] ARM64 architecture with apt (`arch: arm64`, `ASTROPY_USE_SYSTEM_ALL=1`) (@astrofrog opened #11045 )
- [x] Finally, delete `.travis.yml`. (@astrofrog opened #11045 )

From `.circleci/config.yml`:

- [x] 32-bit and parallel (custom `astropy.cfg`, Python 3.7 `-n=4 --durations=50`, Python 3.8 `py38-test-double`). See #11043
- `image-tests-mpl302` (@Cadair says no need to port)
- `image-tests-mpl310` (@Cadair says no need to port)
- `image-tests-mpldev` (@Cadair says no need to port)

Things already in Actions we need to fix:

- [x] Allow remote data job to fail. See #11036
- [x] Separate cron jobs into daily and weekly. See #11036
[...]

PR #11042, opened 2020-11-14
MNT: Ignore divide by zero warning in np.log

pllim:
### Description

This pull request is to fix warnings causing failure in numpy-dev job. `np.log` must be issuing extra warning now in numpy-dev or something.

This is a follow up of #10388 and address part of #11038 .

### DIFF
--- a/astropy/visualization/stretch.py
+++ b/astropy/visualization/stretch.py
@@ -1,5 +1,4 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-
 """
 Classes that deal with stretching, i.e. mapping a range of [0:1] values onto
 another set of [0:1] values with a transformation
@@ -9,7 +8,6 @@ import numpy as np
 
 from .transform import BaseTransform
 from .transform import CompositeTransform
-
 
 __all__ = ["BaseStretch", "LinearStretch", "SqrtStretch", "PowerStretch",
            "PowerDistStretch", "SquaredStretch", "LogStretch", "AsinhStretch",
@@ -23,7 +21,8 @@ def _logn(n, x, out=None):
     if out is None:
         return np.log(x) / np.log(n)
     else:
-        np.log(x, out=out)
+        with np.errstate(divide='ignore', invalid='ignore'):
+            np.log(x, out=out)
         np.true_divide(out, np.log(n), out=out)
         return out
 
@@ -413,7 +412,7 @@ class LogStretch(BaseStretch):
 
         values = _prepare(values, clip=clip, out=out)
         replace_invalid = not clip and invalid is not None
-        with np.errstate(invalid='ignore'):
+        with np.errstate(divide='ignore', invalid='ignore'):
             if replace_invalid:
                 idx = (values < 0)
             np.multiply(values, self.exp, out=values)


pllim:
p.s. If you think we need to back port this all the way to 4.0.x, please change the milestone. Thanks.

astropy/visualization/stretch.py
```diff
@@ -413,7 +412,7 @@ def __call__(self, values, clip=True, out=None, invalid=None):
 
         values = _prepare(values, clip=clip, out=out)
         replace_invalid = not clip and invalid is not None
-        with np.errstate(invalid='ignore'):
+        with np.errstate(divide='ignore', invalid='ignore'):
```
dhomeier, line 415: I don't really have a good feeling about a blanket ignore on all divisions by zero as long as we don't understand the underlying failure. I cannot reproduce it locally (OS X) with the exact same numpy version 1.20.0.dev0+0ffaaf8 , and looking e.g. at https://github.com/astropy/astropy/pull/11043/checks?check_run_id=1400152307
it is incomprehensible how `LogStretch` for input `DATA=[0., 0.25, 0.5, 0.75, 1.] can come up with 
```
1289 ../../.tox/py37-test-devdeps/lib/python3.7/site-packages/astropy/visualization/tests/test_stretch.py:44: 
1290 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
1291
1292 self = <astropy.visualization.stretch.LogStretch object at 0x7f108716cfd0>
1293 values = array([      -inf, 5.52545294, 6.2166061 , 6.62140565, 6.90875478])
1294 clip = False, out = None, invalid = None
```
 – it's as if `DATA[0]` somehow missed the `np.add(values, 1.)` step...
At this point I am more inclined to consider this a buggy numpy build (no idea what best to do about that, though).

mhvk: I think `values` refers to the input values, so first element is `-inf` - this will remain `-inf` after multiplication and addition. I think the problem is that it leads to an invalid division error instead of the invalid error in the test case. That does sounds like a numpy bug, but I cannot reproduce it on my Debian/linux system either: `test-devdeps` just passes (and has the same numpy version, though I have a much more recent python than 3.7; perhaps we should run `devdeps` on python 3.9?)

dhomeier: I first thought the warning was referring to the `np.true_divide` step by mistake, since I cannot see which prior step
https://github.com/astropy/astropy/blob/d2721893accf29015ecfc4e6799ed4b5f163173f/astropy/visualization/stretch.py#L418-L422
could possibly produce `-inf`. Whereas
```
np.log([0, 251., 501., 751., 1001.])
array([      -inf, 5.52545294, 6.2166061 , 6.62140565, 6.90875478])
```
would produce exactly the reported output, so I rather think `out` has already been copied back to `values` when the warning is emitted. 

Running `devdeps` on a newer Python is certainly a good idea, although I suspect this issue is more related to build environment, possibly compilers, than the numpy + python versions as such.

dhomeier: Hmm. In #11048 I just tried to check if running a version with
`values = np.add(values, 1.)`
etc. would avoid the division by zero error. Running numpy-dev, which today has already advanced to `1.20.0.dev0+a645106`, produced no failures. However for e.g. `test_inplace` the `LogStretch` call should not have been affected in any way by my modification, so I am more inclined to mark this as a one-time failure with `0ffaaf8`.
Don't know if there is anything more we could do to retrace this or avoid future failures; but nightly dev builds may of course come back with new bugs at any time.

mhvk: I think it is fine to see if the next "nightly" (really, weekly) still produces the problem... It is very odd I couldn't even reproduce it with exactly the same `tox` setup and same version.

dhomeier: Yes, that's why I was also thinking of a compiler glitch or erratic memory failure (although on the build system it was obviously consistent over calls in `LogScale`). Among the changes between `0ffaaf8` and `a645106` https://github.com/numpy/numpy/pull/17755 looks like it _might_ be related, without having looked any deeper inside; hopefully the next build should bring confirmation.

pllim:
Hah. I like problem that disappears on its own. Thank you very much for checking!

PR #11043, opened 2020-11-14
TST: Move 32-bit and parallel job to Actions

pllim:
### Description

This pull request is to move "32-bit and parallel" job from CircleCI to GitHub Actions. I purposely let the full CI run to make sure I didn't mess up CircleCI config. (CircleCI seems to work as expected in the first commit, so now I will skip it.)

There were talks about we should stop using Docker image or something but I am not sure what to do with that part.

This is a follow up of #10388 and address part of #11038 .

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -109,7 +109,7 @@ jobs:
     - name: Run tests
       run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
 
-  32bit_and_parallel:
+  parallel_and_32bit:
     name: 32-bit and parallel
     runs-on: ubuntu-latest
     container:
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -115,8 +115,9 @@ jobs:
     container:
       image: quay.io/pypa/manylinux1_i686
     steps:
+    # TODO: Use newer checkout actions when https://github.com/actions/checkout/issues/334 fixed
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v1
       with:
         fetch-depth: 0
     - name: Write configuration items to standard location to make sure they are ignored in parallel mode
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -108,3 +108,34 @@ jobs:
       run: python -m pip install --upgrade tox codecov
     - name: Run tests
       run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
+
+  32bit_and_parallel:
+    name: 32-bit and parallel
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_workflows.yml
```diff
@@ -108,3 +108,34 @@ jobs:
...
       run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
+
+  parallel_and_32bit:
+    name: 32-bit and parallel
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/pypa/manylinux1_i686
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
```
pllim, line 119: Looks like this doesn't work with container?
```
/usr/bin/docker exec  .... sh -c "cat /etc/*release | grep ^ID"
OCI runtime exec failed: exec failed: container_linux.go:370: starting container process caused: no such file or directory: unknown
```

pllim: Looks like a known and unresolved issue at actions/checkout#334

pllim:
I think it works. But you should check carefully.

Actions logs: https://github.com/astropy/astropy/pull/11043/checks?check_run_id=1400152329
CircleCI logs: https://app.circleci.com/pipelines/github/astropy/astropy/3246/workflows/1b64c442-a496-4ba1-b769-34509c767e58/jobs/81931

pllim:
Is there anything that is blocking this?

.github/workflows/ci_workflows.yml
```diff
@@ -108,3 +108,35 @@ jobs:
...
+      with:
+        fetch-depth: 0
+    - name: Write configuration items to standard location to make sure they are ignored in parallel mode
+      run: |
+        mkdir -p $HOME/.astropy/config/
+        printf "unicode_output = True\nmax_width = 500" > $HOME/.astropy/config/astropy.cfg
+    # In addition to testing 32-bit, we also use the 3.7 builds to
+    # test the ability to run the test suite in parallel.
+    - name: Install dependencies for Python 3.7
+      run: /opt/python/cp37-cp37m/bin/pip install tox
```
astrojuanlu, line 130: Question: Does

```suggestion
      run: python3.7 -m pip install tox
```

work instead? (Same below with Python 3.8) or are these paths not in `$PATH` by default?

astrojuanlu: Oh, I see! It's a `quay.io/pypa/manylinux1_i686` thing :+1:

astrojuanlu (approved):
The Actions spec looks like a good translation of the Circle CI one. I went over the Actions logs and didn't spot anything suspicious.

.github/workflows/ci_workflows.yml
astrojuanlu: Subscribed!

saimn (approved):
LGTM.

saimn:
I don't see a reason to delay this, so let's merge! Thanks @pllim.

pllim:
Thanks for the reviews, y'all!

PR #11044, opened 2020-11-14
TST: Directives to skip Actions

pllim:
### Description

This pull request is to allow these directives in a PR commit message to skip GitHub Actions (would be nice if Actions support them natively but they don't):

* `[ci skip]`
* `[skip ci]`
* `[skip actions]`
* `[actions skip]`

Follow up of #10388 and addresses part of #11038 .


pllim:
Wow, okay. This is way more painful that it is supposed to be. All the posted solutions that people swear by on the Internet no longer works.

https://dev.to/epassaro/use-skip-ci-in-github-actions-1mnf

https://github.com/marketplace/actions/skip-based-on-commit-message

Something that should have been supported natively. C'mone, Microsoft!

PR #11046, opened 2020-11-15
Updating spinner API, keeping enum

### LINKED ISSUE #11004: Outdated documentation for Spinner

Iamsoto:
Updating the spinner API, keeping the original enum.
See #11015  for the alternative method (removing the enumerator)

Fixes #11004


embray:
LGTM; can you please also add the change from https://github.com/astropy/astropy/pull/11015/files#diff-582a94e0a0f7293634d5bb6531b7bdf8a847ecd2a380ec691a65bac8beaff546L1040 ?

There's also the issue I raised on your other PR about the interpretation of the `step` argument maybe actually being buggy.  If you agree with me there should be a separate issue opened for that though.

Iamsoto:
> LGTM; can you please also add the change from https://github.com/astropy/astropy/pull/11015/files#diff-582a94e0a0f7293634d5bb6531b7bdf8a847ecd2a380ec691a65bac8beaff546L1040 ?
> 
> There's also the issue I raised on your other PR about the interpretation of the `step` argument maybe actually being buggy. If you agree with me there should be a separate issue opened for that though.

Done!
Yeah, I just thought it would be better ettiquete to not needlessly modify code XD

Iamsoto:
Ah! I just read your comment in the other PR about the bug in Step!
I don't mind making that another issue. Should I create the issue and resolve it before we merge this PR?

embray:
@Iamsoto Sorry for the late reply.  Yes, let's open a separate issue for that.

pllim:
Ops, sorry that this feel through the cracks. I'll just clean this up (i.e., fix the change log) and get this in. I also opened a follow-up issue at #11656.

pllim:
Re-reading https://github.com/astropy/astropy/pull/11046#issuecomment-733225534 -- @embray , should we get this in, or move the milestone. Is #11656 blocking this PR?

eteq:

embray:
I think it would be fine to address #11656 in a follow-up.  It's a nice "beginner" issue.

pllim (approved):
I don't think the failures were related.

pllim:
@embray , I'll let you do the merge since I haven't seen any formal approval from you yet.

pllim:
Actually, the timeout might be real. `main` now consistently times out after the merge of this PR. We need a follow-up patch or revert.

embray:
Wow....hard to imagine why

pllim:
Could be a coincidence though I restarted the build a few times and it is consistently timing out. Maybe I'll ping @astrojuanlu to make sure this is not a server-side problem...

embray:
If anything it would be due to #11690 which I also just merged.

pllim:
But the first incident happened with this merge, according to https://readthedocs.org/projects/astropy/builds/

embray:
Yes, and RTD built that PR fine...

PR #11048, opened 2020-11-16
MNT: test LogStretch without all in-place operations

dhomeier:
Experimental PR to probe `RuntimeWarning: divide by zero encountered in log` seen in dumpy-dev Github Actions job and addressed in #11042.
Appears to be related to `values` array not being fully updated before the `np.log` call with only the `devdeps` Numpy version on the GH build host; just testing here if calling numpy functions without in-place operations changes results.
Not sure how to proceed if it does...


dhomeier:
Should be resolved upstream now!

PR #11053, opened 2020-11-16
TST: Move numpy-dev cron from daily to weekly

pllim:
### Description

This pull request is to move numpy-dev cron job from daily to weekly. This is because we use the numpy-dev wheel which is only updated weekly, so I see little point in running it daily. We also run numpy-dev job for push events already.

This is a follow up of #10388 .

xref #11038

p.s. Cancelled CI because not needed.

### DIFF
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -14,16 +14,6 @@ jobs:
       matrix:
         include:
 
-          # We check numpy-dev also in a job that only runs from cron, so that
-          # we can spot issues sooner. We do not use remote data here, since
-          # that gives too many false positives due to URL timeouts. We also
-          # install all dependencies via pip here so we pick up the latest
-          # releases.
-          - name: Python 3.7 with dev version of key dependencies
-            os: ubuntu-latest
-            python: 3.7
-            toxenv: py37-test-devdeps
-
           - name: Bundling with pyinstaller
             os: ubuntu-latest
             python: 3.8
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cron.yml
```diff
@@ -42,6 +32,6 @@ jobs:
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: sudo apt-get install language-pack-de tzdata
     - name: Install Python dependencies
-      run: python -m pip install --upgrade tox codecov
```
pllim, line 45: Also don't think we need `codecov` for cron jobs.

mhvk (approved):
Yes, makes total sense!

bsipocz (approved):

pllim:
Thanks for the reviews!

saimn:
Do we need to run pyinstaller daily ? :thinking:

pllim:
> Do we need to run pyinstaller daily ?

Not sure. It is not run on the PR, so doesn't hurt to make sure changes for the day did not break it?

saimn:
Well, it's just another packaging tool. Running it every day certainly helps to find which change could broke it, but I'm not sure if it's worth the extra computing cost ? (That said, if Github really wanted to saved some computation time on their CI, they should just stop running Actions on forks by default).

pllim:
> if Github really wanted to saved some computation time on their CI

Or support skip ci, etc etc. Yeah.

pllim:
Re: pyinstaller -- @saimn , if you want to pursue moving it to weekly, please open a new issue for discussion. Thanks!

PR #11054, opened 2020-11-16
Restrict builds on push events to master

saimn:
This is to avoid running Actions twice, when pushing a branch and opening a PR.

This double build was also happening sometimes with Travis when pushing a branch directly on the repo and could be avoided with a 
```
branches:
  only:
  - master
```
block.

The difference with Actions is that it is also activated by default on forks, so when pushing a branch to a fork and opening a PR the workflow is run twice, on the fork (push event) and on the main repo (pull_request event).

Note: we should allow the maintenance branch if we backport Actions (#11050).

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -2,6 +2,9 @@ name: CI
 
 on:
   push:
+    branches:
+    - master
+    tags:
   pull_request:
 
 jobs:


pllim (approved):
This would not prevent Actions from running on a fork in the following situations:

* Someone opening PR from their own `master` branch on the fork. It happens more often than I like (sometimes even from seasoned contributors) but since we already tell people not to do that in the docs, fixing it is out of scope here as it would be a dev doc fix or contributor education more than anything else.
* PR to a PR branch on a fork. But maybe we want the CI to run in that case, so nothing to fix here as well.

Thanks!

saimn:
Yes indeed. And another annoyance is that the cron builds run on forks too.

pllim:
I think I remember getting a message from Actions that they automatically cancel cron jobs on forks now.

pllim:
And looks like every contributor who forked has option to disable Actions altogether on their forks: https://github.community/t/stop-github-actions-running-on-a-fork/17965/2

Should we recommend this in contributing guide or something?

saimn:
The daily cron has been running on my fork: https://github.com/saimn/astropy/actions/runs/367311810

Yes, you deactivate it on the fork, but I guess too few people would do it. So maybe instead we can skip explicitly the cron workflows if not running on the main repo.

pllim:
Hmm. Maybe because I haven't sync-ed my fork for some time? 🤷 

> So maybe instead we can skip explicitly the cron workflows if not running on the main repo.

That would be nice, but how? I only find https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository that uses the GitHub UI.

saimn:
#11056 should do it :crossed_fingers:

PR #11055, opened 2020-11-17
Deprecate "accuracy" in all_world2pix

mcara:
Docstring for `WCS.all_world2pix()`  incorrectly uses ``accuracy`` instead of ``tolerance`` in the function signature. Assuming there are any users who might have used ``accuracy`` based on this incorrect documentation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -80,9 +80,6 @@ astropy.wcs
 ^^^^^^^^^^^
 - Add IVOA UCD mappings for some FITS WCS keywords commonly used in solar physics. [#10965]
 
-- Deprecate ``accuracy`` argument in ``all_world2pix`` which was mistakenly
-  *documented*, in the case ``accuracy`` was ever used. [#11055]
-
 API Changes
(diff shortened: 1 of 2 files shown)


pllim (commented):
Just curious: If someone does use `accuracy` accidentally before, what would actually happen?

Also, `accuracy` would have been part of `**kwargs`. Are you sure it is not passed into other stuff internally? With this patch, a legit use of `accuracy` (if allowed) would now be not allowed, if that makes sense.

mcara:
> Just curious: If someone does use `accuracy` accidentally before, what would actually happen?

It would have resulted in ignoring user-supplied value and using `tolerance=1e-4` instead.

> Also, `accuracy` would have been part of `**kwargs`. Are you sure it is not passed into other stuff internally? With this patch, a legit use of `accuracy` (if allowed) would now be not allowed, if that makes sense.

I am pretty sure it is not used anywhere + all tests pass.

nden (approved):
LGTM

PR #11056, opened 2020-11-18
Run cron workflows only on the main repo

saimn:
To avoid running cron jobs on forks (see discussion in https://github.com/astropy/astropy/pull/11054#issuecomment-729001804).

_edit:_ I canceled the normal workflow since it is not affected by this.

### DIFF
--- a/.github/workflows/ci_cron.yml
+++ b/.github/workflows/ci_cron.yml
@@ -9,6 +9,7 @@ jobs:
   tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    if: github.repository == 'astropy/astropy'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci_cron_weekly.yml
+++ b/.github/workflows/ci_cron_weekly.yml
@@ -9,6 +9,7 @@ jobs:
   tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    if: github.repository == 'astropy/astropy'
     strategy:
       fail-fast: false
       matrix:


.github/workflows/ci_cron.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    if: github.repository == ‘astropy/astropy’
```
pllim, line 12:
```suggestion
    if: github.repository == 'astropy/astropy'
```

.github/workflows/ci_cron_weekly.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    if: github.repository == ‘astropy/astropy’
```
pllim, line 12:
```suggestion
    if: github.repository == 'astropy/astropy'
```

pllim (commented):
Thanks! We can certainly give this a try.

Do we also want to limit the runs only to `master` branch?

The quotes look a little weird, so I suggested changes.

saimn:
> Do we also want to limit the runs only to master branch?

Those workflows run only on cron event, so only for master.

> The quotes look a little weird, so I suggested changes.

Good catch, thanks (the problem of copy/pasting from the internet :P)

pllim (commented):
Might wanna squash the commits. 😜 

> Those workflows run only on cron event, so only for master.

If we start to have this cron YAML in release branches, will cron run for those too? Do we want it to?

astrofrog:
In the past I think we do run the cron jobs in releases branches but as regular jobs not cron?

pllim:
> but as regular jobs not cron?

How is that possible? You customized the `.travis.yml` in release branches?

saimn:
Actually I think the workflow will run on the branch where it is triggered, so it should work for any branch.

mhvk:
Once this is in, it may be good to send a message to `astropy-dev` to tell people to pull upstream and push to their own fork, so that the auto-runs stop.

Does this affect those that followed recommendations and changed the name `master` to `placeholder`?

pllim (approved):
I am just going to get this in now, as all the other conversations are more for a follow-up and do not block this patch. Thanks!

pllim:
xref #11038  #10388

saimn:
I don't think many people keep their master branch in sync with the main repo, so it should not be a problem for existing forks. 
It would mostly be a problem for future forks.

mhvk:
True, probably unnecessary to warn people given how short a period this can have been a problem.

PR #11058, opened 2020-11-18
MNT: Pin pytest-astropy-header to 0.1.2 for LTS

pllim:
### Description

This pull request is a follow-up of astropy/pytest-astropy-header#32 . Not even sure if it is really needed, so please review carefully.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,6 +54,7 @@ asdf_extensions =
 
 [options.extras_require]
 test =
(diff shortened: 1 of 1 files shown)


saimn:
I don't think this is needed since the code removed in pytest-astropy-header was for astropy < 4.0

pllim:
Okay then. Sorry for the noise!

Issue #10996, opened 2020-11-06
Cannot directly transform output from GCRS to AltAz / need ITRS location?


mhvk:
Following discussion in #10994, I thought the example below should work, where I use `get_body()` to get a `SkyCoord` with a GCRS position for Mars at a given time and location, and try to transform it to AltAz just using `body.altaz`.

However, it fails, because the temporary `AltAz` frame that was created does not have a `.location` attribute. This is unexpected, since the `SkyCoord` made by `get_body` does have that information - it is just stored in attributes with a different name, `obsgeoloc` and `obsgeovel` (which are `CartesianRepresentation` instead of `EarthLocation` for `.location`). Indeed, if I explicitly set my `body.location` to the relevant value, `body.altaz` does work.

So I guess one question is whether we can just translate between these attributes. In principle, the simplest route might be to just let `AltAz` understand `obsgeopos` as an alias for `location` (and maybe eventially deprecate `AltAz.location` in favour of it).

But perhaps a larger question is why `AltAz` and, e.g., `GCRS` are using different attributes in the first place for storing the observer location.

I should add that I'm not 100% sure either is ideal - for general use, I'd want to have the velocity stored on the position as a derivative (i.e., at least `obsgeovel` as a `.derivaties['s']` on `obsgeopos` -- of course, the derivatives did not exist when `obsgeovel` was first introduced!).  Also, it would seem better to have something that is explicit in what frame the position and velocity are defined. I think the information needed matches exactly what would be an `ITRS` frame instance. So, perhaps one should have a single attribute that replaces `location`, `obsgeoloc` and `obsgeovel`? (and deprecate those three?) Reuse `location`? New `obs_itrs`?

cc @StuartLittlefair, @mkbrewer, @eteq, @adrn, @astrojuanlu

```
from astropy.coordinates import get_body, ITRS
from astropy.time import Time

body = get_body('mars', Time(['J2010'], location=(10.*u.deg, 20*u.deg)))
# The following fails:
body.altaz
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-13-dd22cce62d45> in <module>
----> 1 get_body('mars', Time(['J2010'], location=(10.*u.deg, 20*u.deg))).altaz

/usr/lib/python3/dist-packages/astropy/coordinates/sky_coordinate.py in __getattr__(self, attr)
    785             frame_cls = frame_transform_graph.lookup_name(attr)
[...]

PR #11061, opened 2020-11-19
Use EarthLocation instead of GCRS position for TETE observer.

mhvk:
As discussed in #10994 and #10996, the TETE frame that was introduced for 4.2 should more logically use `EarthLocation` to indicate the location of the observer, instead of `obsgeoloc` and `obsgeovel`. This PR makes that change.

One important question is whether this can still be included in 4.2, even though it is *very* late. The main advantage is that in that case we do not have to deprecate `obsgeoloc` and `obsgeovel`. (The PR is only a bug fix in this regard.)

Input from @astropy/core-release-maintainers much appreciated!!

cc @mkbrewer, @StuartLittlefair

### DIFF
--- a/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
@@ -271,9 +271,8 @@ def hcrs_to_hcrs(from_coo, to_frame):
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, TETE, TETE)
 def tete_to_tete(from_coo, to_frame):
-    if (np.all(from_coo.obstime == to_frame.obstime) and
-            np.all(from_coo.obsgeoloc == to_frame.obsgeoloc) and
-            np.all(from_coo.obsgeovel == to_frame.obsgeovel)):
+    if (np.all(from_coo.location == to_frame.location)
+            and np.all(from_coo.obstime == to_frame.obstime)):
         return to_frame.realize_frame(from_coo.data)
     else:
         # We perform the self-transform through ICRS, since any change in obstime
--- a/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
+++ b/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
@@ -85,21 +85,37 @@ def gcrs_precession_mat(equinox):
     return erfa.fw2m(gamb, phib, psib, epsa)
 
 
+def get_location_gcrs(reference, gcrs_to_reference_matrix):
+    """Create a GCRS frame at the location and obstime of a given reference.
+
+    Helper function that avoids location.get_gcrs (which would trigger
+    infinite recursion in some cases), and uses an already calculated
+    GCRS to Reference matrix to calculate obsgeoloc and obsgeovel
+    required for GCRS.
+    """
+    # TODO: ideally, GCRS and other frames would share how they describe
+    # the location.  See gh-10996.
+    obstime = reference.obstime
+    gcrs_cart = (reference.location.get_itrs(obstime, include_velocity=True)
+                 .transform_to(reference.__class__(obstime=obstime))
+                 .cartesian
+                 .transform(matrix_transpose(gcrs_to_reference_matrix)))
+    return GCRS(obstime=obstime,
+                obsgeoloc=gcrs_cart.without_differentials(),
+                obsgeovel=gcrs_cart.differentials['s'].to_cartesian())
+
+
 # now the actual transforms
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GCRS, TETE)
 def gcrs_to_tete(gcrs_coo, tete_frame):
-    # first apply GCRS self transform to correct time and observatory position/velocity
-    obsgeoloc, obsgeovel = tete_frame.location.get_gcrs_posvel(tete_frame.obstime)
-    gcrs_coo2 = gcrs_coo.transform_to(GCRS(obstime=tete_frame.obstime,
-                                           obsgeoloc=obsgeoloc,
-                                           obsgeovel=obsgeovel))
-
-    jd1, jd2 = get_jd12(tete_frame.obstime, 'tt')
     # Classical NPB matrix, IAU 2006/2000A
     # (same as in builtin_frames.utils.get_cip).
-    rbpn = erfa.pnm06a(jd1, jd2)
-
+    rbpn = erfa.pnm06a(*get_jd12(tete_frame.obstime, 'tt'))
+    # Get GCRS coordinates for the target observer location and time.
+    loc_gcrs = get_location_gcrs(tete_frame, rbpn)
+    gcrs_coo2 = gcrs_coo.transform_to(loc_gcrs)
+    # Now we are relative to the correct observer, do the transform to TETE.
     # These rotations are defined at the geocenter, but can be applied to
     # topocentric positions as well, assuming rigid Earth. See p57 of
     # https://www.usno.navy.mil/USNO/astronomical-applications/publications/Circular_179.pdf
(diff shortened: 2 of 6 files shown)


astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -90,9 +90,10 @@ def gcrs_precession_mat(equinox):
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, GCRS, TETE)
 def gcrs_to_tete(gcrs_coo, tete_frame):
     # first apply GCRS self transform to correct time and observatory position/velocity
+    obsgeoloc, obsgeovel = tete_frame.location.get_gcrs_posvel(tete_frame.obstime)
```
mkbrewer, line 93: This is definitely not the way to go here. You are calculating the BPN matrix twice. This should be done in the same manner as `gcrs_to_cirs()`.

mkbrewer: This is why I liked @StuartLittlefair's `EarthLocation.get_posvel()` method.

mhvk: OK, that makes sense. I'll adjust.  There is also a real failure in comparison with horizons, so I may be missing more than this...

mhvk: Just to be sure, is there still a benefit from having the GCRS<->TETE conversion explicitly defined?

mhvk: (On `get_posvel()`: only useful if we stick with `obsgeoloc, obsgeovel` for GCRS...)

mhvk: Never mind on the benefit - things break greatly if I just remove the transformations...

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -116,8 +117,9 @@ def tete_to_gcrs(tete_coo, gcrs_frame):
     rbpn = erfa.pnm06a(jd1, jd2)
     newrepr = tete_coo.cartesian.transform(matrix_transpose(rbpn))
 
-    gcrs = GCRS(newrepr, obstime=tete_coo.obstime, obsgeoloc=tete_coo.obsgeoloc,
-                obsgeovel=tete_coo.obsgeovel)
+    obsgeoloc, obsgeovel = tete_coo.location.get_gcrs_posvel(tete_coo.obstime)
+    gcrs = GCRS(newrepr, obstime=tete_coo.obstime,
+                obsgeoloc=obsgeoloc, obsgeovel=obsgeovel)
 
     # now do any needed offsets (no-op if same obstime and 0 pos/vel)
```
mkbrewer, line 124: May as well fix this comment also. Drop the zero.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
mkbrewer: How else would we get the apparent position of a solar system body? Those are returned in the GCRS frame. For that matter, as far as I can see any ICRS<->TETE goes through GCRS.

mkbrewer: I do think that you should stick with `obseoloc, obsgeovel` for GCRS as GCRS is the natural Earth centered frame for a non-terrestrial observer location.

mkbrewer: Also, `obsgeloc` is definitely needed in order to calculate the time delay to the observer in `get_body()`.

mkbrewer: This is why I advocated for `get_gcrs_posvel()` to be incorporated into `pos_vel()`. That single `EarthLocation` method could handle all of these cases.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
mhvk: Done

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
mkbrewer: I suppose another possibility would be to add an optional argument to `get_gcrs_posvel(rbpn=None)` then if it is called without and BPN matrix, it would act just as it does currently, but if it was given one it would use it.

mhvk: OK, I think I got it in a more logical state now. @mkbrewer - Would you mind having another look?

StuartLittlefair: I agree with @mkbrewer that we should have something as flexible as `obsgeoloc{vel}` for GCRS. However, we may want to revisit the exact way these are stored. These choices pre-date coordinates having velocities, for example. 

We might want to use a different class to store the observer position, in which case it makes sense for a `pos_vel` routine on `EarthLocation` to return that same class. So having the method there makes sense eventually, but it’s not yet clear it should return a pair of `CartesianRepresentation`s.

mhvk: On `obsgeopos/vel`, I think it is mostly a question of whether it should instead be a `GCRS` instance that holds the position/velocity. Anyway, that is for #10996!

mhvk: p.s. Note that what I'm doing in the new commit is not really quite equivalent to `get_gcrs_posvel(rbpn=None)` -- am I still missing the main speed-up?

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
```diff
@@ -272,8 +272,7 @@ def hcrs_to_hcrs(from_coo, to_frame):
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, TETE, TETE)
 def tete_to_tete(from_coo, to_frame):
     if (np.all(from_coo.obstime == to_frame.obstime) and
```
StuartLittlefair, line 274: Can we swap these round? A common use pattern is for location to be singular and obstime to be a large array. If the conditions are reversed then the faster condition may fail first and avoid the longer check on obstime.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -108,27 +125,21 @@ def gcrs_to_tete(gcrs_coo, tete_frame):
...
-    rbpn = erfa.pnm06a(jd1, jd2)
+    # Compute the pn matrix, and then multiply by its transpose.
+    rbpn = erfa.pnm06a(*get_jd12(tete_coo.obstime, 'tt'))
     newrepr = tete_coo.cartesian.transform(matrix_transpose(rbpn))
-
-    gcrs = GCRS(newrepr, obstime=tete_coo.obstime, obsgeoloc=tete_coo.obsgeoloc,
-                obsgeovel=tete_coo.obsgeovel)
-
-    # now do any needed offsets (no-op if same obstime and 0 pos/vel)
+    # Transform to GCRS but still at the TETE location and obstime.
```
StuartLittlefair, line 131: Super-picky: the transformation has actually been done in the line above. So a different verb is needed - perhaps “Construct a GCRS coord” instead of “Transform to GCRS”.

StuartLittlefair (commented):
Looking good! Some minor nit-picks but the code is fine IMO.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
mkbrewer: Well, it basically is. It's just that it requires another function whereas all of this could be incorporated into a single EarthLocation.get_gcrs_posvel(rpbn=None) method. Anyway, this looks good for now.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
mhvk: Indeed! Copy & pasted too quickly - will change it for CIRS too.

astropy/coordinates/builtin_frames/icrs_cirs_transforms.py
mhvk: Yes, makes sense.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
StuartLittlefair: The main way to speed this up would be to chain matrices together; since we know this is a transform at the same obstime we are just doing rotations to get from ITRS->GCRS so you can avoid some of the flexibility of `transform_to`. 

You could pass a list - in the correct order - of already calculated matrices to transform ITRS into GCRS like so:

```python
matrix_chain = [matrix_transpose(tete_to_itrs_mat(frame.obstime)), rbpn]
loc_gcrs = get_location_gcrs(frame, matrix_chain)
```

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
mhvk: Somewhat changed the wording overall, making it clearer for me at least...

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
mhvk: In fact, that could just be 
```
itrs_to_gcrs = matrix_transpose(tete_to_itrs_mat(frame.obstime) @ rbpn)
loc_gcrs = get_location_gcrs(frame, itrs_to_gcrs))
```
with indeed a simpler `get_location_gcrs`. (I may have swapped things...)

mhvk:
Note to self: this will need a separate PR to backport to 4.2, since I now use stuff from #10994.

mkbrewer:
Well, that could be avoided by including #10994 in the yet to be released 4.2.

mkbrewer:
#10994 and this are really one package, after all.

StuartLittlefair (approved):

mhvk:
I'll merge this and first of all make a back-port of just this to 4.2, since it is super self-contained, not changing any behaviour, and with, I think, next to zero risk of introducing new problems. But I'm happy to make an alternative that includes #10994.

mhvk:
Hmm, but a backport without #10994 breaks accuracy tests for TETE... (the backport can be seen at #11065). So, I think a backport is perhaps just too late (though I made one just in case at #11064)

mhvk:
Backport was tricky so is done separately in #11065

PR #11062, opened 2020-11-19
Fix leap second update when using a non english locale

saimn:
I'm working on a documentation, and for every build I see the leap second update with an expiration warning:
```
Downloading https://hpiers.obspm.fr/iers/bul/bulc/Leap_Second.dat
|==========================================================| 1.3k/1.3k (100.00%)         0s
Downloading https://www.ietf.org/timezones/data/leap-seconds.list [Done]
WARNING: IERSStaleWarning: leap-second file is expired. [astropy.utils.iers.iers]
```
I already saw this in the past, so this is not new, but I never took the time to investigate... I can reproduce with various documentations ([1](https://github.com/GeminiDRSoftware/DRAGONS/tree/master/astrodata/doc/ad_ProgManual), [2](https://github.com/GeminiDRSoftware/DRAGONS/tree/master/astrodata/doc/ad_UserManual), [3](https://github.com/musevlt/origin/tree/master/docs)), but the weird thing is that there is no warning when importing directly astropy.

Putting some prints in `LeapSeconds.auto_open` gives this for a normal import:
```
❯ python -c "import astropy.coordinates.earth_orientation"
open erfa
expires: 2017-06-30 00:00:00.000
open /home/simon/dev/astropy/astropy/utils/iers/data/Leap_Second.dat
expires: 2021-06-28
```
which is fine. And this for a Sphinx build:
```
open erfa
expires: 2017-06-30 00:00:00.000
open /home/simon/dev/astropy/astropy/utils/iers/data/Leap_Second.dat
ERROR: time data '28 June 2021' does not match format '%d %B %Y'
open https://hpiers.obspm.fr/iers/bul/bulc/Leap_Second.dat
ERROR: time data '28 June 2021' does not match format '%d %B %Y'
open https://hpiers.obspm.fr/iers/bul/bulc/Leap_Second.dat
Downloading https://hpiers.obspm.fr/iers/bul/bulc/Leap_Second.dat
|==========================================================| 1.3k/1.3k (100.00%)         0s
ERROR: time data '28 June 2021' does not match format '%d %B %Y'
open https://www.ietf.org/timezones/data/leap-seconds.list
Downloading https://www.ietf.org/timezones/data/leap-seconds.list [Done]
ERROR: time data '28 June 2021' does not match format '%d %B %Y'
WARNING: IERSStaleWarning: leap-second file is expired. [astropy.utils.iers.iers]
```
So when running with Sphinx, there is a parsing error with the `Leap_Second.dat` file, which lead to this:
```
ipdb> Time.strptime('28 June 2021', '%d %B %Y', scale='tai', format='iso', out_subfmt='date')
*** ValueError: time data '28 June 2021' does not match format '%d %B %Y'
```

`%B` is the _Locale's full month name_, but in both case my locale if `fr_FR`.
Diving deeper leads to [_strptime.py](https://github.com/astropy/astropy/blob/master/astropy/extern/_strptime.py) where `_getlang` returns `None` in one case and `fr_FR` in the other, the latter resulting in a regex with French month names :tada:
And I guess Sphinx is setting `LC_TIME` to the default locale, which triggers the error in `_strptime.py`.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -68,9 +68,9 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
-    - name: Install language-pack-de and tzdata
+    - name: Install language-pack-fr and tzdata
       if: startsWith(matrix.os, 'ubuntu')
-      run: sudo apt-get install language-pack-de tzdata
+      run: sudo apt-get install language-pack-fr tzdata
     - name: Install Python dependencies
       run: python -m pip install --upgrade tox codecov
     - name: Run tests
--- a/astropy/utils/iers/iers.py
+++ b/astropy/utils/iers/iers.py
@@ -76,6 +76,9 @@ suppressed by setting the auto_max_age configuration variable to
   conf.auto_max_age = None
 """
 
+MONTH_ABBR = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug',
+              'Sep', 'Oct', 'Nov', 'Dec']
+
 
 def download_file(*args, **kwargs):
     """
@@ -1031,9 +1034,10 @@ class LeapSeconds(QTable):
             for line in lines:
                 match = cls._re_expires.match(line)
                 if match:
-                    expires = Time.strptime(match.groups()[0], '%d %B %Y',
-                                            scale='tai', format='iso',
-                                            out_subfmt='date')
+                    day, month, year = match.groups()[0].split()
+                    month_nb = MONTH_ABBR.index(month[:3]) + 1
+                    expires = Time(f'{year}-{month_nb:02d}-{day}',
+                                   scale='tai', out_subfmt='date')
                     break
             else:
                 raise ValueError(f'did not find expiration date in {file}')
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1110,6 +1110,8 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- Fix leap second update when using a non english locale. [#11062]
+
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
 
--- a/astropy/io/ascii/tests/test_read.py
+++ b/astropy/io/ascii/tests/test_read.py
@@ -1290,9 +1290,9 @@ def test_non_C_locale_with_fast_reader():
 
     try:
         if platform.system() == 'Darwin':
-            locale.setlocale(locale.LC_ALL, 'de_DE')
+            locale.setlocale(locale.LC_ALL, 'fr_FR')
         else:
-            locale.setlocale(locale.LC_ALL, 'de_DE.utf8')
+            locale.setlocale(locale.LC_ALL, 'fr_FR.utf8')
 
         for fast_reader in (True,
                             False,
--- a/astropy/utils/iers/tests/test_leap_second.py
+++ b/astropy/utils/iers/tests/test_leap_second.py
@@ -1,6 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import urllib.request
 import os
+import locale
+import platform
 
 import pytest
 import numpy as np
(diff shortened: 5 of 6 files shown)


pllim:
How interesting! There should be a change log. Not sure about a test, I'll let @mhvk decide.

saimn:
Hmm, this is not yet done... Putting `_set_locale` in `_read_leap_seconds` as I did initially causes a race condition somewhere :scream: . And moving it to `Time.strptime` works, but I don't think we want this here, though having an option to force the locale in `Time.strptime` could be an option ?

astropy/utils/iers/iers.py
```diff
@@ -1025,15 +1026,20 @@ def expires(self):
...
         with get_readable_fileobj(file) as fh:
             lines = fh.readlines()
             for line in lines:
                 match = cls._re_expires.match(line)
                 if match:
-                    expires = Time.strptime(match.groups()[0], '%d %B %Y',
-                                            scale='tai', format='iso',
-                                            out_subfmt='date')
+                    try:
+                        locale.setlocale(locale.LC_TIME, 'C')
```
mhvk, line 1037: Hmm, `locale` is quite explicit about not being thread safe. There is a suggested solution at https://stackoverflow.com/questions/18593661/how-do-i-strftime-a-date-object-in-a-different-locale, though I wonder if we are not better off just parsing it ourselves (as also suggested in various places). I think with a bit smarter regex this may just be possible........

mhvk (commented):
Wow, sorry to have caused what must have been a hard bug to track down! I wish one could just tell `strptime` to use a given locale, but for some reason that is not possible.

I think we need to be a bit careful about the solution (though perhaps the worry about threading problems is overblown in this case...)

.github/workflows/ci_workflows.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         python-version: ${{ matrix.python }}
     - name: Install language-pack-de and tzdata
       if: startsWith(matrix.os, 'ubuntu')
-      run: sudo apt-get install language-pack-de tzdata
+      run: sudo apt-get install language-pack-de language-pack-fr tzdata
```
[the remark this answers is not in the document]
saimn: There is another test using it (https://github.com/astropy/astropy/blob/a68f4564d6c21ef90d82b22bfc6362cf67dd2d70/astropy/wcs/tests/test_wcsprm.py#L865-L876), an old wcs bug, and it's easier for me to test locally that it works as expected ;)

astropy/utils/iers/iers.py
saimn: Yes indeed, the thing is that when using `_set_locale` here the tests are stuck in a deadlock...
Maybe another soution to parse the date would be better: 

> If, when coding a module for general use, you need a locale independent version of an operation that is affected by the locale (such as certain formats used with time.strftime()), you will have to find a way to do it without using the standard library routine.
https://docs.python.org/3/library/locale.html#background-details-hints-tips-and-caveats

.github/workflows/ci_workflows.yml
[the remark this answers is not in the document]
mhvk: Maybe we can get rid of german instead!

astropy/utils/iers/iers.py
mhvk: I think that just changing the `regex` to return the three elements and then doing a simply lookup on the first three characters of the month should do it:

https://github.com/astropy/astropy/blob/a68f4564d6c21ef90d82b22bfc6362cf67dd2d70/astropy/utils/iers/iers.py#L876

astropy/utils/iers/iers.py
```diff
@@ -1031,9 +1034,10 @@ def _read_leap_seconds(cls, file, **kwargs):
             for line in lines:
                 match = cls._re_expires.match(line)
                 if match:
-                    expires = Time.strptime(match.groups()[0], '%d %B %Y',
-                                            scale='tai', format='iso',
-                                            out_subfmt='date')
+                    day, month, year = match.groups()[0].split()
+                    month_nb = MONTH_ABBR.index(month[:3]) + 1
+                    expires = Time(f'{year}-{month_nb:02d}-{day}',
+                                   scale='tai', out_subfmt='date')
```
saimn, line 1040: @mhvk - Something like this ? :top:

saimn:
About locale tests, the one in `test_misc.py` was skipped despite the de locale was installed, should be ok now. And the one in `test_wcsprm.py` was xfailed on master:
```
 ../../.tox/py37-test/lib/python3.7/site-packages/astropy/utils/tests/test_misc.py . [ 94%]
.s..s.  
../../.tox/py37-test/lib/python3.7/site-packages/astropy/wcs/tests/test_wcsprm.py . [ 97%]
........................................................................ [ 98%]
...........................x.....................................        [ 98%]
```

saimn:
With the PR:
```
../../.tox/py37-test/lib/python3.7/site-packages/astropy/utils/tests/test_misc.py . [ 94%]
.s....                                                                   [ 94%]
../../.tox/py37-test/lib/python3.7/site-packages/astropy/io/ascii/tests/test_read.py . [ 22%]
........................................................................ [ 22%]
.......................................................................  [ 23%]
../../.tox/py37-test/lib/python3.7/site-packages/astropy/utils/iers/tests/test_leap_second.py . [ 92%]
.........ss..s...s.............                                          [ 93%]
```

astropy/utils/iers/iers.py
mhvk: Nice!

mhvk (commented):
This looks good!  Does need a changelog entry... And maybe squash the commits?

saimn:
@mhvk - Done!

mhvk (approved):
Great, merging...

PR #11064, opened 2020-11-20
Backport cirs tete precision

mhvk:
@bsiposc - it turned out that the backport of #11061 without #10994 was difficult, but with it straightforward. So, I combined them here and even added a what's new entry, with the logic that #11061 fixes a bug in that our `AltAz` were always off by quite a bit.

But I realize this is really not what release candidates are for, so please do feel free to tell me to let things be and keep #11061 for 4.3.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,16 @@
+4.2RC2 (unreleased)
+===================
+
+Bug Fixes
+---------
+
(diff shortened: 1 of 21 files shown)


mhvk:
Now that we understand why the simpler backport didn't work and fixed that, I'll close this really too large late-in-the-game PR.

PR #6861, opened 2017-11-16
Precise barycentric radial velocity corrections.

StuartLittlefair:
This PR builds on the improvements to ```EarthLocation.get_gcrs_posvel``` in #6699, to offer radial velocity corrections compatible with [Wright & Eastman (2014)](http://adsabs.harvard.edu/abs/2014PASP..126..838W) to 10 mm/s (for sources at infinite distance).

Prior to this PR, radial velocity corrections between astropy and WE14 differed by several metres/second. This is due to ignoring the gravitational redshift at the observatory and a slight bug in ```SkyCoord.radial_velocity_correction```. This PR fixes both of these issues to give much more accurate radial velocity corrections. The plot before shows the situation before (top) and after (below). The yellow star marks the position of the Sun.
barycentric_deltas

As discussed in [5752](https://github.com/astropy/astropy/pull/5752#issuecomment-309268570) improvements to the precision of these corrections should probably be thought of as bug fixes.

The only remaining physics to include is that caused by the proper motion and parallax of the source, but I propose only including those terms once velocity support is added to ```SkyCoord```.

This PR fixes #6697 and also fixes #6680, subject to the precision required by that user.

PR #11069, opened 2020-11-20
Speed up GCRS<->(TETE, CIRS) by smarter location conversions

### LINKED ISSUE #11067: More efficient algorithm for deriving observer's GCRS position and velocity in TETE frame.
### Description

The current method for deriving the observer's GCRS position and velocity in the TETE<->GCRS intermediate rotation transforms is inefficient as it involves a call to `erfa.gst06a()` which in turn calls `erfa.pnm06a()` in order to calculate the equation of the origins. This means that the TETE<->GCRS transforms call the very expensive `erfa.pnm06a()` twice. Fortunately, this can be easily avoided in one of two ways:

- Instead of doing the ITRS->TETE transform directly using `transform_to()`, the required greenwich apparent sidereal time can be inexpensively derived from the BPN already at hand by calling `erfa.gst06()`.

- Simply go through CIRS instead of TETE to derive the GCRS position and velocity. The required celestial to intermediate matrix can be inexpensively derived from the BPN at hand in the same manner as is done in `erfa.apco()` after extracting the CIP coordinates and CIO locator from the BPN.  

### Additional context

Celestial to intermediate matrix from classical BPN. 
```
   # Extract CIP coordinates.
   x, y = erfa.bpn2xy(rnpb)

   # The CIO locator, s. 
   s = erfa.s06(tta, ttb, x, y)

   # CIO based BPN matrix. 
   rc2i = erfa.c2ixys(x, y, s)
```

mhvk:
Following the suggestion by @mkbrewer in #11067, this speeds up TETE<->GCRS considerably by using the `rbpn` matrix also for getting the sidereal time for the conversion of the observer location from ITRS to TETE.

fixes #11067

~Note: need to re-unify code so for now just a draft so we have something to look at.~ (done)

EDIT: quick timing:
```
from astropy.coordinates import SkyCoord, EarthLocation

tete = SkyCoord(10*u.deg, 11*u.deg, frame='tete', location=EarthLocation(11*u.deg, 12*u.deg))
%timeit tete.transform_to('gcrs')
# current master
# 16.6 ms ± 1.13 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# this PR
# 6.51 ms ± 609 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
+++ b/astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
@@ -11,6 +11,7 @@ import erfa
 
 import astropy.units as u
 from astropy.coordinates.baseframe import frame_transform_graph
+from astropy.coordinates.earth import OMEGA_EARTH
 from astropy.coordinates.representation import CartesianRepresentation
 from astropy.coordinates.transformations import FunctionTransformWithFiniteDifference
 from astropy.coordinates.matrix_utilities import matrix_transpose
@@ -32,7 +33,7 @@ def teme_to_itrs_mat(time):
     gst = erfa.gmst82(*get_jd12(time, 'ut1'))
 
     # Polar Motion
-    # Do not include TIO locator s' because it is not used in Vallado 2006
+    # Do not include TIO locator s' because it is not used in Vallado 2006
     xp, yp = get_polar_motion(time)
     pmmat = erfa.pom00(xp, yp, 0)
 
@@ -95,24 +96,29 @@ def gcrs_precession_mat(equinox):
     return erfa.fw2m(gamb, phib, psib, epsa)
 
 
-OMEGA_EARTH = u.Quantity(7.292115855306589e-5, 1./u.s)
-"""
-Rotational velocity of Earth. In UT1 seconds, this would be 2 pi / (24 * 3600),
-but we need the value in SI seconds.
-See Explanatory Supplement to the Astronomical Almanac, ed. P. Kenneth Seidelmann (1992),
-University Science Books.
-"""
+def get_location_gcrs(location, obstime, ref_to_itrs, gcrs_to_ref):
+    """Create a GCRS frame at the location and obstime.
 
+    The reference frame z axis must point to the Celestial Intermediate Pole
+    (as is the case for CIRS and TETE).
 
-def get_location_gcrs(reference, gcrs_to_itrs_matrix):
-    obstime = reference.obstime
-    itrs_cart = reference.location.get_itrs(obstime).cartesian
-    obsgeoloc = itrs_cart.transform(matrix_transpose(gcrs_to_itrs_matrix))
-    vel_x = -OMEGA_EARTH * obsgeoloc.y
-    vel_y = OMEGA_EARTH * obsgeoloc.x
+    This function is here to avoid location.get_gcrs(obstime), which would
+    trigger infinite recursion in some cases, and uses pre-calculated
+    reference to ITRS and GCRS to Reference matrices to calculate obsgeoloc
+    and obsgeovel required for GCRS (as the latter is available where it
+    is used below).
+
+    """
+    itrs_cart = location.get_itrs(obstime).cartesian
+    ref_cart = itrs_cart.transform(matrix_transpose(ref_to_itrs))
+    vel_x = -OMEGA_EARTH * ref_cart.y
+    vel_y = OMEGA_EARTH * ref_cart.x
     vel_z = 0. * vel_x.unit
-    obsgeovel = CartesianRepresentation(vel_x, vel_y, vel_z)
-    return GCRS(obstime=obstime, obsgeoloc=obsgeoloc, obsgeovel=obsgeovel)
+    ref_vel = CartesianRepresentation(vel_x, vel_y, vel_z)
+    ref_to_gcrs = matrix_transpose(gcrs_to_ref)
+    return GCRS(obstime=obstime,
+                obsgeoloc=ref_cart.transform(ref_to_gcrs),
+                obsgeovel=ref_vel.transform(ref_to_gcrs))
 
 
 # now the actual transforms
@@ -123,8 +129,9 @@ def gcrs_to_tete(gcrs_coo, tete_frame):
     # (same as in builtin_frames.utils.get_cip).
     rbpn = erfa.pnm06a(*get_jd12(tete_frame.obstime, 'tt'))
     # Get GCRS coordinates for the target observer location and time.
-    gcrs_to_itrs_mat = tete_to_itrs_mat(tete_frame.obstime, rbpn=rbpn) @ rbpn
-    loc_gcrs = get_location_gcrs(tete_frame, gcrs_to_itrs_mat)
+    loc_gcrs = get_location_gcrs(tete_frame.location, tete_frame.obstime,
+                                 tete_to_itrs_mat(tete_frame.obstime, rbpn=rbpn),
+                                 rbpn)
     gcrs_coo2 = gcrs_coo.transform_to(loc_gcrs)
     # Now we are relative to the correct observer, do the transform to TETE.
     # These rotations are defined at the geocenter, but can be applied to
@@ -141,8 +148,9 @@ def tete_to_gcrs(tete_coo, gcrs_frame):
     newrepr = tete_coo.cartesian.transform(matrix_transpose(rbpn))
     # We now have a GCRS vector for the input location and obstime.
     # Turn it into a GCRS frame instance.
-    gcrs_to_itrs_mat = tete_to_itrs_mat(tete_coo.obstime, rbpn=rbpn) @ rbpn
-    loc_gcrs = get_location_gcrs(tete_coo, gcrs_to_itrs_mat)
+    loc_gcrs = get_location_gcrs(tete_coo.location, tete_coo.obstime,
+                                 tete_to_itrs_mat(tete_coo.obstime, rbpn=rbpn),
+                                 rbpn)
     gcrs = loc_gcrs.realize_frame(newrepr)
     # Finally, do any needed offsets (no-op if same obstime and location)
     return gcrs.transform_to(gcrs_frame)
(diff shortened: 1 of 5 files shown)


mhvk:
Sorry, this is definitely not right: obsgeoloc and obsgeovel become all 0; I'm doing something stupid.

mhvk:
More generally, a sequence of matrix multiplications cannot make the velocity non-zero, so `obsgeovel` is wrong even if I get `obsgeoloc` to be right.

mkbrewer:
That check against Horizons is off by 288 mas. This suggests that there is something wrong at the polar motion level.

mkbrewer:
Oh, never mind. I see you have already discovered the error.

mhvk:
OK, I get tests to pass again by reverting to what used to be done in `get_gcrs_posvel()` before we had velocities in our coordinates -- just calculating the velocity from the earth's rotation rate. So, this avoids the finite difference calls to get the velocity (which is why the speed-up is as large as it is: it not just halves the number of calls to `pnm06`, but quarters it!)

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -85,24 +95,24 @@ def gcrs_precession_mat(equinox):
     return erfa.fw2m(gamb, phib, psib, epsa)
 
 
-def get_location_gcrs(reference, gcrs_to_reference_matrix):
-    """Create a GCRS frame at the location and obstime of a given reference.
+OMEGA_EARTH = u.Quantity(7.292115855306589e-5, 1./u.s)
```
mhvk, line 98: Actually, still defined in `earth.py`, although it is not used there. I wonder if we should revert to the old code there too...

mhvk:
Ah, this is neat: with this change, the GCRS->CIRS->ICRS and GCRS->TETE->ICRS paths again give consistent results within 1 mm! I added a commit with the tightened test.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
```diff
@@ -85,24 +95,24 @@ def gcrs_precession_mat(equinox):
...
-    gcrs_cart = (reference.location.get_itrs(obstime, include_velocity=True)
-                 .transform_to(reference.__class__(obstime=obstime))
-                 .cartesian
-                 .transform(matrix_transpose(gcrs_to_reference_matrix)))
-    return GCRS(obstime=obstime,
-                obsgeoloc=gcrs_cart.without_differentials(),
-                obsgeovel=gcrs_cart.differentials['s'].to_cartesian())
+    itrs_cart = reference.location.get_itrs(obstime).cartesian
+    obsgeoloc = itrs_cart.transform(matrix_transpose(gcrs_to_itrs_matrix))
+    vel_x = -OMEGA_EARTH * obsgeoloc.y
```
mhvk, line 111: @mkbrewer - is there an ERFA routine that does this (maybe even in a better way)?

mhvk: Hmm, looking through why we changed it in `earth.py`, I came back to #6699, which planned to do:
```
        # CIRS velocity
        rotation_vector = CartesianRepresentation(0/u.s, 0/u.s, OMEGA_EARTH)
        cirs_cart = itrs.transform_to(cirs_frame).cartesian
        cirs_vel = rotation_vector.cross(cirs_cart)
        # Transform to GCRS (rotation)
        obsgeovel = cirs_vel.transform(matrix_transpose(gcrs_to_cirs_mat(obstime)))
        return obsgeoloc, obsgeovel
```
But then in #6861 it was changed to its present state. @StuartLittlefair - both were your PRs, so you probably remember better what is right/wrong with it.

mkbrewer:
Since the CIRS  and TETE transforms are now basically the same, I am curious whether or not `test_tete_transforms()` will work in this PR with the tighter tolerance of 1 mm. Would you mind trying that here?

mhvk:
@mkbrewer - are you thinking of a particular test other than the one I just changed? (I.e., did we think about the same possibility again at the same time!?)

mkbrewer:
The ERFA routine is `pvtob()`, but it does the geodetic to geocentric conversion also which not necessary with an `EarthLocation`.

mkbrewer:
Here we go again, you are one comment ahead of me. Yes, we did. I didn't see your comments to that effect until after my previous post which is why I deleted it.

mkbrewer:
Hmm, I wonder why my comment was automatically deleted there. Anyway, yes, we both thought of that at basically the same time.

mhvk:
Ah, from that routine, it looks like I should be calculating the velocity in CIRS, as was done in #6699, so this needs an update.

mhvk:
It is because I deleted it, following your example. Sorry!!

mhvk:
If we need to go to CIRS to add the velocity, maybe it makes more sense after all to go for your second option, i.e., transform the ITRS location to GCRS via CIRS also for TETE positions.

mkbrewer:
I think you are fine there. You are just calculating the velocity after the final rotation. Otherwise, you'd need to rotate the velocity also. See `apco()` for an example of this. In fact, we could optimize `ErfaAstrom.apco()` a bit by using your transform there also.

mkbrewer:
btw, in fact, `pvtob()` works in either CIRS or TETE, see note 3) there.

mhvk:
OK, that's good, so the plan would be to get the location it in ITRS, transform it to either CIRS or TETE using one matrix, then add the velocity as I do now, and finally transform to GCRS with the second matrix (which applies directly to the velocity as well). That should be a straightforward change!

mkbrewer:
I don't see the point of making that change, unless you actually want the velocity in CIRS or TETE, you might as well calculate it in GCRS. That saves one extra rotation.

mhvk:
But I now assume the velocity is perpendicular to the GCRS Z axis, while I think it should be perpendicular to the real Earth's polar axis (and the matrix looked like it had rotations around x and y, but perhaps I misinterpreted c2i06a)

mkbrewer:
Well, it is perpendicular to the CIP. The transformation to GCRS just aligns the CIP with the ICRF frame. That's why `apco()` rotates both the position and velocity into the GCRS. Remember, in that transformation we are going to of date to J2000.

mhvk:
It's not quite perpendicular, though, and for some applications that small difference may matter (which was the reason for #6699 and the subsequent #6861)
```
from astropy.time import Time
from astropy.coordinates import CIRS, GCRS
obstime = Time('J2010')
cirs = CIRS(x=10000*u.km, y=0.*u.km, z=0.*u.km, representation_type='cartesian', obstime=obstime)
cirs.transform_to(GCRS(obstime=obstime)).cartesian
# <CartesianRepresentation (x, y, z) in km
#    (9999.99496686, -0.00014031, -10.03308023)>
```

mkbrewer:
That's actually a quite interesting demonstration of the stability of the CIO. In ten years it only moved by 14 cm while the equator moved by 10 km.

mkbrewer:
Since it seems that I'm not going be able to convince you, here's an experiment that you could do. First calculate the velocity in the CIRS frame and then rotate the position and velocity into the GCRS frame. Then recalculate the velocity from the GCRS position and compare the two.

mhvk:
My lack of being convinced is just because of @StuartLittlefair's old #6699...  But strengthened by the fact that with the current PR, `loc.get_gcrs_posvel()` and my quick routine give inconsistent velocities. If I apply the velocity in the CIRS or TETE frame, then they become consistent. I've just pushed a new commit that includes tests for this - do have a look, since if this is wrong, it affects people trying to get accurate radial velocities to hunt for planets...

mhvk:
If we go this route, we should make sure `OMEGA_EARTH` is exactly consistent with the value used in ERFA's `pvtob` (unfortunately, not exported in their .h file). It might also be a good idea to rewrite `EarthLocation.get_gcrs_posvel()` to use the same method - I had not really realized how by just using the standard transformations, the velocity requires multiple evaluations. Though perhaps it is better to try to rewrite the transformation mechanism itself so that it uses knowledge of how velocities change.

mkbrewer:
OK, you're right. The frame rotates, but the vectors don't. Therefore the spin picks up a z component in the new frame. I was erroneously thinking that the vectors stayed aligned with the CIP. Sorry about that. 

Omega is defined in pvtob.c:
```
/* Earth rotation rate in radians per UT1 second */
   const double OM = 1.00273781191135448 * D2PI / DAYSEC;
```
This is just a sidereal second converted to radians.

StuartLittlefair (approved):

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
StuartLittlefair: This was removed when velocities were added to coordinates purely because it was simpler.

astropy/coordinates/builtin_frames/intermediate_rotation_transforms.py
StuartLittlefair: Yes - let’s leave earth.py alone for now.

StuartLittlefair:
A lot happens when I’m asleep! It looks like you’ve sorted out all the issues; as you say velocity should be added in CIRS or TETE and then rotated from there.

I’m very pleased this also fixes the slight inaccuracies in the ICRS round trip as well as providing a massive speed up.

mkbrewer:
@mhkv You might add that this fixes issue  #11067

mkbrewer:
There is a slight discrepancy between SOFA and Astropy regarding `OMEGA`. Using SOFA's definition of `D2PI`, I'm seeing:
```
OM = 7.292115146706980e-05
```

mhvk:
@mkbrewer - thanks, I think we might as well be consistent with what SOFA has.

I'll make another commit that updates the value and adds a changelog entry about the speed and accuracy improvements.

mhvk:
OK, I changed `OMEGA_EARTH` to match ERFA (and found where they got the value from...), and also added changelog and what's-new entries. @mkbrewer - for the latter, I mention you specifically, as I wanted to ensure your really helpful advice is recognized! (Normally, people are recognized by commits, but that process is not the only one...)

mhvk:
I put some follow-up items in #11072

PR #11074, opened 2020-11-22
EXP: run CI with MacOS X Big Sur

astrofrog:
Currently ``macos-latest`` uses MacOS X 10.15, but this PR is to check that everything works fine with MacOS X 11 (Big Sur). We don't actually have to merge this since I assume ``macos-latest`` will eventually point to MacOS X 11.


astrofrog:
Seems to work! Will close.

PR #11076, opened 2020-11-23
MNT: Move minversion specs out of package __init__.py

### LINKED ISSUE #11059: MNT: Get rid of minversion definitions in __init__.py
### Description

Do we really need all this stuff in `__init__.py`? It duplicates the info in `setup.cfg` anyway. For the usage in `docs/conf.py`, there is no reason we cannot get that info from `setup.cfg` instead?

https://github.com/astropy/astropy/blob/a995665c06b8d145ba811e3cbc59b0dfb8d20bd6/astropy/__init__.py#L15-L30

### Additional context

If the package being checked has no `__version__`, importing `astropy` will fail. For example: liberfa/pyerfa#60

cc @saimn and @astrofrog @embray

pllim:
### Description

This pull request is to move minversion definitions out of `astropy/__init__.py`.

Fixes #11059

### DIFF
--- a/astropy/__init__.py
+++ b/astropy/__init__.py
@@ -12,32 +12,6 @@ from warnings import warn
 
 from .version import version as __version__
 
-__minimum_python_version__ = '3.7'
-__minimum_numpy_version__ = '1.17.0'
-__minimum_erfa_version__ = '1.7'
-__minimum_scipy_version__ = '1.1'
-# ASDF is an optional dependency, but this is the minimum version that is
-# compatible with Astropy when it is installed.
-__minimum_asdf_version__ = '2.6.0'
-# PyYAML is an optional dependency, but this is the minimum version that is
-# advertised to be supported.
-__minimum_yaml_version__ = '3.13'
-# Matplotlib is an optional dependency, but this is the minimum version that is
-# advertised to be supported.
-__minimum_matplotlib_version__ = '3.0'
-# IPython is an optional dependency, but this is the minimum version that is
-# advertised to be supported.
-__minimum_ipython_version__ = '4.2'
-
-
-class UnsupportedPythonError(Exception):
-    pass
-
-
-# This is the same check as the one at the top of setup.py
-if sys.version_info < tuple(int(val) for val in __minimum_python_version__.split('.')):
-    raise UnsupportedPythonError(f"Astropy does not support Python < {__minimum_python_version__}")
-
 
 def _is_astropy_source(path=None):
     """
@@ -65,37 +39,7 @@ else:
     online_docs_root = f'https://docs.astropy.org/en/{__version__}/'
 
 
-def _check_requirement(name, minimum_version):
-    """
-    Check that ``name`` is installed and it is of the ``minimum_version`` we
-    require.
-    """
-    # Note: We could have used distutils.version for this comparison,
-    # but it seems like overkill to import distutils at runtime, and
-    # our own utils.introspection.minversion indirectly needs requirements.
-    requirement_met = False
-    import_fail = ''
-    try:
-        module = __import__(name)
-    except ImportError:
-        import_fail = f'{name} is not installed.'
-    else:
-        version = getattr(module, '__version__')
-        requirement_met = version.split('.') >= minimum_version.split('.')
-
-    if not requirement_met:
-        msg = (f"{name} version {minimum_version} or later must "
-               f"be installed to use Astropy. {import_fail}")
-        raise ImportError(msg)
-
-    return module
-
-
-_check_requirement('numpy', __minimum_numpy_version__)
-_check_requirement('erfa', __minimum_erfa_version__)
-
-
-from . import config as _config
+from . import config as _config  # noqa: E402
 
 
 class Conf(_config.ConfigNamespace):
@@ -132,7 +76,7 @@ conf = Conf()
 
 
 # Define a base ScienceState for configuring constants and units
-from .utils.state import ScienceState
+from .utils.state import ScienceState  # noqa: E402
 
 
 class base_constants_version(ScienceState):
@@ -188,7 +132,7 @@ class astronomical_constants(base_constants_version):
 
 
 # Create the test() function
-from .tests.runner import TestRunner
+from .tests.runner import TestRunner  # noqa: E402
 test = TestRunner.make_test_runner_in(__path__[0])
 
 
@@ -199,7 +143,7 @@ def _initialize_astropy():
     from . import config
 
     try:
-        from .utils import _compiler
+        from .utils import _compiler  # noqa: F401
     except ImportError:
         if _is_astropy_source():
             raise ImportError('You appear to be trying to import astropy from '
@@ -230,26 +174,26 @@ def _get_bibtex():
 
     with open(citation_file, 'r') as citation:
         refs = citation.read().split('@ARTICLE')[1:]
-        if len(refs) == 0: return ''
+        if len(refs) == 0:
+            return ''
         bibtexreference = f'@ARTICLE{refs[0]}'
     return bibtexreference
 
 
 __citation__ = __bibtex__ = _get_bibtex()
 
-import logging
+import logging  # noqa: E402
 
 # Use the root logger as a dummy log before initilizing Astropy's logger
 log = logging.getLogger()
 
-
-from .logger import _init_log, _teardown_log
+from .logger import _init_log, _teardown_log  # noqa: E402, F401
 
 log = _init_log()
 
 _initialize_astropy()
 
-from .utils.misc import find_api_page
+from .utils.misc import find_api_page  # noqa: E402, F401
 
 
 def online_help(query):
@@ -276,13 +220,13 @@ def online_help(query):
     webbrowser.open(url)
 
 
-__dir_inc__ = ['__version__', '__githash__', '__minimum_numpy_version__',
+__dir_inc__ = ['__version__', '__githash__',
                '__bibtex__', 'test', 'log', 'find_api_page', 'online_help',
                'online_docs_root', 'conf', 'physical_constants',
                'astronomical_constants']
 
 
-from types import ModuleType as __module_type__
+from types import ModuleType as __module_type__  # noqa: E402
 # Clean up top-level namespace--delete everything that isn't in __dir_inc__
 # or is a magic attribute, and that isn't a submodule of this package
 for varname in dir():
--- a/astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
+++ b/astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
@@ -3,9 +3,7 @@
 
 import pytest
 
-from astropy.io.misc.asdf.tags.helpers import __minimum_asdf_version__
-
-asdf = pytest.importorskip('asdf', minversion=__minimum_asdf_version__)
+asdf = pytest.importorskip('asdf')
 
 import astropy.units as u
 
--- a/astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
+++ b/astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
@@ -3,7 +3,7 @@
 
 import pytest
 
-from astropy import __minimum_asdf_version__
+from astropy.io.misc.asdf.tags.helpers import __minimum_asdf_version__
 
 asdf = pytest.importorskip('asdf', minversion=__minimum_asdf_version__)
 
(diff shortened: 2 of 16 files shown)


astropy/__init__.py
```diff
@@ -276,13 +220,13 @@ def online_help(query):
     webbrowser.open(url)
 
 
-__dir_inc__ = ['__version__', '__githash__', '__minimum_numpy_version__',
```
pllim, line 279: This is an API breakage but I am not sure why this was provided to begin with. Any advice?

docs/conf.py
```diff
@@ -59,45 +58,68 @@
 
 # To perform a Sphinx version check that needs to be more specific than
 # major.minor, call `check_sphinx_version("X.Y.Z")` here.
-check_sphinx_version("1.2.1")
+check_sphinx_version("1.2.1")  # noqa: F405
```
pllim, line 61: Sorry for the extra PEP 8 stuff here. I needed to clean warnings off my Emacs flake8 linter so I can be sure I didn't accidentally introduce undefined vars, etc.

docs/conf.py
```diff
@@ -59,45 +58,68 @@
...
+exclude_patterns.append('_templates')  # noqa: F405
+exclude_patterns.append('_pkgtemplate.rst')  # noqa: F405
+exclude_patterns.append('**/*.inc.rst')  # .inc.rst mean *include* files, don't have sphinx process them  # noqa: F405, E501
 
 # Add any paths that contain templates here, relative to this directory.
 if 'templates_path' not in locals():  # in case parent conf.py defines it
     templates_path = []
 templates_path.append('_templates')
 
+# Grab minversion from setup.cfg
```
pllim, line 89: If there is a way to make this prettier, please let me know.

docs/conf.py
```diff
@@ -59,45 +58,68 @@
...
 rst_epilog += f"""
-.. |minimum_python_version| replace:: {astropy.__minimum_python_version__}
-.. |minimum_numpy_version| replace:: {astropy.__minimum_numpy_version__}
-.. |minimum_erfa_version| replace:: {astropy.__minimum_erfa_version__}
-.. |minimum_scipy_version| replace:: {astropy.__minimum_scipy_version__}
-.. |minimum_yaml_version| replace:: {astropy.__minimum_yaml_version__}
-.. |minimum_asdf_version| replace:: {astropy.__minimum_asdf_version__}
-.. |minimum_matplotlib_version| replace:: {astropy.__minimum_matplotlib_version__}
-.. |minimum_ipython_version| replace:: {astropy.__minimum_ipython_version__}
+.. |minimum_python_version| replace:: {__minimum_python_version__}
```
mhvk, line 115: Super-fancy might be something like
```
rst_epilog += "\n".join(
    f".. |minimum_{name}_version| replace:: {minvers(name)}"
    for name in ('python', 'numpy', ...))
```
where you define a quick `minvers` function that does the look-up, like
```
def minvers(name):
    if name == 'yaml':
        name = 'PyYaml'
    for cur_req in astropy_req:
       if cur_req.name == name:
           return cur_req.specs[0][1]
    raise ValueError(f"{name} not among requirements!")
```

astropy/__init__.py
```diff
@@ -65,37 +39,7 @@ def _is_astropy_source(path=None):
...
-
-    if not requirement_met:
-        msg = (f"{name} version {minimum_version} or later must "
-               f"be installed to use Astropy. {import_fail}")
-        raise ImportError(msg)
-
-    return module
-
-
-_check_requirement('numpy', __minimum_numpy_version__)
```
mhvk, line 94: Skipping these means we no longer protest if an old version of numpy or erfa is installed. I think that is probably fine...

astropy/io/misc/asdf/tags/helpers.py
```diff
@@ -1,11 +1,16 @@
...
 # -*- coding: utf-8 -*-
 
-import numpy as np
+from pkg_resources import get_distribution
 
+import numpy as np
 
 __all__ = []
 
+for cur_req in get_distribution('astropy').requires(extras=('all', )):
```
mhvk, line 10: If we make this a little helper function in `utils/introspection`, we could do a python-version dependent choice of `pkg_resources` or `importlib.metadata` and thus future-proof ourselves.

It can also raise an error if the name is not found (and perhaps we can use it in `conf.py` as well?)

astropy/__init__.py
astrofrog: pip or conda will complain these days though, so think that's fine

astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 
-from astropy import __minimum_asdf_version__
+from astropy.io.misc.asdf.tags.helpers import __minimum_asdf_version__
 
 asdf = pytest.importorskip('asdf', minversion=__minimum_asdf_version__)
```
saimn, line 8: Why do we need `minversion=__minimum_asdf_version__` ? Since this is the minimum version installed by pip the minversion requirement should always be met.
I also wonder if we could skip all the asdf tests with a ` astropy/io/misc/asdf/conftest.py` file, if asdf is not installed.

pllim: I didn't write this. Maybe @eslavich or @jdavies-st can advise?

eslavich: > Why do we need `minversion=__minimum_asdf_version__` ?

The `asdf` package is in the extras, so it is possible to install astropy into an environment with an incompatible version of asdf and not see a warning.  But that's a weird enough edge case that I don't think we need to handle it in the code.

We don't consistently set `minversion`, for example:

https://github.com/astropy/astropy/blob/v4.3.dev/astropy/io/misc/asdf/tags/coordinates/tests/test_skycoord.py#L11

> I also wonder if we could skip all the asdf tests with a `astropy/io/misc/asdf/conftest.py` file, if asdf is not installed.

This seems way better than adding `pytest.importorskip` to every test file.

saimn: Yes ok. Users should have dependencies that met our requirements. For tests we ask people to `pip install -e .[tests]` so having an older version of whatever dependency should not happen. And I don't think we need to add dozens lines of code to make sure it doesn't happen. It's the same for other dependencies, we cannot check everywhere that some version is not recent enough.

saimn: And btw, that's why tox is so important. It will make sure that the correct versions are installed before running tests :).

docs/conf.py
saimn: Can we add explicit import for `intersphinx_mapping` instead ? Or just `intersphinx_mapping = None` before the import. I agree this is annoying, but we need the `from *` import to get all the other variables.

pllim: Are you saying I should back out all the `# noqa`?

docs/conf.py
pllim: Fancy!

astropy/io/misc/asdf/tags/helpers.py
pllim: Re: this and https://github.com/astropy/astropy/issues/11059#issuecomment-732393971 -- I am not sold on a function in `utils` to enable switching between `pkg_resources` and `importlib.metadata`, at least not for this PR.

Notable differences between the two:

* Using `pkg_resources`, I have to extract Python minversion from `setup.cfg`, which isn't available in the installed version. For `importlib.metadata`, it is given.
* Using `pkg_resources`, the minversion is parsed for you in `.specs` attribute. In `importlib.metadata`, it is given in a string (e.g., `"PyYAML (>=3.13) ; extra == 'docs'"`) that you then have to parse yourselves using regex because the `(>=3.13)` part can be missing if no minversion is specified.

I also don't see much use of this function post-installation, so why maintain one more public API?

astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
pllim: I couldn't find any precedence to skip all tests for a subpackage in `conftest.py` here, so I went the less controversial route of keeping the skip in the module but removing minversion check.

astropy/visualization/mpl_style.py
```diff
@@ -4,13 +4,6 @@
 # plotting style. It is no longer documented/recommended as of Astropy v3.0
 # but is kept here for backward-compatibility.
 
-from astropy import __minimum_matplotlib_version__
-from astropy.utils import minversion
-
-# This returns False if matplotlib cannot be imported
-MATPLOTLIB = minversion('matplotlib', __minimum_matplotlib_version__)
```
pllim, line 11: By @saimn 's logic for `asdf`, I also don't think we need a minversion check here. @astrofrog , what do you think?

pllim:
Thanks for the reviews, y'all. Hopefully this is looking better now?

docs/conf.py
saimn: If the variables are imported explicitly or defined, then you would not need the noqa everywhere.

pllim: Yes... perhaps I'll do that in the next round of edit. 😸

astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
saimn: It seems that tests are already skipped if entry points are not defined, e.g.:
https://github.com/astropy/astropy/blob/master/astropy/io/misc/asdf/tests/__init__.py
I'm not sure if there is a difference between the entry points not installed and asdf itself not installed, but I would guess that tests are already skipped if asdf is not installed, without the `pytest.importorskip` calls.

This is repeated in various directories, so I guess a conftest file in the root asdf module with something inspired from https://stackoverflow.com/a/47700320/4728722 would allow to skip all the tests.

astropy/__init__.py
mhvk: Agreed

astropy/io/misc/asdf/tags/coordinates/tests/test_angle.py
saimn: Actually `pytest.importorskip('asdf')` also does the import, so it's fine to keep it, and it's needed for collecting for tests.

mhvk (approved):
This looks good to me, so I'll approve, though it may be good to wait for a day or so for any possible further inputs.

astropy/io/misc/asdf/tags/helpers.py
saimn: One option with `importlib.metadata` :
```
In [31]: import importlib.metadata
    ...: from packaging.requirements import Requirement
    ...: 
    ...: versions = {}
    ...: for line in importlib.metadata.requires('astropy'):
    ...:     req = Requirement(line.split(';')[0])
    ...:     versions[req.name] = str(req.specifier)
    ...: 

In [32]: versions
Out[32]: 
{'numpy': '>=1.17',
 'pyerfa': '',
 'scipy': '>=1.1',
 'dask': '',
 'h5py': '',
 'beautifulsoup4': '',
 'html5lib': '',
 'bleach': '',
 'PyYAML': '>=3.13',
 'pandas': '',
 'sortedcontainers': '',
 'pytz': '',
 'jplephem': '',
 'matplotlib': '>=3.1',
 'mpmath': '',
 'asdf': '>=2.6',
 'bottleneck': '',
 'ipython': '',
 'pytest': '',
 'sphinx': '',
 'sphinx-astropy': '>=1.3',
 'pytest-astropy': '>=0.8',
 'pytest-xdist': '',
 'objgraph': '',
 'coverage': '',
 'skyfield': '>=1.20',
 'sgp4': '>=2.3'}
```

pllim:
We might want to wait for #11091 first? If we agree to move to `importlib.metadata` altogether, I'll have to update this PR to match.

mhvk:
@saimn - what do you think, it we merge this first, does that make a lot of trouble in #11091? (which I like...) If not, if you want to push the merge button here, do go ahead!

saimn:
No indeed, updating #11091 should not be too difficult (not more than replacing `pkg_resources` here). And I'm also very happy with the simplifications in `__init__` with this PR, so let's move forward and merge it ! ;)

saimn (approved):

pllim:
Thanks, all! 😄

PR #11077, opened 2020-11-24
Do not access uninitialized info when adding a mixin to a table.

### LINKED ISSUE #11066: Accessing `Quantity.info` dramatically decreases performance of indexing
### Description

After accessing `q.info`, `q.__getitem__` gets slower by a factor of 3 for a quantity with 1000 elements (less for larger, even more for smaller).

### Expected behavior

Accessing some meta information should not affect performance.

### Steps to Reproduce

```python
In [22]: import numpy as np

In [23]: import astropy.units as u

In [24]: q = np.random.normal(size=1000) << u.s

In [25]: mask = q > (0 << u.s)

In [26]: %timeit q[mask]
3.42 µs ± 8.12 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [27]: q.info
Out[27]: 
dtype = float64
unit = s
class = Quantity
n_bad = 0
length = 1000

In [28]: %timeit q[mask]
10.3 µs ± 254 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

### System Details
```
Linux-5.8.18-1-MANJARO-x86_64-with-arch-Manjaro-Linux
Python 3.7.7 (default, Mar 26 2020, 15:48:22) 
[GCC 7.3.0]
Numpy 1.19.1
astropy 4.1
Scipy 1.5.2
Matplotlib 3.3.2
```

mhvk:
As noted in #11066, when one adds a quantity (or other mixin column) to a table, it has the strange side effect that subsequent operations on that quantity are substantially slower. This is because `info` has been accessed, which now gets copied in any further operations. 

When something is copied to become part of the table, the expectation should be that original remains unchanged, so this PR avoids accessing `.info` unless it already exists.

In the process, it was also found that in current master, not only does `q.__dict__['info']` get created, it also has its `indices` set to ~an empty list~ `None`. ~Indeed~ Also, in several places, `col.info.indices` is explicitly set to `[]`. This seems unnecessary, so the PR here also just ensures that whenever one accesses `col.info.indices` for the first time, it gets initialized to `[]`. (In principle, the same could be done for `Column`, but there it felt a bit less necessary - can add that if wished.)

fixes #11066

Notes:
- while I think side effects of setting something in a table are undesired and thus labelled this as a bug, it seemed not important enough to backport.
- the first commit is less essential given that 'info' no longer is accessed, but will somewhat speed up performance for mixin column inside a table, so I left it in.
- EDIT: I had to change a couple of very recent tests that checked that `.info.indices is None` for uninitialized info. Would seem fine that those become empty lists.

### DIFF
--- a/astropy/table/column.py
+++ b/astropy/table/column.py
@@ -74,26 +74,15 @@ def col_copy(col, copy_indices=True):
     if isinstance(col, BaseColumn):
         return col.copy()
 
-    # The new column should have None for the parent_table ref.  If the
-    # original parent_table weakref there at the point of copying then it
-    # generates an infinite recursion.  Instead temporarily remove the weakref
-    # on the original column and restore after the copy in an exception-safe
-    # manner.
-
-    parent_table = col.info.parent_table
-    indices = col.info.indices
-    col.info.parent_table = None
-    col.info.indices = []
-
-    try:
-        newcol = col.copy() if hasattr(col, 'copy') else deepcopy(col)
-        newcol.info = col.info
-        newcol.info.indices = deepcopy(indices or []) if copy_indices else []
+    newcol = col.copy() if hasattr(col, 'copy') else deepcopy(col)
+    newcol.info = col.info
+    if copy_indices and col.info.indices:
+        newcol.info.indices = deepcopy(col.info.indices)
         for index in newcol.info.indices:
             index.replace_col(col, newcol)
-    finally:
-        col.info.parent_table = parent_table
-        col.info.indices = indices
+    else:
+        # TODO: bit strange that this is needed. Why not the default?
+        newcol.info.indices = []
 
     return newcol
 
--- a/astropy/table/column.py
+++ b/astropy/table/column.py
@@ -85,9 +85,6 @@ def col_copy(col, copy_indices=True):
             for index in newcol.info.indices:
                 index.replace_col(col, newcol)
 
-    # TODO: bit strange that this is needed. Why not the default?
-    newcol.info.indices = newcol.info.indices or []
-
     return newcol
 
 
--- a/astropy/table/column.py
+++ b/astropy/table/column.py
@@ -75,14 +75,18 @@ def col_copy(col, copy_indices=True):
         return col.copy()
 
     newcol = col.copy() if hasattr(col, 'copy') else deepcopy(col)
-    newcol.info = col.info
-    if copy_indices and col.info.indices:
-        newcol.info.indices = deepcopy(col.info.indices)
-        for index in newcol.info.indices:
-            index.replace_col(col, newcol)
-    else:
-        # TODO: bit strange that this is needed. Why not the default?
-        newcol.info.indices = []
+    # If the column has info defined, we copy it and adjust any indices
+    # to point to the copied column.  By guarding with the if statement,
+    # we avoid side effects (of creating the default info instance).
+    if 'info' in col.__dict__:
+        newcol.info = col.info
+        if copy_indices and col.info.indices:
+            newcol.info.indices = deepcopy(col.info.indices)
+            for index in newcol.info.indices:
+                index.replace_col(col, newcol)
+
+    # TODO: bit strange that this is needed. Why not the default?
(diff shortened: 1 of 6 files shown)


astropy/table/column.py
```diff
@@ -74,26 +74,16 @@ def col_copy(col, copy_indices=True):
     if isinstance(col, BaseColumn):
         return col.copy()
 
-    # The new column should have None for the parent_table ref.  If the
```
mhvk, line 77: I think this was a left-over from before `parent_table` was guarded in `data_info` itself.

astropy/utils/data_info.py
```diff
@@ -346,7 +346,14 @@ def __set__(self, instance, value):
 
         if isinstance(value, DataInfo):
             info = instance.__dict__['info'] = self.__class__(bound=True)
-            for attr in info.attr_names - info.attrs_from_parent - info._attrs_no_copy:
+            attr_names = info.attr_names
+            if value.__class__ is self.__class__:
+                # For same class, attributes are guaranteed to be stored in
+                # _attrs, so speed matters up by not accessing defaults.
+                # Doing this before difference in for loop helps speed.
+                attr_names = attr_names & set(value._attrs)  # NOT in-place!
```
mhvk, line 354: This is the part that I started with, and now helps mostly for columns inside tables. It reduces slicing a simple quantity by 30%:
```
In [2]: q = [1, 2, 3] * u.s

In [3]: t = QTable({'q': q})

In [4]: %timeit t['q'][:2]
# current master
6.14 µs ± 11.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
# this PR
4.25 µs ± 21.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

mhvk:
(coverage failure is a false positive)

taldcroft:
@mhvk - This kind of side effect is definitely not OK, so the concept is good. My first try with this revealed what appears to be a different initialization path that you haven't covered.
```
In [1]: astro                                                                                                                                                                       
astropy=4.3.dev115+g41d4cbcb8

In [2]: q = [1,2,3]*u.m                                                                                                                                                             

In [3]: qt = QTable([q])                                                                                                                                                            

In [4]: 'info' in q.__dict__                                                                                                                                                        
Out[4]: True

In [5]: q = [1,2,3]*u.m                                                                                                                                                             

In [6]: qt = QTable({'q': q})                                                                                                                                                       

In [7]: 'info' in q.__dict__                                                                                                                                                        
Out[7]: False
```

mhvk:
Darn!  Will check.... (funny to see how we both have ipython initializers - mine is `%astropy`)

mhvk:
Hmm, it is trying to find the name...

mhvk:
OK, worked around that. I think it is OK, though I still wonder a bit whether it might be possible to deal with this in `data_info` itself, of not really setting `info` unless there is something in `_attrs`. But it seems tricky and in a way the present PR is fairly self-contained, in that it is just very careful at table/column creation time and otherwise continues to just assume `.info` is present.

taldcroft (approved):
I looked through all the code and don't see anything that is obviously wrong. That said, this part of the code is fairly intricate, so I'm really relying on the test coverage here.

And I'll be honest and say that to this day I don't understand the indexing implementation very well and don't know if there is a difference between `[]` and `None`.

Anyway, this looks OK to me.

mhvk:
Great! I convinced myself that throughout the code `.indices` was always set to `[]` for table work, so I'm reasonably convinced what I did is OK. But I agree that the whole process is a bit unclear... (again noticed your TODO:
```
            if (col if isinstance(col, Column) else col.info).indices:
                # TODO : as far as I can tell the only purpose of setting _copy_indices
                # here is to communicate that to the initial test in `slice_indices`.
                # Why isn't that just sent as an arg to the function?
                col.info._copy_indices = self._copy_indices
                newcol = col.info.slice_indices(newcol, slice_, len(col))
```

Anyway, will merge this as it solves @maxnoe's problem.

PR #11079, opened 2020-11-24
Direct icrs altaz

### LINKED ISSUE #10887: Implement direct ICRS -> Observed transform

StuartLittlefair:
This PR fixes #10887 by implementing a 'direct' transformation between ICRS and AltAz. In fact, this transform routes via CIRS, but by doing it in one routine we avoid the overhead of computing the astrometry context twice, and the overhead in creating a SkyCoord object for each step.

This results in a modest speedup. Running @maxnoe's performance profiling from https://github.com/astropy/astropy/pull/10647#issuecomment-673569823, we find:

```
No Interpolation took 8.8879 s
Interpolation with  10.0 s   resolution took 0.9227 s
Interpolation with   5.0 min resolution took 0.2207 s
Interpolation with  30.0 min resolution took 0.1623 s
```

compared to the current master:

```
No Interpolation took 9.7546 s
Interpolation with  10.0 s   resolution took 1.0634 s
Interpolation with   5.0 min resolution took 0.2984 s
Interpolation with  30.0 min resolution took 0.2053 s
```

The only remaining puzzle is that I am unable to reproduce the high precision for GCRS->AltAz introduced in #11073. What is a mystery is that the transforms in that test are unaltered - they still go through the untouched GCRS->CIRS->AltAz chain. So I am submitting this PR to see if tests pass remotely and the local failure is something to do with my setup.

### DIFF
--- a/astropy/coordinates/builtin_frames/icrs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_observed_transforms.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Contains the transformation functions for getting to "observed" systems from ICRS.
+Currently that just means AltAz.
+"""
+
+import numpy as np
+import erfa
+
+from astropy import units as u
+from astropy.coordinates.builtin_frames.utils import atciqz, aticq
+from astropy.coordinates.baseframe import frame_transform_graph
+from astropy.coordinates.transformations import FunctionTransformWithFiniteDifference
+from astropy.coordinates.representation import (SphericalRepresentation,
+                                                CartesianRepresentation,
+                                                UnitSphericalRepresentation)
+
+from .icrs import ICRS
+from .altaz import AltAz
+from .utils import PIOVER2
+from ..erfa_astrom import erfa_astrom
+
+
+@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, ICRS, AltAz)
+def icrs_to_altaz(icrs_coo, altaz_frame):
+    # if the data are UnitSphericalRepresentation, we can skip the distance calculations
+    is_unitspherical = (isinstance(icrs_coo.data, UnitSphericalRepresentation) or
+                        icrs_coo.cartesian.x.unit == u.one)
+    # first set up the astrometry context for ICRS<->AltAz
+    astrom = erfa_astrom.get().apco(altaz_frame)
+
+    # correct for parallax to find BCRS direction from observer (as in erfa.pmpx)
+    if is_unitspherical:
+        srepr = icrs_coo.spherical
+    else:
+        observer_icrs = CartesianRepresentation(astrom['eb'] * u.au, xyz_axis=-1)
+        srepr = (icrs_coo.cartesian - observer_icrs).represent_as(
+            SphericalRepresentation)
+
+    # convert to topocentric CIRS
+    cirs_ra, cirs_dec = atciqz(srepr, astrom)
+
+    # now perform AltAz conversion
+    az, zen, ha, odec, ora = erfa.atioq(cirs_ra, cirs_dec, astrom)
+    alt = np.pi/2-zen
+    aa_srepr = SphericalRepresentation(az*u.radian, alt*u.radian, srepr.distance)
+    return altaz_frame.realize_frame(aa_srepr)
+
+
+@frame_transform_graph.transform(FunctionTransformWithFiniteDifference, AltAz, ICRS)
+def altaz_to_icrs(altaz_coo, icrs_frame):
+    # if the data are UnitSphericalRepresentation, we can skip the distance calculations
+    is_unitspherical = (isinstance(altaz_coo.data, UnitSphericalRepresentation) or
+                        altaz_coo.cartesian.x.unit == u.one)
+
+    usrepr = altaz_coo.represent_as(UnitSphericalRepresentation)
+    az = usrepr.lon.to_value(u.radian)
+    zen = PIOVER2 - usrepr.lat.to_value(u.radian)
+
+    # first set up the astrometry context for ICRS<->CIRS at the altaz_coo time
+    astrom = erfa_astrom.get().apco(altaz_coo)
+
+    # Topocentric CIRS
+    cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
+    if is_unitspherical:
+        srepr = SphericalRepresentation(cirs_ra*u.deg, cirs_dec*u.deg, 1)
+    else:
+        srepr = SphericalRepresentation(lon=cirs_ra, lat=cirs_dec,
+                                        distance=altaz_coo.distance)
+
+    # BCRS (Astrometric) direction to source
+    bcrs_ra, bcrs_dec = aticq(srepr, astrom)
+
+    # Correct for parallax to get ICRS representation
+    if is_unitspherical:
+        icrs_srepr = SphericalRepresentation(cirs_ra*u.deg, cirs_dec*u.deg, 1)
+    else:
+        icrs_srepr = SphericalRepresentation(lon=bcrs_ra*u.rad, lat=bcrs_dec*u.rad,
+                                             distance=altaz_coo.distance)
+        observer_icrs = CartesianRepresentation(astrom['eb'] * u.au, xyz_axis=-1)
+        newrepr = icrs_srepr.to_cartesian() + observer_icrs
+        icrs_srepr = newrepr.represent_as(SphericalRepresentation)
+
+    return icrs_frame.realize_frame(icrs_srepr)
--- a/astropy/coordinates/builtin_frames/icrs_observed_transforms.py
+++ b/astropy/coordinates/builtin_frames/icrs_observed_transforms.py
@@ -4,8 +4,6 @@
 Contains the transformation functions for getting to "observed" systems from ICRS.
 Currently that just means AltAz.
 """
-
-import numpy as np
 import erfa
 
 from astropy import units as u
@@ -34,7 +32,7 @@ def icrs_to_altaz(icrs_coo, altaz_frame):
     if is_unitspherical:
         srepr = icrs_coo.spherical
     else:
-        observer_icrs = CartesianRepresentation(astrom['eb'] * u.au, xyz_axis=-1)
+        observer_icrs = CartesianRepresentation(astrom['eb'], unit=u.au, xyz_axis=-1, copy=False)
         srepr = (icrs_coo.cartesian - observer_icrs).represent_as(
             SphericalRepresentation)
 
@@ -43,8 +41,11 @@ def icrs_to_altaz(icrs_coo, altaz_frame):
 
     # now perform AltAz conversion
     az, zen, ha, odec, ora = erfa.atioq(cirs_ra, cirs_dec, astrom)
-    alt = np.pi/2-zen
-    aa_srepr = SphericalRepresentation(az*u.radian, alt*u.radian, srepr.distance)
+    alt = PIOVER2-zen
+    if is_unitspherical:
+        aa_srepr = UnitSphericalRepresentation(az << u.radian, alt << u.radian, copy=False)
+    else:
+        aa_srepr = SphericalRepresentation(az << u.radian, alt << u.radian, srepr.distance, copy=False)
     return altaz_frame.realize_frame(aa_srepr)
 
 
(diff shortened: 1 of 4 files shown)


astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+    is_unitspherical = (isinstance(icrs_coo.data, UnitSphericalRepresentation) or
+                        icrs_coo.cartesian.x.unit == u.one)
+    # first set up the astrometry context for ICRS<->AltAz
+    astrom = erfa_astrom.get().apco(altaz_frame)
+
+    # correct for parallax to find BCRS direction from observer (as in erfa.pmpx)
+    if is_unitspherical:
+        srepr = icrs_coo.spherical
+    else:
+        observer_icrs = CartesianRepresentation(astrom['eb'] * u.au, xyz_axis=-1)
```
mhvk, line 37: Likely unimportant but this makes two copies. Avoid both with 
```
observer_icrs = CartesianRepresentation(astrom['eb'], unit=u.au, xyz_axis=-1, copy=False)
```

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+        srepr = (icrs_coo.cartesian - observer_icrs).represent_as(
+            SphericalRepresentation)
+
+    # convert to topocentric CIRS
+    cirs_ra, cirs_dec = atciqz(srepr, astrom)
+
+    # now perform AltAz conversion
+    az, zen, ha, odec, ora = erfa.atioq(cirs_ra, cirs_dec, astrom)
+    alt = np.pi/2-zen
```
mhvk, line 46: Might as well use PIOVER2 here too.

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+    aa_srepr = SphericalRepresentation(az*u.radian, alt*u.radian, srepr.distance)
```
mhvk, line 47: Replace `*` with `<<` and add `copy=False` for avoiding two more copies (all entries are newly created and will be discarded otherwise, so it is safe).

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+
+    usrepr = altaz_coo.represent_as(UnitSphericalRepresentation)
+    az = usrepr.lon.to_value(u.radian)
+    zen = PIOVER2 - usrepr.lat.to_value(u.radian)
+
+    # first set up the astrometry context for ICRS<->CIRS at the altaz_coo time
+    astrom = erfa_astrom.get().apco(altaz_coo)
+
+    # Topocentric CIRS
+    cirs_ra, cirs_dec = erfa.atoiq('A', az, zen, astrom)*u.radian
```
mhvk, line 65: `<< u.radian` to avoid the copy

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+    if is_unitspherical:
+        srepr = SphericalRepresentation(cirs_ra, cirs_dec, 1)
```
mhvk, line 67: `copy=False` here and in the one below.

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        srepr = SphericalRepresentation(lon=cirs_ra, lat=cirs_dec,
+                                        distance=altaz_coo.distance)
+
+    # BCRS (Astrometric) direction to source
+    bcrs_ra, bcrs_dec = aticq(srepr, astrom)*u.radian
```
mhvk, line 73: `<< u.radian`

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Correct for parallax to get ICRS representation
+    if is_unitspherical:
+        icrs_srepr = SphericalRepresentation(bcrs_ra, bcrs_dec, 1)
```
mhvk, line 77: I'd instantialy a `UnitSphericalRepresentation` here (with `copy=False`...)

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        icrs_srepr = SphericalRepresentation(lon=bcrs_ra, lat=bcrs_dec,
```
mhvk, line 79: More `copy=False` and `unit=u.radian, copy=False`

mhvk (commented):
This looks good modulo some comments to avoid copies (which are not going to matter much, but one might as well). It seems surprising that the speed-up is relatively minimal. Are we missing something? Though looking at some quick timings, I guess it is just that CIRS->AltAz is just much more expensive.

p.s. The failures all seem directly related to the PR, but the fact that you get local failures is a bit worrisome.

StuartLittlefair:
Quick work @mhvk! This is a little faster now as a result:

```
No Interpolation took 8.8463 s
Interpolation with  10.0 s   resolution took 0.8960 s
Interpolation with   5.0 min resolution took 0.1998 s
Interpolation with  30.0 min resolution took 0.1333 s
```

The local test failure is quite odd. I see it in the current astropy master as well. I don't know why, but I'm going to start by cleaning out all of my anaconda environments.

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
mhvk: Missed one?

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

astropy/coordinates/builtin_frames/icrs_observed_transforms.py
StuartLittlefair: 👍

mhvk:
Well, hardly worth the 0.5% improvement in timing, but if anybody has very large arrays of `SkyCoord`, they will be thankful...

mkbrewer:
Just a little suggestion. You could call this `icrs_to_observed()` add an argument that takes either `'A'` or `'H'` in order to return either Alt/Az or HA/Dec and then create an `icrs_to_altaz()` function that calls `icrs_to_observed()` with the additional argument set to `'A'`. In this way, this would be all set for the addition of an observed HA/Dec frame (see issue #10239). You could even add an `icrs_to_hadec()` function in anticipation of this, although that may be beyond the scope of this PR.

mkbrewer:
Of course, this applies to the reverse transform also.

StuartLittlefair:
> Just a little suggestion. You could call this `icrs_to_observed()` add an argument that takes either `'A'` or `'H'` in order to return either Alt/Az or HA/Dec and then create an `icrs_to_altaz()` function that calls `icrs_to_observed()` with the additional argument set to `'A'`. In this way, this would be all set for the addition of an observed HA/Dec frame (see issue #10239). You could even add an `icrs_to_hadec()` function in anticipation of this, although that may be beyond the scope of this PR.

This is a neat idea. I'll get to this tomorrow, but JWST deadline needs to take priority for now.

mhvk (approved):
Looks good now! I also like @mkbrewer's idea, but would suggest doing that in a follow-up PR, as this one works well on its own (and doesn't hinder adding `HADec` after). So, OK to merge as far as I'm concerned.

mhvk:
Actually, this needs a changelog entry (or `no-changelog-entry-needed` label if we consider this just an implementation detail)

mkbrewer:
Just trying to kill two birds with one stone here.

mhvk:
Which is a good idea, but even though the code is, as you note, easy, writing the test cases and documentation  will mean it will take a couple of hours, so might as well get this in...

PR #11080, opened 2020-11-25
ENH: switch z_at_value to use minimize_scalar solvers

dhomeier:
## Description
On reviewing some feedback on the `z_at_value` utility function I noticed this is using an old minimiser `fminbound`, which is basically only offered as a legacy option in `scipy.optimize` now.
The original request was to allow the function to directly handle arrays of values, but it turned out there is no obvious alternative in `scipy.optimize` that would allow simultaneous minimisation of vector-value functions.
Still, the recommended method now would be `minimize_scalar`. While this offers several solver methods, the immediate replacement `method='Bounded'` is probably nothing but a different frontend to `fminbound`. Nonetheless this will allow to use the newer interface, and additionally provides the solvers 'Brent' and 'Golden', which offer the `bracket` option as an alternative to control the solution for multi-valued functions (see example and test for application to angular size distance). 'Brent' also is slightly faster than 'Bounded'. A drawback is that the latter two methods require a workaround to respect hard boundaries set in `zmin` and `zmax`, but that is easily implemented.


pllim:
Thanks! If you decide to get it out of experimental stage, would need:

* Change log.
* User facing doc.
* What's new entry? Not sure about this one, will let subpackage maintainer decide.

dhomeier:
> * What's new entry? Not sure about this one, will let subpackage maintainer decide.

There is none for cosmology (which was part of the reason the issue came before the core maintainers list ;-).
Since there are new options, I guess an entry is warranted.

The docs primarily refer back to the function docstring; should I just update and extend the example there?

pllim:
> There is none for cosmology

Or whoever has enough domain expertise to review and approve. 😬 

> should I just update and extend the example there?

I'll leave that to your good judgement. 😄

MSeifert04:
I don't think I'd be very useful here, I'm sorry.

astropy/cosmology/funcs.py
```diff
@@ -113,33 +128,49 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
     other commonly inverted quantities) are monotonic in flat and open
     universes, but not in closed universes.
     """
-    from scipy.optimize import fminbound
+    from scipy.optimize import minimize_scalar
+
+    opt = {'maxiter': maxfun}
+    if method.lower() == 'bounded':
```
MSeifert04, line 134: Just something I noticed: If you allow method to be any callable won't this fail?

dhomeier: Yes. I have not looked into custom methods and how exactly to use them, but I agree if they are mentioned in the docstring, they should not be made to immediately fail at least.

dhomeier:
Looking at the 4.1 and 4.0 docs I'd say this is borderline relevant enough for a whatsnew entry. If it stays in:
The first call will normally issue an
```
~astropy/cosmology/funcs.py:173: UserWarning: fval is not bracketed by func(zmin)=4.430866897759455e-05 Mpc and func(zmax)=13.85203072142112 Mpc. This means either there is no solution, or that there is more than one solution between zmin and zmax satisfying fval = func(z).
  warnings.warn(f"fval is not bracketed by func(zmin)={fval_zmin} and func(zmax)="
```
should this be included in the example?

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+        val = fval
+
+    # 'Brent' and 'Golden' ignore `bounds`, force solution inside zlim
+    def f(z):
+        if z > zmax:
+            return 1.e300 * (1.0 + z - zmax)
+        elif z < zmin:
+            return 1.e300 * (1.0 + zmin - z)
+        else:
+            return abs(Quantity(func(z)).value - val)
```
nstarman, line 189: I think this needs to be fast and Quantity() is not. There probably should be 2 versions of this function, like before.

dhomeier: `func` even less – most (all?) of them are integrators over `z`, and the fastest ones I timed, like `Planck18.efunc`, are still 5-6 times slower than casting the result to Quantity. So this might be premature optimisation; otoh it's not completely insignificant, so it may be best to take care of this now before it's forgotten and might become a bottleneck, if the cosmology functions should be further optimised at some point in the future.

nstarman: I like how z out-of-bounds skips the func() evaluation entirely. That seems a very nice optimization.

astropy/cosmology/funcs.py
```diff
@@ -34,20 +35,49 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+
     maxfun : int, optional
        The maximum number of function evaluations allowed in the
        optimization routine (default 500).
 
+    method : str or callable, optional
+       Type of solver to pass to ``~scipy.optimize.minimize_scalar`` -
+       should be one of 'Brent' (default), 'Golden' or 'Bounded'.
+       Can in theory also be a callable object as custom solver,
+       but this is untested.
```
nstarman, line 61: Maybe a little more verbose, e.g. about case insensitivity and how to implement custom solvers.

```suggestion
       Type of solver to pass to ``~scipy.optimize.minimize_scalar``.
       The built-in options are 'Brent' (default), 'Golden', or 'Bounded',
       and are case insensitive. In principle, custom solvers can also be used
       by passing any callable object that follows the requirements explained
       in section "Custom minimizers" of :fun:`scipy.optimize.minimize_scalar`.
       The use here of custom solvers is currently untested.
```

astropy/cosmology/funcs.py
```diff
@@ -99,8 +129,8 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
     3.19812268
 
     The angular diameter is not monotonic however, and there are two
-    redshifts that give a value of 1500 Mpc. Use the zmin and zmax keywords
-    to find the one you're interested in:
+    redshifts that give a value of 1500 Mpc. You can Use the zmin and
```
nstarman, line 132:
```suggestion
    redshifts that give a value of 1500 Mpc. You can use the zmin and
```

astropy/cosmology/funcs.py
```diff
@@ -99,8 +129,8 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+    zmax keywords to find the one you're interested in:
```
nstarman, line 133:
```suggestion
    zmax keywords to find the one you are interested in:
```

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
     ...            1500 * u.Mpc, zmin=2.5)  # doctest: +FLOAT_CMP
     3.7914913242
 
+    Alternatively the ``bracket`` option can be used to initialise the
+    function solver on a desired region. For the example of angular
+    diameter distance, which has a maximum near a redshift of 1.6 in this
+    cosmology, defining a bracket on either side of this maximum will
+    generally return a solution on the same side.
+
+    >>> z_at_value(Planck13.angular_diameter_distance,
```
nstarman, line 148: I like the example.

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+        opt['xtol'] = ztol
 
     fval_zmin = func(zmin)
     fval_zmax = func(zmax)
     if np.sign(fval - fval_zmin) != np.sign(fval_zmax - fval):
-        warnings.warn("""\
-fval is not bracketed by func(zmin) and func(zmax). This means either
-there is no solution, or that there is more than one solution between
-zmin and zmax satisfying fval = func(z).""")
+        warnings.warn(f"fval is not bracketed by func(zmin)={fval_zmin} and func(zmax)="
```
nstarman, line 173: use an AstropyUserWarning?

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+                      f"{fval_zmax}. This means either there is no solution, or that there is "
+                      "more than one solution between zmin and zmax satisfying fval = func(z).")
 
     if isinstance(fval_zmin, Quantity):
         val = fval.to_value(fval_zmin.unit)
```
nstarman, line 178:
```suggestion
        val = fval.to_value(fval_zmin.unit)
        
        # 'Brent' and 'Golden' ignore `bounds`, force solution inside zlim
        def f(z):
            if z > zmax:
                return 1.e300 * (1.0 + z - zmax)
            elif z < zmin:
                return 1.e300 * (1.0 + zmin - z)
            else:
                return abs(func(z).value - val)
```

astropy/cosmology/funcs.py
nstarman, line 189:
```suggestion
      # 'Brent' and 'Golden' ignore `bounds`, force solution inside zlim
      def f(z):
          if z > zmax:
              return 1.e300 * (1.0 + z - zmax)
          elif z < zmin:
              return 1.e300 * (1.0 + zmin - z)
          else:
              return abs(func(z) - fval)
```

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+        else:
+            return abs(Quantity(func(z)).value - val)
 
-    zbest, resval, ierr, ncall = fminbound(f, zmin, zmax, maxfun=maxfun,
-                                           full_output=1, xtol=ztol)
+    res = minimize_scalar(f, method=method, bounds=(zmin, zmax), bracket=bracket, options=opt)
 
-    if ierr != 0:
-        warnings.warn(f'Maximum number of function calls ({ncall}) reached')
+    if not res.success:
```
nstarman, line 193: What type of warning does scipy emit in this case?

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+        warnings.warn(f"Solver returned {res.status}: {res.message}\n"
+                      f"Precision {res.fun} reached after {res.nfev} function calls.")
 
-    if np.allclose(zbest, zmax):
-        raise CosmologyError("Best guess z is very close the upper z limit.\n"
+    if verbose:
+        print(res)
```
nstarman, line 198: When res.success is False and verbose is True, there is a fair bit of repeat, given the above. But I think that's fine. It's like dumping the trace.

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+
+    if np.allclose(res.x, zmax):
+        raise CosmologyError(f"Best guess z={res.x} is very close to the upper z limit {zmax}.\n"
```
nstarman, line 201: 👍  on the added info.

astropy/cosmology/funcs.py
```diff
@@ -109,37 +139,68 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
                              "Try re-running with a different zmax.")
-    elif np.allclose(zbest, zmin):
-        raise CosmologyError("Best guess z is very close the lower z limit.\n"
+    elif np.allclose(res.x, zmin):
+        raise CosmologyError(f"Best guess z={res.x} is very close to the lower z limit {zmin}.\n"
```
nstarman, line 204: same as above.

astropy/cosmology/tests/test_cosmology.py
```diff
@@ -1582,6 +1582,46 @@ def test_z_at_value():
             z_at_value(cosmo.angular_diameter_distance, 1500*u.Mpc, zmin=4.)
 
 
+@pytest.mark.skipif('not HAS_SCIPY')
+@pytest.mark.parametrize('method', ['Brent', 'Golden', 'Bounded'])
```
nstarman, line 1586: Should an issue be raised for future test writing with a custom method?

astropy/cosmology/tests/test_cosmology.py
```diff
@@ -1582,6 +1582,46 @@ def test_z_at_value():
...
+        else:
+            z = 0.6812777
+            bracket = (1.6, 2.0)
+        with pytest.warns(UserWarning, match=r"Option 'bracket' is ignored"):
+            assert allclose(z_at_value(cosmo.angular_diameter_distance, 1500*u.Mpc, method=method,
+                                       bracket=bracket), z, rtol=1e-6)
+    else:
+        with pytest.warns(UserWarning, match=r'fval is not bracketed'):
+            assert allclose(z_at_value(cosmo.angular_diameter_distance, 1500*u.Mpc, method=method,
+                                       bracket=(0.9, 1.5)), 0.68127769625288614, rtol=1e-6)
```
nstarman, line 1611: with the rtol relatively low... don't need so many decimal places?

astropy/cosmology/tests/test_cosmology.py
```diff
@@ -1582,6 +1582,46 @@ def test_z_at_value():
             z_at_value(cosmo.angular_diameter_distance, 1500*u.Mpc, zmin=4.)
 
 
+@pytest.mark.skipif('not HAS_SCIPY')
+@pytest.mark.parametrize('method', ['Brent', 'Golden', 'Bounded'])
+def test_z_at_value_bracketed(method):
```
nstarman, line 1587: I think some high precision tests would also be valuable. All these are rtol=1e-6.

astropy/cosmology/funcs.py
nstarman: I might have counted spaces wrong...

astropy/cosmology/funcs.py
dhomeier: Yea, I just took that from the scipy docstring, which is just as sloppy about capitalisation.
I've been rather ambivalent about advertising the custom solvers as I did not really find good usage examples, and I guess it's rather unlikely that cosmology users want to bring along their own (at least for now).
But since it is an option in `minimize_scalar` one should probably not hide it. Definitely better with that bit of extra explanation, thanks!
Actually the best documentation I found is probably the tutorial entry https://docs.scipy.org/doc/scipy/reference/tutorial/optimize.html#custom-minimizers , but is this something that could be linked with Xref?

astropy/cosmology/tests/test_cosmology.py
dhomeier: Do you mean raise an issue here somehow or leave a `#TODO:`, or open an issue for a follow-up PR?
As noted above, I don't have a clear idea what custom solver might be used there, or how to test it.
Perhaps one could just take the example from the scipy tutorial, but that would probably mostly test `minimize_scalar`.

astropy/cosmology/tests/test_cosmology.py
dhomeier: Not really, just copied from the old tests.

astropy/cosmology/tests/test_cosmology.py
dhomeier: I think I had made some experiments where the precision did not increase very smoothly with lowering `tol` in `minimize_scalar`, also many of the forward tests don't have higher precision either; so in light of the "different values on different architectures" in the comment it's unclear whether that can be achieved at all. Could try a test run on the different CI architectures perhaps. But rtol=1.e-6 comes down to a few 100 m/s in redshift at most, so this is already well above any precision relevant for practical use cases.

astropy/cosmology/funcs.py
nstarman: That link would be good to include directly. It's funny that the least important feature requires the most explanation.
For something using intersphinx docs reference (https://docs.readthedocs.io/en/stable/guides/intersphinx.html#using-intersphinx ), maybe something along the lines of.

```In principle, custom solvers can also be used, though their use here is currently untested. See section "Custom minimizers" of :fun:`scipy.optimize.minimize_scalar` or of :doc:`scipy:reference/tutorial/optimize` for details.```

File: astropy/cosmology/tests/test_cosmology.py (hunk 13 above)
Comment (nstarman) [reply astropy/cosmology/tests/test_cosmology.py:1586]: I was thinking a TODO or issue. But given that this is mostly a scipy behavior... I'm also not sure what a good test would be.

File: astropy/cosmology/funcs.py (hunk 2 above)
Comment (dhomeier) [reply astropy/cosmology/funcs.py:189]: Your idea above to add an extra case for `abs(func(z).value - val)` seems actually the most elegant solution, if that's what it was intended to be.

File: astropy/cosmology/funcs.py (hunk 9 above)
Comment (dhomeier) [reply astropy/cosmology/funcs.py:193]: None; it's only indicated by the value of `res.success` (and `res.nit` having reached `maxiter`, if you've kept track of that).

File: astropy/cosmology/funcs.py (hunk 10 above)
Comment (dhomeier) [reply astropy/cosmology/funcs.py:198]: Yes, verbose is mostly meant to offer additional info even when successful.
I noted above that 'brent' and 'golden' don't actually include 'status' and 'message', so added a catch for that.

File: astropy/cosmology/tests/test_cosmology.py (hunk 13 above)
Comment (dhomeier) [reply astropy/cosmology/tests/test_cosmology.py:1586]: Perhaps I'll look in the SciPy tests if they have an example for this.

File: astropy/cosmology/tests/test_cosmology.py (hunk 15 above)
Comment (dhomeier) [reply astropy/cosmology/tests/test_cosmology.py:1592]: Tried a bit higher precision now especially in the roundtrip tests; the comment did not really make sense to me especially with the first test that was actually roundtripping too.

File: astropy/cosmology/funcs.py [hunk 16]
```diff
@@ -34,20 +36,52 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+       although their use is currently untested.
+
+       .. versionadded:: 4.3
+
+    bracket : sequence, optional
+       For methods 'Brent' and 'Golden', ``bracket`` defines the bracketing
+       interval and can either have three items (z1, z2, z3) so that z1 < z2 < z3
+       and ``func(z2) < func(z1), func(z3)`` or two items z1 and z3 which are
+       assumed to be a starting interval for a downhill bracket search.
+       For bimodal functions such as angular diameter distance this can be
```
nstarman, line 74: Do you mean non-monotonic functions?

astropy/cosmology/tests/test_cosmology.py
nstarman: doesn't need fixing anyway.

astropy/cosmology/funcs.py
dhomeier: Yes, probably had my head in statistical distributions at the time (I have to reconstruct this stuff myself from last year...)

astropy/cosmology/funcs.py
nstarman: That was my interpretation of the original author's intent. They are very careful throughout the package of stripping units before doing calculations.

astropy/cosmology/funcs.py
nstarman: I know the feeling. I have some PRs languishing that will need a careful read-through to get myself back up to speed. 🔍

astropy/cosmology/tests/test_cosmology.py
dhomeier: But could make it a bit easier to relate to the backwards test for 'Bounded'...

dhomeier:
Don't know what to make of this output in the 3.7 oldest deps job
```
153     >>> z_at_value(Planck18.angular_diameter_distance,
Expected:
    0.68044452                                             # doctest: +SKIP
Got:
    0.6804445231464921
```
- why is this line still executed at all, and why does it apparently want its own `+FLOAT_CMP` despite being told to ignore that code?

astropy/cosmology/funcs.py
```diff
@@ -90,56 +125,118 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+        val = fval
+
+    # 'Brent' and 'Golden' ignore `bounds`, force solution inside zlim
+    def f(z):
+        if z > zmax:
+            return 1.e300 * (1.0 + z - zmax)
+        elif z < zmin:
+            return 1.e300 * (1.0 + zmin - z)
+        elif isinstance(fval_zmin, Quantity):
+            return abs(func(z).value - val)
```
nstarman, line 221: While I like that this has less duplicated code, I think having the prior implementation of two functions might still be best.
Adding the z-bounds checks reduces calls to ``func``, which is great, but this adds an extra check every time. Without a side-by-side performance comparison, I image that having one function for non-quantity and one with quantity is faster.

nstarman:
> why is this line still executed at all, and why does it apparently want its own +FLOAT_CMP despite being told to ignore that code?

In my experience, some of these flags are finicky and really want to be 2 spaces after the content. Does restoring the spacing fix this?

astropy/cosmology/funcs.py
dhomeier: The z-bounds checks are inevitable, since some minimisers ignore the `bounds` settings, so they must be forced to reject those solutions, but the speedup is a nice side effect.

I must have misunderstood your earlier comment on the interpretation of the `return abs(func(z).value - val)` branch; but anyway I had checked that the conditional overhead is at the sub-percent level here, e.g. on my old notebook:

- `Planck18.angular_diameter_distance(z)`: 75 µs
- `Planck18.efunc(z)`: 25 µs
- `u.Quantity(x).value`:  4.5 µs
- `return x.value if isinstance(x, u.Quantity) else x`: 720 ns / 200 ns (for quantities / scalars)
- `return x` : 90 ns

so 100-120 ns for evaluating the condition is hardly worth the cost in readability and maintainability imo, since the function has also grown a bit beyond a simple `lambda z:`. And the function evaluations are not that numerous here, typically around 20.

nstarman: SGTM.

dhomeier:
Don't know what to make of this output in the 3.7 oldest deps job
```
153     >>> z_at_value(Planck18.angular_diameter_distance,
Expected:
    0.68044452                                             # doctest: +SKIP
Got:
    0.6804445231464921
```
- is 

> > why is this line still executed at all, and why does it apparently want its own +FLOAT_CMP despite being told to ignore that code?
> 
> In my experience, some of these flags are finicky and really want to be 2 spaces after the content. Does restoring the spacing fix this?

No, it failed already before I reformatted the lines; seems rather that the +FLOAT_CMP and +SKIP flags can interact which each other in weird ways, and completely different on three different installations (it seems the 3.9 run was happy all the time; the 3.8 one I think worked with just the +FLOAT_CMP but broke when I added the +SKIP, and I haven't found any way to get this through on 3.7 other than marking the entire code block +SKIP).

nstarman:
Very odd. I've never encountered that. Maybe we should ask for a 3rd opinion?

nstarman:
Why doctest skip these at all? why not just float_cmp?

dhomeier:
> Why doctest skip these at all? why not just float_cmp? (and skip the other weirdness)

Which other weirdness? Can't just fall back on float_cmp, since the particular installation on 3.7 + oldest deps never seems to find the high-z solution no matter what starting bracket is given (thus at least confirming the point that one should not rely on the 2-parameter version of `bracket`, but that was a result I could not reproduce with any version from SciPy 1.0). And once I marked that line (rather it's both L158 and L166 giving the unexpected result), the entire mess started.
If we can get an expert's tip for a smarter solution, I am happy to change it back; but debugging this by waiting 10 minutes for a test to fail on some machines seems worth neither the effort nor the CI time.

astropy/cosmology/funcs.py
dhomeier: Though I realised the alternative
`return 1.e300 * (1.0 + z - zmax) if z > zmax else 1.e300 * (1.0 + zmin - z) if z < zmin else abs(func(z).value - val`
just barely does not fit on one line... ;-)

nstarman:
> Which other weirdness?

Of combining with "doctest SKIP"

> If we can get an expert's tip for a smarter solution, I am happy to change it back; but debugging this by waiting 10 minutes for a test to fail on some machines seems worth neither the effort nor the CI time.

Completely fair. If this is a consistent behavior in py3.7, then it should be noted in the docs. Also if this happens every time, then it can also have a test. When we drop py3.7 it'll be very obvious and we can fix the docs.

**Maybe one more try**, I've been reading the pytest documentation and found this https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-doctest.

```
... doctest:
    :pyversion: > 3.7

    >>> z_at_value(Planck18.angular_diameter_distance,
   0.68044452
```

astropy/cosmology/funcs.py
nstarman: That would be nice.

astropy/cosmology/funcs.py
nstarman: This appears resolved. 👍

astropy/cosmology/funcs.py
nstarman: This appears resolved. 👍

astropy/cosmology/funcs.py
nstarman: This appears resolved. 👍

astropy/cosmology/funcs.py
nstarman: This appears to be resolved. 👍

astropy/cosmology/funcs.py
dhomeier: Yes, I thought about cutting the values to `1e99` and further abbreviating the parameters, but I must say I did not find the one-liner actually more readable!

dhomeier:
> 
> Completely fair. If this is a consistent behavior in py3.7, then it should be noted in the docs. Also if this happens every time, then it can also have a test. When we drop py3.7 it'll be very obvious and we can fix the docs.
> 
> **Maybe one more try**, I've been reading the pytest documentation and found this https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-doctest.
> 
> ```
> ... doctest:
>     :pyversion: > 3.7
> 
>     >>> z_at_value(Planck18.angular_diameter_distance,
>    0.68044452
> ```

I like that, thanks! Perhaps as close to the conditional skip some people are asking for in the pytest-doctest issues as it gets.
It's worth keeping in mind that there are two unrelated issues present:

1. `z_at_value(Planck18.angular_diameter_distance(1500 * u.Mpc)` has a non-deterministic result, and that particular combination of Scipy version, architecture and Python version happens to be the only one so far to produce the "less likely" outcome. I rather suspected one of the first two factors, but either way there is no guarantee this may not occur again on some other present or future setup. In fact I just checked that nearly the same test in the regular tests does return the expected result on all architectures:
https://github.com/astropy/astropy/blob/f526a0a33294391c1c8f514f9e1d912d21efa8f5/astropy/cosmology/tests/test_cosmology.py#L1616-L1617
whereas this returns the alternative value to the doctest:
https://github.com/astropy/astropy/blob/f526a0a33294391c1c8f514f9e1d912d21efa8f5/astropy/cosmology/tests/test_cosmology.py#L1618-L1619
the only differences being the wider bracket in the first case, and the regular tests still using `Planck13` rather than `Planck18`. So I think this is anything but consistent. Still could be useful to check for any changes once a new environment is added; the doctest options could still be updated then.

2. The non-helpful interplay of the doctest flags, apparently also different for various Python/pytest/pytest-doctestplus combinations. But `:pyversion: > 3.7` seems a good way to control that, at least for the present case.

dhomeier:
Giving up on the `pyversion` stuff for now – without a working example it's entirely opaque to me how to use this correctly, or if it is even supposed to work within code (rather than .rst files). And anyway in the last run it was at least on 3.7 _and_ 3.8 that produced different results. In light of this I have rewritten the docstring a bit to advise more clearly against just using two bracketing values. Alternatively that whole section might be shortened radically as it is covering some rather fringe use cases.

astropy/cosmology/funcs.py
```diff
@@ -90,56 +125,118 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
-zmin and zmax satisfying fval = func(z).""")
+        if bracket is None:
+            nobracket = True
+        else:
+            fval_brac = func(bracket)
+            if np.sign(fval - fval_brac[0]) != np.sign(fval_brac[-1] - fval):
+                nobracket = True
+            else:
+                zmin, zmax = bracket[0], bracket[-1]
+                fval_zmin, fval_zmax = fval_brac[[0, -1]]
```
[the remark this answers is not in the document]
nstarman: @dhomeier, I think this is still an issue. ``func`` does not need to return a numpy array if given a numpy vector input.

astropy/cosmology/funcs.py
```diff
@@ -90,56 +125,130 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+    res = minimize_scalar(f, method=method, bounds=(zmin, zmax), bracket=bracket, options=opt)
+
+    # Scipy docs state that `OptimizeResult` always has 'status' and 'message' attributes,
+    # but only `_minimize_scalar_bounded()` seems to have really implemented them.
+    if not res.success:
+        warnings.warn(f"Solver returned {res.get('status')}: {res.get('message', 'Unsuccessful')}\n"
+                      f"Precision {res.fun} reached after {res.nfev} function calls.")
+
+    if verbose:
+        print(res)
```
nstarman, line 246: In the tests, this line can be covered by

```
import sys
original_stdout = sys.stdout # Save a reference to the original standard output
with open('filename.txt', 'w') as f:
    sys.stdout = f # Change the standard output to the file we created.
    print('This message will be written to a file.')
    sys.stdout = original_stdout # Reset the standard output to its original value
```
(https://www.stackabuse.com/writing-to-a-file-with-pythons-print-function/)

Of course, we'd want a temporary file instead.

astropy/cosmology/funcs.py
```diff
@@ -90,56 +125,130 @@ def z_at_value(func, fval, zmin=1e-8, zmax=1000, ztol=1e-8, maxfun=500):
...
+            return abs(func(z).value - val)
+        else:
+            return abs(func(z) - val)
+
+    res = minimize_scalar(f, method=method, bounds=(zmin, zmax), bracket=bracket, options=opt)
+
+    # Scipy docs state that `OptimizeResult` always has 'status' and 'message' attributes,
+    # but only `_minimize_scalar_bounded()` seems to have really implemented them.
+    if not res.success:
+        warnings.warn(f"Solver returned {res.get('status')}: {res.get('message', 'Unsuccessful')}\n"
```
nstarman, line 242: Perhaps a good use of a custom minimizer is one that tests this line! or maybe set the maximum number of evaluations to be super low and the starting point really far away?

nstarman (commented):
Looks good except for one indexing issue. The line coverage would be nice, but maybe a fair bit of work to get the error messages tested.

astropy/cosmology/funcs.py
dhomeier: > @dhomeier, I think this is still an issue. func does not need to return a numpy array if given a non-scalar input.

Which one doesn't? Every single one I've checked start's right away with
```python
        if isiterable(z):
            z = np.asarray(z)
```
– I've probably not looked into all of them for all possible cosmology implementations, but this seemed quite clearly to define the standard API.

nstarman: I agree that it's a reasonable expectation that ``func(bracket)`` is amenable to numpy broadcasting and returns an ndarray. It's just that doesn't appear to be assumed in the old code.
In particular,
``fval_brac = func(bracket)`` can  be a list, not an ndarray. 
This passes all the new code except
``fval_zmin, fval_zmax = fval_brac[[0, -1]]``, which assumes an ndarray.
Changing it to 
``fval_zmin, fval_zmax = fval_brac[0],  fval_brac[-1]`` would also work on lists.

dhomeier: > I agree that it's a reasonable expectation that `func(bracket)` is amenable to numpy broadcasting and returns an ndarray. It's just that doesn't appear to be assumed in the old code.

As I have yet to see an example of `func` that does not follow this convention (the tests are now covering all single-parameter functions of all cosmologies in `core`), this probably comes down to the question "what kind of cosmological functions is `z_at_value` supposed to handle". If this might include any hypothetical cosmology a user might provide, or even any generic function (the description "A function that takes redshift as input" is of course extremely generic), then it should be prepared to get lists, tuples or any kind of sequences returned. But my feeling is that a) this is opening a can or worms (might be talking about oscillating cosmological functions we have no guarantee how the solvers will behave for), and b) downstream may well expect to get an ndarray returned on any input that can instantiate one, so it would not hurt to detect non-conforming functions straight away.

astropy/cosmology/funcs.py
nstarman: especially with the nested ternary. Hard to quickly parse.

astropy/cosmology/funcs.py
nstarman: 🆗 . Then this looks good! Right in time for v4.3 feature freeze! 🎉

nstarman (approved):
Looks great! 
(did you mean to rebase on main, not merge?)

dhomeier:
Thanks a lot for the thorough review @nstarman and the opportunity to polish up my intersphinx experiences a little bit!

> (did you mean to rebase on main, not merge?)

I was just fixing the merge conflicts here on GitHub, but now rebased to make sure everything runs and tests.

I don't know the official word on the feature freeze right now, but maybe @taldcroft or @pllim can chime in about merging.

pllim:
As of writing of this comment, v4.3.x branching has not happened yet, so it is @eteq 's call. But we have yet to get a formal approval from core maintainers... Not sure who should review this... @adrn or @mhvk ?

eteq:
Oops, I did the branching before seeing this one. I can probably merge and backport given its current status, though - it probably *should* have been merged before hand and feel bewteen the cracks with `cosmology`

eteq (approved):

eteq:
backported to v4.3.x in 97656f43e829b3738033dfe1a3a5cbb45c5c0508

dhomeier:
Thanks for the reviews and getting this in!

Issue #11063, opened 2020-11-19
RST Substitutions in docstrings


nstarman:
## Description

In ``docs/conf.py`` it is possible to add often used substitutions to the ``rst_epilog`` variable. These substitutions can **also be used in docstrings**.  I propose we replace many of the docstrings with their RST substitution counterparts, at least for very obvious substitutions like ``Unit`` and ``Quantity`` and ``SkyCoord``, etc.

As @mhvk  points out in https://github.com/astropy/astropy/pull/10662#discussion_r527127005, we can define things like "Quantity-like" where a Column with a unit will work where a Quantity would too. Some of these, like where a string can replace a unit, might be a multiple RST replacement (see example below).

- - -

## Example

docs/conf.py
```
rst_epilog += """

.. |Unit| replace:: :class:`~astropy.units.Unit`
.. |Unit.to| replace:: :meth:`~astropy.units.Unit.to`  # descriptive
.. |Quantity| replace:: :class:`~astropy.units.Quantity`
.. |SkyCoord| replace:: :class:`~astropy.coordinates.SkyCoord`

.. |find_api_page| replace:: :func:`~astropy.find_api_page`  # function

.. |Representation| replace:: :class:`~astropy.coordinates.BaseRepresentation`  # simplification
.. |Differential| replace:: :class:`~astropy.coordinates.BaseDifferential`  # simplification

.. |Quantity-like| replace:: Quantity-like  # replacement invariant
.. |unit-like| replace:: :class:`~astropy.units.Unit` or str  # multiple subs

"""
```

some_file.py
```python
def func(x):
    """function

    Parameters
    ——————
    x : |Quantity|
    """
    pass

```

Cadair:
I don't think the `.. class::` replacements are worthwhile. Given I am pretty sure  ``` `.Quantity` ``` alone will find the right class. (I recently discovered that shortcut, it finds the matching class in the obj inventory).

nstarman:
> Given I am pretty sure `.Quantity` alone will find the right class. 

does that also work for functions? in particular are they correctly appended with ``()``?

Cadair:
No, the default is `.. obj::` so they don't get the `()` unless you do ```:meth:`.Quantity.to` ``` I would guess.

nstarman:
Still. That's a cool simplification. I wish they would advertise that on the ReST docs.
[...]

PR #11086, opened 2020-11-25
Add geodetic representations

astrojuanlu:
### Description

In line with discussions at #4261, this pull request aims to add geodetic representations.

I started with `WGS84GeodeticRepresentation` to touch base and get a few eyes on my approach. In particular, I don't think we can cleanly have a generic `GeodeticRepresentation` where the ellipsoid is an extra parameter, retained in roundtrip transformations, but I would be glad to be proven wrong.

The tests are almost a copy-paste of `TestSphericalRepresentation`, with a few changes:

- `height` instead of `distance`
- `height` can only be a `Quantity`, not a `Distance`
- I didn't add differentials (not sure if worth it)

Besides, I also adapted `test_gc2gd` and `test_gd2gc` from `test_earth.py` to make sure we get the same results.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,9 @@ astropy.coordinates
 - Adds a direct transformation from ``ICRS`` to ``AltAz`` frames. This
   provides a modest speedup of approximately 10 percent. [#11079]
 
+- Adds new ``WGS84GeodeticRepresentation``, ``WGS72GeodeticRepresentation``,
+  and ``GRS80GeodeticRepresentation``. [#11086]
+
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,7 +24,9 @@ astropy.coordinates
   provides a modest speedup of approximately 10 percent. [#11079]
 
 - Adds new ``WGS84GeodeticRepresentation``, ``WGS72GeodeticRepresentation``,
-  and ``GRS80GeodeticRepresentation``. [#11086]
+  and ``GRS80GeodeticRepresentation``. These are mostly for use inside
+  ``EarthLocation`` but can also be used to convert between geocentric
+  (cartesian) and different geodetic representations directly. [#11086]
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -882,6 +882,17 @@ class BaseGeodeticRepresentation(BaseRepresentation):
             height = Distance(0, 'm')
         super().__init__(lon, lat, height, copy=copy)
 
+        try:
+            self._height = Distance(self._height, copy=False)
+        except ValueError as e:
+            if e.args[0].startswith('Height must be >= 0'):
+                raise ValueError("Height must be >= 0. To allow negative "
+                                 "height values, you must explicitly pass"
+                                 " in a `Distance` object with the the "
+                                 "argument 'allow_negative=True'.")
+            else:
+                raise
+
     def to_cartesian(self):
         """
         Converts WGS84 geodetic coordinates to 3D rectangular (geocentric)
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -17,7 +17,7 @@ from astropy.units.quantity import QuantityInfoBase
 from astropy.utils.exceptions import AstropyUserWarning
 from .angles import Angle, Longitude, Latitude
 from .distances import Distance
-from .representation import BaseRepresentation, CartesianRepresentation, CartesianDifferential
+from .representation import BaseRepresentation, CartesianRepresentation, CartesianDifferential, REPRESENTATION_CLASSES
 from .matrix_utilities import matrix_transpose
 from .errors import UnknownSiteException
 from astropy.utils import data
@@ -288,22 +288,18 @@ class EarthLocation(u.Quantity):
         ``gd2gc`` is used.  See https://github.com/liberfa/erfa
         """
         ellipsoid = _check_ellipsoid(ellipsoid, default=cls._ellipsoid)
-        # We use Angle here since there is no need to wrap the longitude -
-        # gd2gc will just take cos/sin anyway.  And wrapping might fail
-        # on readonly input.
-        lon = Angle(lon, u.degree, copy=False)
+        # As wrapping fails on readonly input, we do so manually
+        lon = Angle(lon, u.degree, copy=False).wrap_at(180 * u.degree)
         lat = Latitude(lat, u.degree, copy=False)
         # don't convert to m by default, so we can use the height unit below.
         if not isinstance(height, u.Quantity):
             height = u.Quantity(height, u.m, copy=False)
         # get geocentric coordinates. Have to give one-dimensional array.
-        xyz = erfa.gd2gc(getattr(erfa, ellipsoid),
-                         lon.to_value(u.radian),
-                         lat.to_value(u.radian),
-                         height.to_value(u.m))
+        xyz = REPRESENTATION_CLASSES[ellipsoid.lower() + "geodetic"](
+            lon, lat, Distance(height, allow_negative=True), copy=False,
+        ).represent_as(CartesianRepresentation).get_xyz(xyz_axis=-1)
         self = xyz.ravel().view(cls._location_dtype,
                                 cls).reshape(xyz.shape[:-1])
-        self._unit = u.meter
         self._ellipsoid = ellipsoid
         return self.to(height.unit)
 
@@ -597,13 +593,13 @@ class EarthLocation(u.Quantity):
         ``gc2gd`` is used.  See https://github.com/liberfa/erfa
         """
         ellipsoid = _check_ellipsoid(ellipsoid, default=self.ellipsoid)
-        self_array = self.to(u.meter).view(self._array_dtype, np.ndarray)
-        lon, lat, height = erfa.gc2gd(getattr(erfa, ellipsoid), self_array)
+        llh = CartesianRepresentation(self.x, self.y, self.z).represent_as(
+            REPRESENTATION_CLASSES[ellipsoid.lower() + "geodetic"])
         return GeodeticLocation(
-            Longitude(lon * u.radian, u.degree,
-                      wrap_angle=180.*u.degree, copy=False),
-            Latitude(lat * u.radian, u.degree, copy=False),
-            u.Quantity(height * u.meter, self.unit, copy=False))
+            llh.lon,
+            llh.lat,
+            u.Quantity(llh.height, self.unit, copy=False))
+
 
     @property
     def lon(self):
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -25,7 +25,8 @@ from .matrix_utilities import matrix_transpose
 from .errors import UnknownSiteException
 
 
-__all__ = ['EarthLocation', 'WGS84GeodeticRepresentation', 'WGS72GeodeticRepresentation', 'GRS80GeodeticRepresentation']
+__all__ = ['EarthLocation', 'BaseGeodeticRepresentation',
+           'WGS84GeodeticRepresentation', 'WGS72GeodeticRepresentation', 'GRS80GeodeticRepresentation']
 
 GeodeticLocation = collections.namedtuple('GeodeticLocation', ['lon', 'lat', 'height'])
 
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -19,18 +19,21 @@ from astropy.utils.decorators import format_doc
 from astropy.utils.exceptions import AstropyUserWarning
 
 from .angles import Angle, Longitude, Latitude
-from .representation import BaseRepresentation, CartesianRepresentation, CartesianDifferential, REPRESENTATION_CLASSES
+from .representation import (BaseRepresentation, CartesianRepresentation,
+                             CartesianDifferential)
 from .matrix_utilities import matrix_transpose
 from .errors import UnknownSiteException
 
 
 __all__ = ['EarthLocation', 'BaseGeodeticRepresentation',
-           'WGS84GeodeticRepresentation', 'WGS72GeodeticRepresentation', 'GRS80GeodeticRepresentation']
+           'WGS84GeodeticRepresentation', 'WGS72GeodeticRepresentation',
+           'GRS80GeodeticRepresentation']
 
 GeodeticLocation = collections.namedtuple('GeodeticLocation', ['lon', 'lat', 'height'])
 
-# Available ellipsoids (defined in erfam.h, with numbers exposed in erfa).
-ELLIPSOIDS = ('WGS84', 'GRS80', 'WGS72')
+ELLIPSOIDS = {}
+"""Available ellipsoids (defined in erfam.h, with numbers exposed in erfa)."""
+# Note: they get filled by the creation of the geodetic classes.
 
 OMEGA_EARTH = ((1.002_737_811_911_354_48 * u.cycle/u.day)
                .to(1/u.s, u.dimensionless_angles()))
@@ -297,8 +300,7 @@ class EarthLocation(u.Quantity):
         if not isinstance(height, u.Quantity):
             height = u.Quantity(height, u.m, copy=False)
         # get geocentric coordinates.
-        geodetic = REPRESENTATION_CLASSES[ellipsoid.lower() + "geodetic"](
-            lon, lat, height, copy=False)
+        geodetic = ELLIPSOIDS[ellipsoid](lon, lat, height, copy=False)
         xyz = geodetic.to_cartesian().get_xyz(xyz_axis=-1) << height.unit
         self = xyz.view(cls._location_dtype, cls).reshape(geodetic.shape)
         self._ellipsoid = ellipsoid
@@ -596,11 +598,10 @@ class EarthLocation(u.Quantity):
         ellipsoid = _check_ellipsoid(ellipsoid, default=self.ellipsoid)
         xyz = self.view(self._array_dtype, u.Quantity)
         llh = CartesianRepresentation(xyz, xyz_axis=-1, copy=False).represent_as(
-                REPRESENTATION_CLASSES[ellipsoid.lower() + "geodetic"])
+                ELLIPSOIDS[ellipsoid])
         return GeodeticLocation(
             Longitude(llh.lon, u.deg, wrap_angle=180*u.deg, copy=False),
             llh.lat << u.deg, llh.height << self.unit)
-
 
     @property
     def lon(self):
@@ -874,6 +875,11 @@ class BaseGeodeticRepresentation(BaseRepresentation):
     attr_classes = {'lon': Longitude,
                     'lat': Latitude,
                     'height': u.Quantity}
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        if '_ellipsoid' in cls.__dict__:
+            ELLIPSOIDS[cls._ellipsoid] = cls
 
     def __init__(self, lon, lat=None, height=None, copy=True):
         if height is None and not isinstance(lon, self.__class__):
--- a/astropy/coordinates/earth.py
+++ b/astropy/coordinates/earth.py
@@ -888,10 +888,8 @@ class BaseGeodeticRepresentation(BaseRepresentation):
         cartesian coordiantes.
         """
         xyz = erfa.gd2gc(getattr(erfa, self._ellipsoid),
-                         self.lon.to_value(u.radian),
-                         self.lat.to_value(u.radian),
-                         self.height.to_value(u.m))
-        return CartesianRepresentation(xyz, unit=u.m, xyz_axis=-1, copy=False)
+                         self.lon, self.lat, self.height)
+        return CartesianRepresentation(xyz, xyz_axis=-1, copy=False)
 
     @classmethod
     def from_cartesian(cls, cart):
@@ -900,12 +898,8 @@ class BaseGeodeticRepresentation(BaseRepresentation):
         WGS84 geodetic coordinates.
         """
         lon, lat, height = erfa.gc2gd(getattr(erfa, cls._ellipsoid),
-                                      cart.get_xyz(xyz_axis=-1).to_value(u.m))
-        return cls(
-            Longitude(lon << u.radian, u.degree,
-                      wrap_angle=180. << u.degree, copy=False),
-            Latitude(lat << u.radian, u.degree, copy=False),
-            u.Quantity(height << u.meter, copy=False))
+                                      cart.get_xyz(xyz_axis=-1))
+        return cls(lon, lat, height, copy=False)
 
 
 @format_doc(geodetic_base_doc)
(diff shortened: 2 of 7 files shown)


astropy/coordinates/representation.py
```diff
@@ -2222,6 +2222,62 @@ def to_cartesian(self):
...
+    @classmethod
+    def from_cartesian(cls, cart):
+        """
+        Converts 3D rectangular cartesian coordinates (assumed geocentric) to
+        WGS84 geodetic coordinates.
+        """
+        # TODO: Check ret?
+        lon, lat, height, ret = erfa_ufunc.gc2gd(1, cart.get_xyz(xyz_axis=-1).to_value(u.m))
+        return cls(
+            Longitude(lon * u.radian, u.degree,
```
mhvk, line 2275: While you are at it: do change this to `lon << u.radian`, etc., that avoids an extra copy.

mhvk (commented):
@astrojuanlu - this looks good! I'm actually pleasantly surprised to see how relatively simple it is to add - the work is mostly in the tests!

Beyond the small in-line comments: I think these representations belong in `earth.py` as they are not useful as general representations. Similarly, I'd put the tests in `test_earth`. (Or perhaps a new `earth_representations` or even an `earth/` module. Though seems a bit overkill right now.)

I agree that for now we can ignore velocities - we can add the ability to mimic observations from airplanes when the need arises!

How were you thinking of proceeding? A first step might simply be to remove the transformation machinery in `EarthLocation` and instead go through these representations inside `from_geodetic` and `lon, lat, heigh` - `to_geodetic()` may need some kind of deprecation warning as right now it returns a `namedtuple` and people may count on it. (I think that is OK, we should move to `.represent_as` anyway, making it more frame-like.)

p.s. I'll change this to draft right now just to be sure we don't accidentally merge.

astropy/coordinates/representation.py
```diff
@@ -2222,6 +2222,62 @@ def to_cartesian(self):
...
+    def __init__(self, lon, lat=None, height=None, copy=True):
+        super().__init__(lon, lat, height, copy=copy)
+
+    def to_cartesian(self):
+        """
+        Converts WGS84 geodetic coordinates to 3D rectangular (geocentric)
+        cartesian coordiantes.
+        """
+        # TODO: Check ret?
+        xyz, ret = erfa_ufunc.gd2gc(1,
```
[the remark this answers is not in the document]
astrojuanlu: Good point about the class attribute, I'll explore that!

About `erfa.gd2gc1` though: I can't seem to find it :question:

astropy/coordinates/representation.py
```diff
@@ -2222,6 +2222,62 @@ def to_cartesian(self):
...
+
+    copy : bool, optional
+        If `True` (default), arrays will be copied. If `False`, arrays will
+        be references, though possibly broadcast to ensure matching shapes.
+
+    """
+
+    attr_classes = {'lon': Longitude,
+                    'lat': Latitude,
+                    'height': u.Quantity}
```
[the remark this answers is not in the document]
astrojuanlu: Will do! I was dissuaded from doing it because I didn't want the parallax, cosmology, etc But it's good to enforce the unit here.

astropy/coordinates/representation.py
astrojuanlu: Good catch!

astrojuanlu:
Thanks for the prompt review @mhvk !

I agree with moving this to `earth.py` and with all your comments. I wanted to think about the way forward with `EarthLocation` today but I couldn't, so I don't have any ideas yet - will give it some thought tomorrow.

astropy/coordinates/representation.py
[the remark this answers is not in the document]
mhvk: That was a type (1 instead of back tick), i.e., just use `erfa.gd2gc`. Sorry!

astropy/coordinates/representation.py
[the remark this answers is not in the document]
mhvk: Indeed, maybe we should have a `Length` class or so... But my suggestion was indeed only to enforce the unit, which seems easiest through the class.

mkbrewer:
> I agree that for now we can ignore velocities - we can add the ability to mimic observations from airplanes when the need arises!

Don't forget balloons!  https://askdrphilastro.wordpress.com/2019/12/15/blast-antarctica-2019-2020/

mhvk:
> Don't forget balloons!

How can I forget my own colleague!  So, we will have to get there...

astropy/coordinates/representation.py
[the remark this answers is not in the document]
astrojuanlu: Besides, notice that in other representation classes, `Distance` is not used in `attr_classes`, but in the initializer. See for example:

https://github.com/astropy/astropy/blob/0ff46cb5305d288af47bd3de4432c7dc941b33cb/astropy/coordinates/representation.py#L1829-L1848

astropy/coordinates/representation.py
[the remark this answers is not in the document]
astrojuanlu: Oh, got it!

astropy/coordinates/tests/test_earth.py
```diff
@@ -7,16 +7,43 @@
...
+        return False
+    if getattr(rep1, '_differentials', False):
+        if rep1._differentials.keys() != rep2._differentials.keys():
+            return False
+        for key, diff1 in rep1._differentials.items():
+            result &= components_equal(diff1, rep2._differentials[key])
+    elif getattr(rep2, '_differentials', False):
+        return False
+
+    return result & components_equal(rep1, rep2)
```
astrojuanlu, line 44: These are copy-pasted from `test_representations.py`. Should we move them to `astropy.utils` or `astropy.tests.helper`, or is it not too bad to have them duplicated?

astrojuanlu:
Addressed the first batch of comments, going to look into `EarthLocation` now.

astropy/coordinates/representation.py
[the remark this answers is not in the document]
astrojuanlu: ...which, interestingly, doesn't work: https://github.com/astropy/astropy/issues/11104

astrojuanlu:
What I did in the end was to deprecate `.to_geodetic` and `.to_geocentric` without changing their return values, and added a new `.represent_as`.

This is ready for a second pass!

astrojuanlu:
Many test failures due to the `AstropyDeprecationWarning` I introduced - I guess no internal calls to deprecated methods are allowed. Will fix it.

mkbrewer:
I would just note that many locations on Earth have heights that are below that of the ellipsoid. I seems a bit unreasonable to force a different method for creating and `EarthLocation` in these instances.

mhvk:
Darn, we'd need `allow_negative`. I think for now it would be best to just proceed with a `Quantity` - we can deal with insisting it is a length differently/later.

astropy/coordinates/earth.py
```diff
@@ -595,14 +593,28 @@ def to_geodetic(self, ellipsoid=None):
         For the conversion to geodetic coordinates, the ERFA routine
         ``gc2gd`` is used.  See https://github.com/liberfa/erfa
         """
+        warn("Method .to_geodetic(ellipsoid) is deprecated, "
```
mhvk, line 596: Agree with the deprecation message, but in analogy with a frame or sky coord, I think there is little against still having the properties for particular representations, i.e., I think we should allow (and also refer to in the warning message),
```
el.{wgs84,...}geodetic
el.cartesian
```

astropy/coordinates/earth.py
```diff
@@ -595,14 +593,28 @@ def to_geodetic(self, ellipsoid=None):
...
         return GeodeticLocation(
-            Longitude(lon * u.radian, u.degree,
-                      wrap_angle=180.*u.degree, copy=False),
-            Latitude(lat * u.radian, u.degree, copy=False),
-            u.Quantity(height * u.meter, self.unit, copy=False))
+            llh.lon,
+            llh.lat,
+            u.Quantity(llh.height, self.unit, copy=False))
+
+    def represent_as(self, other_class):
```
mhvk, line 608: Again seeing `EarthLocation` as more like a frame/skycoord rather than a representation, I think we can allow strings. I'd suggest following the nomenclature in their `represent_as` method and calling it `base`.

astropy/coordinates/earth.py
```diff
@@ -848,3 +864,84 @@ def _to_value(self, unit, equivalencies=[]):
...
         return new_array.view(self.dtype).reshape(self.shape)
+
+
+class BaseGeodeticRepresentation(BaseRepresentation):
+    """
+    Representation of points in 3D geodetic coordinates.
+
+    Parameters
+    ----------
+    lon, lat : `~astropy.units.Quantity`
```
mhvk, line 875: The classes are `Longitude` and `Latitude` - might as well be precise...

astropy/coordinates/earth.py
```diff
@@ -848,3 +864,84 @@ def _to_value(self, unit, equivalencies=[]):
             equivalencies = self._equivalencies
         new_array = self.unit.to(unit, array_view, equivalencies=equivalencies)
         return new_array.view(self.dtype).reshape(self.shape)
+
+
+class BaseGeodeticRepresentation(BaseRepresentation):
```
mhvk, line 869: I think this will get registered in `REPRESENTATION_CLASSES`, which is not ideal... Pinging @adrn for ideas; should we allow a `name = False` explicit way to avoid registration? (Note: this should not have to hold up this PR!)

astropy/coordinates/earth.py
```diff
@@ -848,3 +864,84 @@ def _to_value(self, unit, equivalencies=[]):
...
+                                      cart.get_xyz(xyz_axis=-1).to_value(u.m))
+        return cls(
+            Longitude(lon << u.radian, u.degree,
+                      wrap_angle=180. << u.degree, copy=False),
+            Latitude(lat << u.radian, u.degree, copy=False),
+            u.Quantity(height << u.meter, copy=False))
+
+
+class WGS84GeodeticRepresentation(BaseGeodeticRepresentation):
+    """
```
mhvk, line 927: One disadvantage of setting this docstring is that you lose the nice information you added in the `Base...`.  For a possible solution, see https://docs.astropy.org/en/latest/api/astropy.utils.decorators.format_doc.html#astropy.utils.decorators.format_doc

astropy/coordinates/tests/test_earth.py
mhvk: I'd just import then from `test_representation` for now (also done in `test_representation_methods`...)

astropy/coordinates/tests/test_earth.py
```diff
@@ -410,3 +437,236 @@ def test_info():
     EarthLocation._get_site_registry(force_builtin=True)
     greenwich = EarthLocation.of_site('greenwich')
     assert str(greenwich.info).startswith('name = Royal Observatory Greenwich')
+
+
+class TestWGS84GeodeticRepresentationRepresentation:
```
mhvk, line 442: I would shrink this class substantially - many of the tests are done on `Spherical` to just have them done once, no need to do them again. Also, do re-align for the longer name...

astropy/coordinates/tests/test_earth.py
```diff
@@ -410,3 +437,236 @@ def test_info():
...
+        """ This is a regression test: calling represent_as() and passing in the
+            same class as the object shouldn't round-trip through cartesian.
+        """
+
+        sph = WGS84GeodeticRepresentation(1*u.deg, 2*u.deg, np.nan*u.m)
+        new_sph = sph.represent_as(WGS84GeodeticRepresentation)
+        assert quantity_allclose(new_sph.lon, sph.lon)
+        assert quantity_allclose(new_sph.lat, sph.lat)
+
+        # dif = SphericalCosLatDifferential(1*u.mas/u.yr, 2*u.mas/u.yr,
```
mhvk, line 619: Probably good to put a TODO on top that we do not yet have differentials (and then remove the commented out code here)

astropy/coordinates/tests/test_earth.py
```diff
@@ -410,3 +437,236 @@ def test_info():
...
+
+
+def test_geodetic_to_geocentric():
+    """Test that we reproduce erfa/src/t_erfa_c.c t_gd2gc"""
+    gd = WGS84GeodeticRepresentation(3.1 * u.rad, -0.5 * u.rad, 2500.0 * u.m)
+    gc = gd.to_cartesian()
+
+    assert quantity_allclose(gc.x, -5599000.5577049947 * u.m, atol=1e-7 * u.m)
+    assert quantity_allclose(gc.y, 233011.67223479203 * u.m, atol=1e-7 * u.m)
+    assert quantity_allclose(gc.z, -3040909.4706983363 * u.m, atol=1e-7 * u.m)
```
mhvk, line 672: Missing from the tests is the new `represent_as` method.

astropy/coordinates/tests/test_earth.py
mhvk: Thinking more, since all new tests are (very nicely) just focussed on the representations, how about a new `test_geodetic_representations`?

mhvk (commented):
@astrojuanlu - this is starting to look very nice.  Bit of a mix of larger and probably overly detailed comments.

More general one is where we stop. One option is actually a bit *before* where you are now, i.e., not add deprecation messages yet, but just introduce the new representations and use them *internally* only for the transformations. This has the advantage that no tests will fail yet. Arguably, in that case, we would also not yet add the `represent_as` method (which is not tested yet, so that is fine, I think).

The other is perhaps a bit further than you are now, whether `EarthLocation` stops becoming a `Quantity` and instead stores the representation in `.data` like a frame (or perhaps as an intermediate suggestion still shares the data with a `CartesianRepresentation`).

But my sense would be to take the smaller step suggested above, so we can get this in and follow-up on it.

astrojuanlu:
> But my sense would be to take the smaller step suggested above, so we can get this in and follow-up on it.

I agree - I will remove the deprecation warnings and remove `represent_as` for now, and follow up in a future PR. Addressing your comments now.

astropy/coordinates/earth.py
astrojuanlu: Removed the deprecation message for now, but saving this for the future.

astropy/coordinates/earth.py
astrojuanlu: Removed `represent_as` for now, but saving this for the future.

astropy/coordinates/tests/test_earth.py
astrojuanlu: Removed `represent_as` for now, will test it when we finally add it.

astropy/coordinates/earth.py
astrojuanlu: Tried this trick https://github.com/astropy/astropy/issues/11063#issuecomment-732212333 let's see if it works!

astropy/coordinates/earth.py
astrojuanlu: I ended up removing the docstrings, after all the name should be explicit enough.

However, I did a local `tox -e build_docs` and I didn't see these classes anywhere in the docs, so I added them to `__all__` in `earth.py`.

astropy/coordinates/tests/test_earth.py
astrojuanlu: I moved these tests to `test_geodetic_representations.py`, but I'm not sure which ones to remove.

astropy/coordinates/earth.py
astrojuanlu: I was wrong, removing the docstrings leaves the class documentation empty:

Screenshot_2020-12-01 WGS84GeodeticRepresentation — Astropy v4 3 dev322+g7f34535cc d20201201

astrojuanlu: `@format_doc` proved to be the best solution :+1:

astropy/coordinates/tests/test_earth.py
astrojuanlu: Done!

astropy/coordinates/tests/test_earth.py
astrojuanlu: Done!

mkbrewer:
While supporting alternate ellipsoids is fine, you should realize that there is more work involved in doing so. As currently written, `ErfaAstrom.apco()` calls `erfa.apco()` which calculates the observer's position and velocity based on the WGS-84 ellipsoid. Thus geodetic  coordinates based on an alternate ellipsoid will yield incorrect results in any ICRS<->CIRS or ICRS<->AltAz transform.

mkbrewer:
I should note that this relates to task 1) in issue #11072

astrojuanlu:
Warnings are legitimate:

```
/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/11086/lib/python3.8/site-packages/astropy/coordinates/earth.py:docstring of astropy.coordinates.GRS80GeodeticRepresentation:1: WARNING: py:class reference target not found: astropy.coordinates.earth.BaseGeodeticRepresentation
writing output... [ 16%] api/astropy.coordinates.SphericalCosLatDifferential .. api/astropy.io.ascii.Latex
/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/11086/lib/python3.8/site-packages/astropy/coordinates/earth.py:docstring of astropy.coordinates.WGS72GeodeticRepresentation:1: WARNING: py:class reference target not found: astropy.coordinates.earth.BaseGeodeticRepresentation
/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/11086/lib/python3.8/site-packages/astropy/coordinates/earth.py:docstring of astropy.coordinates.WGS84GeodeticRepresentation:1: WARNING: py:class reference target not found: astropy.coordinates.earth.BaseGeodeticRepresentation
...
[changelog_links] Adding changelog links to "changelog"
<rst_epilog>:14: WARNING: py:class reference target not found: astropy.coordinates.earth.BaseGeodeticRepresentation
```

fixing

astrojuanlu:
> As currently written, `ErfaAstrom.apco()` calls `erfa.apco()` which calculates the observer's position and velocity based on the WGS-84 ellipsoid. Thus geodetic coordinates based on an alternate ellipsoid will yield incorrect results in any ICRS<->CIRS or ICRS<->AltAz transform.

I don't understand @mkbrewer , isn't `ErfaAstrom.apco()` always converting whatever coordinates to WGS84?

https://github.com/astropy/astropy/blob/9680cd546aa9063758f2c23c836ca79a7c8f1eb1/astropy/coordinates/erfa_astrom.py#L45

Unless I missed anything, the `EarthLocation.to_geodetic` return value didn't change with this PR.

astrojuanlu:
All CI checks passed now :tada:

mkbrewer:
> I don't understand @mkbrewer , isn't ErfaAstrom.apco() always converting whatever coordinates to WGS84?

Oh, you're right. I missed that conversion. Sorry.

CHANGES.rst
```diff
@@ -23,6 +23,9 @@ astropy.coordinates
 - Adds a direct transformation from ``ICRS`` to ``AltAz`` frames. This
   provides a modest speedup of approximately 10 percent. [#11079]
 
+- Adds new ``WGS84GeodeticRepresentation``, ``WGS72GeodeticRepresentation``,
+  and ``GRS80GeodeticRepresentation``. [#11086]
```
mhvk, line 27: Should we note what these are good for? E.g., "These are mostly for use inside ``EarthLocation`` but can also be used to convert between geocentric (cartesian) and different geodetic representations directly."

astropy/coordinates/earth.py
```diff
@@ -287,22 +291,18 @@ def from_geodetic(cls, lon, lat, height=0., ellipsoid=None):
...
             height = u.Quantity(height, u.m, copy=False)
         # get geocentric coordinates. Have to give one-dimensional array.
-        xyz = erfa.gd2gc(getattr(erfa, ellipsoid),
-                         lon.to_value(u.radian),
-                         lat.to_value(u.radian),
-                         height.to_value(u.m))
+        xyz = REPRESENTATION_CLASSES[ellipsoid.lower() + "geodetic"](
+            lon, lat, Distance(height, allow_negative=True), copy=False,
+        ).represent_as(CartesianRepresentation).get_xyz(xyz_axis=-1)
         self = xyz.ravel().view(cls._location_dtype,
```
mhvk, line 304: Might be clearer to change the unit of `xyz` here instead of `self` at the end - mostly as we can do it in-place.

astropy/coordinates/earth.py
```diff
@@ -596,13 +596,12 @@ def to_geodetic(self, ellipsoid=None):
...
+        llh = CartesianRepresentation(self.x, self.y, self.z).represent_as(
+            REPRESENTATION_CLASSES[ellipsoid.lower() + "geodetic"])
         return GeodeticLocation(
-            Longitude(lon * u.radian, u.degree,
-                      wrap_angle=180.*u.degree, copy=False),
-            Latitude(lat * u.radian, u.degree, copy=False),
-            u.Quantity(height * u.meter, self.unit, copy=False))
+            llh.lon,
+            llh.lat,
+            u.Quantity(llh.height, self.unit, copy=False))
```
mhvk, line 604: Can just be `llh.height << self.unit`.

astropy/coordinates/representation.py
```diff
@@ -1810,6 +1810,8 @@ def __init__(self, lon, lat=None, distance=None, differentials=None,
             try:
                 self._distance = Distance(self._distance, copy=False)
             except ValueError as e:
+                # FIXME: This will raise no matter what,
```
mhvk, line 1813: Maybe keep outside of this PR?

mhvk (commented):
This is looking good! It made me realize that really `gc2gd` and `gd2gc` should be made to work with `Quantity` - so I took the liberty of adding a commit that does so.

Also, many of the other comments are a bit nit-picky, so while I was at it, I addressed them. Main thing to perhaps think about is that I removed the use of `Distance` - I find it just too unhandy, not really suitable for a height, and it really is more useful for converting to redshift, etc...

Will push those 2 commits to your branch...

mhvk:
Oops, some failures - will fix (should have tested all of `coordinates`...

astrojuanlu:
Thanks @mhvk ! I can leave the rest to you or address the comments tomorrow, as you prefer

astropy/coordinates/earth.py
```diff
@@ -848,3 +847,82 @@ def _to_value(self, unit, equivalencies=[]):
...
+        be references, though possibly broadcast to ensure matching shapes.
+
+"""
+
+
+@format_doc(geodetic_base_doc)
+class BaseGeodeticRepresentation(BaseRepresentation):
+    """Base geodetic representation."""
+
+    attr_classes = {'lon': Longitude,
```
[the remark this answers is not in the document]
mhvk: Let's leave this for follow-up...

mhvk:
@astrojuanlu - OK, I also changed the tests now. Plus one thing that you should sign off on: to make `ELLIPSOIDS` a dict with the actual representations as values. I think it works quite nicely, mostly because #11090 already replaced the meta class with `__init_subclass__`, so one can now easily chain things. See 4612ca1.

mhvk (approved):
Approving, but, as noted, @astrojuanlu, would be best if you approve the changes I made!

astropy/coordinates/earth.py
```diff
@@ -875,6 +876,11 @@ class BaseGeodeticRepresentation(BaseRepresentation):
                     'lat': Latitude,
                     'height': u.Quantity}
 
+    def __init_subclass__(cls, **kwargs):
```
astrojuanlu, line 879: Nice trick!

astrojuanlu (commented):
Can't approve my own pull request. This looks good to me, thanks @mhvk for the final tweaks!

Issue #6623, opened 2017-09-28
Use entry points to allow packages to register plugins/extensions in various parts of Astropy


astrofrog:
There are currently several places where we allow or should allow packages to register extensions to Astropy. For example:

- The unified I/O interface
- Custom WCS to frame/frame to WCS mappings (https://github.com/astropy/astropy/blob/master/astropy/wcs/utils.py#L141)
- Custom serializations (https://github.com/astropy/astropy/pull/6619)
- Coordinate frames and transformations

The current workflow is for a package to append something to a list or call a register function. However, this is not ideal because it means the user has to know to explicitly import the package first. For example, let's say sunpy defined a ``.solar`` file format (I'm making this up, but hey).

If I just did:

```python
from astropy.table import Table
t = Table(...)
t.write('myfile.solar')
```

this wouldn't work. I'd need to do:

```python
from astropy.table import Table
import sunpy.io  # making this up
t = Table(...)
t.write('myfile.solar')
```

The standard way around this kind of issue is to consider using entry points to set up a plugin system. With entry points, sunpy could then say it has plugins for file formats and provide the entry point to set up those file formats. This then allows astropy to discover plugins without the user having to import other packages explicitly (installing is enough).

To avoid a proliferation of ways of doing this, we should probably decide if we want to do this and then come up with standard entry point categories, e.g. ``[astropy.table.serialization]``, ``[astropy.table.formats]``, ``[astropy.coordinates.frames]`` and so on.

Note that if we had one entry point per Table format, we could also use entry points in the core package itself to register reader/writers instead of doing this ugly hack: https://github.com/astropy/astropy/blob/58fdc0115d0204c5bb64a36b5a24939190e4c396/astropy/table/__init__.py#L54

I'm opening this here to get some initial feedback - not sure if this should become and APE, or if this + astropy-dev discussion would be enough. Maybe we can see how controversial this idea is 😆

MSeifert04:
The main problem is documentation I guess. I don't mean documenting the entry points but updating the documentation of `Table.write` depending on the entry points. I'm not sure if that's easy to overcome but that was one of the main incentives for `delay_doc_updates`...

astrofrog:
[...]

PR #11091, opened 2020-11-25
Replace pkg_resources with importlib.metadata

saimn:
Following discussions on `pkg_resources` vs `importlib.metadata`: `pkg_resources` is slow ([1](https://github.com/astropy/astropy/issues/4598#issuecomment-404657295), [2](https://github.com/astropy/astropy/pull/7647), [3](https://github.com/astropy/astropy/issues/6623#issuecomment-626183284)), which is the reason why [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html) was added to the stdlib in Python 3.8 (with a backport for 3.7).

For instance, importing `astropy.modeling`:
image

There is a better alternative, so let's use it ?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,12 +25,14 @@
 # be accessible, and the documentation will not build correctly.
 # See sphinx_astropy.conf for which values are set there.
 
-from datetime import datetime
 import os
 import sys
 
 import configparser
-from pkg_resources import get_distribution
+import importlib.metadata
+from datetime import datetime
+from packaging.requirements import Requirement
+
 
 try:
     from sphinx_astropy.conf.v1 import *  # noqa
@@ -95,25 +97,17 @@ setup_cfg = configparser.ConfigParser()
 setup_cfg.read(os.path.join(os.path.pardir, 'setup.cfg'))
 __minimum_python_version__ = setup_cfg['options']['python_requires'].replace('>=', '')
 project = u'Astropy'
-astropy_dist = get_distribution(project)
-astropy_req = astropy_dist.requires(extras=('all', ))
 
-
-def minvers(name):
-    if name == 'yaml':
-        name = 'PyYAML'
-    elif name == 'erfa':
-        name = 'pyerfa'
-    for cur_req in astropy_req:
-        if cur_req.name == name:
-            return cur_req.specs[0][1]
-    raise ValueError(f"{name} not among requirements!")
+min_versions = {}
+for line in importlib.metadata.requires('astropy'):
+    req = Requirement(line.split(';')[0])
+    min_versions[req.name] = str(req.specifier)
 
 
 # This is added to the end of RST files - a good place to put substitutions to
 # be used globally.
 rst_epilog += "\n".join(
-    f".. |minimum_{name}_version| replace:: {minvers(name)}"
+    f".. |minimum_{name}_version| replace:: {min_versions[name]}"
     for name in ('numpy', 'erfa', 'scipy', 'yaml', 'asdf', 'matplotlib', 'ipython')) + f"""
 .. |minimum_python_version| replace:: {__minimum_python_version__}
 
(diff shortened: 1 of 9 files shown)


docs/conf.py
```diff
@@ -115,7 +117,7 @@
 # built documents.
 
 # The full version, including alpha/beta/rc tags.
-release = get_distribution(project).version
+release = version(project)
```
astrojuanlu, line 120: :open_mouth: I'm going to use this everywhere now!

pllim:
This touches `modeling` and `io.misc.asdf` entry points, so maybe one of @nden , @perrygreenfield , @jdavies-st , or @eslavich can have a look?

I guess adding it as a required external dependency for just Python 3.7 is acceptable annoyance if it has clear performance gain.

nden:
Sorry I'm just noticing this now. 
Will review shortly.

eslavich (approved):
This looks great to me, seems like we should adopt importlib.metadata in asdf as well.  Thanks for showing us the way!

nden:
@saimin Thanks! Feel free to merge after a rebase.

embray:
:+1: :confetti_ball: 

Python packaging has come a long way since this was last worked on.

pllim:
Just to be sure, can you please rebase? All the CI should be green now.

pllim:
Also, since this introduces a new required dependency, it needs a change log. Alas, the towncrier check doesn't work right now, so you won't get reminded in the status checks.

embray:
FWIW the importlib-metadata dependency is just a backport for Python 3.7, but I agree it merits a changelog mention.

saimn:
All green now :+1: and I added the changelog. I also had to rename the erfa/yaml imports in `conf.py` to pyerfa/pyyaml (the package names), not sure how that was working before.

docs/conf.py
```diff
@@ -25,12 +25,17 @@
...
 # See sphinx_astropy.conf for which values are set there.
 
-from datetime import datetime
 import os
 import sys
 
 import configparser
-from pkg_resources import get_distribution
+from datetime import datetime
+from packaging.requirements import Requirement
```
pllim, line 33: Is `packaging` a new dependency too?

docs/conf.py
```diff
@@ -25,12 +25,17 @@
 # be accessible, and the documentation will not build correctly.
 # See sphinx_astropy.conf for which values are set there.
 
-from datetime import datetime
 import os
 import sys
 
 import configparser
-from pkg_resources import get_distribution
+from datetime import datetime
```
pllim, line 32: Not sure why you moved `datetime`. Shouldn't it be grouped with stdlib imports?

docs/conf.py
saimn: We are already using it but I see now that it's only for test modules. And as it is a dependency of pytest it's not listed explicitely. It's also a dependency for Sphinx, so it's similar. I'm fine to add it explicitely to `docs` and `tests` if you think that's useful.

```
❯ pip show packaging
...
Required-by: tox, Sphinx, pytest, jupyterlab, jupyterlab-server, jupyterlab-git, jupyter-packaging, build, bokeh, bleach, pytest-doctestplus
```

docs/conf.py
saimn: I don't remember, I guess I was using isort but not sure why `datetime` moved here since indeed it's from stdlib.

saimn: And btw isort is broken with out config:  #11496

docs/conf.py
pllim: Naw, it's okay. It is a can of worms I don't want to open (listing deps of deps).

saimn: Well if we use them I think we should list them explicitly. Hypothetically Sphinx or pytest may stop using `packaging` in the future. But that can be for another fight ;)

pllim (approved):
LGTM. Thanks!

docs/conf.py
pllim: Oh, ops. I merged this PR. If you feel strongly about adding `packaging`, please open a follow up PR. I am neutral on this. Thanks!

PR #11098, opened 2020-11-27
Remove use of OrderedDict from astropy.modeling

embray:
Part of #10888.

### DIFF
--- a/astropy/modeling/core.py
+++ b/astropy/modeling/core.py
@@ -22,7 +22,7 @@ import operator
 import types
 import warnings
 
-from collections import defaultdict, OrderedDict, deque
+from collections import defaultdict, deque
 from inspect import signature
 from itertools import chain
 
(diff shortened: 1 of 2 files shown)


astropy/modeling/core.py
```diff
@@ -69,9 +69,6 @@ class _ModelMeta(abc.ABCMeta):
     Currently just handles auto-generating the param_names list based on
     Parameter descriptors declared at the class-level of Model subclasses.
     """
-    @classmethod
-    def __prepare__(mcls, name, bases):
-        return OrderedDict()
```
nden, line 72: @nstarman  in #11017 modifies this method to add `**kwargs`. My understanding is that Python returns a dict by default if this method is not defined, so removing it here is fine. #11017 will need to be rebased on top of this before merging.

nden (approved):
lgtm

astropy/modeling/core.py
embray: Thanks for pointing that out.  I hand't seen that one.  Indeed, with Python 3.6 this `__prepare__` is no longer needed.

PR #11099, opened 2020-11-27
[#11094] Deprecate and remove use of OrderedDescriptor

### LINKED ISSUE #11094: Simplify OrderedDescriptor[Container]
The work to remove metaclasses in favour of `__init_subclass__` at #11090 reminded me that @embray [had mentioned](https://groups.google.com/g/astropy-dev/c/gpRTvftTe_E/m/tIFi82ToAwAJ) that this would become possible, and that similar changes would also allow one to simplify `OrderedDecriptor`:
```
Once Python 3.6 is out I would encourage looking for bits in Astropy
that can be simplified by these PEPs. The most obvious that comes to
mind is the OrderedDescriptor class, and its associated
OrderedDescriptorContainer class (that a class must subclass in order
for OrderedDescriptors to work on it). PEP 520 makes
OrderedDescriptor trivial, and obviates the need for it.
```

This is a reminder for us to look at this.

**Details**

The combination of `OrderedDescriptor` and `OrderedDescriptorContainer` provided two functionalities:

It allowed setting a "name" attribute on each `OrderedDescriptor` when the class is defined.  This can be replaced with the `__set_name__` magic method introduced in Python 3.6 by [PEP 487](https://www.python.org/dev/peps/pep-0487/).  For example, given:

```python
class ExampleDecorator(OrderedDescriptor):
    _name_attribute_ = 'name'
```

this can now be written:

```python
class ExampleDescriptor:
    def __set_name__(self, owner, name):
        self.name = name
```

when defining a class using this descriptor, e.g. like

```python
class MyClass:
    descr1 = ExampleDescriptor()
    descr2 = ExampleDescriptor()
```

at class definition time `__set_name__` will be called on each `ExampleDescriptor` with the name it was assigned to in the class body, resulting in:

```python
>>> MyClass.descr1.name
'descr1'
>>> MyClass.descr2.name
'descr2'
```

The other functionality of `OrderedDescriptor` + `OrderedDescriptorContainer` is that it would set an attribute on the class containing an `OrderedDict` of all instances of some `OrderedDescriptor` subclass.  For example, given

```python
class ExampleDescriptor(OrderedDescriptor):
    _class_attribute_ = '_examples_'
```

this can now be implemented in the class which contains the `ExampleDescriptor`s (without using `OrderedDescriptor` or `OrderedDescriptorContainer` at all) by implementing `__init_subclass__` on a base class:

```python
class MyBaseClass:
    _examples_ = {}

    def __init_subclass__(cls):
        cls._examples_ = {k, v for k, v in cls.__dict__.items()
                                      if isinstance(v, ExampleDescriptor)}
```

Now use this base class and put some `ExampleDescriptor`s in the class body:

```python
class MyClass(MyBaseClass):
    descr1 = ExampleDescriptor()
    descr2 = ExampleDescriptor()
```

Resulting in:

```python
>>> MyClass._examples_
{'descr1': <__main__.ExampleDescriptor at 0x7f4e2bfb44e0>,
 'descr2': <__main__.ExampleDescriptor at 0x7f4e2bfb4ef0>}
```

Finally, a more advanced feature of `OrderedDescriptorContainer` was if you set `_inherit_descriptors_ = True` in classes which used it.  This would cause the descriptors in its `_class_attribute_` to be inherited from base classes.

If this functionality is needed it can also be achieved with a slightly more advanced `__init_subclass__`:

```python
class MyBaseClass:
    _examples_ = {}

    def __init_subclass__(cls):
        for basecls in cls.__mro__:
            if hasattr(basecls, '_examples_') and isinstance(basecls._examples_, dict):
                cls._examples_ = dict(basecls._examples_)
                break
        else:
            cls._examples_ = {}

        cls._examples_.update(
            {k, v for k, v in cls.__dict__.items()
             if isinstance(v, ExampleDescriptor)})
```

Then given

```python
class MyClass(MyBaseClass):
    descr1 = ExampleDescriptor()
    descr2 = ExampleDescriptor()
```

and 

```python
class MySubClass(MyClass):
    descr3 = ExampleDescriptor()
```

The result is:

```python
>>> MySubClass._examples_
{'descr1': <__main__.ExampleDescriptor at 0x7f4e2bfb44e0>,
 'descr2': <__main__.ExampleDescriptor at 0x7f4e2bfb4ef0>,
 'descr3': <__main__.ExampleDescriptor at 0x7f4e2bf9b550>}
```

embray:
Fixes #11094.  Builds on top of #11098 but they could be separated I think.

~~Note, this does not attempt to remove any metaclasses in favor of `__init_subclass__` as discussed somewhat in #11090.  That should be handled elsewhere for coordinates~~

Completes the work of #11090 to remove `FrameMeta` since it is no longer used.

* [x] Should probably have a changelog entry.
* [x] Fix the case brought up by @ayshih https://github.com/astropy/astropy/pull/11099#issuecomment-735829157

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -100,6 +100,10 @@ astropy.utils
 
 - Add new ``utils.parsing`` module to with helper wrappers around ``ply``. [#11227]
 
+- Deprecated ``astropy.utils.OrderedDescriptor`` and
+  ``astropy.utils.OrderedDescriptorContainer``, as new features in Python 3
+  make their use less compelling. [#11094, #11099]
+
 astropy.visualization
 ^^^^^^^^^^^^^^^^^^^^^
 
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -21,8 +21,7 @@ from astropy.utils.compat.misc import override__dir__
 from astropy.utils.decorators import lazyproperty, format_doc
 from astropy.utils.exceptions import AstropyWarning, AstropyDeprecationWarning
 from astropy import units as u
-from astropy.utils import (OrderedDescriptorContainer, ShapedLikeNDArray,
-                           check_broadcast)
+from astropy.utils import ShapedLikeNDArray, check_broadcast
 from .transformations import TransformGraph
 from . import representation as r
 from .angles import Angle
@@ -197,13 +196,12 @@ _components = """
 # TODO: This is only needed so that BaseCoordinateFrame can have
 # OrderedDescriptorContainer as a metaclass. Trying to use
 # metaclass=OrderedDescriptorContainer directly causes a metaclass conflict.
-class FrameMeta(OrderedDescriptorContainer, abc.ABCMeta):
+class FrameMeta(abc.ABCMeta):
     pass
 
 
 @format_doc(base_doc, components=_components, footer="")
-class BaseCoordinateFrame(ShapedLikeNDArray,
-                          metaclass=FrameMeta):
+class BaseCoordinateFrame(ShapedLikeNDArray):
     """
     The base class for coordinate frames.
 
@@ -251,8 +249,6 @@ class BaseCoordinateFrame(ShapedLikeNDArray,
     # attributes.
     frame_specific_representation_info = {}
 
-    _inherit_descriptors_ = (Attribute,)
-
     frame_attributes = {}
     # Default empty frame_attributes dict
 
@@ -262,6 +258,7 @@ class BaseCoordinateFrame(ShapedLikeNDArray,
         default_repr = getattr(cls, 'default_representation', None)
         default_diff = getattr(cls, 'default_differential', None)
         repr_info = getattr(cls, 'frame_specific_representation_info', None)
+        frame_attrs = getattr(cls, 'frame_attributes', None)
 
         # Then, to make sure this works for subclasses-of-subclasses, we also
         # have to check for cases where the attribute names have already been
@@ -285,6 +282,19 @@ class BaseCoordinateFrame(ShapedLikeNDArray,
         cls._create_readonly_property('default_differential', default_diff)
         cls._create_readonly_property('frame_specific_representation_info',
                                       copy.deepcopy(repr_info))
+
+        # Set the frame attributes
+        # TODO: Should this be made to use readonly_prop_factory as well or
+        # would it be inconvenient for getting the frame_attributes from
+        # classes?
+        if frame_attrs is None:
+            frame_attrs = {}
+        else:
+            frame_attrs = frame_attrs.copy()
+
+        frame_attrs.update({k: v for k, v in cls.__dict__.items()
+                            if isinstance(v, Attribute)})
+        cls.frame_attributes = frame_attrs
 
         # Deal with setting the name of the frame:
         if not hasattr(cls, 'name'):
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -284,8 +284,14 @@ class BaseCoordinateFrame(ShapedLikeNDArray):
         else:
             frame_attrs = frame_attrs.copy()
 
-        frame_attrs.update({k: v for k, v in cls.__dict__.items()
-                            if isinstance(v, Attribute)})
+        for basecls in (cls,) + cls.__bases__:
+            if not issubclass(basecls, BaseCoordinateFrame):
+                continue
+
+            for k, v in basecls.__dict__.items():
+                if isinstance(v, Attribute):
+                    frame_attrs.setdefault(k, v)
+
         cls.frame_attributes = frame_attrs
 
         # Deal with setting the name of the frame:
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -7,7 +7,6 @@ classes.
 
 
 # Standard library
-import abc
 import copy
 import inspect
 from collections import namedtuple, defaultdict
@@ -191,13 +190,6 @@ _components = """
     *args, **kwargs
         Coordinate components, with names that depend on the subclass.
 """
-
-
-# TODO: This is only needed so that BaseCoordinateFrame can have
-# OrderedDescriptorContainer as a metaclass. Trying to use
-# metaclass=OrderedDescriptorContainer directly causes a metaclass conflict.
-class FrameMeta(abc.ABCMeta):
-    pass
 
 
 @format_doc(base_doc, components=_components, footer="")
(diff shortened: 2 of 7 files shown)


embray:
Looks like I forgot to update my master branch because I didn't have #11090 incorporated yet.  I'll rebase.

mhvk (commented):
Looks all good to me - nice to have things simpler and less hidden (well, really there are hidden inside python now, of course...).

Does indeed need an entry under "API changes"

adrn:
Great - thanks @embray! Agreed that this looks good and just requires a changelog entry.

ayshih:
Not a huge deal, but I believe this PR results in incorrect behavior when a frame class subclasses two different subclasses of `BaseCoordinateFrame`.  I'll try to craft a minimal example.

ayshih:
Here you go:
```python
>>> from astropy.coordinates import BaseCoordinateFrame, Attribute

>>> class Frame1(BaseCoordinateFrame):
>>>     attr1 = Attribute()
>>> class Frame2(BaseCoordinateFrame):
>>>     attr2 = Attribute()

>>> class Frame3(Frame1, Frame2):
>>>     pass
>>> Frame3.frame_attributes
{'attr1': <astropy.coordinates.attributes.Attribute at 0x1bfded96248>}
```
Note that `attr2` is not picked up.  I presume this is because `__init_subclass__()` – which is intended to accumulate all of the frame attributes – isn't called multiple times.

For comparison, on `master`:
```python
>>> Frame3.frame_attributes
Frame3.frame_attributes
Frame3.frame_attributes
OrderedDict([('attr1',
              <astropy.coordinates.attributes.Attribute at 0x29bb9e74348>),
             ('attr2',
              <astropy.coordinates.attributes.Attribute at 0x29bbad9dbc8>)])
```

embray:
(Oops, accidentally pushed commits for a different branch)

embray:
Thanks for the example @ayshih.  I had thought about this case, but since it didn't happen to come up in any of the Astropy tests I assumed it maybe wasn't a usecase that was needed.  Maybe it just wasn't tested because this case was tested implicitly by the use of OrderedDescriptor.  But I completely agree with you on how this should work.  I'll fix it and add a test.

astrojuanlu:
Out of curiosity @ayshih, what would be the use case of subclassing two different frames?

astrojuanlu:
(Also, gh-11098 was merged, so this could be rebased)

embray:
@astrojuanlu thanks for the heads up, will do.

I'm also curious about @ayshih's use case of multiple inheritance of frames, but nevertheless they're right that it's a regression over the previous version that should be fixed.  I'll try to finish that now.

embray:
Rebased and fixed the multiple inheritance case.

embray:
Fixed the style check and added changelog entry.

embray:
In the changelog I referenced the original issue number, not the PR number.  Is that what we're doing now?

CHANGES.rst
```diff
@@ -98,6 +98,10 @@ astropy.utils
   function ``_find_pkg_data_path``) for obtaining file paths without checking if the
   file/directory exists, as long as the package and module do. [#11006]
 
+- Deprecated ``astropy.utils.OrderedDescriptor`` and
+  ``astropy.utils.OrderedDescriptorContainer``, as new features in Python 3
+  make their use less compelling. [#11094]
```
pllim, line 103:
```suggestion
  make their use less compelling. [#11094, #11099]
```

embray: Ah ok, I can do that.

adrn (approved):
Looks good to me! (apart from the changelog edit by @pllim). Very happy to be rid of the metaclass stuff in coordinates after this 😄.

astropy/coordinates/baseframe.py
```diff
@@ -286,6 +275,25 @@ def __init_subclass__(cls, **kwargs):
         cls._create_readonly_property('frame_specific_representation_info',
                                       copy.deepcopy(repr_info))
 
+        # Set the frame attributes
+        # TODO: Should this be made to use readonly_prop_factory as well or
```
adrn, line 279: The current functionality allows overriding `.frame_attributes`. Not that it's a good idea to allow this, but I think it's fine to leave functionality as is for now and reconsider this in a separate issue/PR.

pllim:
There are failures in `modeling` though.

adrn:
> There are failures in `modeling` though.

I don't think those failures are caused by this PR!

pllim:
You are right! Seems to be from  #11108 that just got merged before this was rebased.

pllim:
Wait... Ugh, I think it is due to Scipy 1.6.0 that just got released.

mhvk:
Just a :fireworks: from the sidelines (will light some real ones later...)

ayshih:
@astrojuanlu 
> Out of curiosity @ayshih, what would be the use case of subclassing two different frames?

@embray 
> I'm also curious about @ayshih's use case of multiple inheritance of frames

In principle, this could be useful for using class inheritance to manage common frame attributes.  For example, there could be a base class that adds the `equinox` frame attribute rather than defining that attribute separately within each frame definition (and possibly introducing an inconsistency).  It'd be particularly useful if there are any convenience methods that'd be useful to any frame with an `equinox` frame attribute, but not useful to any other frame.

In practice, Astropy *already* has cases of subclassing two different frames, although this may be a design oversight.  Any frame created by `SkyOffsetFrame` subclasses both `SkyOffsetFrame` (which has the `rotation` and `origin` frame attributes) and the frame of the `origin` argument (which has its normal frame attributes, e.g., `obstime`).  For example:
```python
>>> import astropy.units as u
>>> from astropy.coordinates import SkyCoord, SkyOffsetFrame

>>> sc = SkyCoord(0*u.deg, 0*u.deg, frame='hcrs')
>>> so = SkyOffsetFrame(origin=sc)
>>> so
<SkyOffsetHCRS Frame (obstime=J2000.000, rotation=0.0 deg, origin=<HCRS Coordinate (obstime=J2000.000): (ra, dec) in deg
    (0., 0.)>)>

>>> so.__class__.__bases__
(astropy.coordinates.builtin_frames.skyoffset.SkyOffsetFrame,
 astropy.coordinates.builtin_frames.hcrs.HCRS)
>>> so.frame_attributes
OrderedDict([('obstime',
              <astropy.coordinates.attributes.TimeAttribute at 0x18959ce1e88>),
             ('rotation',
              <astropy.coordinates.attributes.QuantityAttribute at 0x18959cfb708>),
             ('origin',
              <astropy.coordinates.attributes.CoordinateAttribute at 0x18959df5b88>)])
```

ayshih:
Re `SkyOffsetFrame`, I say "design oversight" because for a similar frame in SunPy (`RotatedSunFrame`), I have recently decided to remove the inheritance of the frame class of the supplied coordinate.  `SkyOffsetFrame` may not be able to make the same decision, of course.

embray:
Fixed changelog conflict.

mhvk:
OK, just checked again and this all looks great. Let's get it in!

Issue #11096, opened 2020-11-27
Problem with base classes for coordinate-frame classes needing to have `name`


ayshih:
A recent code change in #11090 is causing errors in SunPy:

https://github.com/astropy/astropy/blob/87df980318eac43850a3add1e5f32286dad9d723/astropy/coordinates/baseframe.py#L289-L297

We have a base class for a frame that does not have a `name` attribute, so the `base.name` access produces an error.  It's easy enough to add a dummy `name` attribute to "fix" this on our side, but presumably the Astropy code needs some protection?

adrn:
Ah, yes, an oversight on my part - oops! I'll submit a follow-up PR now, and add a regression test.

adrn:
I'm assuming this is `RotatedSunFrame`? I'm a little confused about why it misses the first `if` clause then, if it doesn't have a defined `name` attribute?

ayshih:
The error does not occur directly with our `RotatedSunFrame` metaframe, but rather when it tries to dynamically create a coordinate frame.  For example, `RotatedSunFrame` can create `RotatedSunHeliographicStonyhurst`, which has two base classes: the coordinate frame `HeliographicStonyhurst` and the non-coordinate-frame `RotatedSunFrame`.  The problem is that the latter doesn't have the `name` attribute.

Astropy's `SkyOffsetFrame` doesn't have an issue because that metaframe itself subclasses `BaseCoordinateFrame`, so will automatically have `.name`.  I found that I needed `RotatedSunFrame` to *not* subclass `BaseCoordinateFrame` to work properly, although I don't currently recall why.

adrn:
[...]

PR #11101, opened 2020-11-28
Follow-up to #11090 to address #11096

adrn:
Soon after merging #11090, @ayshih noticed that sunpy tests were failing, as reported in #11096, because the new `BaseCoordinateFrame.__init_subclass__` expected any coordinate frame parent class to have a `.name` attribute. This is a simple follow-up to use `getattr()` instead.

### DIFF
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -338,7 +338,8 @@ class BaseCoordinateFrame(ShapedLikeNDArray,
 
         if kwargs:
             raise TypeError(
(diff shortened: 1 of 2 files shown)


adrn:
This is a follow-up bug fix to #11090, so I think it doesn't need a changelog entry?

mhvk (approved):
Looks good to me, but perhaps best if @ayshih checks that it solves the problem.

ayshih:
:+1: Yup, this fixes the issue

mhvk:
Great, will merge now! Thanks, @adrn and @ayshih!

PR #11103, opened 2020-11-30
Update precision test of AltAz computation.

mhvk:
Follow-up of https://github.com/astropy/astropy/pull/11073#issuecomment-735486271, where @mkbrewer found that astropy and his code are now consistent down to well below 100 nas. Updated the test accordingly, fixing a minor precision-reducing bug in the process.

### DIFF
--- a/astropy/coordinates/tests/test_intermediate_transformations.py
+++ b/astropy/coordinates/tests/test_intermediate_transformations.py
@@ -681,8 +681,7 @@ def test_straight_overhead():
 
 @pytest.mark.remote_data
 def test_aa_high_precision():
-    """
-    These tests are provided by @mkbrewer - see issue #10356.
+    """These tests are provided by @mkbrewer - see issue #10356.
 
     The code that produces them agrees very well (<0.5 mas) with SkyField once Polar motion
     is turned off, but SkyField does not include polar motion, so a comparison to Skyfield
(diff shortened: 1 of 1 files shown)


StuartLittlefair:
Hi @mhvk 

Thanks for this update. I'm wondering if we are doing the right thing by tightening the precision so much. The two codes are both implementations of SOFA algorithms that include various treatments of precession etc. They agree very precisely which is wonderful, but should these tests tie us to these algorithmic choices?

Perhaps I haven't explained myself very well?

astropy/coordinates/tests/test_intermediate_transformations.py
```diff
@@ -690,22 +690,29 @@ def test_aa_high_precision():
 
     The absence of polar motion within Skyfield and the disagreement between Skyfield and Horizons
     make high precision comparisons to those codes difficult.
+
+    Updated 2020-11-29, after the comparison between codes became even better,
+    down to 100 nas (reflecting consistency in approach at least).
```
mhvk, line 695: @StuartLittlefair - your comment is why I added this to the docstring. I think it is OK to tie ourselves this tightly, so that if an error starts happening, we are forced to look into what changed. Maybe I should make that more explicit for our future selves?

StuartLittlefair: I think so, what I'm hoping to avoid is future us implementing a new precession model from the IAU (for example) and wondering why this test fails and what we should do about it.

mhvk: How about if I change the comment to read as follows (passing by you first to avoid re-running CI needlessly - `[ci skip]` no longer works...)
```
    NOTE: the agreement reflects consistency in approach between two codes,
    not necessarily absolute precision.  If this test starts failing, the
    tolerance can and should be weakened *if* it is clear that the change is
    due to an improvement (e.g., a new IAU precession model).
```

StuartLittlefair: Perfect!

mkbrewer:
Well, my objective here has been to get Astropy up to the full accuracy of the bare SOFA model, which we have now accomplished. There are several improvements that you could make to this model that would invalidate this test, the first of which would be to include the effects of free core nutation provided by the `dX` and `dY` values given in the IERS tables. For a good introduction to these effects see Chapter 5 here:

https://www.iers.org/IERS/EN/Publications/TechnicalNotes/tn36.html

mhvk:
@mkbrewer - yes, good point. I started a new issue about possible further improvements: #11110.

@StuartLittlefair - I pushed the PR with the revised note.

mhvk:
Annoying that the remote-data run fails because of a numpy-dev problem...

mhvk:
But to be clear, the test passed locally and (I think...) also passed before here.

StuartLittlefair:
Looks like a problem reading the IERS table?

mhvk:
The test failed very quickly, with
```
E   RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 80 from C header, got 88 from PyObject
```

mkbrewer:
Should these tests be rerun now that #11116 has been merged?

mhvk:
Rebased to see if tests now pass. @StuartLittlefair - let me know if you have any remaining comments.

StuartLittlefair (approved):

StuartLittlefair:
Thanks @mhvk!

Issue #8091, opened 2018-11-08
Constructing a SphericalRepresentation object with a negative distance is impossible


Cadair:
The explicit kwarg in the error dosen't seem to work.
```python
In [2]: from astropy.coordinates import Distance, SphericalRepresentation; import astropy.units as u                                                                                                                                 
In [3]: SphericalRepresentation(10*u.deg, 10*u.deg, Distance(-100*u.m, allow_negative=True))    
```
```                                                                                                                                     
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/Git/astropy/astropy/coordinates/representation.py in __init__(self, lon, lat, distance, differentials, copy)
   1569             try:
-> 1570                 self._distance = Distance(self._distance, copy=False)
   1571             except ValueError as e:

~/Git/astropy/astropy/coordinates/distances.py in __new__(cls, value, unit, z, cosmology, distmod, parallax, dtype, copy, order, subok, ndmin, allow_negative)
    183         if not allow_negative and np.any(distance.value < 0):
--> 184             raise ValueError("Distance must be >= 0.  Use the argument "
    185                              "'allow_negative=True' to allow negative values.")

ValueError: Distance must be >= 0.  Use the argument 'allow_negative=True' to allow negative values.

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-3-bc5d4ae3542b> in <module>
----> 1 SphericalRepresentation(10*u.deg, 10*u.deg, Distance(-100*u.m, allow_negative=True))

~/Git/astropy/astropy/coordinates/representation.py in __init__(self, lon, lat, distance, differentials, copy)
   1571             except ValueError as e:
   1572                 if e.args[0].startswith('Distance must be >= 0'):
-> 1573                     raise ValueError("Distance must be >= 0. To allow negative "
   1574                                      "distance values, you must explicitly pass"
   1575                                      " in a `Distance` object with the the "

ValueError: Distance must be >= 0. To allow negative distance values, you must explicitly pass in a `Distance` object with the the argument 'allow_negative=True'.
```

Cadair:
@adrn @eteq This is actually only an issue in SunPy's test suite (which was using unphysical numbers) but as far as I can tell it is actually a real bug that should be fixed anyway.

mhvk:
Yes, that's a bug: in the initializer, it should skip the conversion to a `Distance` if the `distance` is already of that class, ie.,
```
[...]

PR #11113, opened 2020-12-01
Allow explicit negative distances in SphericalRepresentation

### LINKED ISSUE #11104: Cannot use negative distance in spherical representation despite error message
```
>>> from astropy.coordinates import SphericalRepresentation
>>> SphericalRepresentation(10 * u.deg, 45 * u.deg, 10 * u.m)
<SphericalRepresentation (lon, lat, distance) in (deg, deg, m)
    (10., 45., 10.)>
>>> SphericalRepresentation(10 * u.deg, 45 * u.deg, -10 * u.m)
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/representation.py", line 1840, in __init__
    self._distance = Distance(self._distance, copy=False)
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/distances.py", line 187, in __new__
    raise ValueError("Distance must be >= 0.  Use the argument "
ValueError: Distance must be >= 0.  Use the argument 'allow_negative=True' to allow negative values.

  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/representation.py", line 1843, in __init__
    raise ValueError("Distance must be >= 0. To allow negative "
ValueError: Distance must be >= 0. To allow negative distance values, you must explicitly pass in a `Distance` object with the the argument 'allow_negative=True'.
>>> from astropy.coordinates import Distance
>>> SphericalRepresentation(10 * u.deg, 45 * u.deg, Distance(-10 * u.m, allow_negative=True))
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/representation.py", line 1840, in __init__
    self._distance = Distance(self._distance, copy=False)
  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/distances.py", line 187, in __new__
    raise ValueError("Distance must be >= 0.  Use the argument "
ValueError: Distance must be >= 0.  Use the argument 'allow_negative=True' to allow negative values.

  File "/home/juanlu/Personal/Astropy/astropy/astropy/coordinates/representation.py", line 1843, in __init__
    raise ValueError("Distance must be >= 0. To allow negative "
ValueError: Distance must be >= 0. To allow negative distance values, you must explicitly pass in a `Distance` object with the the argument 'allow_negative=True'.
```

mhvk:
Fixes #11104, #8091

The problem was that any input `Distance` was turned into a `Quantity` instead of just passed through. It really should pass it through, as it is a subclass, so this PR allows it (which is what makes it an API change). But to fix that, a small change had to be made to `Angle.__new__` to allow it to pass on keyword arguments like `subok=True` to `Quantity.__new__`.

Also, it turned out `_combine_xyz` was not safe for some operations with this change, but since that private function that was no longer needed anyway now our minimum numpy version is 1.17, so the first commit just turns that into a simple `np.stack`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -102,6 +102,9 @@ astropy.convolution
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
 
+- For input to representations, subclasses of the class required for a
+  given attribute will now be allowed in. [#11113]
+
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 
@@ -187,6 +190,11 @@ astropy.convolution
 
 astropy.coordinates
 ^^^^^^^^^^^^^^^^^^^
+
+- Allow ``Distance`` instances with negative distance values as input for
+  ``SphericalRepresentation``.  This was always noted as allowed in an
+  exception message when a negative ``Quantity`` with length units was
+  passed in, but was not actually possible to do. [#11113]
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
--- a/astropy/coordinates/angles.py
+++ b/astropy/coordinates/angles.py
@@ -108,7 +108,7 @@ class Angle(u.SpecificTypeQuantity):
     _equivalent_unit = u.radian
     _include_easy_conversion_members = True
 
-    def __new__(cls, angle, unit=None, dtype=None, copy=True):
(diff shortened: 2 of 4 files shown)


astrojuanlu (approved):
LGTM!

mhvk:
Approved a few days ago, so I'll go ahead and merge.

PR #11115, opened 2020-12-01
Mention type size for floating point precision

### LINKED ISSUE #11071: suggestion to include simpler/ common language to refer to a format type for FITS table file creation
I suggest that in this documentation, https://docs.astropy.org/en/stable/io/fits/usage/table.html, for FITS format code descriptions to add the "float32,float64" explicitly somewhere in the docs, since "single precision.... " and "double precision..." are less commonly known to refer to those data types in tables and less obfuscating terminology for the otherwise simple concept.
The reason for this suggestion is to make things more accessible and not add difficulty with less commonly used technical terminology.
For example,
"note that single precision floating point is equivalent to float32 and double precision floating point is equivalent to float64" or something like that

saimn:
Fix  #11071.

### DIFF
--- a/docs/io/fits/usage/table.rst
+++ b/docs/io/fits/usage/table.rst
@@ -290,8 +290,8 @@ name and format. Here is a summary of all allowed formats for a binary table:
     J                        32-bit integer                  4
     K                        64-bit integer                  8
     A                        character                       1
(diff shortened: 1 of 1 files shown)


saimn:
(Cancelled CI as it's a documentation only change)

PR #11117, opened 2020-12-01
Remove dead code (from Python 2)

saimn:
This code is never reached because of the
```
        if not isinstance(value,
                          (str, int, float, complex, bool, Undefined,
                           np.floating, np.integer, np.complexfloating,
                           np.bool_)):
```
above.


mhvk (approved):
Yes, nice! (I double-checked your logic just in case)

saimn:
Thanks for the check @mhvk :)

pllim:
Belated 👍 . Good catch!

PR #11121, opened 2020-12-02
Fix link in io.ascii RDB docstring

taldcroft:
### Description

This pull request is to fix a broken / dead link.  I'm not sure what happened to the original, but this one is the next best thing.

Fixes #11112

PR #11116, opened 2020-12-01
Ignore runtimewarning about ndarray size change.

mhvk:
I think these warnings, like for the dtype, can be ignored. 

Note: Only merge if the `*dev` job passes again (it does locally, but one never knows...)

EDIT: Fix #11119

PR #11089, opened 2020-11-25
TST: Set CI in Azure Pipelines

pllim:
### Description

This pull request is to allow Azure Pipelines to skip flaky tests that are already skipped in the CI environment. My Azure-fu isn't strong, please check syntax carefully.

I'll cancel Actions.

Follow up of https://github.com/astropy/astropy/issues/10914#issuecomment-733856067

PR #11122, opened 2020-12-02
TST: Sync Actions between 4.2.x and master branch

pllim:
### Description

This pull request is to address the 4.2.x release branch for #11050 .

#### Expected failures:

* linkcheck -- Fix at #11121 (job later removed)
* remote-data/numpy-dev -- Temporarily ignored at #11116 
* Azure -- Fix at #11089
* ppc -- Known issues #11025, #11026, and #11027


pllim:
Not sure why it says `Running tests with Astropy version 4.3.dev4+g4a54fc7d4.` Maybe `setuptools_scm` is just confused. 🤷

saimn:
setuptools_scm guesses the next version number from the previous tag, so the version after 4.2 is 4.3, not 4.2.1

.github/workflows/ci_cron_weekly.yml
```diff
@@ -0,0 +1,118 @@
...
+          # releases.
+          #- name: Python 3.7 with dev version of key dependencies
+          #  os: ubuntu-latest
+          #  python: 3.7
+          #  toxenv: py37-test-devdeps
+
+          - name: Documentation link check
+            os: ubuntu-latest
+            python: 3.7
+            toxenv: linkcheck
```
pllim, line 38: Maybe we don't need linkcheck for release branches either?

Cadair:
> setuptools_scm guesses the next version number from the previous tag, so the version after 4.2 is 4.3, not 4.2.1

one of many reasons one shouldn't drop digits from versions when they are 0...

.github/workflows/ci_cron_weekly.yml
saimn: I think we can remove it.

.github/workflows/ci_cron_daily.yml
```diff
@@ -0,0 +1,44 @@
...
+    if: github.repository == 'astropy/astropy'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+
+          - name: Bundling with pyinstaller
+            os: ubuntu-latest
+            python: 3.8
+            toxenv: pyinstaller
```
saimn, line 27: Maybe move this to the main workflow as well ? It seems useless to run it daily as there are not commits every day pushed to those release branches.

pllim: All the cron files are actually push events now. I added comments to explain that I keep the files as-is for ease of backporting (in case we do a 4.2.x release).

.github/workflows/ci_cron_weekly.yml
pllim: It is commented out.

pllim:
I cancelled Actions for the commit to remove only `.travis.yml`. See the CI results for the commit before that one.

saimn (approved):
Looks good! Maybe the azure failure deserves a new issue ?

pllim:
I am pretty confident that Azure failure will go away once #11089 is merged, I just need someone to review it to make sure `CI` is defined correctly as per Azure syntax.

pllim:
Let's get this in then. Thanks for the review!

mhvk:
Just noticed that with this one we have exactly 10000 closed issues!

pllim:
I also snatched #11111 . _Pats self on back._ 😹

PR #11124, opened 2020-12-03
Correct docs for number of terms in 2D Polynomials

mcara:
Docstring for `Polynomial2D` says that number of coefficients is:
```
degree: int
    highest power of the polynomial, the number of terms is degree+1
```
which is wrong. This PR corrects documentation for `degree`.

### DIFF
--- a/astropy/modeling/polynomial.py
+++ b/astropy/modeling/polynomial.py
@@ -969,7 +969,7 @@ class Polynomial1D(_PolyDomainWindow1D):
 
 
 class Polynomial2D(PolynomialModel):
-    """
+    r"""
     2D Polynomial  model.
 
     Represents a general polynomial of degree n:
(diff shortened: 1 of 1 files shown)


astropy/modeling/polynomial.py
```diff
@@ -985,8 +985,10 @@ class Polynomial2D(PolynomialModel):
     Parameters
     ----------
     degree : int
-        highest power of the polynomial,
-        the number of terms is degree+1
+        Polynomial degree: largest sum of exponents (:math:`i + j`) of
+        variables in each monomial term of the form :math:`x^i y^j`. The
+        number of terms in 2D polynomial of degree ``n`` is given by binomial
+        coefficient :math:`C(n + 2, 2) = (n + 2)! / (2!\,n!)`.
```
pllim, line 991: If you want math notation here, I think you need to make the docstring a raw string.

pllim: I think using raw string would fix your error:
```
E   SyntaxError: invalid escape sequence \,
```

mcara:
@pllim Thanks!

nden (approved):

PR #11125, opened 2020-12-03
TST: Add Python 3.9 to test matrix

pllim:
I realized we have no job for Python 3.9 in the main matrix.

Also moved a comment to the correct place.


saimn (approved):
Good catch :+1:

PR #11134, opened 2020-12-04
Set time default location to geocenter for scale changes to/from TDB

### LINKED ISSUE #11132: Incorrect position used in call to erfa.dtdb when location=None
https://github.com/astropy/astropy/blob/4da65447b13483dbc9b2aa7f1c4a5d133773be86/astropy/time/core.py#L1973-L1975

When a `Time()` has location=None, it is supposed to do the TT to TDB conversion without the topocentric term, which corresponds to passing the coordinates of the geocenter to to `erfa.dtdb`. However, this bit of code is setting the location to (0,0,0) in *geodetic* coordinates, which is at the equator on the surface of the Earth!  This should be changed to `from_geocentric` instead of `from_geodetic`.  This was the cause of this issue in PINT https://github.com/nanograv/PINT/issues/246

mhvk:
Fixes #11132 

The existing default seems just a mistake (or I didn't think about this hard enough when I implemented `location` back in 2013...). The only logical choice for no location would seem to be to ignore it and thus assume geocenter, and from #11132 it is clear that this was thought to be the case in PINT, which is probably the only group that really cares at the ~1 us level (and caused what I'm sure was a difficult-to-track discrepancy, sorry!). I think back-porting is the right thing to do here...

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -1460,6 +1460,8 @@ class Time(TimeBase):
         length for geocentric coordinates, or contain a longitude, latitude,
         and an optional height for geodetic coordinates.
         Can be a single location, or one for each input time.
+        If not given, assumed to be the center of the Earth for time scale
+        transformations to and from the solar-system barycenter.
     copy : bool, optional
         Make a copy of the input values
     """
@@ -1971,17 +1973,17 @@ class Time(TimeBase):
             ut = day_frac(njd1 - 0.5, njd2)[1]
 
             if self.location is None:
-                from astropy.coordinates import EarthLocation
-                location = EarthLocation.from_geodetic(0., 0., 0.)
+                # Assume geocentric.
+                self._delta_tdb_tt = erfa.dtdb(jd1, jd2, ut, 0., 0., 0.)
             else:
                 location = self.location
-            # Geodetic params needed for d_tdb_tt()
-            lon = location.lon
-            rxy = np.hypot(location.x, location.y)
-            z = location.z
-            self._delta_tdb_tt = erfa.dtdb(
-                jd1, jd2, ut, lon.to_value(u.radian),
-                rxy.to_value(u.km), z.to_value(u.km))
+                # Geodetic params needed for d_tdb_tt()
+                lon = location.lon
+                rxy = np.hypot(location.x, location.y)
+                z = location.z
+                self._delta_tdb_tt = erfa.dtdb(
+                    jd1, jd2, ut, lon.to_value(u.radian),
+                    rxy.to_value(u.km), z.to_value(u.km))
 
         return self._delta_tdb_tt
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1123,6 +1123,9 @@ astropy.time
 
 - Fix leap second update when using a non english locale. [#11062]
 
+- Fix default assumed location to be the geocenter when transforming times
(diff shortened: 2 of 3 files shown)


paulray:
Thanks for the quick fix @mhvk !

kerrm:
It was a fun one to track down, yes :)

astropy/time/core.py
```diff
@@ -1971,17 +1973,17 @@ def _get_delta_tdb_tt(self, jd1=None, jd2=None):
             ut = day_frac(njd1 - 0.5, njd2)[1]
 
             if self.location is None:
-                from astropy.coordinates import EarthLocation
-                location = EarthLocation.from_geodetic(0., 0., 0.)
+                # Assume geocentric.
+                self._delta_tdb_tt = erfa.dtdb(jd1, jd2, ut, 0., 0., 0.)
```
astrojuanlu, line 1977: I assume this is done instead of `location = EarthLocation.from_geocentric(0., 0., 0.)` to short-circuit the calculation and avoid the extra operations?

mhvk: Yes, indeed. I probably should note that...

paulray: Yes, I like this simplification

astrojuanlu (approved):
I am not an `astropy.time` maintainer, but I think this is good to go. Perhaps @mhvk you want to add a comment about the simplification you introduced for the default case, but I don't think it's worth the 40 minutes of CI :)

mhvk:
@taldcroft - want to have a look at this PR, which corrects the behaviour for changes to/from TDB for `location=None`?

taldcroft (approved):
Looks good to me.  Thanks @astrojuanlu for the review as well.

pllim:
Either this patch has introduced this failure or it is pure coincidence with changes in numpy-dev but it is not obvious from numpy-dev commit history that they did anything to cause this failure:

```
_________________________ [doctest] spectralcoord.rst __________________________
441           redshift=0.0)
442       [115.27577801, 115.28177883, 115.28277896] GHz>
443 
444 The frequencies are very close to the rest frequency of the 12CO J=1-0 molecular line transition,
445 which is 115.2712018 GHz. However, they are not exactly the same, so if the features we see are
446 indeed from 12CO, then they are Doppler shifted compared to what we consider the rest frame of
447 T Tau. We can convert these frequencies to velocities assuming the Doppler shift equation
448 (in this case with the radio convention)::
449 
450     >>> sc_feat_rest.to(u.km / u.s, doppler_convention='radio', doppler_rest=115.27120180 * u.GHz)  # doctest: +FLOAT_CMP +REMOTE_DATA
    @@ -1,3 +1,3 @@
    -<SpectralCoord
    +<SpectralCoord 
        (observer: <ICRS Coordinate: (x, y, z) in m
                       (-1.25867767e+11, -7.48979688e+10, -3.24757657e+10)
    @@ -13,3 +13,3 @@
         doppler_rest=115.2712018 GHz
         doppler_convention=radio)
    -  [-11.90160347, -27.50828539, -30.10939904] km / s>
    +  [-11.90160353, -27.50828545, -30.1093991 ] km / s>

/home/runner/work/astropy/astropy/docs/coordinates/spectralcoord.rst:450: DocTestFailure
```

mhvk:
Arrgg, caught out by `remote-data` tests again...

PR #11135, opened 2020-12-04
TST: Actions cannot have empty tags

pllim:
### Description

I had the Actions YAML open via GitHub edit UI on the browser for another thing and it complained about `tags:` cannot be empty. This should fix it.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -5,6 +5,7 @@ on:
     branches:
     - master
     tags:
+    - '*'
   pull_request:
 
 env:


saimn (approved):
Thanks @pllim  :+1:

pllim:
Somehow this does not apply to the version in v4.2.x, so I re-milestoned this to 4.3. What is in v4.2.x at the time of writing:

https://github.com/astropy/astropy/blob/95f34053fdc3b14ad77e4e40ad08ea11038412e6/.github/workflows/ci_workflows.yml#L3-L9

cc @eteq

PR #11137, opened 2020-12-06
Allow documentation tests in FITS verification page

### LINKED ISSUE #6686: FITS verification documentation is still untested
The page on Verification of `io.fits` (http://docs.astropy.org/en/latest/io/fits/usage/verification.html) wasn't changed in #6664 and many parts don't work at all.

Related: #6010

cmarmo:
### Description
This pull request removes the `.. doctest-skip` directive from the [FITS verification page](http://docs.astropy.org/en/latest/io/fits/usage/verification.html). I could not find a better way to manage warnings than inserting a [doctest.IGNORE_EXCEPTION_DETAIL](https://docs.python.org/3/library/doctest.html#doctest.IGNORE_EXCEPTION_DETAIL).

Fixes #6686.


docs/io/fits/usage/verification.rst
```diff
@@ -246,51 +253,57 @@ Unfixable Cards:
...
     >>> h = fits.PrimaryHDU()  # create a PrimaryHDU
     >>> # Try to add an non-standard FITS keyword 'P.I.' (FITS does no allow
     >>> # '.' in the keyword), if using the update() method - doesn't work!
-    >>> h['P.I.'] = 'Hubble'
-    ValueError: Illegal keyword name 'P.I.'
+    >>> h.header['P.I.'] = 'Hubble' # doctest: +IGNORE_EXCEPTION_DETAIL
+    Traceback (most recent call last):
+     ...
+    VerifyWarning: Keyword name 'P.I.' contains characters not allowed
+        by the FITS standard; a HIERARCH card will be created.
```
saimn, line 263: Hmm, it's a bit unfortunate that we turn warnings into errors (for unit tests) here, since for users this should give only a warning by default. I see two solutions:
- we could maybe use the `testsetup` and `testcleanup` directives to remove the 'error' filter before running this file,
- or just wrap this line with `warning.catch_warnings` to make sure warnings are not turn into exceptions here.

docs/io/fits/usage/verification.rst
```diff
@@ -246,51 +253,57 @@ Unfixable Cards:
...
     >>> # non-standard FITS file
     >>> h.writeto('pi.fits', output_verify='ignore')
     >>> # Now reading a non-standard FITS file
     >>> # astropy.io.fits is magnanimous in reading non-standard FITS files
     >>> hdus = fits.open('pi.fits')
-    >>> hdus[0].header
+    >>> hdus[0].header # doctest: +IGNORE_EXCEPTION_DETAIL
+    Traceback (most recent call last):
+     ...
+    VerifyWarning: Unfixable error: Illegal keyword name 'P.I. '
```
saimn, line 288: Same issue with warnings.

docs/io/fits/usage/verification.rst
```diff
@@ -337,42 +350,40 @@ Examples
...
-     >>> hdu.writeto('out.fits', checksum=True)
-
-.. doctest-skip::
-
-     >>> # Append a new HDU constructed from array data to the end of
-     >>> # the file existingfile.fits with only the appended HDU
-     >>> # containing both CHECKSUM and DATASUM cards.
-     >>> fits.append('existingfile.fits', data, checksum=True)
+    >>> # Open the file checksum.fits verifying the checksum values for all HDUs
+    >>> fits_tobeverified_filename = \
```
saimn, line 354:
```suggestion
    >>> filename = ...
```
And same below.

saimn (commented):
Thanks for your work on this @cmarmo, very useful since obviously some examples were out of date. I'm not how to best handle the warnings/exception issue (some ideas in the comments). It would be better if the users can see the warnings as they would get them by default.

pllim:
Re: Warnings -- If you want them to show but don't want doctest to fail, use https://github.com/astropy/pytest-doctestplus#ignoring-warnings

`# doctest: +IGNORE_WARNINGS`

saimn:
@pllim - `IGNORE_WARNINGS` is to hide the warning isn't it ? I would see to see it, since that's the result of the FITS verification.

pllim:
Yes, correct that `IGNORE_WARNINGS` does hide it... 🤔

docs/io/fits/usage/verification.rst
```diff
@@ -169,21 +169,23 @@ following keywords:
...
-    Output verification result:
-    'BITPIX' card at the wrong place (card 2). Fixed by moving it to the right
-    place (card 1).
-    >>> hdu.header                 # voila!
-    SIMPLE =                     T / conforms to FITS standard
-    BITPIX =                     8 / array data type
-    NAXIS  =                     0
-
+    >>> from astropy.io import fits
+    >>> fits_tobeverified_filename = \
```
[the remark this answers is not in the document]
cmarmo: Indeed... :)

cmarmo:
Thanks @saimn and @pllim for your suggestions. I have recorded the warnings and printed them: note that one of the warnings is thrown as an exception by default, I've kept the `IGNORE_EXCEPTION_DETAIL` directive for that one. I'm not sure this is the most readable solution...

cmarmo:
I've tried to clean a bit, but now to avoid `ResourceWarning: unclosed file _io.FileIO` warnings all fits files should be closed.

saimn:
@cmarmo : I've hijacked a bit your PR, sorry, trying to find a better solution for warnings. To keep the docs clear and focused on the initial topic, it would be better if we could avoid all the warnings related code. So currently I have factored the catching and printing in a `showwarnings` context manager, which works well. And then I think we could move this to [pytest-doctestplus](https://github.com/astropy/pytest-doctestplus) in a new doctest flag.

cmarmo:
> @cmarmo : I've hijacked a bit your PR, sorry, trying to find a better solution for warnings. To keep the docs clear and focused on the initial topic, it would be better if we could avoid all the warnings related code. So currently I have factored the catching and printing in a `showwarnings` context manager, which works well.

@saimn no problem, this kind of refactoring was above my python skills... always happy to learn... :)

> And then I think we could move this to [pytest-doctestplus](https://github.com/astropy/pytest-doctestplus) in a new doctest flag.

Do you mind if I add a comment in the testsetup block, to document that the function is accessory to the warning management in the example? and that it could one day be replaced by a `SHOW_WARNING` options?

saimn:
> Do you mind if I add a comment in the testsetup block, to document that the function is accessory to the warning management in the example? and that it could one day be replaced by a SHOW_WARNING options?

@cmarmo - I managed to make a PR on doctestplus (https://github.com/astropy/pytest-doctestplus/pull/136), so if you don't mind waiting a bit we can update the PR once it's merged and released ? (hopefully soon!)

cmarmo:
> update the PR once it's merged and released ?

Arrgggh, I should have wait for the release too... :(

saimn:
I released the new version of pytest-doctestplus, and closed/reopened the PR to trigger the actions builds.

saimn (approved):
Seems all good now with the new doctestplus version. Thanks @cmarmo !

cmarmo:
Thank you @saimn .

PR #11139, opened 2020-12-07
Bugfix for user defined projection point in fit_wcs_from_points, issue 10991

### LINKED ISSUE #10991: Issue with fit_wcs_from_points when proj_point is passed as a SkyCoord object
### Description

When calling fit_wcs_from_points() with proj_point=<SkyCoord> object, the following exception is raised:  
```AttributeError: 'str' object has no attribute '_extra_frameattr_names'```

This used to work in Astropy 4.0.

  File "wcs_from_points_bug.py", line 25, in <module>
    projection='TAN'
  File "C:\Users\chioettop\Miniconda3\envs\test1\lib\site-packages\astropy\wcs\utils.py", line 1026, in fit_wcs_from_points
    if (proj_point != 'center') and (type(proj_point) != type(world_coords)):
  File "C:\Users\chioettop\Miniconda3\envs\test1\lib\site-packages\astropy\coordinates\sky_coordinate.py", line 379, in __ne__
    return np.logical_not(self == value)
  File "C:\Users\chioettop\Miniconda3\envs\test1\lib\site-packages\astropy\coordinates\sky_coordinate.py", line 368, in __eq__
    for attr in self._extra_frameattr_names | value._extra_frameattr_names:
AttributeError: 'str' object has no attribute '_extra_frameattr_names'

### Expected behavior

A wcs object should be returned.

### Actual behavior

  File "wcs_from_points_bug.py", line 25, in <module>
    projection='TAN'
  File "C:\Users\chioettop\Miniconda3\envs\test1\lib\site-packages\astropy\wcs\utils.py", line 1026, in fit_wcs_from_points
    if (proj_point != 'center') and (type(proj_point) != type(world_coords)):
  File "C:\Users\chioettop\Miniconda3\envs\test1\lib\site-packages\astropy\coordinates\sky_coordinate.py", line 379, in __ne__
    return np.logical_not(self == value)
  File "C:\Users\chioettop\Miniconda3\envs\test1\lib\site-packages\astropy\coordinates\sky_coordinate.py", line 368, in __eq__
    for attr in self._extra_frameattr_names | value._extra_frameattr_names:
AttributeError: 'str' object has no attribute '_extra_frameattr_names'

### Steps to Reproduce

This script reproduces the error:

```python
import numpy as np
from astropy.wcs.utils import fit_wcs_from_points

xy = np.array([[1766.88276168,  662.96432257,  171.50212526,  120.70924648],
               [1706.69832901, 1788.85480559, 1216.98949653, 1307.41843381]])

world_coords = SkyCoord([(66.3542367 , 22.20000162), (67.15416174, 19.18042906),
                         (65.73375432, 17.54251555), (66.02400512, 17.44413253)],
                        frame="icrs", unit="deg")

proj_point = SkyCoord(64.67514918, 19.63389538,
                      frame="icrs", unit="deg")

fit_wcs_from_points(
    xy=xy,
    world_coords=world_coords,
    proj_point=proj_point,
    projection='TAN'
)
```

### System Details

Windows-10-10.0.19041-SP0
Python 3.7.9 (default, Aug 31 2020, 17:10:11) [MSC v.1916 64 bit (AMD64)]
Numpy 1.19.2
astropy 4.3.dev82+gfeee7e5ab
Scipy 1.5.2

karlwessel:
### Description

This pull request is to address an error when a user defined projection point is passed to `fit_wcs_from_points`.

It adds tests for the issue and fixes them by avoiding comparison of `SkyCoords` with `strings`.

Fixes #10991

### DIFF
--- a/astropy/wcs/tests/test_utils.py
+++ b/astropy/wcs/tests/test_utils.py
@@ -1043,9 +1043,7 @@ def test_pixel_to_pixel_1d():
     assert unbroadcast(x).shape == (10,)
 
 
-@pytest.mark.skipif('not HAS_SCIPY')
-def test_fit_wcs_from_points():
-    header_str_linear = """
+header_str_linear = """
 XTENSION= 'IMAGE   '           / Image extension
 BITPIX  =                  -32 / array data type
 NAXIS   =                    2 / number of array dimensions
@@ -1067,8 +1065,7 @@ CD1_2   =   -0.001042099258152
 CD2_1   =     0.00118144146585
 CD2_2   =   -0.005590816683583
 """
-
-    header_str_sip = """
+header_str_sip = """
 XTENSION= 'IMAGE   '           / Image extension
 BITPIX  =                  -32 / array data type
 NAXIS   =                    2 / number of array dimensions
@@ -1112,8 +1109,7 @@ BP_1_1  =  -1.711876336719E-05
 A_DMAX  =    44.72893589844534
 B_DMAX  =    44.62692873032506
 """
-    # A known header that failed before
-    header_str_prob = """
+header_str_prob = """
 NAXIS   =                    2 / number of array dimensions
 WCSAXES =                    2 / Number of coordinate axes
 CRPIX1  =               1024.5 / Pixel coordinate of reference point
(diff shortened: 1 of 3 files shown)


pllim:
Thanks! Might need a squash.

astropy/wcs/utils.py
```diff
@@ -1023,11 +1023,14 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
         lon, lat = unit_sph.lon.deg, unit_sph.lat.deg
 
     # verify input
-    if (proj_point != 'center') and (type(proj_point) != type(world_coords)):
+    if (type(proj_point) != type(world_coords)) and (proj_point != 'center'):
```
mcara, line 1026: What is the reason for using `type` instead of `isinstance`?

Regardless of the answer, however, I am not sure this is the right place to fix this. Shouldn't the fix actually be done in `SkyCoord.__eq__()`? CC: @eteq

mcara (commented):
I am not sure this is the right fix (in the right place). Maybe the fix should be in `SkyCoord.__eq__()`.

karlwessel:
> I am not sure this is the right fix (in the right place). Maybe the fix should be in `SkyCoord.__eq__()`.

I was thinking the same, however changing the behaviour of `SkyCoord` itself might have a lot of side effects which I can't oversee. Therefore I opted for this simple fix for now.

astropy/wcs/utils.py
karlwessel: > What is the reason for using `type` instead of `isinstance`?

Not sure, that is what was used before and I wanted to keep the changes for the fix minimal. I thought refactoring might be better done in a separate issue/PR.

astropy/wcs/tests/test_utils.py
```diff
@@ -1213,6 +1213,11 @@ def test_fit_wcs_from_points():
     fit_wcs_prob = fit_wcs_from_points((x, y), world_pix_prob,
                                        proj_point='center', sip_degree=None)
 
+    # Fitting with user defined projection point
```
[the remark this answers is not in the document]
cshanahan1: i would suggest maybe just merging this for now - if i understand correctly, wouldn't parameterize be used to add 'projection_point' as an arg to the test function, and iterate over several values? i dont think all the tests that currently use 'center' as the projection point also need to be run with a user defined point. i think this test needs to be split up further too, not just separating out this new addition, and that could be done in a new pr?

cshanahan1 (commented):
this makes sense and looks good for the most part, i just added a few comments and asked for clarification on something.

astropy/wcs/utils.py
```diff
@@ -1023,11 +1023,14 @@ def fit_wcs_from_points(xy, world_coords, proj_point='center',
...
         raise ValueError("proj_point must be set to 'center', or an" +
                          "`~astropy.coordinates.SkyCoord` object with " +
                          "a pair of points.")
-    if proj_point != 'center':
+
+    use_center_as_proj_point = (str(proj_point) == 'center')
+
+    if not use_center_as_proj_point:
         assert proj_point.size == 1
```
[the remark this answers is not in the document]
ceb8: You can get a skycoord with a larger size. For example:
```

a = SkyCoord(["34 23","43 54"], unit='deg')
print(a.size)
2
```

astropy/wcs/utils.py
ceb8: I don't understand why the change in this line, I wouldn't have thought it mattered what order the checks are in since it's an `and` statement. What am I missing?

karlwessel: > I don't understand why the change in this line, I wouldn't have thought it mattered what order the checks are in since it's an `and` statement. What am I missing?

Because of lazy evaluation: If `proj_point` has the same type as the `world_coords` - so probably a `SkyCoord` - the second check `proj_point != 'center'` isn't executed and therefore won't throw the error anymore.

mcara: > > What is the reason for using `type` instead of `isinstance`?
> 
> Not sure, that is what was used before and I wanted to keep the changes for the fix minimal. I thought refactoring might be better done in a separate issue/PR.

I would prefer to improve things, if we can... I was thinking of replacing this line and the next couple of lines with

```
    if isinstance(proj_point, str) and proj_point == 'center':
        use_center_as_proj_point = True
    elif isinstance(proj_point, SkyCoord):
        use_center_as_proj_point = False
        if proj_point.size != 1:
            raise ValueError("'proj_point' must contain exactly one point")
    else:
            raise ValueError("proj_point must be set to 'center', or an"
                             "`~astropy.coordinates.SkyCoord` object with "
                             "a pair of points.")
```

But looking at the rest of the code, I see that there are other instances that need refactoring and so I would agree with you.

mcara (changes requested):
Please, either parametrize the unit test or split it into two different tests. Thanks

karlwessel:
> Please, either parametrize the unit test or split it into two different tests. Thanks

Sorry for taking so long, but now I did both:
- I separated the explicit test for issue #10991 as well as the test case for the CRPIX bounds requirement into different functions.
- I parametrized the `test_fit_wcs_from_points` method to handle the common tests for the different testsets 
  - linear, 
  - sip, 
  - problematic,
  - user defined projection
  - and all of the above for the 360 -> 0 degree wrap around case.

mcara:
This is great! Thank you very much! Now one last thing (from me): can you please squash all commit? Note that in  https://docs.astropy.org/en/stable/development/workflow/development_workflow.html#how-to-squash `astropy` is used as the name of the remote (I personally use `upstream`).

mcara (approved):
Looks good to me.

karlwessel:
Thanks for the quick review and approval. I now squashed all commits.

PR #11142, opened 2020-12-07
Move non-astronomy units to misc.py.

### LINKED ISSUE #10769: Move non-astronomy units out of astrophys.py to new misc.py or so

mnijhuis-tos:
### Description

This pull request addresses the pollution of astrophys.py with miscellaneous units. It moves all non-astrophyiscal units to a new misc.py file.

Fixes #10769

This pull request was done during the astropy hackathon in December 2020. Note to the reviewer: My knowledge of python is limited. I will not be surprised if you encounter beginner mistakes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -249,6 +249,8 @@ astropy.uncertainty
 astropy.units
 ^^^^^^^^^^^^^
 
+- Move non-astronomy units from astrophys.py to a new misc.py file. [#11142]
+
 astropy.utils
 ^^^^^^^^^^^^^
 
--- a/astropy/units/__init__.py
+++ b/astropy/units/__init__.py
@@ -20,6 +20,7 @@ from . import si
 from . import cgs
 from . import astrophys
 from . import photometric
+from . import misc
 from .function import units as function_units
 
 from .si import *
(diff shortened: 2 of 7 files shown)


astrojuanlu (changes requested):
Thanks @mnijhuis-tos for this pull request! The code looks good, and this needs a changelog entry.

mhvk (commented):
Looks good indeed!  A changelog entry is a good suggestion (though really how we store the units is something of an implementation detail, but doesn't hurt to mention it). 

One other thing to add is an entry in `docs/units/index.rst` in the `Reference/API` section - I'd add it just under `.. automodapi:: astropy.units.astrophys`.

mhvk (approved):
Thanks, this looks all ok now!  The failures are unrelated, so I'll go ahead and merge.

PR #11143, opened 2020-12-07
Add TimeUnixPTP to time formats

maxnoe:
### Description

This adds a new time format `TimeUnixPTP` that has epoch `1970-01-01T00:00:00 TAI` as apposed to `1970-01-01T00:00:08 TAI` of the `TimeUnixTAI` format.

This format is used by the Precision Time Protocol IEEE standard and by extension the White Rabbit hardware / software in high energy physics.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,6 +68,9 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- Add new time format ``unix_ptp`` (Precision Time Protocol) which provides 
+  the number of SI seconds since ``1970-01-01T00:00:00 TAI``. [#11143]
+
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
 
--- a/astropy/time/formats.py
+++ b/astropy/time/formats.py
@@ -745,6 +745,52 @@ class TimeUnixTai(TimeUnix):
     epoch_scale = 'tai'
 
 
+class TimeUnixPTP(TimeUnix):
+    """
+    Precision Time Protocol: SI seconds elapsed from 1970-01-01 00:00:00 TAI.
+
+    This convention matches the definition of the Precision Time Protocol,
+    https://en.wikipedia.org/wiki/Precision_Time_Protocol and
+    is also used by the White Rabbit protocol in High Energy Physics:
+    https://white-rabbit.web.cern.ch/
+
+    For more information on this time format definition, see:
+    https://www-acc.gsi.de/wiki/Timing/TimingSystemButisInterface#Time
+
+    This format is basically the same as the ``unix_tai`` with an offset of
+    8 seconds due to the different epoch (1970-01-01 00:00:00 UTC vs. TAI).
+
+    This will generally differ from Unix time by the cumulative integral number
+    of leap seconds since 1970-01-01 UTC plus the initial offset of 8 seconds.
+
+    Caveats:
+    - Before 1972, fractional adjustments to UTC were made, so the difference
+      between ``unix`` and ``unix_ptp`` time is no longer an integer.
+    - Because of the fractional adjustments, to be very precise, ``unix_ptp``
+      is the number of seconds since ``1970-01-01 00:00:00 TAI`` or equivalently
+      ``1969-12-31 23:59:51.999918 UTC``.  The difference between TAI and UTC
+      at that epoch was 8.000082 sec.
+    - On the day of a leap second the difference between ``unix`` and ``unix_ptp``
+      times increases linearly through the day by 1.0.  See also the
+      documentation for the `~astropy.time.TimeUnix` class.
+
+    Examples
+    --------
+
+      >>> from astropy.time import Time
+      >>> t = Time('2020-01-01', scale='utc')
+      >>> t.unix_ptp - t.unix
+      37.0
+
+      >>> t = Time('1970-01-01', scale='utc')
+      >>> t.unix_ptp - t.unix  # doctest: +FLOAT_CMP
+      8.000082
+    """
+    name = 'unix_ptp'
+    epoch_val = '1970-01-01 00:00:00'
+    epoch_scale = 'tai'
+
+
 class TimeCxcSec(TimeFromEpoch):
     """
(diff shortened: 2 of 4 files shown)


astropy/time/formats.py
```diff
@@ -710,7 +710,7 @@ class TimeUnix(TimeFromEpoch):
 
 class TimeUnixTai(TimeUnix):
     """
-    Seconds from 1970-01-01 00:00:08 TAI (see notes), including leap seconds.
+    Counts the number of SI seconds elapsed since 1970-01-01 00:00:08 TAI.
```
taldcroft, line 713: Please revert this unrelated change, the original is consistent in style with the other format class descriptions.

astropy/time/formats.py
```diff
@@ -725,8 +725,9 @@ class TimeUnixTai(TimeUnix):
       ``1969-12-31 23:59:59.999918 UTC``.  The difference between TAI and UTC
       at that epoch was 8.000082 sec.
     - On the day of a leap second the difference between ``unix`` and ``unix_tai``
-      times increases linearly through the day by 1.0.  See also the
-      documentation for the `~astropy.time.TimeUnix` class.
+      times increases linearly through the day by 1.0.
```
taldcroft, line 728: The original here was also fine, please revert.

astropy/time/formats.py
```diff
@@ -745,6 +746,52 @@ class TimeUnixTai(TimeUnix):
     epoch_scale = 'tai'
 
 
+class TimeUnixPTP(TimeUnix):
+    """
+    Counts the number of SI seconds elapsed since 1970-01-01 00:00:00 TAI.
```
taldcroft, line 751:
```suggestion
    Precision Time Protocol:  SI seconds elapsed from 1970-01-01 00:00:00 TAI.
```

astropy/time/formats.py
```diff
@@ -745,6 +746,52 @@ class TimeUnixTai(TimeUnix):
...
+
+    This is basically the same as the ``unix_tai`` with an offset of 8 seconds
+    due to the different epoch (1970-01-01 00:00:00 UTC vs. TAI).
+
+    This will generally differ from Unix time by the cumulative integral number
+    of leap seconds since 1970-01-01 UTC plus the initial offset of 8 seconds.
+
+    This convention matches the definition of the Precision Time Protocol,
```
taldcroft, line 759: Put this paragraph first in the docstring as the fundamental definition of the format, rather than starting with the format just being an 8-second derivative of `unix_tai`.

astropy/time/formats.py
```diff
@@ -745,6 +746,52 @@ class TimeUnixTai(TimeUnix):
...
+    https://en.wikipedia.org/wiki/Precision_Time_Protocol and
+    also used by the White Rabbit protocol in High Energy Physics:
+    https://white-rabbit.web.cern.ch/
+
+    For a definition of the time scale, see:
```
taldcroft, line 764: This is a bit confusing since one doesn't know the context of "the time scale", and this class is about a time format.  Maybe you mean "For more information about White Rabbit Based Clocks and Timestamps see https://www-acc.gsi.de/wiki/Timing/TimingSystemButisInterface#Clocks"?

CHANGES.rst
```diff
@@ -68,6 +68,13 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- Add new time format ``unix_ptp``, that monotonically
+  counts number of seconds (no leap seconds are introduced)
+  since ``1970-01-01T00:00:00 TAI``.
+  This format is similar but different to ``unix_tai`` which uses an epoch
+  of ``1970-01-01T00:00:08 TAI`` and ``unix`` which uses an epoch of
+  ``1970-01-01T00:00:00 UTC`` and which is subject to leap seconds. [#11143]
```
taldcroft, line 76:
```suggestion
- Add new time format ``unix_ptp`` (Precision Time Protocol) which provides 
  the number of SI seconds since ``1970-01-01T00:00:00 TAI``. [#11143]
```

taldcroft (changes requested):
Looks basically good @maxnoe, but I have a number of documentation suggestions.

astropy/time/formats.py
maxnoe: The point is that the formulation `including leap seconds` is wrong.

Monotonic timescales do not have leap seconds, that's the point of them.

Leap seconds are a property of UTC.

astropy/time/formats.py
maxnoe: fixed

astropy/time/formats.py
maxnoe: done

astropy/time/formats.py
maxnoe: I reverted the unrelated docs changes and will open a second PR for those.

astropy/time/formats.py
maxnoe: I reverted the changes here and will add a second PR for those

taldcroft (approved):
Thanks @maxnoe!

taldcroft:
@maxnoe - please see #11245. It turns out we miffed it with the original definition of `unix_tai`. The preferred plan now is to change `unix_tai` to be exactly what you implemented here for `unix_ptp`. 

Are you OK with backing out this PR in favor of just using `unix_tai`?

maxnoe:
I am against changing the defintion of an already released time format. That will really confuse a lot of people I think.

taldcroft:
Can you go over to #11245 and weigh in? The basic thought there is that `unix_tai` as defined was just wrong from the start and that likely nobody is really using it for that reason. (You are a perfect example.)  But you have a very valid point.

maxnoe:
Sure. If it turns out that there is no place where the current definition of `unix_tai` is used, I am all for just calling it a bug and fixing it, at which point `unix_ptp` is just unnecessary.

taldcroft:
It's not definitive, but there is not any application code on GitHub actually using `unix_tai` based searching GitHub for that term.

maxnoe:
Ah, right.  That's indeed weird.

PR #11144, opened 2020-12-07
Extend description of fits scripts

### LINKED ISSUE #3009: Add source information to executable scripts
I discovered that I have a program called "fitscheck" on my computer ($PATH), but I had no idea where it came from. The help message didn’t say. A `which fitscheck` listed `/usr/local/anaconda/bin/fitscheck`, but that didn’t really help. One Google search later, and I see it's part of the Astropy package (among other similar files).

I’d like to suggest that the source of these programs be listed in the `--help` or usage header, something along the lines of:

```
This script is part of the Astropy package.
```

along with a version number. Ideally, URL to the documentation page would be fantastic (http://astropy.readthedocs.org/en/latest/io/fits/usage/scripts.html). This covers fitsheader, fitscheck, and fitsdiff, and I don't know what other executables in Astropy.

jmmaljaars:
### Description

When invoking `fits[check/diff/header/info] --help` from the command line, the Astropy version is printed as well as a link to the documentation.

Let me know if there are any other executable scripts in Astropy that might benefit from these changes, I am really fresh to Astropy.

fixes #3009

### DIFF
--- a/astropy/io/fits/scripts/fitscheck.py
+++ b/astropy/io/fits/scripts/fitscheck.py
@@ -46,15 +46,19 @@ import argparse
 import warnings
 
 from astropy.io import fits
+from astropy import version
 
 log = logging.getLogger('fitscheck')
 
-_DESCRIPTION = """
+_DESCRIPTION = f"""
 e.g. fitscheck example.fits
 
 Verifies and optionally re-writes the CHECKSUM and DATASUM keywords
 for a .fits file.
 Optionally detects and fixes FITS standard compliance problems.
+
+This script is part of the Astropy package, version {version.major}.{version.minor}.{version.bugfix}. See
+https://docs.astropy.org/en/latest/io/fits/usage/scripts.html#module-astropy.io.fits.scripts.fitscheck for further documentation.
 """
 
 
--- a/astropy/io/fits/scripts/fitscheck.py
+++ b/astropy/io/fits/scripts/fitscheck.py
@@ -46,7 +46,7 @@ import argparse
 import warnings
 
 from astropy.io import fits
-from astropy import version
+from astropy import __version__
 
 log = logging.getLogger('fitscheck')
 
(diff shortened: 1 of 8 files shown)


astropy/io/fits/scripts/fitscheck.py
```diff
@@ -46,15 +46,19 @@
...
 
-_DESCRIPTION = """
+_DESCRIPTION = f"""
 e.g. fitscheck example.fits
 
 Verifies and optionally re-writes the CHECKSUM and DATASUM keywords
 for a .fits file.
 Optionally detects and fixes FITS standard compliance problems.
+
+This script is part of the Astropy package, version {version.major}.{version.minor}.{version.bugfix}. See
```
pllim, line 60: Might be simpler if you use `from astropy import __version__`? Not sure if a `try ... except ...` for `ImportError` is necessary or not.

pllim:
Thanks! Does seem like a very useful info to have.

astropy/io/fits/scripts/fitscheck.py
jmmaljaars: Now using `from astropy import __version__`, which indeed is simpler.

saimn:
Adding the documentation URL seems useful, thanks.
For the version it would be better to use the standard `--version` flag (https://docs.python.org/3/library/argparse.html#action), e.g.:
```
parser.add_argument('--version', action='version', version='%(prog)s 2.0')
```

astropy/io/fits/scripts/fitsheader.py
```diff
@@ -401,7 +401,10 @@ def main(args=None):
                      'Optional arguments allow the desired extension(s), '
                      'keyword(s), and output format to be specified. '
                      'Note that in the case of a compressed image, '
-                     'the decompressed header is shown by default.'))
+                     'the decompressed header is shown by default. \n\n'
+                     f'This script is part of the Astropy package, version {__version__}. '
+                     'See https://docs.astropy.org/en/latest/io/fits/usage/scripts.html#module-astropy.io.fits.scripts.fitsheader for further documentation.'),
+                     formatter_class=argparse.RawDescriptionHelpFormatter)
```
saimn, line 407: Would be great if we can improve consistency between the scripts, so please:
- Move the description in a global variable as fitscheck and fitsdiff,
- Use a multi-line string (`""" ... """`)
- Always put this message at the end of the description and split long lines (e.g. after the URL).

Thanks!

jmmaljaars: Thanks for the feedback, should be fixed in commit https://github.com/astropy/astropy/pull/11144/commits/a915758e804fecaf1d884d24ee639b8c7f028d30

saimn (approved):
Looks perfect, thanks @jmmaljaars !

PR #11145, opened 2020-12-07
Use proper precision when format not specified

### LINKED ISSUE #11131: quantity.to_string(precision=<int>) does not produce a string of the quantity with <int> decimals of precision
### Description

The precision argument to the to_string method of a Quantity object doesn't appear to do what it should.

### Expected behavior

I expect quantity.to_string(precision=<int>) to produce a string of the quantity with <int> decimals of precision.

### Actual behavior

quantity.to_string(precision=<int>) produces a string with the full precision of quantity.

### Steps to Reproduce

```
q = 10.12345 * u.km
print("%s"%q.to_string(precision=1))  # the full precision is printed
```

### System Details

Linux-5.4.0-53-generic-x86_64-with-glibc2.29
[GCC 9.3.0]
Numpy 1.17.4
astropy 4.1rc2
Scipy 1.3.3
Matplotlib 3.1.2

jmmaljaars:
### Description

The precision of a `Quantity` was not properly printed in case the precision was specified and `format` was empty. This should be fixed by defining `format_spec` a bit earlier, and using this when parsing the "unadorned" string.

Fixes #11131

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1175,6 +1175,9 @@ astropy.units
 - Ensure ``keepdims`` works for taking ``mean``, ``std``, and ``var`` of
   ``Quantity``. [#11198]
 
+- For ``Quantity.to_string()``, ensure that the precision argument is also
+  used when the format is not latex. [#11145]
+
 astropy.utils
 ^^^^^^^^^^^^^
 
--- a/astropy/units/quantity.py
+++ b/astropy/units/quantity.py
@@ -1169,14 +1169,20 @@ class Quantity(np.ndarray):
         if format not in formats:
             raise ValueError(f"Unknown format '{format}'")
         elif format is None:
-            return f'{self.value}{self._unitstr:s}'
+            if precision is None:
+                # Use default formatting settings
+                return f'{self.value}{self._unitstr:s}'
+            else:
+                # np.array2string properly formats arrays as well as scalars
+                return np.array2string(self.value, precision=precision, floatmode="fixed") + self._unitstr
 
         # else, for the moment we assume format="latex"
 
         # need to do try/finally because "threshold" cannot be overridden
         # with array2string
-        pops = np.get_printoptions()
 
+        # Set the precision if set, otherwise use numpy default
+        pops = np.get_printoptions()
         format_spec = f".{precision if precision is not None else pops['precision']}g"
 
         def float_formatter(value):
--- a/astropy/units/tests/test_quantity.py
+++ b/astropy/units/tests/test_quantity.py
@@ -913,6 +913,10 @@ class TestQuantityDisplay:
         res = 'Quantity as KMS: 150000000000.0 km / s'
         assert f"Quantity as KMS: {qscalar.to_string(unit=u.km / u.s)}" == res
 
+        # With precision set
+        res = 'Quantity as KMS: 1.500e+11 km / s'
+        assert f"Quantity as KMS: {qscalar.to_string(precision=3, unit=u.km / u.s)}" == res
+
         res = r'$1.5 \times 10^{14} \; \mathrm{\frac{m}{s}}$'
         assert qscalar.to_string(format="latex") == res
 


pllim:
Thanks! Would need a test and a change log.

@mhvk or anyone who reviews, feel free to adjust milestone if I set it wrongly.

jmmaljaars:
I converted it to `Draft`, since it doesn't properly handle the case where `self.value` is a `numpy` array.

astropy/units/quantity.py
```diff
@@ -1166,18 +1166,19 @@ def to_string(self, unit=None, precision=None, format=None, subfmt=None):
             },
         }
 
+        # Set the precision, if set, otherwise use numpy default
+        pops = np.get_printoptions()
```
mhvk, line 1170: Yes, it makes sense that this doesn't work with arrays. I fear we'll need to explicitly set one of the numpy formatting arguments (maybe just precision itself? but a different default?), since those will be used if one just expands `{self.value}`. I remember the machinery changed a little while ago, so probably good to start by looking what it presently allows.

jmmaljaars: Should be fixed - while being backwards compatible - by using `np.array2string` in the conditional

```
        if format not in formats:
            raise ValueError(f"Unknown format '{format}'")
        elif format is None:
            if precision is None:
                # Use default formatting settings
                return f'{self.value}{self._unitstr:s}'
            else:
                # np.array2string properly formats arrays as well as scalars
                return f'{np.array2string(self.value, precision=precision, floatmode="fixed")}{self._unitstr:s}'
```

The `CI/Python3.8` job is failing for an unknown reason. Locally, it works perfectly fine. In any case, I will mark this PR as ready.

pllim:
Failure does look transient. I restarted the CI.

jmmaljaars:
Would this PR need a changelog entry?

pllim:
> Does this PR need a changelog entry?

This PR is marked as a bug fix, so yes, please add a change log entry. Thanks!

jmmaljaars:
Done! (sorry that it took a while...)

CHANGES.rst
```diff
@@ -370,6 +370,9 @@ astropy.uncertainty
 astropy.units
 ^^^^^^^^^^^^^
 
+- Use proper precision when printing a ``Quantity`` in case precision format not
```
mhvk, line 373: Maybe "when creating a string from a ``Quantity``"? It affects `to_string()` and `__str__`.

mhvk (commented):
Looks good, thanks so much! Two comments left, both nitpicky...

mhvk:
p.s. Need to move the changelog entry to 4.0.5 to match the milestone...

CHANGES.rst
jmmaljaars: Should be fixed (and migrated to milestone 4.0.5) in https://github.com/astropy/astropy/pull/11145/commits/acb7cc342df621b71899ff96c02843f6c31ff4bd

CHANGES.rst
```diff
@@ -1175,6 +1175,9 @@ astropy.units
 - Ensure ``keepdims`` works for taking ``mean``, ``std``, and ``var`` of
   ``Quantity``. [#11198]
 
+- Use proper precision when creating a string from a ``Quantity`` in case precision format is not
```
mhvk, line 1178: Sorry to have another nitpick, but I think this suggestion of mine doesn't actually convey to a reader what changed. Maybe
```
  - For ``Quantity.to_string()``, ensure that the precision argument is also
    used when the format is not latex. [#11145]
```

mhvk (commented):
Thanks, looks good now, but I did manage to find one more nitpick. Sorry! When you change that, could you also squash the commits? (If not sure how, see https://docs.astropy.org/en/latest/development/workflow/development_workflow.html#how-to-squash -- am also happy to do it for you).

CHANGES.rst
jmmaljaars: Fixed in https://github.com/astropy/astropy/commit/179bf320e769dcc6acba921a762bb83cf56f6313 , also squashed this bug fix into one commit.

mhvk (approved):
Thanks! Merging...

PR #11146, opened 2020-12-07
TST: Prevent weekly cron running on fork

pllim:
### Description

I noticed that some weekly cron jobs still kicked off on my fork. This should disable them.

I am going to cancel CI as this only affects cron job.


saimn (approved):
LGTM, thanks @pllim !

PR #11149, opened 2020-12-08
Astropy#2632 multiple fits tables

### LINKED ISSUE #2632: Save multiple Table objects to one FITS file
I have multiple `astropy.table.Table` objects that I'd like to save to one FITS file.

From the docs (see [here](http://astropy.readthedocs.org/en/latest/io/unified.html#fits)) I couldn't figure out how to do this ... for now I create `BinTableHDU` objects by initialising from `table._data` and put those in a `HDUList` ... but there must be a better way for this common task!?

@taldcroft @embray Can you implement and / or document how to do this? (Apologies if I missed it in the docs.)

stvguest:
### Description

This pull request is to address feature request #2632 on saving multiple tables to one FITS file. The implementation follows the original suggestion by the submitter of the request. The performance of appending multiple tables is comparable to writing the same table to a file multiple times.

Fixes #2632

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,7 @@ astropy.io.fits
 - Check that the SIMPLE card is present when opening a file, to ensure that the
   file is a valid FITS file and raise a better error when opening a non FITS
   one. ``ignore_missing_simple`` can be used to skip this verification. [#10895]
+
 - Added ``append`` keyword to append table objects to an existing FITS file [#2632, #11149]
 
 astropy.io.misc
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,7 +38,7 @@ astropy.io.fits
 - Check that the SIMPLE card is present when opening a file, to ensure that the
   file is a valid FITS file and raise a better error when opening a non FITS
   one. ``ignore_missing_simple`` can be used to skip this verification. [#10895]
-- Added ``append`` keyword to append table objects to an existing FITS file [#2632]
+- Added ``append`` keyword to append table objects to an existing FITS file [#2632, #11149]
 
 astropy.io.misc
 ^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,7 @@ astropy.io.fits
 - Check that the SIMPLE card is present when opening a file, to ensure that the
   file is a valid FITS file and raise a better error when opening a non FITS
   one. ``ignore_missing_simple`` can be used to skip this verification. [#10895]
+- Added ``append`` keyword to append table objects to an existing FITS file [#2632]
 
 astropy.io.misc
 ^^^^^^^^^^^^^^^
--- a/astropy/io/fits/connect.py
+++ b/astropy/io/fits/connect.py
@@ -14,7 +14,7 @@ from astropy.time import Time
 from astropy.utils.data_info import MixinInfo, serialize_context_as
 from astropy.utils.exceptions import (AstropyUserWarning,
                                       AstropyDeprecationWarning)
-from . import HDUList, TableHDU, BinTableHDU, GroupsHDU
+from . import HDUList, TableHDU, BinTableHDU, GroupsHDU, append as fits_append
 from .column import KEYWORD_NAMES, _fortran_to_python_format
 from .convenience import table_to_hdu
 from .hdu.hdulist import fitsopen as fits_open, FITS_SIGNATURE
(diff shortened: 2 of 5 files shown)


pllim:
cc @taldcroft @embray @cdeil from the original discussions

astropy/io/fits/tests/test_connect.py
```diff
@@ -267,6 +267,24 @@ def test_oned_single_element(self, tmpdir):
         assert read['x'].shape == (2, 1)
         assert len(read['x'][0]) == 1
 
+    def test_write_append(self, tmpdir):
+        filename = str(tmpdir.join('test_write_append.fits'))
+        t = Table(self.data)
+        t.write(filename, append=True)
+        t.write(filename, append=True)
+        with fits.open(filename) as hdu_list:
+            assert len(hdu_list) == 3
```
pllim, line 276: Should we also check the content? Although maybe that is pointless here given this test is appending the same stuff over and over again. Should we modify the test to append different data to better mimic real life usage?

pllim (commented):
Need a change log and possibly what's new (the latter needs input from reviewers).

astropy/io/fits/connect.py
```diff
@@ -424,10 +426,14 @@ def write_table_fits(input, output, overwrite=False):
     if isinstance(output, str) and os.path.exists(output):
         if overwrite:
             os.remove(output)
-        else:
+        elif not append:
             raise OSError(f"File exists: {output}")
```
saimn, line 430: Not sure if we should allow `(overwrite=True, append=True)`. It seems a weird combination of parameters.

astropy/io/fits/tests/test_connect.py
saimn: Yes, seems useful to check that the HDUs are correct.

stvguest: I have updated the tests to check the content.

astropy/io/fits/connect.py
stvguest: It does seem a bit weird, but I don't think the behaviour is in any way surprising, and it's not clear to me that there's any benefit in explicitly forbidding it. It could be that there's a use case where it's handy, although I can't think of one. The hdf5 version also has both parameters, although the behaviour is a bit different.

stvguest:
The CI test failed in spectralcoord.rst, not what I did. I also added an entry to CHANGES.rst, am I missing something?

CHANGES.rst
```diff
@@ -38,6 +38,7 @@ astropy.io.fits
 - Check that the SIMPLE card is present when opening a file, to ensure that the
   file is a valid FITS file and raise a better error when opening a non FITS
   one. ``ignore_missing_simple`` can be used to skip this verification. [#10895]
+- Added ``append`` keyword to append table objects to an existing FITS file [#2632]
```
pllim, line 41: This should fix the change log check.

```suggestion
- Added ``append`` keyword to append table objects to an existing FITS file [#2632, #11149]
```

pllim:
You can ignore the spectralcoord doctest failure. I think it has something to do with numpy-dev, as I see that on development branch too.

CHANGES.rst
```diff
@@ -38,6 +38,7 @@ astropy.io.fits
 - Check that the SIMPLE card is present when opening a file, to ensure that the
   file is a valid FITS file and raise a better error when opening a non FITS
   one. ``ignore_missing_simple`` can be used to skip this verification. [#10895]
+- Added ``append`` keyword to append table objects to an existing FITS file [#2632, #11149]
```
embray, line 41: Annoying nitpick, but this should have a blank line before the previous entry.  This can be fixed up just before merging though.

astropy/io/fits/connect.py
embray: Yeah, I think it's fine to allow.  It's one of those cases where "there's probably not a strong reason to do it intentionally" but the behavior is at least consistent and unsurprising.

embray (changes requested):
Other than a minor nitpick it looks good to me.  Thanks for taking care of this 6 year old issue.

astropy/io/fits/connect.py
saimn: Yeah, ok, works for me.

CHANGES.rst
saimn, line 41:
```suggestion

- Added ``append`` keyword to append table objects to an existing FITS file [#2632, #11149]
```

saimn (approved):
Thanks for the contribution @stvguest !

PR #11150, opened 2020-12-08
add example for getting the timezone from location

### LINKED ISSUE #676: Feature request: convert longitude values to timezones
In an Astrodynamics course I am following I had the need to do some operations on times and dates: handle time zones, find out sidereal times, position observers according to time measurements, etc. One of the key operations I have to perform is: given the longitude (we usually neglect irregular timezones, hence no need for latitude) of an observer, find out its time zone (also given a particular timezone, find out the longitude of the central meridian). Anyone else thinks Astropy would be suitable for this? I found some web services with this purpose, but having this functionality in this Python package would be ideal. I think it is also available in a package named astral (http://pypi.python.org/pypi/astral/).

EDIT: the `good first issue` refers to the `docs` part of this issue - what would be needed is to find where best to put the example given in https://github.com/astropy/astropy/issues/676#issuecomment-623138161

jbmorgado:
### Description

This pull request is to address the lack of an example in documentation, on how to find out
the local timezone for an observer given his position.

It uses the `zoneinfo` standard module from python. This module is provided by default on python >=3.9,
the example adds an workaround for python<3.9 using the `backports` package.

Fixes #676

### DIFF
--- a/docs/coordinates/timezones.rst
+++ b/docs/coordinates/timezones.rst
@@ -1,0 +1,84 @@
+.. include:: references.txt
+
+.. _working_with_timezones:
+
+Working with Timezones
+**********************
+
+It is sometimes needed to find the timezone of an observer from their
+coordinates on Earth. Since timezones are defined legally for a site, they
+often don't simply follow the longitude of that site, as such, in order
+to figure out the timezone we need to consult a database.
+
+For this purpose we are going to use the IANA time zone database, which is
+part of Python's standard library with the `zoneinfo` module.
+
+.. _prerequisites:
+
+Prerequisites
+=============
+This functionality depends on the the `zoneinfo` module. This module is part
+of Python >= 3.9 standard library. In case you are using a lower Python version,
+you will need to install the `zoneinfo` module using the backports namespace
+package:
+
+.. code-block:: bash
+
+  pip install backports.zoneinfo
+
+Example
+=======
+
+..
+  EXAMPLE START
+  Determining the altitude/azimuth of a celestial object using the local observation time for a given geographical location
+
+Start by importing the necessary modules:
+
+.. code-block:: python
+
+  import datetime
+  from timezonefinder import TimezoneFinder
+  # in case we aren't on python 3.9 import zoneinfo from backports
+  try:
+    from zoneinfo import ZoneInfo
+  except ImportError:
+    from backports.zoneinfo import ZoneInf
+
+  import astropy.coordinates as coord
+  from astropy.time import Time
+  from astropy import units as 
+
+  import numpy as np
+
+
+Define the target object:
+
+.. code-block:: python
+
+  target = coord.SkyCoord.from_name('albireo')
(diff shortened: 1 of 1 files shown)


docs/coordinates/timezones.rst
```diff
@@ -0,0 +1,84 @@
...
+Prerequisites
+=============
+This functionality depends on the the `zoneinfo` module. This module is part
+of Python >= 3.9 standard library. In case you are using a lower Python version,
+you will need to install the `zoneinfo` module using the backports namespace
+package:
+
+.. code-block:: bash
+
+  pip install backports.zoneinfo
```
pllim, line 27: We declare dependencies in `setup.cfg`.

pllim (changes requested):
I think this file needs to be listed in a TOC somewhere? Or does it really have to be a separate file or can it be a new section in an existing file instead? Or maybe this is better suited for `astropy-tutorials`?

I'll let subpackage maintainer decide on the decision to introduce a new optional dependency for this documentation.

Need some changes to follow https://docs.astropy.org/en/latest/development/style-guide.html .

I don't think `doctest` is running this without the `>>>`  prompt. `doctest` should run, right?

docs/coordinates/timezones.rst
jbmorgado: Thank you for the feedback.

I added the `>>>` prompt (although it impairs readability, are we supposed to use those everywhere?) and removed the explicit pip install of `backports`.

I don't know about the TOC. Maybe someone else should decide where this goes.

I also went through the style guide, by my poor grammar knowledge doesn't allow me to see where the style errors lie on the regard

jbmorgado:
The CI pipeline is failing at unexpected points. This all run in the previous commit and all I did was to change some lines in the documentation: they shouldn't have any impact on those tests.

pllim:
The failure is indeed related. You import `timezonefinder`, which is not part of Python standard library nor any existing dependency.

pllim:
I am started to think perhaps this is not a suitable example for core lib since this one short example introduces two new dependencies but other maintainers may think differently.

jbmorgado:
True, I missed that timezonefinder wasn't a part of astropy requirements as well.

astrojuanlu:
I am not sure the extra two dependencies are worth adding either. Left a comment https://github.com/astropy/astropy/issues/676#issuecomment-748587680 about this.

pllim:
Okay then, I added a "Close?" label, which means the bot will close this PR without merge the next time it runs unless there are objections. Let's continue the discussion over at #676 if desired (until it is also closed by the bot).

@jbmorgado , we still thank you very much for this PR even if we don't end up merging it, as it has provided much needed insight to move the discussions forward. We hope you would continue to contribute in the future! 🙇‍♀️

adrn:
To close the loop here: I added an (explicitly) untested (so no extra dependencies) example to the main coordinates page in #11529

PR #11151, opened 2020-12-08
Improve AltAz documentation

ludwigschwardt:
### Description

This pull request is to address the `astropy.coordinates.AltAz` documentation.

Mention that the local horizon is defined with respect to the WGS84 reference ellipsoid, since that is what ERFA does underneath (see `astropy.coordinates.erfa_astrom` for more details). This is useful to know because other libraries might make different choices. For example, GSFC Calc uses the IERS 2003 ellipsoid.

Mention that altitude == elevation angle, AltAz == AzEl and zenith angle == zenith distance, as those terms are in common use in astronomy.

Also improve grammar and fix a typo.


pllim:
Tentatively milestoning to LTS but feel free to re-milestone if direct backport is not possible. Thanks!

adrn (approved):
Thanks  @ludwigschwardt - this looks good to me!

PR #11156, opened 2020-12-09
TST: Fix spectralcoord doctest

pllim:
### Description

This pull request is to address doctest failure seen on `master` branch. Also see https://github.com/astropy/astropy/pull/11134#issuecomment-741890062


mhvk:
I think #11134 indeed caused the difference. My sense would be to adjust the result in the example since it now would be more precise (if indeed it is meant to be for an observed at GCRS; otherwise, an explicit location should be passed in). But maybe best for @astrofrog to have a look at the actual example (cannot right now myself, have to host a talk...).

pllim:
I opted to update the results as you suggested, @mhvk . If you are satisfied, please approve so we can have a green CI again. Thanks!

mhvk (approved):
Yay!

PR #11157, opened 2020-12-09
BUG: VOTable NumericArray to broadcast mask

### LINKED ISSUE #11087: Saving a VOTable with an array of fixed size to a file only saves first element
### Description

I am trying to add metadata to a VOTable that I'm creating. This metadata corresponds to a tuple of known length. The data seems to be ok when inserted into the table but, when I save the table to a file, only the first element is saved.

### Expected behavior

What I expect to see is the entire tuple saved in the file. The tuple is `(1.0, 2.0, 3.0)`, what I see when I print `params` is  `[<PARAM ID="sampling" arraysize="3" datatype="double" name="sampling" value="(1.0, 2.0, 3.0)"/>]`

### Actual behavior

Only the first element is saved in the file. When I open it, I see: `PARAM` section I see: `<PARAM ID="sampling" arraysize="3" datatype="double" name="sampling" value="1"/>`

### Steps to Reproduce

What I'm doing to create the table and add the metadata is the following:
```
# Create a new VOTable file
votable = VOTableFile()
# Add a resource
resource = Resource(); votable.resources.append(resource)
# Add a table for the spectra (and add the sampling as metadata)
spectra_table = Table(votable); resource.tables.append(spectra_table)
# Add sampling as param
sampling = tuple([1.0, 2.0, 3.0])
# I'm using a list because, later, I could have more than one param
params = [Param(votable, name='sampling', datatype='double', arraysize='3', value=sampling)]
spectra_table.params.extend(params)
```

When I print `params`, I see `[<PARAM ID="sampling" arraysize="3" datatype="double" name="sampling" value="(1.0, 2.0, 3.0)"/>]` which seems to be ok. However, when I save this to a file: 
```
votable.to_xml('table_with_params.xml')
```
In the `PARAM` section I see: `<PARAM ID="sampling" arraysize="3" datatype="double" name="sampling" value="1"/>`.

I tried sampling being a tuple, list and numpy array, but they all lead to the same result. If I use `*` instead of the fixed known length, the entire array is saved.

### System Details

Linux-5.4.0-52-generic-x86_64-with-glibc2.29
[GCC 9.3.0]
Numpy 1.18.2
astropy 4.3.dev268+g883a7c4f5
Scipy 1.4.1
Matplotlib 3.2.1

pllim:
### Description

This pull request is to handle mask broadcasting for `NumericArray` converter in `io.votable`.

Fixes #11087 

cc @druzmieres

### DIFF
--- a/astropy/io/votable/converters.py
+++ b/astropy/io/votable/converters.py
@@ -598,7 +598,7 @@ class NumericArray(Array):
     def parse(self, value, config=None, pos=None):
         if config is None:
             config = {}
-        elif 'version_1_3_or_later' in config and value == '':
+        elif config['version_1_3_or_later'] and value == '':
             return np.zeros(self._arraysize, dtype=self._base.format), True
         parts = self._splitter(value, config, pos)
         if len(parts) != self._items:
@@ -633,8 +633,12 @@ class NumericArray(Array):
         base_output = self._base.output
         value = np.asarray(value)
         mask = np.asarray(mask)
+        if mask.size <= 1:
+            func = np.broadcast
+        else:  # When mask is already array but value is scalar, this prevents broadcast
+            func = zip
         return ' '.join(base_output(x, m) for x, m in
-                        np.broadcast(value.flat, mask.flat))
+                        func(value.flat, mask.flat))
 
     def binparse(self, read):
         result = np.frombuffer(read(self._memsize),
--- a/astropy/io/votable/converters.py
+++ b/astropy/io/votable/converters.py
@@ -598,7 +598,7 @@ class NumericArray(Array):
     def parse(self, value, config=None, pos=None):
         if config is None:
             config = {}
-        elif config['version_1_3_or_later'] and value == '':
+        elif 'version_1_3_or_later' in config and value == '':
             return np.zeros(self._arraysize, dtype=self._base.format), True
         parts = self._splitter(value, config, pos)
         if len(parts) != self._items:
@@ -634,7 +634,7 @@ class NumericArray(Array):
         value = np.asarray(value)
         mask = np.asarray(mask)
         return ' '.join(base_output(x, m) for x, m in
-                        zip(value.flat, mask.flat))
+                        np.broadcast(value.flat, mask.flat))
 
     def binparse(self, read):
         result = np.frombuffer(read(self._memsize),
--- a/astropy/io/votable/tests/converter_test.py
+++ b/astropy/io/votable/tests/converter_test.py
@@ -147,19 +147,17 @@ def test_float_mask_permissive():
 
 
 def test_double_array():
-    config = {'verify': 'exception'}
+    config = {'verify': 'exception', 'version_1_3_or_later': True}
     field = tree.Field(None, name='c', datatype='double', arraysize='3',
                        config=config)
     data = (1.0, 2.0, 3.0)
     c = converters.get_converter(field, config=config)
     assert c.output(1.0, False) == '1'
-    assert c.output(1.0, [False, False]) == '1 1'
+    assert c.output(1.0, [False, False]) == '1'
     assert c.output(data, False) == '1 2 3'
     assert c.output(data, [False, False, False]) == '1 2 3'
     assert c.output(data, [False, False, True]) == '1 2 NaN'
-
-    with pytest.raises(ValueError, match='shape mismatch'):
-        c.output(data, [False, False])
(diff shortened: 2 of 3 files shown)


astropy/io/votable/tests/converter_test.py
```diff
@@ -146,6 +146,41 @@ def test_float_mask_permissive():
...
 
 
+def test_double_array():
+    config = {'verify': 'exception'}
+    field = tree.Field(None, name='c', datatype='double', arraysize='3',
+                       config=config)
+    data = (1.0, 2.0, 3.0)
+    c = converters.get_converter(field, config=config)
+    assert c.output(1.0, False) == '1'
+    assert c.output(1.0, [False, False]) == '1 1'
```
pllim, line 156: Is this desirable?

astropy/io/votable/tests/converter_test.py
```diff
@@ -146,6 +146,41 @@ def test_float_mask_permissive():
...
+    assert c.output(data, False) == '1 2 3'
+    assert c.output(data, [False, False, False]) == '1 2 3'
+    assert c.output(data, [False, False, True]) == '1 2 NaN'
+
+    with pytest.raises(ValueError, match='shape mismatch'):
```
pllim, line 161: I think it is okay to raise a standard Python exception here. I can't think of anything that would cause this to actually happen in the real world? And if it does, then someone should fix it so doesn't really matter what type of exception is raised.

astropy/io/votable/tests/converter_test.py
```diff
@@ -146,6 +146,41 @@ def test_float_mask_permissive():
...
+        c.output(data, [False, False])
+
+    a = c.parse("1 2 3", config=config)
+    assert_array_equal(a[0], data)
+    assert_array_equal(a[1], False)
+
+    with pytest.raises(exceptions.E02):
+        c.parse("1", config=config)
+
+    with pytest.raises(AttributeError):
```
pllim, line 171: This is probably okay to error out like this rather than emitting `E02` as warning and then assume things for the output?

astropy/io/votable/tests/converter_test.py
```diff
@@ -146,6 +146,41 @@ def test_float_mask_permissive():
...
+        c.parse("1")
+
+    with pytest.raises(exceptions.E02):
+        c.parse("2 3 4 5 6", config=config)
+
+    with pytest.warns(exceptions.E02):
+        a = c.parse("2 3 4 5 6")
+
+    assert_array_equal(a[0], [2, 3, 4])
```
pllim, line 180: Is this desirable?

astropy/io/votable/tests/converter_test.py
tomdonaldson: No.  Good point.  We really only want the broadcast for the case where a single mask needs to be broadcast over multiple values.

While I can't think of how this case could happen in the wild, it might be worth adding a check to do the broadcast only if `mask.size == 1`, and use the original `zip()` otherwise.

tomdonaldson (commented):
Thanks for tackling this @pllim.  The new test cases were very helpful to me in thinking through the possibilities.

I think the broadcast change is fine as is, but regarding your question about the broadcast actually replicating values, you could consider adding a conditional to only do the broadcast if there is a single mask.  I don't think it would affect the end user either way, but makes the output more correct by itself.  Your call.

The other thing I noticed after the fact is that the change to the version check seems to have broken some tests.

astropy/io/votable/converters.py
```diff
@@ -598,7 +598,7 @@ def __init__(self, field, base, arraysize, config=None, pos=None):
     def parse(self, value, config=None, pos=None):
         if config is None:
             config = {}
-        elif config['version_1_3_or_later'] and value == '':
+        elif 'version_1_3_or_later' in config and value == '':
```
tomdonaldson, line 601: It looks like this change broke some tests.

astropy/io/votable/tests/converter_test.py
pllim: Great idea, thanks!

astropy/io/votable/converters.py
pllim: Yeah, I don't know why it is so confident that the config here would have `'version_1_3_or_later'`. Rather than diving into a rabbit hole, I modified the test to include this keyword instead. 🤷

tomdonaldson (approved):
This all looks good to me now.  Thanks again for doing it @pllim!

PR #11168, opened 2020-12-15
TST: Skip CI in Action

pllim:
### Description

This pull request is to implement a workaround to enable skipping CI in Actions portion of #11038 .

This uses custom Action at [action-skip-ci](https://github.com/pllim/action-skip-ci) and [styfle/cancel-workflow-action](https://github.com/styfle/cancel-workflow-action).

If this works correctly, only one Action job would run and it will ~fail~ set output value, preventing other jobs from running. What is not showcased here is that duplicate runs would also be cancelled if the CI is not skipped (I tested this on my own repo).

Screenshot 2021-01-07 183342

If this needs backporting, please re-milestone.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -14,23 +14,31 @@ env:
 
 jobs:
   initial_checks:
-    name: Failure means CI is skipped on purpose
+    name: Mandatory checks before CI
     runs-on: ubuntu-latest
+    outputs:
+      run_next: ${{ steps.skip_ci_step.outputs.run_next }}
     steps:
     - name: Check skip CI
       uses: pllim/action-skip-ci@main
+      id: skip_ci_step
       with:
+        NO_FAIL: true
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     # This should only run if we did not skip CI
     - name: Cancel previous runs
       uses: styfle/cancel-workflow-action@ce17749
+      if: steps.skip_ci_step.outputs.run_next == 'true'
       with:
         access_token: ${{ secrets.GITHUB_TOKEN }}
+
+  # The rest only run if above are done
 
   tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
     needs: initial_checks
+    if: needs.initial_checks.outputs.run_next == 'true'
     strategy:
       fail-fast: true
       matrix:
@@ -106,6 +114,7 @@ jobs:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
     needs: initial_checks
+    if: needs.initial_checks.outputs.run_next == 'true'
     strategy:
       fail-fast: false
       matrix:
@@ -137,6 +146,7 @@ jobs:
     name: 32-bit and parallel
     runs-on: ubuntu-latest
     needs: initial_checks
+    if: needs.initial_checks.outputs.run_next == 'true'
     container:
       image: quay.io/pypa/manylinux1_i686
     steps:
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -13,9 +13,24 @@ env:
   IS_CRON: "false"
 
 jobs:
+  initial_checks:
+    name: Failure means CI is skipped on purpose
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check skip CI
+      uses: pllim/action-skip-ci@main
+      with:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    # This should only run if we did not skip CI
+    - name: Cancel previous runs
+      uses: styfle/cancel-workflow-action@ce17749
+      with:
+        access_token: ${{ secrets.GITHUB_TOKEN }}
+
   tests:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    needs: initial_checks
     strategy:
       fail-fast: true
       matrix:
@@ -90,6 +105,7 @@ jobs:
   allowed_failures:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    needs: initial_checks
     strategy:
       fail-fast: false
       matrix:
@@ -120,6 +136,7 @@ jobs:
   parallel_and_32bit:
     name: 32-bit and parallel
     runs-on: ubuntu-latest
+    needs: initial_checks
     container:
       image: quay.io/pypa/manylinux1_i686
     steps:


pllim:
p.s. No idea why CircleCI ran, so I manually cancelled it.

saimn:
I suppose the `CI / ${{ matrix.name }} (pull_request) ` builds are because you use dependencies between builds to not run them if the ci-skip one fails.
It's also a bit annoying to have to run a docker for this, with the time it takes. But I'm not ready to learn typescript right now to avoid this ;), and couldn't find third party actions that do it better, so I guess it's fine for now.

pllim:
I was about to add https://github.com/pllim/actions-cancel-workflows here too (tested at kali-maa/astropy#42) but now I am not sure with your comment about Docker being slow. Should I just use the third-party Action (https://github.com/styfle/cancel-workflow-action)? For that third-party one, I couldn't vet the code because its `dist/index.js` has over 6000 LOC, but I see other repos using it, including Apache and Google (though those repos might just be some employee pet projects or something, as they don't look high profile). 🤷

pllim:
Now that I know how to cancel the entire workflow with PyGithub, I can possibly look into making [action-skip-ci](https://github.com/pllim/action-skip-ci) simply cancels instead of failing first. Let me turn this PR into draft for now.

saimn:
Well, it's just that needing a docker build which takes ~1min just to check some string in a commit message seems a bit overkill, but I don't have a better solution for now ...
For the "skip workflow" one, the source for https://github.com/pllim/actions-cancel-workflows in typescript is ~80 lines and relatively easy to understand (https://github.com/styfle/cancel-workflow-action/blob/main/src/index.ts), and I guess runs much faster than in a docker. If we can avoid multiple builds with docker taking minutes before the actual test builds, it seems better, or maybe we can combine all our custom actions in a unique docker build that would skip ci, kill running workflows, etc ?

pllim:
I was talking about this file, which I think `index.ts` calls: https://github.com/styfle/cancel-workflow-action/blob/main/dist/index.js

Speed wise, there are two possible solutions but none of them pretty:

* Like you said, combine everything into one giant custom action -- This will really complicate the logic, especially if you want to custom pieces of them in the future. Also might render it unusable for anything else other than core lib (e.g., smaller package might not want all the bells and whistles).
* Move the custom actions into core lib as scripts (can exclude them from manifest and so on) and have the CI run them as steps instead of individual jobs. This way, I don't think you need Docker as theoretically it would just use the Python you installed in the previous step. This would definitely only make them usable only in core lib.

pllim:
(LoL... Where did it go? Either I got blacklisted or they finally support it natively... I am guessing the former.)

pllim:
FWIW, it seems to work at https://github.com/pllim/playpen/runs/1566774010?check_suite_focus=true but I'll have to wait till Actions un-blacklist me here to find out for sure.

Also, I have it pull the Docker image now instead of creating from scratch, so the "build" part went down from over a minute to 18s. Maybe it is more bearable this way?

```
"[ci skip]" found in "fix yaml [ci skip]", cancelling this workflow
Cancelled WorkflowRun(url=..., id=426714081)
```

Let's defer the cancel duplicate runs to another PR. One thing at a time.

pllim:
No, something is still not right. It tried to cancel but it didn't work. Maybe it is grabbing the token from my fork, which does not have the power to cancel jobs here upstream? If that is the case, we cannot know if the cancelling will work until this is merged. 🤷 

https://github.com/astropy/astropy/pull/11168/checks?check_run_id=1566876617

```
"[skip ci]" found in "tst: skip ci in action [skip ci]", cancelling this workflow
Cancelled WorkflowRun(url=..., id=426750801)
```

pllim:
Cancelling does not work at kali-maa/astropy#43 . I need to understand this inconsistency.

saimn:
> I was talking about this file, which I think index.ts calls: https://github.com/styfle/cancel-workflow-action/blob/main/dist/index.js

The javascript file is generated from the typescript one ;)

> Also, I have it pull the Docker image now instead of creating from scratch, so the "build" part went down from over a minute to 18s. Maybe it is more bearable this way?

Yes, seems much better :+1:

pllim:
Anyway, I think this is the best I can do right now. A bit ugly but a bonus is that you can use `[skip action]` to skip Action but not CircleCI (not sure if people need that kind of skipping but it is possible).

> The javascript file is generated from the typescript one

Huh!!! How did 80 LOC expand into over 6k? I couldn't tell. 😨

pllim:
As for cancelling duplicate jobs, I did get my other custom action to work, though I still don't understand why cancelling works on PR for that action, but not this one. Proving it is harder because the cancelled job does not tell you who or what cancelled it. The only way you can prove to yourself is try it out over at https://github.com/kali-maa/astropy .

Anyways, I think that is a battle for another day. Perhaps you can try out https://github.com/styfle/cancel-workflow-action (80 lines, allegedly) and then we can compare notes against https://github.com/pllim/actions-cancel-workflows (43 lines 😛 ).

pllim:
I rewrote it in TypeScript. Now it takes 3s to run. I still think we should implement the cancel duplicate workflows one as a separate PR. I am quite done with this one until I hear back via reviews. Thanks!

pllim:
Thanks to brilliant idea from @eslavich over at spacetelescope/jwst#5591 , I updated the Action to not fail, so now you get your nice green check mark and also cancelled jobs.

mhvk:
Yay for keeping at it! It sounds very nice now, and also extensible. Will you make the javascript part of the PR? In any case, do change the message actually shown ("Failure means CI is skipped on purpose").

pllim:
Yeah, need to fix that message. Not sure what you mean by JS part of the PR.

Hmm looks like there is a bug on push event, will keep you updated.

mhvk:
I meant whatever is in "pllim/action-skip-ci@main" and the other one!

pllim:
I fixed the bug upstream in the action.

> whatever is in "pllim/action-skip-ci@main" and the other one

Ah, I thought I linked them above. But here they are again. This time, I'll link to the actual TypeScript where the implementation is.

* https://github.com/pllim/action-skip-ci/blob/main/src/main.ts
* https://github.com/styfle/cancel-workflow-action/blob/ce177499ccf9fd2aded3b0426c97e5434c2e8a73/src/index.ts (note that I pin this one to the hash of its 0.6.0 release, so we don't need to vet the code again every time a change is pushed there)

saimn:
So this PR seems ready right ? It looks great and I think we should start to use it and see how it goes.

pllim:
Yeah, I have been using it myself. If we move the code to some org, I'll batch PR to update all affected packages.

saimn (approved):
Ok, let's move forward and get this in. Thanks @pllim!

mhvk:
Hurray! :fireworks: 

I immediately used in #11222 where @saimn had asked for a small change to the changelog entry. Saved the world ~10x6minx100W ~0.1 kWhr.

PR #11174, opened 2020-12-17
Expose Header.strip as a public method

saimn:
This method is currently private though it can be useful to remove the most common structural keywords.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,6 +41,9 @@ astropy.io.fits
 
 - Added ``append`` keyword to append table objects to an existing FITS file [#2632, #11149]
 
+- Expose ``Header.strip`` as a public method, to remove the most common
+  structural keywords. [#11174]
+
 astropy.io.misc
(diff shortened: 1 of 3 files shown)


pllim (commented):
Was there a reason for it to be private in the first place? Is there a reason why it should be public now? Will people misuse it and then think it is a bug instead of a feature?

saimn:
> Is there a reason why it should be public now?

Because it's useful to remove the FITS structural keywords before updating another header with the stripped one.

I don't see how it could be a "bug". Also it's currently possible to call `.copy(strip=True)` which uses this method, it would just be useful to call it in-place.

embray:
LGTM.  I think maybe a nice follow-up would be to change the list of keywords to remove to a global list (including some "patterns" e.g. for NAXIS<n> type keywords) and allow this method to take a list of additional keywords to either remove or keep besides the defaults.

embray (approved):

PR #11182, opened 2020-12-20
Implement power method for NDArithmetic.

DanRyanIrish:
### Description

This PR implements a `.pow` method on the `NDArithmetic` mixin that raises the `NDData` to a positive integer exponent.

Before writing tests etc. I want to get some feedback as to whether this is something that `NDArithmetic` should support or whether it has been implemented elsewhere.

Fixes #<Issue Number>

### DIFF
--- a/astropy/nddata/mixins/ndarithmetic.py
+++ b/astropy/nddata/mixins/ndarithmetic.py
@@ -536,6 +536,33 @@ class NDArithmeticMixin:
         return self._prepare_then_do_arithmetic(np.true_divide, operand,
                                                 operand2, **kwargs)
 
+    def pow(self, exponent, **kwargs):
+        """Raises ``self`` to the power of an exponent.
+
+        Non-integer and non-positive exponents are not supported.
+        All kwargs are passed to ``self.multiply``.
+
+        Parameters
+        ----------
+        exponent: `numbers.Integral`
+            The exponent to which ``self`` is raised.  Must be > 0
+
+        Returns
+        -------
+        result : `~astropy.nddata.NDData`-like
+            The resulting dataset
+        """
+        if not (isinstance(exponent, numbers.Integral) and exponent > 0):
+            raise TypeError("Non-positive and non-integer exponents not supported.")
+        if exponent == 1:
+            return self
+        result = deepcopy(self)
+        i = 1
+        while i < exponent:
+            result = result.multiply(self, **kwargs)
+            i += 1
+        return result
+
     @sharedmethod
     def _prepare_then_do_arithmetic(self_or_cls, operation, operand, operand2,
                                     **kwargs):
--- a/astropy/nddata/mixins/ndarithmetic.py
+++ b/astropy/nddata/mixins/ndarithmetic.py
@@ -536,32 +536,66 @@ class NDArithmeticMixin:
         return self._prepare_then_do_arithmetic(np.true_divide, operand,
                                                 operand2, **kwargs)
 
-    def pow(self, exponent, **kwargs):
+    def pow(self, exponent, propagate_uncertainties=True, **kwargs):
         """Raises ``self`` to the power of an exponent.
 
-        Non-integer and non-positive exponents are not supported.
-        All kwargs are passed to ``self.multiply``.
+        Non-integer and non-positive exponents are not supported
+        if uncertainties must be propagated.
+        **kwargs are passed to NDData constructor.
 
         Parameters
         ----------
         exponent: `numbers.Integral`
             The exponent to which ``self`` is raised.  Must be > 0
 
+        propagate_uncertainties : `bool` or ``None``, optional
+            If ``None`` the result will have no uncertainty. If ``False`` the
+            result will have a copied version of the first operand that has an
+            uncertainty. If ``True`` the result will have a correctly propagated
+            uncertainty from the uncertainties of the operands but this assumes
+            that the uncertainties are `NDUncertainty`-like. Default is ``True``.
+
         Returns
         -------
         result : `~astropy.nddata.NDData`-like
             The resulting dataset
         """
-        if not (isinstance(exponent, numbers.Integral) and exponent > 0):
-            raise TypeError("Non-positive and non-integer exponents not supported.")
-        if exponent == 1:
-            return self
-        result = deepcopy(self)
-        i = 1
-        while i < exponent:
-            result = result.multiply(self, **kwargs)
-            i += 1
-        return result
+
+        if (self.uncertainty is not None and propagate_uncertainties is True and
+                not (isinstance(exponent, numbers.Integral) and exponent > 0)):
+            raise TypeError("Non-positive and non-integer exponents not supported "
+                            "if propagate_uncertainties is True.")
+        # If exponent is 1, NDData is unchanged.
+        if exponent == 1 and isinstance(propagate_uncertainties, bool):
+            if kwargs.get("copy", False) is True:
+                return deepcopy(self)
+            else:
+                return self
+        # Raise data and uncertainty to power of exponent.
+        data = self.data ** exponent
+        if self.unit is None:
+            unit = None
+        else:
+            unit = u.Unit(self.unit) ** exponent
+        # Propagate uncertainties if user desires.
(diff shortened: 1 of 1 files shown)


astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
         return self._prepare_then_do_arithmetic(np.true_divide, operand,
                                                 operand2, **kwargs)
 
+    def pow(self, exponent, propagate_uncertainties=True, **kwargs):
```
pllim, line 539: Why not add 2 more chars and make the name clearer?
```suggestion
    def power(self, exponent, propagate_uncertainties=True, **kwargs):
```

DanRyanIrish: Thanks for the suggestion @pllim.  We could even name it `raise_to_power`.  It's certainly more characters but is much clearer and in keeping with the convention of the other arithmetic operation method names of using a verb.

pllim: `raise_to_power` might be overkill... but I'll leave that to the subpackage maintainer.

Talking about consistency, do you plan to add these too?

```python
    @sharedmethod
    @format_doc(_arit_doc, name="power", op="**")
```

DanRyanIrish: > Talking about consistency, do you plan to add these too?
> 
> ```python
>     @sharedmethod
>     @format_doc(_arit_doc, name="power", op="**")
> ```

I was initially going to.  But then I realised that the `@format_doc` template wasn't suitable as `power` has only one operand while the other methods for which the docstring template was written have two.  Furthermore, the fact that there is only one operand meant that the same WCS, mask and uncertainty handling options aren't valid.  It seemed easier to just write a unique docstring.

As for the `@sharedmethod`, this could be a class method as well.  But due to how the infrastructure is written, the API would have to be written as `result = NDArithmetic.power(exponent, operand)`, with the args in that order.  I am open to this but it is a bit different from the other methods APIs, e.g. `result = NDArithmetic.add(operand1, operand2)`.  So I wanted to get some feedback as to whether this was a good idea.

What do you think @pllim?

pllim: Ah... okay. Thanks for the clarifications!

Unfortunately, I can't make design decisions in this subpackage as I haven't been involved in its prior development. Perhaps @mwcraig et al. can advise.

mwcraig:
@DanRyanIrish -- I should have time to look at this and the issue you opened tomorrow.

DanRyanIrish:
> @DanRyanIrish -- I should have time to look at this and the issue you opened tomorrow.

OK great!  Looking forward to your thoughts.

DanRyanIrish:
ping @mwcraig

mwcraig:
@DanRyanIrish -- this looks very close to ready to go.  Do you have any time to write tests this week? 

If not, I think I can get the tests written. I can also take care of changing the changelog entry to the new format.

Some detail comments (long overdue):

+ I think `power` is a better name than `pow`. `raise_to_power` seems a little long but I'm fine with it if you prefer that.
+ Making it a `@classmethod` with signature `NDArithmetic.power(exponent, operand)` seems fine, even if it is a bit awkward. I think most people who use it will use the instance method not the class method.

astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
...
+        """Raises ``self`` to the power of an exponent.
+
+        Non-integer and non-positive exponents are not supported
+        if uncertainties must be propagated.
+        **kwargs are passed to NDData constructor.
```
mwcraig, line 544: I would move the description of `kwargs` to Parameters. Also, please wrap it in double-backticks if you keep the `**` there because otherwise Sphinx thinks you have started, but not ended, a block of bold text.

astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
...
+            that the uncertainties are `NDUncertainty`-like. Default is ``True``.
+
+        Returns
+        -------
+        result : `~astropy.nddata.NDData`-like
+            The resulting dataset
+        """
+
+        if (self.uncertainty is not None and propagate_uncertainties is True and
+                not (isinstance(exponent, numbers.Integral) and exponent > 0)):
```
mwcraig, line 565: I think you need to import `numbers` up at the top.

astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
...
+            raise TypeError("Non-positive and non-integer exponents not supported "
+                            "if propagate_uncertainties is True.")
+        # If exponent is 1, NDData is unchanged.
+        if exponent == 1 and isinstance(propagate_uncertainties, bool):
+            if kwargs.get("copy", False) is True:
```
mwcraig, line 570: I think this could just be:

```suggestion
            if kwargs.get("copy", False):
```

astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
...
+        # Raise data and uncertainty to power of exponent.
+        data = self.data ** exponent
+        if self.unit is None:
+            unit = None
+        else:
+            unit = u.Unit(self.unit) ** exponent
+        # Propagate uncertainties if user desires.
+        if self.uncertainty is None or propagate_uncertainties is None:
+            uncertainty = None
+        elif propagate_uncertainties is False:
```
mwcraig, line 583:
```suggestion
        elif not propagate_uncertainties:
```

astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
...
+            uncertainty = self.uncertainty
+        elif propagate_uncertainties is True:
```
mwcraig, line 585: I suppose this and the previous comment are more stylistic than substantive. Ignore as desired.
```suggestion
        elif propagate_uncertainties:
```

astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
...
+            uncertainty = deepcopy(self.uncertainty)
+            i = 1
+            while i < exponent:
+                result_data = self.data**i  # The result_data arg to propagate method.
```
mwcraig, line 589:
```suggestion
                result_data = self.data ** i  # The result_data arg to propagate method.
```

mwcraig (changes requested):
The changes in this review are minor....the bigger thing is some tests.

astropy/nddata/mixins/ndarithmetic.py
```diff
@@ -536,6 +536,67 @@ def divide(self, operand, operand2=None, **kwargs):
...
+            if kwargs.get("copy", False) is True:
+                return deepcopy(self)
+            else:
+                return self
+        # Raise data and uncertainty to power of exponent.
+        data = self.data ** exponent
+        if self.unit is None:
+            unit = None
+        else:
+            unit = u.Unit(self.unit) ** exponent
```
mwcraig, line 579: I think that this needs an import too -- right now we import a couple of things from `astropy.unit` but we don't have the usual `from astropy import units as u`. I'm fine with adding that or explicitly importing `Unit`.

mwcraig:
@DanRyanIrish -- I'll contact you about setting up a short meeting in mid-May once my semester is done to talk through this and the issue you opened. I want to get these changes in for 5.0

mwcraig:
Remilestone-ing as nddata discussions are ongoing/

dhomeier:
@mwcraig, @DanRyanIrish is this still undergoing implementation discussions, and should it be remilestoned again?

PR #11183, opened 2020-12-20
Added separation functions for Latitude and Longitude objects. Overridden is_within_bounds function for Longitude objects

AD-TheDev:
### Description

For simple calculations involving Latitude and Longitude objects, one has to create a SkyCoord object e.g. calculating separation between two Longitude objects, checking if Longitude lies in a given sector. I have added these features so that it won't be necessary to create SkyCoord objects for such simple calculations.

Calculating latitudinal and longitudinal distances are very trivial but very commonly used functions by sky-watchers, astro-photographers etc. Adding these functions would make their job much easier

I have also added a very rudimentary documentation for the added functions (couldn't find the documentation format for writing function specifics). Please review this part.

### DIFF
--- a/astropy/coordinates/angles.py
+++ b/astropy/coordinates/angles.py
@@ -703,21 +703,25 @@ class Longitude(Angle):
 
     def separation(self,lon):
         """
-        Calculates the angular separation of two Longitude objects i.e the measure of the smaller angle subtended by the longitudes
+        Calculates the angular separation of two Longitude objects i.e the measure of the smaller angle subtended by the longitudes.
         Raises `TypeError` if `lat` is not of instance :class:`~astropy.coordinates.Longitude`
         """
+        if not isinstance(lon,Longitude):
+            raise TypeError("Inappropriate argument type.")
         separation=Angle(abs(self-lon))
         if(separation > Angle(180 * u.deg)):
             separation=Angle(360 * u.deg) - separation
         return separation
 
-    def is_within_bounds(self, lower=None, upper=None):
+    def is_within_bounds(self, lower, upper):
         """
-        Check if the Longitude lies in the minor segment described by the 'lower' and 'upper' longitudes 
+        Check if the Longitude lies in the minor segment described by the 'lower' and 'upper' longitudes.
+        Raises `TypeError` if `lower` and `upper` are not instances of :class:`~astropy.coordinates.Longitude`
         """
+        if not isinstance(lower,Longitude) or not isinstance(upper,Longitude):
+            raise TypeError("Inappropriate argument type.")
         if(self.separation(lower) + self.separation(upper) > lower.separation(upper)):
-            return True
-        return False
-
+            return False
+        return True
     
 
--- a/astropy/coordinates/angles.py
+++ b/astropy/coordinates/angles.py
@@ -696,3 +696,11 @@ class Longitude(Angle):
         results = super().__array_ufunc__(*args, **kwargs)
         return _no_angle_subclass(results)
 
+    def separation(self,lon):
+        separation=Angle(abs(self-lon))
+        if(separation > Angle(180 * u.deg)):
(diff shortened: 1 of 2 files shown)


mhvk:
@AD-TheDev - thanks very much for the PR! Looking at it, though, I am not convinced that this addition is needed: the functions are very simple and I think most people would just write them out, never even knowing the methods existed.

But I am also a bit confused about the use case: for `SkyCoord`, a proper angular separation is calculated - what is the use for differences in longitude and latitude separately? If you do want angular separations between two longitude/latitude pairs, there is also the `angle_utilities.angular_separation` function - we've had discussion before to expose this and could revisit that.

Separately, what is the purpose of overriding `is_within_bounds`? (Aside: that is also a method I'm not sure is an example I'd want to emulate, as it is used nowhere within astropy itself...)

AD-TheDev:
> @AD-TheDev - thanks very much for the PR! Looking at it, though, I am not convinced that this addition is needed: the functions are very simple and I think most people would just write them out, never even knowing the methods existed.
> 
> But I am also a bit confused about the use case: for `SkyCoord`, a proper angular separation is calculated - what is the use for differences in longitude and latitude separately? If you do want angular separations between two longitude/latitude pairs, there is also the `angle_utilities.angular_separation` function - we've had discussion before to expose this and could revisit that.
> 
> Separately, what is the purpose of overriding `is_within_bounds`? (Aside: that is also a method I'm not sure is an example I'd want to emulate, as it is used nowhere within astropy itself...)

@mhvk Thanks for responding.  I just thinking that readily giving the logic for longitude separation at least would be beneficial to novice users (like me) who can't intuitively write the logic for that. I know it's very simple, just thought it might actually be of some use as it is used a lot

I was doing an internship wherein I used this package. I had divided the night sky into 12 sections of 30 degree longitude and had to find in which of these section a particular star or planet lies.  I tried the `is_within_bounds` method but it did not serve the purpose(obviously). That is why, I added this override function.

As mentioned, there can be some cases where only the longitudinal separation is required i.e some projection of the proper angular separation. We can create two SkyCoord objects with the same longitudes and 0 latitude and calculate separation, but that introduces redundancy. 

I did not know that `angle_utilities.angular_separation` already exists. If it does, then according to me, it must be surely exposed to the users. That might prove pretty useful in my opinion

mhvk:
@AD-TheDev - the more I look at `Angle.is_within_bouds`, the less I understand its purpose... For most cases I can think of, a direct comparison of angles is less useful than one with wrapping, so I now understand why put an override in `Longitude`. (Your code, though, is OK only for scalar longitudes, not longitude arrays.)

The current code of `Array.is_within_bounds` is: https://github.com/astropy/astropy/blob/9c79f08d72cdd7982ea990d9b01553ac9c4d657b/astropy/coordinates/angles.py#L448-L453

I think it may be possible to this more useful generally, including for subclasses. For starters, I would replace initializer calls of `Angle(lower)` with `self.__class__(lower)` (so that also for Latitude, one cannot compare with, say, an angle larger than 90 degrees). But it does need a bit of care to ensure that wrapping of the angles is done correctly for `Longitude` - I see now that there your `.separation` method came in handy!

If you are interested in pursuing this, I would suggest to start by putting together some test cases of what you think the behaviour should be and then we can see how best to achieve that behaviour...

p.s. On array vs scalar, given numpy precedence I would have hoped that `is_within_bounds` returns an array of bool, and that there would be an `all_within_bounds` which tests all angles. Anyway, it is what it is and we should not break existing code that uses it.
p.s.2 For the separation in `Longitude`, I think I'd generally do something like:
```
from astropy.coordinates import Angle

one = Angle(10., u.deg)
other = Angle([25., 355.], u.deg)

abs((one-other).wrap_at(180*u.degree))
# <Angle [15., 15.] deg>
```

eteq:
not 4.3-critical, so re-milestoning

pllim:

PR #11187, opened 2020-12-22
Added Docker integration for Dev Container

JaydenB0:
### Description

This pull request is to address the lack of docker integration and this commit allows developers to set up a development environment in VSCode quickly.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,7 @@
+// Licensed under a 3-clause BSD style license - see LICENSE.rst
+{
+	"name": "AstroPy",
+	"build": { "dockerfile": "../Dockerfile" } ,
+	"settings": { "terminal.integrated.shell.linux": "/bin/bash" },
+	"extensions": ["ms-python.python"]
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+FROM python:slim
+
+# Install build dependencies
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends build-essential git
+
+# Install test dependencies
+RUN pip install hypothesis flake8
+
+# Copy source to image and install development version
+COPY . .
+RUN pip install -e .[test]


pllim:
Hello and thank you for your interest in contributing! Unfortunately, this PR is out of scope for the core library, as it opens up a can of worms of what other configs we should store if we accept this (e.g., Dockerfile for other dev environments, `.emacs`, AWS config, etc). This would introduce a path to extra maintenance burden on top of the core maintainers, who are already stretched very thin. 

However, I wonder if a new section in dev docs about using Docker is a good alternative. cc @astrofrog @saimn @astropy/devops-maintainers

JaydenB0:
Hey! Glad to be contributing to an amazing project like this (this is actually my first contribution to an open source library). I agree with your concerns regarding the possibility of new configs. Should I push a commit for the docs adding a section regarding how to set up a dev container with Docker using the provided code instead?

pllim:
@JaydenB0 , let's wait a bit and see what other core maintainers think. Given the holiday season followed by American Astronomical Society meeting thereafter, this might take a while. Thank you for your patience!

embray:
I'm not so sure.  The problem with putting things like this in the documentation is it tends to fall out of date before long if no one is paying attention, and we don't even have a particular section in the docs for setting up different development environments (i.e. different IDEs) perhaps in part for this reason.  It is kept fairly generic.

Furthermore, there doesn't seem to be much in this configuration that is particularly unique to Astropy; if there was some particular challenge, e.g., to creating a Docker image that can build and run Astropy that might be a little different.

However, I wouldn't object to creating a wiki page for things like this.

JaydenB0:
@embray I like your suggestion of a wiki page and I think a section for using Docker would be suitable under the ["For Developers"](https://github.com/astropy/astropy/wiki#for-developers) section. However, I think an addition to the [virtual environments section](https://docs.astropy.org/en/stable/development/workflow/virtual_pythons.html#virtual-envs) of the documentation might seem more appropriate as the Dockerfile essentially creates a virtual environment under Docker. Regarding different configurations, getting rid of the .devcontainer file in this commit will keep it generic for most IDEs. What are your thoughts on placing information about the Dockerfile in the virtual environment section?

pllim:

pllim:
I think the decision is to document this in Wiki instead of commiting a docker file.

PR #11191, opened 2020-12-24
MNT: Add CodeQL to catch security issues

pllim:
### Description

I am curious if CodeQL is useful for us. Once merged, it would report at https://github.com/astropy/astropy/security/code-scanning . Also see https://github.com/github/codeql

I am going to cancel Actions CI.

**TODO**

- [x] Squash commits if we decide to merge.
- [x] Remove temporary code to test in this PR.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,77 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  schedule:
+    # run every Wednesday at 6am UTC
+    - cron: '0 6 * * 3'
+  # REMOVE ME
+  pull_request:
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ['cpp', 'python']
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
+        # Learn more:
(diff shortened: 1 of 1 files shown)


pllim:
Got it to run. Takes 10 minutes. Not very useful until merged.

embray:
I'm a little confused about what this is doing.  Does "cpp" language support include C?  Also it doesn't look like it ran any analysis on the Python code.

I've encountered a few corner cases with CodeQL on Python before, but mostly just in unusual stuff with metaclasses.

pllim:
As for how useful this really is for this package, I can't say. I just thought maybe it doesn't hurt to include anyway for best practices, but if there are objections, I am okay with closing this without merge too.

> Does "cpp" language support include C?

I think so, given they lumped C and C++ together in this doc at https://codeql.github.com/codeql-query-help/cpp/ ?

pllim:
I added the ability to skip/cancel CodeQL (also see #11168) given that its run time isn't insignificant.

saimn:
I'm not sure how useful this would be for something like astropy. But if there is no way to see before merging in master, maybe the best is to try ?

pllim:
> way to see before merging in master

I enabled it on a fork. It actually found 34 things: https://github.com/kali-maa/astropy/security/code-scanning

saimn:
@pllim - I get a 404 - Page not found with your link ?

pllim:
I updated the YAML to only run this check weekly as cron job. I'll remove temporary code to test on this PR once I can prove to myself that it can still run.

pllim:
As for the check results, I think GitHub lock down the results to people with write access to the repo. In the case of this unmerged PR, write access to my fork. But once merged, anyone with write access to `astropy/astropy` should be able to view the report, theoretically.

pllim:
p.s. If you dig into the log, here is the [link to report](https://github.com/astropy/astropy/security/code-scanning?query=ref%3Arefs%2Fpull%2F11191%2Fmerge+tool%3ACodeQL) but not sure who can see it.

pllim:
There are a few from `astropy/wcs` (not extern) if you squint a little. 😸

embray:
> However it seems warnings concern only cfitsio and wcslib 

I was going to remark on this as well--they are all minor issues though it would be worth fixing, and maybe submitting the fixes upstream.

saimn (approved):
Looks good to me, running this once a week seems a good compromise :+1:

pllim:
Thanks, all!

PR #11192, opened 2020-12-24
docs: fix simple typo, explcitly -> explicitly

timgates42:
There is a small typo in astropy/io/fits/hdu/compressed.py, astropy/io/fits/hdu/image.py.

Should read `explicitly` rather than `explcitly`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


pllim (approved):
Thanks! I cancelled the CI since it is not needed.

PR #11195, opened 2020-12-28
Fix inability to write masked times with ``formatted_value``.

### LINKED ISSUE #11193: Roundtrip fails for masked mjd Time to ecsv
Found while trying to get masked `Time` to write to a file with my new `Masked` class. cc @taldcroft.

```
import numpy as np
from astropy.time import Time

ma = np.ma.MaskedArray([1., 2.], mask=[True, False])
Table([Time(50000, ma, format='mjd')], names=['mt']).write('a.ecsv', overwrite=True)
Table.read('a.ecsv')
ValueError: Input values did not match the format class mjd:
TypeError: Input values for mjd class must be finite doubles
```

Contents of `a.ecsv`:
```
# %ECSV 0.9
# ---
# datatype:
# - {name: mt, datatype: float64}
# meta: !!omap
# - __serialized_columns__:
#     mt:
#       __class__: astropy.time.core.Time
#       format: mjd
#       in_subfmt: '*'
#       out_subfmt: '*'
#       precision: 3
#       scale: utc
#       value: !astropy.table.SerializedColumn {name: mt}
# schema: astropy-2.0
mt
nan
50002.0
```

mhvk:
The very simple solution is to use `MaskedColumn` for serialization rather than `Column` - this propagates any mask wherever `MaskedColumn` knows how to do it.

fixes #11193

### DIFF
--- a/astropy/table/serialize.py
+++ b/astropy/table/serialize.py
@@ -1,10 +1,13 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
 from importlib import import_module
 import re
 from copy import deepcopy
 from collections import OrderedDict
 
+import numpy as np
+
 from astropy.utils.data_info import MixinInfo
-from .column import Column
+from .column import Column, MaskedColumn
 from .table import Table, QTable, has_info_class
 from astropy.units.quantity import QuantityInfo
 
@@ -128,7 +131,9 @@ def _represent_mixin_as_column(col, name, new_cols, mixin_cols,
             new_name = name + '.' + data_attr
 
         if not has_info_class(data, MixinInfo):
-            new_cols.append(Column(data, name=new_name, **info))
+            col_cls = MaskedColumn if (hasattr(data, 'mask')
+                                       and np.any(data.mask)) else Column
+            new_cols.append(col_cls(data, name=new_name, **info))
             obj_attrs[data_attr] = SerializedColumn({'name': new_name})
         else:
             # recurse. This will define obj_attrs[new_name].
--- a/astropy/time/tests/test_mask.py
+++ b/astropy/time/tests/test_mask.py
@@ -5,6 +5,7 @@ import functools
 import numpy as np
 import pytest
 
+from astropy import units as u
 from astropy.utils import iers
 from astropy.time import Time
 from astropy.table import Table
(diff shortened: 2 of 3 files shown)


astropy/time/tests/test_mask.py
```diff
@@ -212,6 +211,6 @@ def test_serialize_ecsv_masked(tmpdir):
 
     assert t2['col0'].masked
     assert np.all(t2['col0'].mask == [False, True, False])
-    # Serializing floats to ASCII loses some precision so use allclose
-    # and 1e-7 seconds tolerance.
-    assert np.allclose(t2['col0'].value, t['col0'].value, rtol=0, atol=1e-7)
+    # Serializing formatted_value loses some precision.
+    atol = 0.1*u.us if serialize_method == 'formatted_value' else 1*u.ps
+    assert np.all(abs(t2['col0'] - t['col0']) <= atol)
```
mhvk, line 216: Am not using `.isclose()` since we might as well backport this to 4.0

mhvk:
Failures unrelated.

astropy/table/serialize.py
```diff
@@ -128,7 +128,9 @@ def _represent_mixin_as_column(col, name, new_cols, mixin_cols,
             new_name = name + '.' + data_attr
 
         if not has_info_class(data, MixinInfo):
-            new_cols.append(Column(data, name=new_name, **info))
+            # We initialize as MaskedColumn so that we keep a possible mask.
+            # If there is no mask, this will behave as a regular column.
+            new_cols.append(MaskedColumn(data, name=new_name, **info))
```
taldcroft, line 133: I'm hesitant to make this global change and always go to `MaskedColumn` here. I've just had such a long history of being surprised by MaskedArray vs np.array, so I'd prefer to keep the original behavior for the 99% use-case of not-masked attributes.

What about checking for a `mask` attribute first and then `np.any(data.mask)`:
```
col_cls = MaskedColumn if (hasattr(data, 'mask') and np.any(data.mask)) else Column
new_cols.append(col_cls(data, name=new_name, **info))
```

mhvk: Nominally, this is of course what `MaskedColumnInfo` does but I agree initialization itself could cause problems too, so I've changed it to your suggestion. With that, the comment I inserted could be removed since the code became obvious, so it clearly is the better way!

mhvk:
Code coverage failure can be ignored.

mhvk:
@taldcroft - any chance for a minute to get this one in? I'm trying to ensure that the `Masked` PR touches as few things as possible... (and with your comment addressed, I think this is ready...)

taldcroft (approved):
OK, looks good now!

PR #4769, opened 2016-04-11
Angular diameter distance for positive curvature (#4723)

pllim:
Squashed #4723 into one nice commit.

PR #4723, opened 2016-03-23
Angular diameter distance for positive curvature

joergdietrich:
This is the PR promised in #4661 to make `angular_diameter_distance_z1z2` work reliably in positive curvature models. Along the way it creates the new methods `comoving_distance_z1z2` and `comoving_transverse_distance_z1z2` on which `angular_diameter_distance_z1z2` is now based instead of Peebles' (1993) eq. 13.70.

`comoving_distance` and `comoving_transverse_distance` are now special cases of the corresponding `_z1z2` methods. The performance impact is one additional function call for these methods and amounts to 2-4 µs on my laptop for a total execution time of 78 µs for a scalar input. For array like input, the perfomance difference is negligible because the code spends most of the time in the for loop of the vectorized function call.

`vectorize_if_needed` now deals with functions with more than one input argument (arbitrary combinations of scalar and array like)

`angular_diameter_distance_z1z2` is 50% faster on scalar input because the integral over 1/E(z) is computed only once instead of twice. Again, the performance difference vanishes if the function needs to be vectorized.

Issue #4661, opened 2016-03-02
Angular diameter distance between redshifts and negative curvature


joergdietrich:
`angular_diameter_distance_z1z2` contains

```
        # does not work for negative curvature
        Ok0 = self._Ok0
        if Ok0 < 0:
            raise CosmologyError('Ok0 must be >= 0 to use this method.')
```

presumably because 

```
            out = ((dm2 * np.sqrt(1. + Ok0 * dm1 ** 2 / dh_2) -
                    dm1 * np.sqrt(1. + Ok0 * dm2 ** 2 / dh_2)) /
                   (1. + z2))
```

comes from Hogg (1999, arXiv:astro-ph/9905116), which claims "Unfortunately,
the above formula is not correct for Ωk < 0 (Phillip Helbig, 1998, private communication)." However, the derivation in Peebles (1993, p. 336) seems to indicate that this result is independent of the sign of curvature. 

There is no numeric problem because Ok0 \* (dm/dh)**2 evaluates to -sin^2[sqrt(|Ok0| dc/dm], which has a lower bound of -1.
Using the expression above and comparing it to Dm(z1, z2)/(1+z2), where Dm(z1, z2) is computed by performing the integral over 1/E(z) from z1 to z2, gives numerically identical results. 

@nhmc  Am I missing something or is the claim in Hogg (1999) wrong and the code should allow negative Ok0?

joergdietrich:
PR in case I'm right.

astrofrog:
@nhmc @aconley - can you comment on this?

nhmc:
I just copied the result from Hogg et al. So it sounds likely that the expression is indeed valid for Omega_k < 0, but it would be nice to have an independent check. I won't have time to check for the next couple of days though - maybe @aconley can take a look sooner?

aconley:
Hmm.. looks like my email started swallowing everything from github about a month ago.  So I just saw this (and a few other things!).  I will try to investigate this tomorrow-ish.

aconley:
As far as I can tell, @joergdietrich is right.  Unfortunately, this seems to be very seldom discussed in cosmology books.  But the derivation I found in Weinberg 'Cosmology' (I threw my Peebles away a long time ago... ugh, what a horrible book), indeed, this doesn't depend on the sign of the curvature.

I think the next thing to do is to ask Dave (Hogg) if he has any more details to back this up.

joergdietrich:
[...]

PR #4593, opened 2016-02-10
Allow scalar/iterable combinations in angular_diameter_distance_z1z2

joergdietrich:
A common use case in gravitational lensing is to compute the angular diameter distance between one lens and many background galaxies. This PR allows such scalar/iterable combinations in angular_diameter_distance and speeds up resulting code because the distance to the lens does not have to be recomputed for every lens-source pair:

```
In [1]: from astropy import cosmology

In [2]: cosmo = cosmology.FlatLambdaCDM(70, 0.3)

In [3]: z1 = 0.3

In [4]: z2 = np.random.uniform(0.5, 1, 5000)

In [5]: # Old version

In [6]: %timeit -n5 cosmo.angular_diameter_distance_z1z2(np.repeat(z1, z2.size), z2)
5 loops, best of 3: 154 ms per loop

In [7]: 

In [7]: # New version

In [8]: %timeit -n5 cosmo.angular_diameter_distance_z1z2(z1, z2)
5 loops, best of 3: 75.9 ms per loop

In [9]: 

In [9]: res1 = cosmo.angular_diameter_distance_z1z2(np.repeat(z1, z2.size), z2)

In [10]: res2 = cosmo.angular_diameter_distance_z1z2(z1, z2)

In [11]: np.allclose(res1.value, res2.value)
Out[11]: True

```

PR #11197, opened 2020-12-29
Check input to cosmo.angular_diameter_distance_z1z2, that z2>z1

janerigby:
### Description

This pull request is to address a problem in cosmo.angular_diameter_distance_z1z2().  A comment in the function states that of the input redshifts z1, z2, "z2 must be larger than z1".  However, the code itself does not check the input for this.  As such, the code will return a nonsensical negative distance if z2<z1.  I have added one line of error checking of the input values.  I also fixed the typo in "larger" in the comment.   This is only my second contribution to astropy, so I would appreciate polite, constructive feedback on whether I did it right.  I followed the tutorial in https://docs.astropy.org/en/stable/development/workflow/git_edit_workflow_examples.html

### DIFF
--- a/astropy/cosmology/core.py
+++ b/astropy/cosmology/core.py
@@ -1382,7 +1382,7 @@ class FLRW(Cosmology, metaclass=ABCMeta):
         ----------
         z1, z2 : array_like, shape (N,)
           Input redshifts. For most practical applications such as gravitational lensing,
-          z2 should be larger than z1.  The method will work for z2<z1; however, it may
+          z2 should be larger than z1.  The method will work for z2<z1; however, this will
           return negative distances.
 
         Returns
--- a/astropy/cosmology/core.py
+++ b/astropy/cosmology/core.py
@@ -14,7 +14,8 @@ from astropy import constants as const
 from astropy import units as u
 from astropy.utils import isiterable
 from astropy.utils.state import ScienceState
-from astropy.utils.exceptions import AstropyDeprecationWarning
+from astropy.utils.exceptions import AstropyDeprecationWarning, AstropyUserWarning
+
 
 from . import parameters
 
@@ -1374,23 +1375,26 @@ class FLRW(Cosmology, metaclass=ABCMeta):
 
     def angular_diameter_distance_z1z2(self, z1, z2):
         """ Angular diameter distance between objects at 2 redshifts.
-        Useful for gravitational lensing.
+        Useful for gravitational lensing, for example computing the angular diameter
+        distance between a lensed galaxy and the foreground lens.
 
         Parameters
         ----------
         z1, z2 : array_like, shape (N,)
-          Input redshifts. z2 must be large than z1.
+          Input redshifts. For most practical applications such as gravitational lensing,
+          z2 should be larger than z1.  The method will work for z2<z1; however, it may
+          return negative distances.
 
         Returns
         -------
         d : `~astropy.units.Quantity`, shape (N,) or single if input scalar
-          The angular diameter distance between each input redshift
-          pair.
+          The angular diameter distance between each input redshift pair.
 
         """
-
         z1 = np.asanyarray(z1)
         z2 = np.asanyarray(z2)
+        if np.any(np.less(z2, z1)) :
+            warnings.warn(f"Second redshift(s) z2 ({z2}) is less than first redshift(s) z1 ({z1}).", AstropyUserWarning)
         return self._comoving_transverse_distance_z1z2(z1, z2) / (1. + z2)
 
     def absorption_distance(self, z):
--- a/astropy/cosmology/core.py
+++ b/astropy/cosmology/core.py
@@ -1393,8 +1393,9 @@ class FLRW(Cosmology, metaclass=ABCMeta):
         """
         z1 = np.asanyarray(z1)
         z2 = np.asanyarray(z2)
-        if np.any(np.less(z2, z1)) :
-            warnings.warn(f"Second redshift(s) z2 ({z2}) is less than first redshift(s) z1 ({z1}).", AstropyUserWarning)
+        if np.any(np.less(z2, z1)):
+            warnings.warn(f"Second redshift(s) z2 ({z2}) is less than first "
+                          f"redshift(s) z1 ({z1}).", AstropyUserWarning)
         return self._comoving_transverse_distance_z1z2(z1, z2) / (1. + z2)
 
     def absorption_distance(self, z):
--- a/astropy/cosmology/tests/test_cosmology.py
+++ b/astropy/cosmology/tests/test_cosmology.py
@@ -9,6 +9,7 @@ from astropy.cosmology import core, funcs
 from astropy.units import allclose
 from astropy import constants as const
 from astropy import units as u
+from astropy.utils.exceptions import AstropyUserWarning
 
 try:
     import scipy  # pylint: disable=W0611  # noqa
@@ -1233,11 +1234,16 @@ def test_angular_diameter_distance_z1z2():
     assert allclose(tcos.angular_diameter_distance_z1z2(1, 2),
                     646.22968662822018 * u.Mpc)
 
-    z1 = 0, 0, 2, 0.5, 1
-    z2 = 2, 1, 1, 2.5, 1.1
+    z1 = 2  # Separate test for z2<z1, returns negative value with warning
+    z2 = 1
+    results = -969.34452994 * u.Mpc
+    with pytest.warns(AstropyUserWarning, match='less than first redshift'):
+        assert allclose(tcos.angular_diameter_distance_z1z2(z1, z2), results)
+
+    z1 = 0, 0, 0.5, 1
+    z2 = 2, 1, 2.5, 1.1
     results = (1760.0628637762106,
                1670.7497657219858,
-               -969.34452994,
                1159.0970895962193,
                115.72768186186921) * u.Mpc
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -131,6 +131,10 @@ astropy.coordinates
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
 
+- Clarified definition of inputs to ``angular_diameter_distance_z1z2``.
+  The function now emits ``AstropyUserWarning`` when ``z2`` is less than ``z1``.
+  [#11197]
+
 astropy.extern
 ^^^^^^^^^^^^^^
 
@@ -221,7 +225,6 @@ astropy.coordinates
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
-- Clarified definition of inputs to `angular_diameter_distance_z1z2` Added warning if z2<z1. [#11197]
 
 astropy.extern
 ^^^^^^^^^^^^^^
@@ -337,7 +340,6 @@ astropy.cosmology
 
 - Fixed an issue where specializations of the comoving distance calculation
   for certain cosmologies could not handle redshift arrays. [#10980]
-
 
 astropy.extern
 ^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -221,6 +221,7 @@ astropy.coordinates
 
 astropy.cosmology
 ^^^^^^^^^^^^^^^^^
+- Clarified definition of inputs to `angular_diameter_distance_z1z2` Added warning if z2<z1. [#11197]
 
 astropy.extern
 ^^^^^^^^^^^^^^
@@ -336,6 +337,7 @@ astropy.cosmology
 
 - Fixed an issue where specializations of the comoving distance calculation
   for certain cosmologies could not handle redshift arrays. [#10980]
+
 
 astropy.extern
 ^^^^^^^^^^^^^^


astropy/cosmology/core.py
```diff
@@ -1388,7 +1388,7 @@ def angular_diameter_distance_z1z2(self, z1, z2):
           pair.
 
         """
-
+        if (z2 <= z1) : raise ValueError("Second redshift z2 must be larger than first redshift z1")
```
pllim, line 1391: Just nitpick to make it slightly more legible and informative.
```suggestion
        if (z2 <= z1):
            raise ValueError(f"Second redshift z2 ({z2}) must be larger than first redshift z1 ({z1})")
```

pllim (commented):
Hello and thank you! I think the fix is correct; I left a minor comment below. I did think about whether the check should be at a lower level method here:

https://github.com/astropy/astropy/blob/7fb60275e84c9e8a0ea0d3ebef0959cbbc035c94/astropy/cosmology/core.py#L1231

But the docstrings of other methods that calls it did not explicitly say z2 should be larger, so this should be okay, but I'll leave it open a bit longer in case other maintainers disagree.

Please add the following:

* Change log under 4.3 -> bug fix -> cosmology (I decided not to backport because at some level, this is also a behavior change, but it is open to discussion)
* A test to ensure that this method now raises the expected error when z2 <= z1

pllim:
Ah, it appears that your check also needs to be fixed to support numpy broadcasting. The test failure is real.

```
>       if (z2 <= z1) : raise ValueError("Second redshift z2 must be larger than first redshift z1")
E       TypeError: '<=' not supported between instances of 'tuple' and 'float'
```

janerigby:
I've fixed the broadcasting issue, and I've added a test.  However, in updating CHANGES.rst, I came across a baffling comment:  
`- ``angular_diameter_distance_z1z2`` now supports the computation of
  the angular diameter distance between a scalar and an array like
  argument. [#4593] The method now supports models with negative
  Omega_k0 (positive curvature universes) [#4661] and allows z2 < z1.`
However, the code still says that z2 cannot be less than z1.  And if you try, it returns a negative distance, which makes no sense.  How to I contact the person that wrote that z2<z1 is now allowed, to figure out what's going on?

Indeed, I had to delete one element in one of the test arrays, because it expects z2<z1 and then checks that you indeed get a (nonsensical) negative distance.

pllim:
Ah, good catch. The changes you were asking about were introduced in #4769 , which was a cleaned up version of #4723 . Looks like there were plenty of discussions between @joergdietrich , @aconley , and @nhmc all those years ago.

janerigby:
Thanks for that link. But now I’m even more confused about what those developed intended by having the method spit out negative distances. This exchange seems contradictory: “ Doing things this way would also allow us to keep the old behavior of raising ValueError for angular_diameter_distance_z1z2 if any z2 < z1 and going ahead with the new, more sensible approach to return negative distances if z2 < z1 in angular_diameter_distance(z1, z2) calls.”

joergdietrich:
Hi, It's indeed been a while since I wrote that code and I left the field over two years ago. It's a little difficult for me to piece everything back together. Nevertheless, I remember that it was a very conscious decision to take out the ValueError in angular_diameter_distance_z1z2 and allow z2 to be smaller than z1. You'll even find tests for the private method `_comoving_distance_z1z2` that ensure it is symmetric under redshift inversion, which so far is not addressed in your pull request.

I unearthed a lot of the email exchanges but cannot reconstruct the exact arguments anymore and I certainly don't understand anymore what I meant by that contradictory statement cited above. However, 

1. I know I wrote code that depends on z2<z1 being allowed and I assume there is other code out there that would be broken by this PR.
2. Negative distances are not necessarily meaningless. E.g. in (very) closed universes image a lens at the antipodal point from the observer (i.e. at angular_diameter_distance 0) and the background source beyond that. Convention has it that when the background source is viewed beyond the lens it has negative ADD.

In my opinion the correct fix would be to remove the old comment we apparently missed back then instead of regressing to a behavior that we deliberately abandoned. I'm not too invested in this anymore, though, and will not have the time to participate in an in-depth discussion.

dhomeier:
> I unearthed a lot of the email exchanges but cannot reconstruct the exact arguments anymore and I certainly don't understand anymore what I meant by that contradictory statement cited above. However,
> 
Thanks for investigating.
My understanding, before just reading your comment here, was that the statement
> Doing things this way would also allow us to keep the old behavior of raising ValueError for angular_diameter_distance_z1z2 if any z2 < z1 and going ahead with the new, more sensible approach to return negative distances if z2 < z1 in angular_diameter_distance(z1, z2) calls.

referred to the option to keep the old interface (including the `ValueError` for `angular_diameter_distance_z1z2`, but enabling the other functions to accept two input redhifts as well, and allow `z2 < z1` for the alternative function `angular_diameter_distance(z1, z2)` . This second, new interface did not materialise, but `z2 < z1` obviously became permitted.

> 1. I know I wrote code that depends on z2<z1 being allowed and I assume there is other code out there that would be broken by this PR.
> 2. Negative distances are not necessarily meaningless. E.g. in (very) closed universes image a lens at the antipodal point from the observer (i.e. at angular_diameter_distance 0) and the background source beyond that. Convention has it that when the background source is viewed beyond the lens it has negative ADD.

The distance function at the core of `_integral_comoving_distance_z1z2` is strictly antisymmetric in its arguments, since it is just integrating in the opposite direction. So if one accepts the convention of considering distances from higher to lower redshift as negative, `d(z2, z1) = -d(z1, z2)` can certainly make sense. The only thing that changes this antisymmetry in `angular_diameter_distance_z1z2` is that the comoving distance is always scaled to the second redshift. I am not sure how much sense this makes in the context of gravitational lensing, but I'd agree that negative distances probably should not automatically raise an exception.

janerigby:
Thanks for all that backstory!  If I may summarize: Past developers intentionally changed the functions so that z2<z1 does not throw an error and instead returns negative distances.  Negative distances may make sense in certain cosmologies (though not for my practical uses of the function in gravitational lensing.)  These changes were not documented in the code.  The comment "z2 must be larger than z1" was left in, which contradicts these changes.  

Given that new information, I believe I need to withdraw the pull request.  Instead, I believe what should change is to replace the ""z2 must be larger than z1" comment with more helpful documentation, which states that for the common use case of gravitational lensing, z2 should be > z1.  And that there are use cases where z2<z1, in which case the the tool will return a negative distance.

Comments?  Thoughts?

pllim:
Yes, I believe updating the documentation is the way forward. You could choose to re-purpose this PR for the documentation update **or** close this without merge and open a new one. Thank you!

dhomeier:
> Given that new information, I believe I need to withdraw the pull request. Instead, I believe what should change is to replace the ""z2 must be larger than z1" comment with more helpful documentation, which states that for the common use case of gravitational lensing, z2 should be > z1. And that there are use cases where z2<z1, in which case the the tool will return a negative distance. Personally, as a user I'd prefer to get a Warning if the returned distance is negative, but I don't know how that meshes with astropy practices.

Generally agreed. It's still not clear to me there actually _are_ use cases for `z2 < z1` even outside gravitational lensing applications – the concept of angular diameter distance to me always seems to relate to a source at `z2` observed from `z1`, so how the source could be associated to a later time than the observer is puzzling.
But I am not an expert, so if there still is a possibility that an application for this exists, it would OK to return the negative distance with an `AstropyUserWarning` IMO.

pllim:
Issuing warning might break some downstream usage, especially if they set their CI to fail on warning.

janerigby:
Let's keep everything in this Issue.  I undid my changes.  z2<z1 is again allowed.  I updated the method's comments to include the following, which I hope is a non-judgemental summary of the above discussion:
```
For most practical applications such as gravitational lensing, 
          z2 should be larger than z1.  The method will work for z2<z1; however, it may 
          return negative distances.
```

Re @dhomeier 's comment above about `AstropyUserWarning` if z2<z1, I would very much want that behavior as a user.  It's why I opened the Issue.  It's far more likely that I've swapped the order of the inputs, than I intended the not-our-universe edge cases glossed above.  I will send that over as an additional commit.

dhomeier:
> Re @dhomeier 's comment above about `AstropyUserWarning` if z2<z1, I would very much want that behavior as a user. It's why I opened the Issue. It's far more likely that I've swapped the order of the inputs, than I intended the not-our-universe edge cases glossed above. I will send that over as an additional commit.

+1 And about possible breakage downstream, if someone is using that functionality seriously enough to test it in their CI, I think it is worth taking the risk to find out.

janerigby:
I added an AstropyUserWarning if z2<z1, and split out the test case for it. @pllim do you agree that the new merge closes this issue?

astropy/cosmology/core.py
```diff
@@ -14,7 +14,8 @@
 from astropy import units as u
 from astropy.utils import isiterable
 from astropy.utils.state import ScienceState
-from astropy.utils.exceptions import AstropyDeprecationWarning
+from astropy.utils.exceptions import AstropyDeprecationWarning, AstropyUserWarning
+
```
pllim, line 18:
```suggestion
```

astropy/cosmology/core.py
```diff
@@ -1374,23 +1375,26 @@ def luminosity_distance(self, z):
...
         d : `~astropy.units.Quantity`, shape (N,) or single if input scalar
-          The angular diameter distance between each input redshift
-          pair.
+          The angular diameter distance between each input redshift pair.
 
         """
-
         z1 = np.asanyarray(z1)
         z2 = np.asanyarray(z2)
+        if np.any(np.less(z2, z1)) : #
```
pllim, line 1396:
```suggestion
        if np.any(np.less(z2, z1)):
```

astropy/cosmology/tests/test_cosmology.py
```diff
@@ -1233,11 +1234,16 @@ def test_angular_diameter_distance_z1z2():
     assert allclose(tcos.angular_diameter_distance_z1z2(1, 2),
                     646.22968662822018 * u.Mpc)
 
-    z1 = 0, 0, 2, 0.5, 1
-    z2 = 2, 1, 1, 2.5, 1.1
+    z1 = 2  # Separate test for z2<z1, returns negative value with warning
+    z2 = 1
+    results = -969.34452994 *u.Mpc
```
pllim, line 1239:
```suggestion
    results = -969.34452994 * u.Mpc
```

astropy/cosmology/tests/test_cosmology.py
```diff
@@ -1233,11 +1234,16 @@ def test_angular_diameter_distance_z1z2():
...
+    with pytest.warns(AstropyUserWarning):
```
pllim, line 1240: Partial warning message match will ensure that this warning is not coming from somewhere else. I didn't include full warning because I want to avoid regex but you are free to make it stricter if you wish:
```suggestion
    with pytest.warns(AstropyUserWarning, match='less than first redshift'):
```

pllim (commented):
Looks like there are some trailing whitespace. For future reference, some editors have options to automatically remove that for you.

```
astropy/cosmology/core.py:1384:90: W291 trailing whitespace
astropy/cosmology/core.py:1385:88: W291 trailing whitespace
astropy/cosmology/core.py:1397:121: W291 trailing whitespace
astropy/cosmology/tests/test_cosmology.py:1262:1: W293 blank line contains whitespace
4
```

This PR commits need to be squashed due to presence of merge commit in the history. If you need help with this, please let me know.

Some minor comments below. Otherwise, I agree with this approach. Thank you for your continued patience!

janerigby:
@pllim  I believe I have addressed all your comments including the trailing whitespace. 

I don't understand what this means: "This PR commits need to be squashed due to presence of merge commit in the history."  So Yes, I'd like help.

pllim:
@janerigby , I have sent you a direct message on Slack for more information.

pllim (approved):
@janerigby , I took care of the commits for you. I moved the change log from "bug fix" to "API change" as well, and expanded it a little. Thanks!

@dhomeier and @joergdietrich , do you want to have one more look before merge?

astropy/cosmology/core.py
```diff
@@ -1374,23 +1375,27 @@ def luminosity_distance(self, z):
...
-        Useful for gravitational lensing.
+        Useful for gravitational lensing, for example computing the angular diameter
+        distance between a lensed galaxy and the foreground lens.
 
         Parameters
         ----------
         z1, z2 : array_like, shape (N,)
-          Input redshifts. z2 must be large than z1.
+          Input redshifts. For most practical applications such as gravitational lensing,
+          z2 should be larger than z1.  The method will work for z2<z1; however, it may
```
dhomeier, line 1385:
```suggestion
          z2 should be larger than z1.  The method will work for z2<z1; however, this will
```
might as well be unambiguous here – at least I am fairly sure it will _always_ be negative in this case, some comments on possible very exotic cosmologies in the original thread nonwithstanding.
As a matter of fact all existing `scalar_inv_efuncs` return a `sqrt`, so `_integral_comoving_distance_z1z2` integrating in negative direction will definitely be negative.

dhomeier (approved):
Looks good, thanks!

pllim:
Done, @dhomeier . Thanks for the review! I'll leave this open overnight and plan to merge tomorrow unless I hear otherwise. 😸

joergdietrich:
Looks good to me, too. Thanks for discovering and fixing this!

dhomeier:
With the night passed I've been pondering if we should not be even more explicit by stating right away that
```
this will return a negative distance d(z2, z1) = -(1+z1)/(1+z2) d(z1, z2).
```
It's quite obvious from the code though, so maybe no need to further bloat the docs.

pllim:
I think it is too specific to include the equation for a rare case like this. People can always read the code on GitHub. Therefore, I am merging as-is, but @dhomeier , feel free to open a follow up PR if necessary.

Thanks again for your contribution and patience, @janerigby , and thank you all for the reviews!

dhomeier:
Thanks @plim – no, if there is consensus I don't see the need to further expand the documentation.
Users don't necessarily need to know the formula, it just might make it clearer that they do not simply get the negative of the "forward distance".

PR #11200, opened 2020-12-31
Expand and modify physical types

namurphy:
### Description

This pull request adds new physical types and generalizes two existing physical types.

The physical types I added are:
 - Power density (W m⁻³; e.g., something like a heating or cooling power per volume)
 - Volumetric rate (m⁻³ s⁻¹; e.g., collisions per cubic meter per second)
 - Energy flux (J m⁻² s⁻¹; e.g., a heat flux)

The physical types changes are:
 - Pascals are described as units of `"pressure and energy density"` instead of just `"pressure"`, since pressure has units of energy per unit volume
 - The physical type of m⁻² s⁻¹ is broadened from `"kinematic viscosity"` to `"diffusivity"`, since those units are used for multiple reasons in addition to kinematic viscosity.
 
I put more detail about these in each commit message, occasionally with analogies involving wombats.  I still need to add some tests and presumably a change log entry.

### DIFF
--- a/astropy/units/physical.py
+++ b/astropy/units/physical.py
@@ -84,7 +84,7 @@ for unit, name in [
     (si.N, 'force'),
     (si.J, 'energy'),
     (si.J * si.m ** -2 * si.s ** -1, 'energy flux'),
-    (si.Pa, 'pressure'),
+    (si.Pa, 'pressure or energy density'),
     (si.W, 'power'),
     (si.W / si.m ** 3, 'power density'),
     (si.kg / si.m ** 3, 'mass density'),
--- a/astropy/units/physical.py
+++ b/astropy/units/physical.py
@@ -86,6 +86,7 @@ for unit, name in [
     (si.J * si.m ** -2 * si.s ** -1, 'energy flux'),
     (si.Pa, 'pressure'),
     (si.W, 'power'),
+    (si.W / si.m ** 3, 'power density'),
     (si.kg / si.m ** 3, 'mass density'),
     (si.m ** 3 / si.kg, 'specific volume'),
     (si.m ** -3 * si.s ** -1, 'volumetric rate'),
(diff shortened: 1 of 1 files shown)


pllim:
Thank you! I feel like perhaps this PR could be split into 2 PRs -- One for new addition, and one for changing things.

For the part where you change things, I worry about backward incompatibility, especially given that such changes do not emit deprecation warning first (and probably impossible to do so). Does the benefit of these changes justify the possibility of breaking existing code? Maybe @mhvk could advise.

> I still need to add some tests and presumably a change log entry.

Yes, please.

mhvk:
Thanks very much! I agree with @pllim that we should split the PR, since the additions I'd be happy to just get in, but for the changes, I also worry that they could break existing code. For instance, if someone has decorated a function with `@quantity_input` using `pressure`.

Note that I do agree very much with the suggested changes in principle, the example of pressure and energy density is a very good one. I wonder if somehow we can allow a unit to be associated with multiple physical types?! (Redefining equality to make things still work?) Could you raise an issue where we can discuss this? What is needed also depends a little on how exactly the types get used. For instance, for `quantity_input` we just need the reverse dict where there is no problem for multiple physical types associated with a given unit. cc @adrn who has thought more about them.

namurphy:
> Thanks very much! I agree with @pllim that we should split the PR, since the additions I'd be happy to just get in, but for the changes, I also worry that they could break existing code. For instance, if someone has decorated a function with `@quantity_input` using `pressure`.

That makes sense — I was also thinking about backward compatibility for conditionals where `unit.physical_type == 'pressure'`.  I'll remove the backward incompatible PRs here and add some unit units tests (or do we call them unit<sup>2</sup> tests?) shortly.

>  Could you raise an issue where we can discuss this? What is needed also depends a little on how exactly the types get used. For instance, for `quantity_input` we just need the reverse dict where there is no problem for multiple physical types associated with a given unit. cc @adrn who has thought more about them.

I was just thinking I should raise an issue about this!

namurphy:
After adding the tests, I decided to parametrize the tests for `physical_type` to simplify the process for adding even more tests.  Then I added even more tests to increase test coverage a bit more.

PR #11118, opened 2020-12-01
X-Ref in Docstrings

nstarman:
numpydoc x-ref allows for python, numpy, and custom-defined terms to be automatically cross-referenced and linked in RTD. It's what numpy uses for things like ``array_like`` and so that terms like `None` and `list` and `str` never need to be wrapped in backticks.

For more info, see section ``numpydoc_xref_param_type`` in  https://numpydoc.readthedocs.io/en/latest/install.html,
or #11063 for initial discussion.

## Plan

- [x] turn on `numpydoc_xref_param_type`
- [x] ignore terms until the docs compile (`numpydoc_xref_ignore`)
- [x] add term glossary (`numpydoc_xref_aliases` and `glossary.rst`)
- [x] edit docs to not have to ignore things
- [x] edit docs to use glossary
- [x] get everything approved
~- [ ] squash commits.~

See #11555 for followups!

### Subpackage review sign-off sheet

@adonath @adamginsburg @StuartLittlefair @astrojuanlu ~@hamogu~ ~@dhomeier~ ~@saimn~ ~@matteobachetti~ @eslavich @tboch @tomdonaldson @perrygreenfield @mwcraig @taldcroft ~@adrn~ ~@mhvk~ @pllim @bsipocz @eteq @larrybradley @Cadair ~@astrofrog~ @mcara ~@nden~

please check and add your github handle after an approving review
Thanks @adrn for suggestion.
- [x] constants: @mhvk 
- [x] convolution: @mhvk
- [x] coordinates: @adrn
- [x] io.ascii: @hamogu (change requested)
- [x] io.fits: @saimn 
- [x] io.misc: @matteobachetti 
- [x] io.votable: @adrn 
- [x] modeling: (@nden initial look) @adrn
- [x] nddata: @adrn
- [x] stats: @adrn
- [X] table: @mhvk final review
- [X] time: @mhvk
- [X] timeseries: @mhvk
- [X] uncertainties: @mhvk
- [X] units: @mhvk
- [X] utils: @mhvk
- [X] visualization: @mhvk final review
- [X] wcs: @mhvk final review

Issue #6864, opened 2017-11-16
Physical types in units could be improved


adrn:
For other contexts (specifically in `astropy.coordinates`, where we the `Unit*` representation and differential classes), it would be useful to have a concept of `dimensional_unscaled` units (compare
 to `dimensionless_unscaled`). But it occurred to us that this is actually a physical type! It may be that we want to make the physical types in astropy.units into objects themselves that behave like dimensional but unscaled units that cancel and combine properly.

cc @mhvk

adrn:
A little more context: in coordinates, there is currently a conceptual difference between not passing a distance and not passing a velocity component (e.g., radial velocity). For the former, the representation is perceived as "`Unit*`", meaning the distance is assumed to be `1 dimensionless`. If a radial velocity is not passed, but the velocity is re-represented to cartesian, it interprets the radial velocity as being `0 km/s` (or 0 with specific velocity units). That is, with the above example, if you re-represent the representation to cartesian, it has dimensionless units. But, if you re-represent the velocity to cartesian, the resulting object has scaled velocity units.

I think ultimately we'd really like the unit distance to be something like `1 length(unscaled)` and the radial velocity (in the case where it is not passed) to be `0 velocity(unscaled)` so that the resulting cartesian velocity object becomes `velocity(unscaled)` as a way to be more explicit to the user.

adrn:
Also, BTW, this is related to #6293

mhvk:
This is still a very good idea! Also related to @nstarman's problems with having a catalogue in which only for some objects the distance is known.

nstarman:
That might solve the UnitSpherical case!

mhvk:
Indeed, the "abitrary_length" unit would solve `UnitSpherical` - but one would not be able to mix and match.

nstarman:
@adrn, close?

adrn:
Solved with the new `PhysicalType` machinery!

PR #11204, opened 2021-01-04
Create a class to represent physical types

### LINKED ISSUE #11202: Allow units to have multiple physical types
### Description

The `physical_type` attribute for `UnitBase` instances currently returns a string that describes what a particular unit measures:
```Python
>>> import astropy.units as u
>>> (u.barn * u.Mpc).physical_type
'volume'
```
However, sometimes a set of units corresponds to multiple physical types.  For example, a 1 Pa ≡ 1 J · m<sup>-3</sup>, so a pascal can represent _both_ pressure _and_ energy density. In addition, a [kessel run](https://github.com/astropy/astropy/pull/5927) can be retconned from distance to time.  

Units to physical type is not a one-to-one mapping.  Each set of units can map to multiple physical types.  Each physical type should usually map to one set of units, though there can be some exceptions to this (for example, the `temperature_energy` equivalency that lets eV represent energy as well as temperature).  

I propose that we enable each unit to correspond to multiple physical types instead of limiting them to a single physical type.  This would change backward compatibility.

### Draft design and brainstorm for the API

A possibility would be to create a `PhysicalType` class.  Instances of `PhysicalType` could be created for each set of units and become the `physical_type` attribute, and contain a collection of all of the possible physical types that each set of units could have.  

#### Essential behavior of `physical_type` attribute

I'd like to allow the following behavior.
```Python
>>> import astropy.units as u
>>> u.Pa.physical_type 
"pressure, energy density"
>>> u.Pa.physical_type  # not sure if this should be a list or str
["pressure", "energy density"]
>>> u.Pa.physical_type == "pressure"
True
>>> u.Pa.physical_type == "energy_density"
True
```
The last two lines would require defining `PhysicalType.__eq__` to be able to handle multiple physical types (e.g., by making sure `"pressure"` or `"energy"` is in the collection of physical types).  

#### Handling equivalencies

The physical type will also depend on which equivalencies are enabled.  

```Python
>>> u.eV.physical_type == "temperature"
False
>>> u.set_enabled_equivalencies(u.temperature_energy())
>>> u.eV.physical_type == "temperature"
True
```

#### Allowing users to specify the physical type for units or quantities

It would also be helpful for users to be able to specify the physical type in more detail.  I'm not sure if this would be best done for units or quantities, and I have a feeling this could be more difficult to implement.  A possible API for the unit approach would be:

```Python
>>> specific_pressure_unit = create_unit_with_specific_physical_type(u.Pa, physical_type="pressure")
>>> specific_pressure_unit.physical_type = "pressure"
>>> specific_pressure_unit.physical_type == "energy density"
False
>>> u.Pa.physical_type == "energy density"  # we don't want to change the physical type of the original unit
True
```

For a `Quantity` instance, this could be something like
```Python
>>> pressure = u.Quantity(1.0, u.Pa, physical_type="pressure")
>>> pressure.physical_type == "pressure"
True
>>> pressure.physical_type == "energy density"
False
```
It appears that `Quantity` does not have a `physical_type` attribute so this would be particularly new.

namurphy:
This pull request creates a class, `units.physical.PhysicalType`, that represents the physical type(s) associated with a set of units.  The overall purpose is to allow there to be multiple physical types associated with a particular set of units.  This functionality could allow for pressure having equivalent units to energy density.  Instances of this class will become the `physical_type` attribute of `Unit` instances.  The dunder methods were written to mostly preserve backward compatibility, and also allow for some dimensional analysis.    

The motivation and general API for this is described in #11202.  Closes #11202.  The idea for this came about while working on #11200 and https://github.com/PlasmaPy/PlasmaPy/pull/859.

I did notice and fix one minor bug: the physical type for moles per cubic meter was corrected to `"molar concentration"`.  It was previously `"molar volume"`, which was its inverse.  

Things I still need to do:
 - [x] Expand tests (situations with unknown or multiple physical types)
 - [x] Expand the docstring of `PhysicalType` (parameters, examples including of dunder methods, etc.)
 - [x] Write a changelog entry ~~in iambic pentameter~~
 - [x] Define multiple physical types ~~(perhaps mostly in a follow-up pull request that will supersede #11200)~~

In #11202 the idea came up to have this work with equivalencies.  I'm not quite sure if or how we should deal with that since it could get kind of messy and/or ambiguous.  Oh well, that's for another day.


mhvk:
@namurphy - this sounds interesting. But before I dive in, is it ready for review, or are you still changing things around?

namurphy:
@mhvk — thanks!  It's ready for review.  At this point the only likely changes are expanding the physical types...in particular for cases where the same set of units can correspond to multiple physical types.

Some things to look out for in code review:
 - **Physical types for temperature units**: Though K, °C, °F, K, and °Ra are not directly convertible, they all correspond to the same physical type.  
 - **Definitions of new physical types**: The last ~20 physical types defined in `_units_and_physical_types` in `physical.py` are new.  I put in absement and absity in there which are integrals of the position vector, but for the moment I decided against including even higher order integrals for which names have been proposed but aren't in wide use.  I'm guessing the higher order integrals will probably only be used by integral kinematics hobbyists.
 - **How equality is handled**: I tried to keep the existing API for physical types as much as possible, while also expanding it.  
 - **Should the objects in `physical.py` be made private?**:  This module states that it is not intended for use by the user, in which case it might make sense to deprecate them.  At the same time, the documentation for `PhysicalType` especially is probably worth having public.

All in all, this was a pretty fun pull request to write to give me an excuse to procrastinate doing other things.  Thanks again!

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+            return self._get_physical_type_id() == other._get_physical_type_id()
+        else:
+            return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __contains__(self, item):
+        return item in self.as_set
+
```
mhvk, line 313: I'd also define `__iter__` so that one can do `set(physical_type)`:
```
def __iter__(self):
    yield from self._as_set
```

(With that, in principle you do not even need `__contains__` any more, though one might as well keep it, I think)

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+        The unit to be represented by the physical type.
+
+    name : `str` or `set` of `str`
+        A `str` representing the name of the physical type of the unit,
+        or a `set` containing strings that represent one or more names
+        of physical types.
+    """
+
+    if name in ("unknown", {"unknown"}) or "unknown" in name:
+        raise ValueError("Unable to uniquely define an unknown physical type")
```
mhvk, line 376: Maybe try to cover? Also total nitpick: python style seems to be to start exception messages with lower case (since they follow a colon when written out)

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+            f"{physical_type_id!r} ({name!r}) already defined as "
+            f"{_physical_unit_mapping[physical_type_id]!r}"
+        )
+
+    physical_type = PhysicalType(unit, name)
 
+    _physical_unit_mapping[physical_type_id] = physical_type
 
-def get_physical_type(unit):
+    for ptype in physical_type.as_set:
```
mhvk, line 389: With `__iter__` defined as above, you can just iterate directly `in physical_type`.

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+        if ptype in _unit_physical_mapping.keys():
+            raise ValueError(f"{ptype} has already been defined")
```
mhvk, line 391: Try to cover this?

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
-]:
-    def_physical_type(unit, name)
+    if not isinstance(unit, core.UnitBase):
+        raise TypeError("The input to get_physical_type must be a unit.")
+
+    unit = _replace_temperatures_with_kelvin(unit)
+
+    physical_type_id = unit._get_physical_type_id()
+
+    known_unit = physical_type_id in _physical_unit_mapping
```
mhvk, line 416: I'd either do
```
ptype = _physical_unit_mapping.get(physical_type_id)
if ptype is not None:
   ...
```
or put the expression directly in the `if` statement.

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+    def __eq__(self, other):
+        """
+        Return `True` if ``other`` represents a physical type that is
+        consistent with the physical type of the `PhysicalType` instance.
+        """
+        if isinstance(other, (str, set)):
+            other = self._standardize_physical_type_names(other)
+
+        if isinstance(other, set):
+            return other.issubset(self.as_set)
```
mhvk, line 302: Isn't `other.intersection(self.as_set)` good enough?

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+        elif isinstance(other, PhysicalType):
```
mhvk, line 303: One of those totally trivial things, but I'd put this first as it is the comparison that should be easiest to deal with.

mhvk (commented):
@namurphy - This looks very nice indeed!  Quite a number of comments in-line. In terms of implementation, the main ones are to let `str(physical_type)` and `set(physical_type)` be the simple ways for the current `as_set` and `as_string` properties (which I do not like so much).

More general:
- Would it be possible not to use typing yet? I've no strong opinion myself but feel it is something the project should decide on how to do it; for instance, it may be that we decide to go the numpy route of keeping this out of the code proper and work with stubs instead. (Also, there is a more specific problem that the units are not in fact instances of `u.Unit`...)
- I prefer EAFP over LBYL myself, so would not worry so much about all the isinstance checks. This especially since it is code used internally. But it is just a preference!

On your specific points:
1. I think your work-around for temperature is quite good. For now, it is nicely an implementation detail - perhaps we will eventually think of something better!
2. I like the units you added, though am a bit less sure about 'count', 'pixel', etc., as "number of ..." does not really feel like a physical type.
3. I like how you handle equality!
4. I don't think the objects should be made private. The two functions were already public and I think it is fine to add the new class. It also means we can refer to `PhysicalType` elsewhere if needed. Regarding that, do you think we should add some of your examples in its docstring to the actual docs? Looking at the docs, we basically have next to nothing right now... Overall, my sense would be to do this as a follow-up PR, leaving this one as a supernice one about implementation.

mhvk:
Ah, and of course this needs a changelog entry in `CHANGES.rst`, under 4.3, units

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+Defines the physical types that correspond to different units.
 
 This module is not intended for use by user code directly.  Instead,
-the physical unit name of a `Unit` can be obtained using its `ptype`
-property.
+the physical unit name of a `Unit` can be obtained using its
+`physical_type` property.
 """
 
+__all__ = ["def_physical_type", "get_physical_type", "PhysicalType"]
```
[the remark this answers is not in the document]
namurphy: Oops, force of habit!  In PlasmaPy we ended up having the `__all__` before as a PEP 8 thing.  I'll move it back.

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,417 @@
...
+def momentum():
+    return (u.kg.physical_type) * (u.m / u.s).physical_type
 
-@raises(ValueError)
-def test_redundant_physical_type():
-    physical.def_physical_type(u.m, 'utter craziness')
 
+def test_physical_type_as_set(length, pressure):
+    """Test the `as_set` attribute of `physical.PhysicalType`."""
+    assert length.as_set == {"length"}
```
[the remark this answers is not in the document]
namurphy: Good idea!  I suppose the best way to do this is to set `__iter__` and `__next__` and make it iterable, unless there's another way I'm unaware of.

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,417 @@
...
-    assert u.m.physical_type == 'length'
+unit_physical_type_pairs = [
+    (u.m, "length"),
+    (u.cm ** 3, "volume"),
+    (u.km / u.h, "speed"),
+    (u.barn * u.Mpc, "volume"),
+    (u.m * u.s ** 8, "unknown"),
+    (u.m / u.m, "dimensionless"),
+    (hbar.unit, "angular momentum"),
+    (hbar.unit, "angular momentum"),
```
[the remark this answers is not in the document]
namurphy: :scream_cat: Thanks!

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+
+    >>> length_to_19th_power, length_to_20th_power
+    ('unknown', 'unknown')
+
+    >>> length_to_20th_power / length_to_19th_power
+    'length'
+
+    A string representation of a physical type can include underscores
+    instead of spaces, and multiple physical types in a single string
+    can be separated by a comma or backslash.
```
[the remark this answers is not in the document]
namurphy: > It is a regular slash.

I've been getting this wrong probably my whole life!

> But I wonder if it isn't better to just given one way only to indicate multiple possibilities (if so, I'd go with comma, though I could live with `|` - which I see you allow too).

The regular slash was there for backwards compatibility, since some physical types were defined like `"momentum/impulse"`.  I'm wondering if it'd be best to leave that in there for backwards compatibility, but not to describe it as part of the API so it's okay to change later.  In the interest of there being one and only one way to obviously do it, perhaps it'd be best to say that it has to be put in as a `set`.  I'll keep thinking...

> Similarly, I'm not sure about allowing underscores, but even if you do, I'd not mention it.

The underscores are something that could make my life a little bit easier later on.  I'm in the process of developing classes to represent normalizations for different sets of equations that describe plasmas.  The attributes of these classes will have the same names as some of the physical types, except with underscores instead of spaces, so allowing underscores would allow having a variable name represent a physical type.  I'd still like to mention it since there's a use case, though I can make the mention brief.

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
+
+        if isinstance(physical_type_input, str):
+            physical_type_input = {physical_type_input}
+
+        delimiter_pattern = ",|/"
+        standardized_physical_types = set()
+
+        for ptype_input in physical_type_input:
+            if not isinstance(ptype_input, str):
+                raise TypeError(f"Expecting a string, but got {ptype_input}")
```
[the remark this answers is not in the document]
namurphy: Yeah, I was wondering about this.  This does seem more like a `ValueError`, doesn't it?

astropy/units/physical.py
namurphy: Ah!  Ok.  That answers another question I had.  I'll check if `__contains__` is needed or not.

astropy/units/physical.py
namurphy: Yeah...that makes a bit more sense, come to think of it.  Like, if someone is looking for physical types of `"mass"` and `"force"` they might want to check if the physical type is either of those things.  I should also document this.

astropy/units/physical.py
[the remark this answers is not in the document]
mhvk: Definitely not a big deal! If you have a possible use for it, go for it. For the slash, I think backwards compatibility is definitely important. We could simply state that `/` is supported for backwards compatibility.

astropy/units/physical.py
mhvk: I did check and `__contains__` is not needed for testing `x in physical_type`, though presumably it is faster. Since we have the nice equality, it is a bit less obvious it is strictly needed, though, so perhaps remove it?

astropy/units/physical.py
mhvk: Ha, I did not actually think of that, somehow! Just that they might have a set of names for the same unit too. I now wonder if we're not being too generous here. Perhaps we should not allow comparison with a set of strings at all?

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,430 @@
...
-    assert (u.cm ** 3).physical_type == 'volume'
+@pytest.mark.parametrize("unit, physical_type", unit_physical_type_pairs)
+def test_physical_types(unit, physical_type):
+    """
+    Test that the `physical_type` attribute of `u.Unit` objects provides
+    the expected physical type for various units.
 
+    Many of these tests are used to test backwards compatibility.
+    """
+    if unit.physical_type != physical_type:
```
mhvk, line 135: Why not just `assert unit.physical_type == physical_type`? Pytest will then list essentially the failure message added here, no?

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,430 @@
...
+
+
+def test_physical_type_iteration():
+    """Test iterating through different physical type names."""
+    pressure = u.Pa.physical_type
+    ptypes = [ptype for ptype in pressure]
+    assert ptypes == ["energy density", "pressure", "stress"]
+
+
+def test_redundant_physical_type():
```
mhvk, line 424: You already have a function with this name above (though it does a different tests). Rename and move together?

astropy/units/physical.py
```diff
@@ -1,133 +1,423 @@
...
+
+        if not isinstance(unit, core.UnitBase):
+            raise TypeError(f"Expecting a unit, not {unit}.")
+
+        self._unit = _replace_temperatures_with_kelvin(unit)
+        self._physical_type_id = self._unit._get_physical_type_id()
+        self._as_set = self._standardize_physical_type_names(physical_types)
+        self._as_list = sorted(self._as_set)
+
+    def __iter__(self):
```
mhvk, line 270: I tried the much simpler
```
def __iter__(self):
    yield from self.as_set
```
which did not seem to need a `__next__`. Was anything wrong with that?

astropy/units/physical.py
```diff
@@ -1,133 +1,423 @@
...
+
+    def __next__(self):
+        if self._n < len(self._as_list):
+            result = self._as_list[self._n]
+            self._n += 1
+            return result
+        else:
+            raise StopIteration
+
+    def _get_physical_type_id(self):
```
mhvk, line 282: I'm a bit confused why the private function is needed (instead of just using the attribute as needed)

mhvk (commented):
Few more comments, though when I started going I realized it was not quite ready for the next round...

mhvk:
@namurphy - the test failure on 32-bit looks related. I have not looked in detail, but maybe something got changed and not put back or so. 

Otherwise, let me know when ready for the next round of review!

namurphy:
@mhvk — thanks!  I addressed most or all of your code review suggestions, so it's ready for another round of review.  

One change is that I decided to let `def_physical_type` expand the names of physical types for a `PhysicalType` instance that has already been defined for a unit.   I did this thinking about backward compatibility, since there are a few uses of `def_physical_type` that I found on GitHub that define physical types that are being added in this PR, though with different names.  This change will make it less likely to break code in those packages.

I also made `PhysicalType` instances hashable, since I figured I might end up using them as `dict` keys here and there.

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,457 @@
...
+
+def test_physical_type_iteration():
+    """Test iterating through different physical type names."""
+    pressure = u.Pa.physical_type
+    physical_type_names = [physical_type_name for physical_type_name in pressure]
+    assert physical_type_names == ["energy density", "pressure", "stress"]
+
+
+def test_expanding_names_for_physical_type():
+    """
```
mhvk, line 399: The docstring is wrong here.  Also, you need to be sure to remove the definition from the physical types dict, i.e., some `try/finally` construct or `setup/teardown` (it may be an idea to move some functions in a test class with `setup_class` and `teardown_class` methods).

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,457 @@
...
+    physical.def_physical_type(weird_unit, "new physical type")
+
+    assert set((u.s ** 42).physical_type) == {
+        "original physical type",
+        "new physical type",
+    }
+
+
+def test_multiple_same_physical_type_names():
+    """Test that def_physical_type does not define a """
```
mhvk, line 414: Incomplete docstring.

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+    (si.Ohm, {"electrical resistance", "impedance", "reactance"}),
+    (si.Ohm * si.m, "electrical resistivity"),
+    (si.S, "electrical conductance"),
+    (si.S / si.m, "electrical conductivity"),
+    (si.F, "electrical capacitance"),
+    (si.C * si.m, "electrical dipole moment"),
+    (si.A / si.m ** 2, "electrical current density"),
+    (si.V / si.m, "electrical field strength"),
+    (
+        si.C / si.m ** 2,
```
mhvk, line 70: Why so weirdly offset? I'd just indent the set by 4 spaces, and keep the closing parenthesis behind it (or break after the set items)

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+    """
+
+    if isinstance(physical_type_input, str):
+        physical_type_input = {physical_type_input}
+
+    standardized_physical_types = set()
+
+    for ptype_input in physical_type_input:
+        if not isinstance(ptype_input, str):
+            raise ValueError(f"expecting a string, but got {ptype_input}")
```
mhvk, line 185: Might as well cover this...

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+        input_set = set(ptype_input.split("/"))
+        processed_set = {s.strip().replace("_", " ") for s in input_set}
+        standardized_physical_types |= processed_set
+
+    return standardized_physical_types
+
+
+def _identify_unit_from_unit_or_physical_type(obj):
```
mhvk, line 193: I think I'd just make it a method on the class, since it is only used there, but obviously no big deal.

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+    >>> length_to_19th_power, length_to_20th_power
+    ('unknown', 'unknown')
+    >>> length_to_20th_power / length_to_19th_power
+    'length'
+    """
+
+    def __init__(self, unit, physical_types):
+
+        if not isinstance(unit, core.UnitBase):
+            raise TypeError(f"Expecting a unit, not {unit}.")
```
mhvk, line 295: Cover this? (Or, given that this is for internal use, just omit the test? Why stop someone who writes a Unit ducktype?)

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+
+        self._unit = _replace_temperatures_with_kelvin(unit)
+        self._physical_type_id = self._unit._get_physical_type_id()
+        self._physical_type = _standardize_physical_type_names(physical_types)
+        self._physical_type_list = sorted(self._physical_type)
```
mhvk, line 300: Hmm, this makes me wonder a bit. It seems more reasonable to either leave it unsorted or keep user input (e.g., "pressure" is more common than "energy density", and both are more common than "shear").

Actually, I'm a bit confused why this is here; is it just for `__iter__`? I don't really see why that should be ordered.

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return repr(str(self)) if len(self._physical_type) == 1 else str(self)
+
+    def __str__(self):
+        if len(self._physical_type) == 1:
+            return list(self._physical_type)[0]
+        else:
+            return "{" + str(sorted(self._physical_type))[1:-1] + "}"
```
mhvk, line 328: If you do stick to the `_physical_type_list` then use it here and above.

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+        return new_unit.physical_type
+
+    def __rtruediv__(self, other):
+        other_unit = _identify_unit_from_unit_or_physical_type(other)
+        other_unit = _replace_temperatures_with_kelvin(other_unit)
+        new_unit = other_unit / self._unit
+        return new_unit.physical_type
+
+    def __pow__(self, power):
+        if not isinstance(power, numbers.Real):
```
mhvk, line 352: Why not let `unit.__pow__` do the work of warning about wrong numbers?

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+    """
+    If a unit is passed in, return that unit.  If a physical type is
+    passed in, return a unit that corresponds to that physical type.
+    """
+    if isinstance(obj, core.UnitBase):
+        return _replace_temperatures_with_kelvin(obj)
+    elif isinstance(obj, PhysicalType):
+        return obj._unit
+    else:
+        raise TypeError("Expecting a unit or a physical type")
```
mhvk, line 203: Actually, since this is used in methods, you should not raise but return `NotImplemented` (after all, it is possible `other` knows how to deal with a `PhysicalType` even though we do not know how to deal with `other`). Probably another reason to do it on the class...

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+        Return `True` if ``other`` represents a physical type that is
+        consistent with the physical type of the `PhysicalType` instance.
+        """
+        if isinstance(other, PhysicalType):
+            return self._physical_type_id == other._physical_type_id
+        elif isinstance(other, str):
+            other = _standardize_physical_type_names(other)
+            return other.issubset(self._physical_type)
+        else:
+            return False
```
mhvk, line 316: Strictly, you should return `NotImplemented` here (and adjust `__ne__` to take care of that as needed)

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
+    def __repr__(self):
+        return repr(str(self)) if len(self._physical_type) == 1 else str(self)
+
+    def __str__(self):
+        if len(self._physical_type) == 1:
+            return list(self._physical_type)[0]
+        else:
+            return "{" + str(sorted(self._physical_type))[1:-1] + "}"
+
+    def __mul__(self, other):
```
mhvk, line 330: To be able to return `NotImplemented` as needed, I would write a new method, like
```
def _dimensional_analysis(self, other, op):
    # Maybe easiest to just include function in here.
    other_unit = _identify_unit_from_unit_or_physical_type(other)
    if other_unit is NotImplemented:
        return NotImplemented
    other_unit = _replace_temperatures_with_kelvin(other_unit)
    new_unit = getattr(self._unit, op)(other_unit)
    # Don't think we can get NotImplemented here any more
    return new_unit.physical_type
```
and then
```
def __mul__(self, other):
    return self._dimensional_analysis(other, '__mul__')
def __rmul__....
def __truediv__....
def __rtruediv__....
```

astropy/units/physical.py
mhvk: Given my logic above
```
    new_unit = self._unit.__pow__(power)
    return new_unit.physical_type if new_unit is not NotImplemented else NotImplemented
```

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
-    _unit_physical_mapping[name] = r
+            f"the following physical type names are already in use: "
+            f"{names_already_in_use}."
+        )
+
+    physical_type_id = unit._get_physical_type_id()
+    unit_already_in_use = physical_type_id in _physical_unit_mapping
+    if unit_already_in_use:
+        old_physical_type = _physical_unit_mapping[physical_type_id]
+        name |= set(old_physical_type)
```
mhvk, line 416: Are you instantiating a new physical type to ensure they are immutable?

astropy/units/physical.py
```diff
@@ -1,133 +1,461 @@
...
-    (astrophys.R, 'photon flux'),
-    (misc.bit, 'data quantity'),
-    (misc.bit / si.s, 'bandwidth'),
-    (cgs.Franklin, 'electrical charge (ESU)'),
-    (cgs.statampere, 'electrical current (ESU)'),
-    (cgs.Biot, 'electrical current (EMU)'),
-    (cgs.abcoulomb, 'electrical charge (EMU)')
-]:
-    def_physical_type(unit, name)
+    if not isinstance(unit, core.UnitBase):
```
mhvk, line 444: I'm still not sold on these checks for non-user code - just let the replacement fail?

mhvk (commented):
@namurphy - I like the idea to allow adding new physical types to existing units, as well as making them hashable.

Still quite a lot of comments. For the operators, mostly to ensure you return `NotImplemented` instead of raising (should have caught that earlier); for tests, nitpicks, but perhaps good to organize those a bit more.

The other things are probably more philosophical: I'm very much a "Easier to Ask Forgiveness than Permission" coder, and avoid "Look before you leap" `isinstance` checks - I like duck-typing, making the most minimal assumptions possible! And in non-user facing code it seems just not very necessary.

astropy/units/physical.py
namurphy: Oh!  I like this!

astropy/units/physical.py
namurphy: The main reason I'm sorting them is because [doctests require the line of output to match exactly](https://docs.python.org/3/library/doctest.html#warnings)...which can cause failures when the thing being printed out is a set or a string representation of a set.  This also happens when the physical types are printed out in error messages.  If the results are unordered, then we could run into situations where `str(physical_type) == str(physical_type)` returns `False`.

I like the idea of keeping the order that the items were inputted in, with the most common ones used first.

astropy/units/physical.py
namurphy: Oh...I got in the habit of using [black](https://black.readthedocs.io/en/stable/) for formatting and have gotten accustomed to that style.  I'll change this to match the Astropy style.

astropy/units/tests/test_physical.py
namurphy: Oops!  I have a bad habit of starting sentences and then forgetting to

astropy/units/physical.py
mhvk: Hmm, I don't really like that doctests drive design, but I can see how it is easier.

On second thought, I'm not sure any more I like my own suggestion of keeping order, but we're starting to risk overthinking it, so maybe best for you to just decide something and stick to it!

astropy/units/physical.py
mhvk: Yes, I noticed it is taking python programmers by storm...  I'm still a bit on the fence, feeling the urge to minimize line changes is coming at the cost of readability. Though I can see the case for consistency. Anyway, some of the others are quite allergic to it!

astropy/units/physical.py
namurphy: I didn't have any particular reason.  I ended up changing it so that it's the same `PhysicalType` instance, only re-initialized with the new names.  Seems like this will be a slight improvement.

astropy/units/physical.py
namurphy: In that case I'll stick with it being sorted, since it probably isn't going to make much difference one way or another when there are ≲ 4 physical type names, and most of the time there's only one name anyway.

That doctests are driving design is indeed irksome, though there's also the more general principle that it's harder to test code that is not deterministic...

astropy/units/core.py
```diff
@@ -1328,14 +1327,38 @@ def cgs(self):
...
         Examples
         --------
         >>> from astropy import units as u
-        >>> print(u.m.physical_type)
-        length
+        >>> u.m.physical_type
+        'length'
+
+        >>> u.Pa.physical_type
+        {'energy density', 'pressure', 'stress'}
```
maxnoe, line 1344: I strongly disagree. A Pascal is not a unit of energy density (J / m³). It just decomposes to the same SI base units.

maxnoe:
I have to issues with this:

* By going through decomposition, special units loose their semantic meaning, which should be avoided
* `physical_type` is now sometimes a string, sometimes a set. This should be consistent (give a set of length one if there is only one possibility). This avoids the need for checking what you got and won't break code if we realize we forgot a meaning.

mhvk:
@maxnoe - just to be clear, there is the reverse problem that this is trying to solve:
```
u.Unit("kg / (m s2)").physical_type
'pressure'
```

Internally, `physical_type` now already always is a `set` - is your concrete suggestion to let `str` and `repr` reflect that? The reason that for length 1 they look like strings is for backwards compatibility. Similarly, for that reason `__eq__` is really defined as whether there is an intersection.

maxnoe:
@mhvk ok, good, then I was only confusing the `repr` with the actual thing in the examples.

Which brings about another point then: the common python way is that `repr(thing)` identifies really what an object is, and only `str` should simplify this for nice printing.

So `repr` should probably be something like `PhysicalType({length, })`, not `'length'`.

mhvk:
True, main problem of course is that changing `__repr__` that way will break doctests, etc. But perhaps worth it for the enhanced clarity...

namurphy:
> the common python way is that `repr(thing)` identifies really what an object is, and only `str` should simplify this for nice printing.

My intention was to maintain backwards compatibility with the previous API of having `physical_type` be a string, though I personally would prefer it to have `repr` be something more descriptive like you suggested.  As @mhvk mentioned, this could end up breaking doctests, in particular in downstream packages.  This change would probably be straightforward enough for other packages to fix, so I'd be okay with a minor break in backwards compatibility to match Python conventions.  If we have agreement, I'll go ahead and change `__repr__` when next I have a chance to work on this.

mhvk:
On balance, I think a `PhysicalType(...)` repr is better. I think most bigger packages that have documentation would test against astropy-dev, so we'd hopefully find out about problems before an actual release.

namurphy:
I changed `PhysicalType.__repr__` to return strings like `PhysicalType('length')` for things with one physical type, and strings like `PhysicalType({'absement', 'sustained displacement'})` when there is more than one physical type.  I'd be okay with having both cases come back with set notation, or both without the set notation.  I put a note in the changelog that stringifying physical types is not backwards compatible.  Updating the doctests was pretty straightforward.  This still could break doctests in downstream packages, but the fix for the broken doctests is pretty straightforward.  I like the improved clarity.

I changed the docstrings to mention that `PhysicalType` instances represent the physical types of "dimensionally compatible" units, which provides context for units like pascals.  

@mhvk — when you have a chance, could you partake in another round of code review?  Thanks in advance!

astropy/units/physical.py
```diff
@@ -1,133 +1,458 @@
...
+        other = self._identify_unit_from_unit_or_physical_type(other)
+        if other is NotImplemented:
+            return NotImplemented
+        return other.physical_type._dimensional_analysis(self, "__truediv__")
+
+    def __pow__(self, power):
+        return (self._unit ** power).physical_type
+
+    def __hash__(self):
+        return hash(self._physical_type_id)
```
namurphy, line 357: I might or might not be looking forward to using physical type instances as keys in dictionaries. :smile_cat:

namurphy:
Update: I fell down a rabbit hole and am fixing a few more things, so it'd be best to review this January 21 or later.

astropy/units/physical.py
```diff
@@ -1,133 +1,472 @@
...
+
+    def __len__(self):
+        return len(self._physical_type)
+
+    def __array_ufunc__(self):
+        # This prevents operations like Unit(...) * PhysicalType(...)
+        # from returning a Quantity with a PhysicalType instance as the
+        # Quantity's value.
+        pass
```
namurphy, line 371: My motivation for adding this method was to try ways to prevent things like this from happening:
```Python
>>> u.m * u.m.physical_type
<Quantity PhysicalType('length') m>
```
I wanted `Unit.__mul__` to return `NotImplemented` here so that this operation would fall back on `PhysicalType.__rmul__`, ideally without making any changes to `Unit` or `Quantity`.  I speculated that having `PhysicalType.__array_ufunc__` raise a `TypeError` would prevent this from happening, but it appears to only be necessary for `PhysicalType.__array_ufunc__` to exist without actually doing anything.  It's a bit confusing why this works, so I'm curious if there is a more understandable way to ensure this behavior.

mhvk: I'll think about this - I think the underlying problem is `np.array(physical_type)` creating an object array. `__array__ = None` may be better.

namurphy: Thanks!  I'll go with that for the time being.

astropy/units/core.py
```diff
@@ -1328,14 +1327,38 @@ def cgs(self):
     @property
     def physical_type(self):
         """
-        Return the physical type on the unit.
+        Return the physical type(s) that are dimensionally compatible
```
mhvk, line 1330: Maybe keep as one line as "Physical type(s) dimensionally compatible with the unit."

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,436 @@
...
+    supplied names is not a string, while making sure that the physical
+    type for the unit remains unknown.
+    """
+    obscure_unit = u.s ** 87
+    with pytest.raises(ValueError):
+        physical.PhysicalType(obscure_unit, invalid_input)
+    assert obscure_unit.physical_type == "unknown"
+
+
+def _undef_physical_type(unit):
```
mhvk, line 379: I think the way to do this is with `teardown_function` - that way it is guaranteed to run also if the test fails.

(I think all this is easier with a test class as suggested in https://github.com/astropy/astropy/pull/11204/files#r558759530 below)

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
 from . import cgs
-from . import imperial
 from . import misc
 
+__all__ = ["def_physical_type", "get_physical_type", "PhysicalType"]
 
-__all__ = ['def_physical_type', 'get_physical_type']
-
+_units_and_physical_types = [
+    (core.Unit(1), "dimensionless"),
```
mhvk, line 22: just `core.dimensionless_unscaled`, I think.

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+    The preferred method to access a physical type is by accessing the
+    ``physical_type`` attribute of a unit.
+
+    >>> import astropy.units as u
+    >>> u.meter.physical_type
+    PhysicalType('length')
+
+    Some units are dimensionally compatible with multiple physical types.
+    A pascal is intended to represent pressure or stress, but the unit
+    decomposition is equivalent to that of energy density.
```
mhvk, line 232: and "and stress"

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+            return other.issubset(self._physical_type)
+        else:
+            return NotImplemented
+
+    def __ne__(self, other):
+        equality = self.__eq__(other)
+        return not equality if isinstance(equality, bool) else NotImplemented
+
+    @property
+    def _name_string_as_ordered_set(self):
```
mhvk, line 302: Absolute nitpick that you should feel free to ignore, but I'd keep this a method rather than a property (just to avoid the indirection).

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+        If a unit is passed in, return that unit.  If a physical type is
+        passed in, return a unit that corresponds to that physical type.
+        If a real number is passed in, return a dimensionless unit.
+        Otherwise, return `NotImplemented`.
+        """
+        if isinstance(obj, core.UnitBase):
+            return _replace_temperatures_with_kelvin(obj)
+        elif isinstance(obj, PhysicalType):
+            return obj._unit
+        elif isinstance(obj, numbers.Real):
```
mhvk, line 330: I'm a bit unsure why we allow numbers here; for dimensional analysis, that should not be useful.

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+            return core.dimensionless_unscaled
+        else:
+            return NotImplemented
```
mhvk, line 333: Maybe add a test that something like `length.__mul__([])` returns `NotImplemented`.

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+        return self._dimensional_analysis(other, "__mul__")
+
+    def __rmul__(self, other):
+        return self.__mul__(other)
+
+    def __truediv__(self, other):
+        return self._dimensional_analysis(other, "__truediv__")
+
+    def __rtruediv__(self, other):
+        other = self._dimensionally_compatible_unit(other)
```
mhvk, line 353: Couldn't this just be `return self._dimensional_analysis(other, "__rtruediv__")`?

astropy/units/physical.py
mhvk: Nice!

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+
+def _get_names_in_use_except_from(unit):
+    """
+    Get a `set` containing all of the physical type names in use, except
+    for the names that correspond to ``unit``.
+    """
+    all_names_in_use = set(_unit_physical_mapping.keys())
+    id = unit._get_physical_type_id()
+    if id in _physical_unit_mapping.keys():
+        names_for_this_unit = set(_physical_unit_mapping[id])
```
mhvk, line 383: Just `return all_names_in_use - set(...)` here and then `return all_names_in_use` in the `else` clause.

Or, maybe better, call it `names_in_use` and here do `names_in_use -= set(_physical_unit_mapping[id])`

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
-    _physical_unit_mapping[r] = name
-    _unit_physical_mapping[name] = r
+            f"the following physical type names are already in use: "
+            f"{names_already_in_use}."
+        )
+
+    physical_type_id = unit._get_physical_type_id()
+    unit_already_in_use = physical_type_id in _physical_unit_mapping
+    if unit_already_in_use:
+        old_physical_type = _physical_unit_mapping[physical_type_id]
```
mhvk, line 431: Why not just get `physical_type` here in one go?

mhvk (commented):
OK, went for another round. It is looking good. The only more major comment is that I'm not sure why in the dimensional analysis numbers are allowed.

Also, maybe time to squash the commits?

astropy/units/physical.py
namurphy: That's a good question!  

My first use case for allowing numbers is that there might be times when someone would want to do an operation like:
```Python
>>> length = u.m.physical_type
>>> 1 / length
PhysicalType('wavenumber')
```
which would be equivalent to `length ** -1`.

My second use case is that if I were introducing the idea of dimensional analysis to someone, I might use an example like
```Python
>>> length * 42
PhysicalType('length')
```
to show that when doing dimensional analysis, multiplying by a constant will not change the physical type.  This would also be equivalent to `length ** 1`.  

I was also thinking this would be reasonable since SI defines "one" as a unit that is implicitly present.  (This make me wonder if an alternative acceptable name for `u.dimensionless_unscaled` would be `u.one`, but it's a few years too late to suggest that.  Nevermind, I just looked and `u.one` is defined as `u.dimensionless_unscaled`.  😹 )

In any case, are these use cases sufficient motivation to allow these operations?

astropy/units/core.py
namurphy: @mhvk — Thank you!  Your comments have been really helpful in improving this pull request on both the small scale code changes as well as the larger overall structure.  I'll hopefully have a chance to implement the suggestions by early this week.

> Also, maybe time to squash the commits?

That does seem like an increasingly good idea.  We could either squash and merge when the PR is ready, or I could do a rebase and squash commits individually.

namurphy:
@mhvk — Thank you!  Your comments have been really helpful in improving this pull request on both the small scale code changes as well as the larger overall structure.  I'll hopefully have a chance to implement the suggestions by early this week.

> Also, maybe time to squash the commits?

That does seem like an increasingly good idea.  We could either squash and merge when the PR is ready, or I could do a rebase and squash commits individually.

namurphy:
It's taking me a while longer to get to this since I've been dealing with the words "unexpected", "plumbing", and "situation" this week.  (On the plus side, the open air plumbing is now giving my kitchen the decor and feel of the interior of the _Millenium Falcon_.)

I was thinking (and mentioned on Slack) that it might be helpful to have a couple of people from a few different research areas look through the list that defines physical types.  That would be particularly helpful because people from different research areas would be most familiar with different physical types, and could potentially spot errors or omissions that I'd have a hard time finding.

pllim:
@namurphy , posting on https://groups.google.com/g/astropy-dev is also an option if you don't hear enough feedback on Slack. Sorry to hear about the plumbing and good luck!

p.s. Silver lining is that maybe you can redecorate the house to be Mario Brothers themed? 😬

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+    Represents the physical type(s) that are dimensionally compatible
+    with a set of units.
+
+    Instances of this class should be accessed using the ``physical_type``
+    property of `Unit` instances.  This class is not intended to be
+    instantiated directly in user code.
+
+    Parameters
+    ----------
+    unit : u.Unit
```
adrn, line 200:
```suggestion
    unit : `~astropy.units.Unit`
```

astropy/units/physical.py
```diff
@@ -1,133 +1,473 @@
...
+    If a physical type already exists for a unit, add new physical type
+    names so long as those names do not correspond to other units.
+
+    This function is not intended to be called directly in user code.
 
     Parameters
     ----------
-    unit : `~astropy.units.UnitBase` instance
-        The unit to map from.
+    unit : u.Unit
```
adrn, line 400:
```suggestion
    unit : `~astropy.units.Unit`
```

adrn (commented):
Thank you immensely @namurphy for this PR! At first look, this is **great**, and resolves a longstanding wishlist item for many of us :) (relevant: #6864). I haven't yet taken a close look at the tests, but will aim to do that over the next few days. Thanks again!

mhvk:
@adrn - indeed, neat it is!

@namurphy - at some point, do rebase and squash in the process (you can keep particularly important points if you want). The squash and merge feature of github has caused problems before, so we do not use it.

As for other physical types, yes, would be good to have others check! (Though can of course be a separate issue...)

astropy/units/physical.py
namurphy: Good catch!

astropy/units/physical.py
```diff
@@ -1,133 +1,425 @@
...
 _unit_physical_mapping = {}
 
 
-def def_physical_type(unit, name):
+def _replace_temperatures_with_kelvin(unit: core.Unit) -> core.Unit:
+    """
+    If a unit contains a temperature unit besides kelvin, then replace
+    that unit with kelvin.
+
+    Temperatures cannot be converted directly between K, °F, °C, and
```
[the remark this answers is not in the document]
maxnoe: Most other unit frsmeworks I know solve this bei having absolute Temperaturen and Temperaturen difference. Which you need to know to be able to handle this.

astropy/units/physical.py
namurphy: Just to confirm that I understood your suggestion correctly, I changed `to represent pressure or stress` to `to represent pressure and stress`.

astropy/units/physical.py
namurphy: No, it turns out that there are a few tests that end up failing when using that one-liner.  I'm curious about why that was happening, but I'm content enough with the current implementation since it's working and still pretty short.

namurphy:
I finally got around to the rebase and addressing the last round of code review suggestions, so I believe this is ready for another round of review.  Thank you again!

My to do list after this PR is complete includes:
 - [ ] Creating an issue asking for independent verification and additional review of the physical types
 - [ ] Creating an issue about adding physical types to the narrative documentation
 - [ ] Making a pull request to resolve the issue regarding the narrative documentation

astropy/units/physical.py
adrn: Seems like good enough motivation to me!

adrn (commented):
This is really quite nice @namurphy -- thank you for all of the time you have put into this! I'm very excited to see this become a part of Astropy 🎉! The implementation looks quite clean and your tests look well structured. After reading through the code changes, I am happy with this. I've just installed your branch locally to do some testing, and it would be good to have @mhvk take another look at this.

adrn:
PS: one other thing to add to our list of things to consider (after this PR! so may need an issue): It might make sense to review the `quantity_input()` decorator and related functionality. I think we should consider making use of this new machinery instead of the current dependence on passing in `Quantity` objects.

namurphy:
> This is really quite nice @namurphy -- thank you for all of the time you have put into this! I'm very excited to see this become a part of Astropy

This PR has been a textbook example of procrastination driven development!  😸  Not that PDD actually shows up in textbooks...

> The implementation looks quite clean and your tests look well structured. 

The thoughtful code reviews from @mhvk have helped quite a lot!

> It might make sense to review the `quantity_input()` decorator and related functionality. I think we should consider making use of this new machinery instead of the current dependence on passing in `Quantity` objects.

I'll add that to the list!  I've only looked a little bit into `@quantity_input` so it'd be helpful if you or someone else would be able to create the issue.
 - [ ] Make an issue for ⬆️

mhvk:
On `quantity_input`, @nstarman has been refactoring that already, to ensure we could use something like `Quantity[u.m]` (as would be logical for typing as well) - see #10662 and #10655. I think it might be relatively straightforward to extend it to `Quantity[length]` where `length` is a physical type (and/or a string 'length').

I hope to get to further review (though if I don't get to it soonish, I think it is fine if @adrn signs off on it; I think the implementation has become very nice).

nstarman:
> On `quantity_input`, @nstarman has been refactoring that already, to ensure we could use something like `Quantity[u.m]` (as would be logical for typing as well) - see #10662 and #10655. I think it might be relatively straightforward to extend it to `Quantity[length]` where `length` is a physical type (and/or a string 'length').

It should be quite straightforward. Internally we just run all physical types through the machinery to convert them from strings to the PhyicalType class instance. When this is merged I can rebase.

namurphy:
> On `quantity_input`, @nstarman has been refactoring that already, to ensure we could use something like `Quantity[u.m]` (as would be logical for typing as well) - see #10662 and #10655. I think it might be relatively straightforward to extend it to `Quantity[length]` where `length` is a physical type (and/or a string 'length').

I've been following those PRs and am excited about them!

CHANGES.rst
```diff
@@ -215,6 +229,11 @@ astropy.units
 
 - Calling ``Unit()`` with no argument now returns a dimensionless unit,
   as was documented but not implemented. [#11295]
+- The ``physical_type`` attribute of units now returns an instance of
```
mhvk, line 232: Needs an empty line beforehand (likely a rebase issue...)

astropy/units/physical.py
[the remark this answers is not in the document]
mhvk: Raised #11386 so I can call this resolved...

astropy/units/physical.py
mhvk: I meant to add it at the end "of energy density and stress" - I don't think Pa is meant to be used for anything but pressue.

astropy/units/physical.py
mhvk: I must admit I'm not convinced, especially not with the multiplication with a number: really, what is meant by `42 * PhysicalType('length')`? To me, the physical type is very strictly tied to units, and for units, `1/unit` is treated differently from `unit**-1` (quantities do not have physical type).

That said, the `1/phystype` example is something I could see myself doing...  So,  while my tendency would still be to start strict and not allow it, I am also OK with for now allowing just that one (and only integer 1, i.e., `if type(obj) is int and obj == 1`)

mhvk: p.s. In that case, `1` should also be allowed as input for `PhysicalType`. Indeed, I could live with allowing string inputs here, and thus just using the initializer to parse `obj`.  This would be consistent with what we do with units, where `u.m / 's'` is allowed.

astropy/units/physical.py
mhvk: I'm still curious what failed... Could you try?

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,455 @@
...
+    dictionary = {length: 42}
+    assert dictionary[length] == 42
+
+
+def test_physical_type_multiplication():
+    """
+    Test that multiplication of a physical type returns `NotImplemented`
+    when attempted for an invalid type.
+    """
+    assert length.__mul__([]) is NotImplemented
```
mhvk, line 357: Maybe test what this should cause? I.e., that `length * []` raises `TypeError`?

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,455 @@
...
+        """
+        Test that calling `def_physical_type` on an existing physical
+        type adds a new physical type name.
+        """
+        weird_name = "weird name"
+        strange_name = "strange name"
+
+        physical.def_physical_type(self.weird_unit, weird_name)
+        assert (
+                self.weird_unit.physical_type == weird_name
```
mhvk, line 418: Indent wrong?

mhvk (commented):
@namurphy - finally some time for what should be the last review. Small comments only, with perhaps only the argument against numbers something worth discussing.

astropy/units/physical.py
maxnoe: According to the wikipedia article, Pa is used for stress: https://en.wikipedia.org/wiki/Stress_(mechanics)

namurphy: Yeah, I ended up finding a lot of inconsistency on the World Wide Web™ about what pascals correspond to.  Some sites only said pressure, others said pressure & stress.  I decided to go with the [SI standard](https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9.pdf), which has the pascal representing "pressure, stress" (pdf page 137).

astropy/units/physical.py
namurphy: > That said, the `1/phystype` example is something I could see myself doing... So, while my tendency would still be to start strict and not allow it, I am also OK with for now allowing just that one (and only integer 1, i.e., `if type(obj) is int and obj == 1`)

Sure, that sounds like a fair compromise.  Definitely something I can live with!  On the chance that there are more use cases that come up in the future, this could be revisited.

> p.s. In that case, `1` should also be allowed as input for `PhysicalType`. Indeed, I could live with allowing string inputs here, and thus just using the initializer to parse `obj`. This would be consistent with what we do with units, where `u.m / 's'` is allowed.

Good ideas!

astropy/units/physical.py
mhvk: OK, I stand corrected!

astropy/units/physical.py
```diff
@@ -1,133 +1,469 @@
...
+    (si.m * si.s ** -5, "crackle"),
+    (si.m * si.s ** -6, {"pop", "pounce"}),
+    (si.K / si.m, "temperature gradient"),
+    (si.J / si.kg, "specific energy"),
+    (si.mol * si.m ** -3 * si.s ** -1, "reaction rate"),
+    (si.kg * si.m ** 2, "moment of inertia"),
+    (si.mol / si.s, "catalytic activity"),
+    (si.J * si.K ** -1 * si.mol ** -1, "molar heat capacity"),
+    (si.mol / si.kg, "molality"),
+    (si.m * si.s, ["absement", "sustained displacement"]),
```
nstarman, line 107: Should this be a set, for consistency?

nstarman (commented):
I'm really looking forward to this PR.
Minorest of nitpicks as I was skimming.

mhvk:
@namurphy - would be good to wrap this up! Note that it needs a rebase and also instead of the entry in `CHANGES.rst`, we'll need a changelog fragment in `docs/changes/units/11204.api.rst`

astropy/units/physical.py
```diff
@@ -484,7 +485,7 @@ def get_physical_type(unit):
 
     Returns
     -------
-    physical_type : PhysicalType
+    physical_type : `PhysicalType`
```
nstarman, line 488: Absolute paths are better for xref.

```suggestion
    physical_type : `~astropy.units.PhysicalType`
```

namurphy:
@mhvk — thank you for the reminder.  I've been having a bit more pandemic era brain blah than usual but this was good to get back to.  

Here are some substantive changes in response to code review.
 1. Operations between physical types and numbers are only allowed if the number is one.
 2. Similarly, `get_physical_type(1)` now returns a dimensionless physical type.
 3. Operations between `PhysicalType` instances and names of physical types are now allowed:
```Python
>>> length = u.meter.physical_type
>>> length * "time"
PhysicalType({'absement', 'sustained displacement'})
```
The main other substantive change is that `get_physical_type` will now work with any of the following: 

 - Units
 - Quantities
 - Strings representing names of physical types (e.g., `"length"`)
 - The number one
 - Objects with a `unit` attribute

I did this because to me it's more readable to do `get_physical_type("length")` than `u.m.physical_type`, and there are times when I would want to get the physical type of a `Quantity`.  This is probably how I'll approach this in the future.

If you have a chance for another code review, that would be much appreciated.  Thanks!

astropy/units/physical.py
namurphy: I think this is the last older comment I still need to address.  I'll try to get to it tomorrow.

astropy/units/physical.py
```diff
@@ -1,133 +1,536 @@
...
+
+
+def _get_names_in_use_except_from(unit):
+    """
+    Get a `set` containing all of the physical type names in use, except
+    for the names that correspond to ``unit``.
+    """
+    names_in_use = set(_unit_physical_mapping.keys())
+    id = unit._get_physical_type_id()
+    known_physical_type = id in _physical_unit_mapping.keys()
```
mhvk: This line and the new two could be combined in:
```
return names_in_use.difference(_physical_unit_mapping.get(id, {}))
```

astropy/units/physical.py
```diff
@@ -1,133 +1,536 @@
...
+        if attempting to name a unit as ``"unknown"``.
     """
-    r = unit._get_physical_type_id()
-    if r in _physical_unit_mapping:
+    physical_type_names = _standardize_physical_type_names(name)
+
+    if "unknown" in physical_type_names:
+        raise ValueError("cannot uniquely define an unknown physical type")
+
+    names_for_other_units = _get_names_in_use_except_from(unit)
```
mhvk: Actually, I think this could just become a one-liner instead of an additional function.

astropy/units/physical.py
```diff
@@ -1,133 +1,536 @@
...
-    (cgs.Franklin, 'electrical charge (ESU)'),
-    (cgs.statampere, 'electrical current (ESU)'),
-    (cgs.Biot, 'electrical current (EMU)'),
-    (cgs.abcoulomb, 'electrical charge (EMU)')
-]:
-    def_physical_type(unit, name)
+    if isinstance(obj, str):
+        return _physical_type_from_str(obj)
+
+    if hasattr(obj, "unit"):
```
mhvk, line 504: One alternative to checking for a `unit` attribute is simply to pass `obj` through `u.Quantity` and then take the `.unit` attribute. The net effect is to allow all numbers/arrays, but in this case I think that is fine: here, the user specifically asks for the physical type, so the input can be quite broad.  Anyway, concrete suggestion:
```
unit = u.Quantity(obj).unit
```

astropy/units/tests/test_physical.py
```diff
@@ -1,60 +1,476 @@
...
+    (u.K, u.imperial.deg_F),
+    (u.deg_C, u.imperial.deg_F),
+    (u.m ** 18, u.pc ** 18),
+]
+
+
+@pytest.mark.parametrize("unit1, unit2", equivalent_unit_pairs)
+def test_physical_type_id(unit1, unit2):
+    """Test that the physical type IDs of equivalent units match."""
+    id1 = physical.get_physical_type(unit1)._physical_type_id
```
mhvk, line 284: Do we really want to test the private `_physical_type_id` attribute rather than just the `PhysicalType` itself? I'm not quite sure what this adds to the tests two functions down.

mhvk (commented):
@namurphy - really nice! I push a quick extra commit with a simplification (which instantly invalidates two comments), but otherwise only a nitpick about the tests, and a suggestion for `get_physical_type`. Also, could you add some tests to complete the coverage?

Finally, I wondered whether we should not add something to the documentation proper - your docstring for `PhysicalType` would be a good start. It may make most sense, however, to leave that for a follow-up PR.

nstarman:
Nitpick and feel free to ignore. Many if not all the sets in this PR are never mutated. In the case, they could be `frozenset`.

astropy/units/physical.py
```diff
@@ -465,15 +467,17 @@ def get_physical_type(obj):
 
     Parameters
     ----------
-    obj : `~astropy.units.UnitBase`, `str`, number, or `~astropy.units.Quantity`
-        The object for which to retreive the physical type.  This object
-        should be a unit, the name of a physical type, the number one, a
-        `~astropy.units.Quantity`, or an `object` that has a ``unit``
-        attribute that is a unit.
+    obj
```
nstarman, line 470: when #11118 is merged, this can be 
```suggestion
    obj : quantity-like or `~astropy.units.PhysicalType`-like
```

astropy/units/physical.py
```diff
@@ -493,24 +497,27 @@ def get_physical_type(obj):
...
         return _physical_type_from_str(obj)
 
-    if hasattr(obj, "unit"):
-        unit = obj.unit
-    elif isinstance(obj, numbers.Real) and obj == 1:
-        unit = core.dimensionless_unscaled
-    else:
-        unit = obj
+    try:
+        unit = obj if isinstance(obj, core.UnitBase) else quantity.Quantity(obj).unit
```
nstarman, line 513:
```suggestion
        unit = obj if isinstance(obj, core.UnitBase) else quantity.Quantity(obj, copy=False).unit
```

astropy/units/physical.py
namurphy: Good idea!  Is there any problem with merging this pre-emptively?

astropy/units/physical.py
mhvk: @namurphy - did you ever figure out why one cannot use `return self._dimensional_analysis(other, "__rtruediv__")` here?

astropy/units/physical.py
mhvk: Should be no problem to just do this already - current sphinx would just render if a little differently.

astropy/units/physical.py
namurphy: I'm still trying to wrap my head around it, but one thing I noticed is that if we do: 
```Python
>>> u.m.__rtruediv__(1)
<Quantity 1. 1 / m>
```
then we get a `Quantity` which does not have a physical type defined.  In `PhysicalType._dimensional_analysis` the underlying operations are unit operations, and the code is expecting to get the physical type from the resulting unit, and it isn't able to do that from a `Quantity`.  I _think_ that's the reason why we need a workaround.

mhvk: Ah, yes, that's right: one has to do `unit ** -1` to get the inverse unit.

This may be solvable, but does not have to be in this PR! So, I'll resolve.

mhvk (approved):
I think this looks all OK now!  Let's get it in.  Thanks so much, @namurphy, both a nice addition and a nice collaboration!!

namurphy:
Agreed — this was really fun!  I learned a lot during the code review too.

pllim:
I think this broke the `pyinstaller` cron job. Example log: https://github.com/astropy/astropy/runs/2367552126?check_suite_focus=true

```
___ ERROR collecting .pyinstaller/astropy_tests/units/tests/test_physical.py ___
astropy_tests/units/tests/test_physical.py:27: in <module>
    (u.imperial.mi / u.week, "speed"),
E   AttributeError: module 'astropy.units' has no attribute 'imperial'
___ ERROR collecting .pyinstaller/astropy_tests/units/tests/test_physical.py ___
astropy_tests/units/tests/test_physical.py:27: in <module>
    (u.imperial.mi / u.week, "speed"),
E   AttributeError: module 'astropy.units' has no attribute 'imperial'
```

nstarman:
@pllim. Alternatively, for backward compatibility we could mod ``__contains__`` in PhysicalType..

pllim:
If I am the only one affected, there is no need, as I already patched it downstream. Not sure about other users. @karllark , are you affected by this change?

nstarman:
something like:

```python
def __contains__(self, other):
    contains = super().__contains__(other)
    if not contains:
        warnings.warn("deprecated", AstropyDeprecationWarning)  # ?
        contains |= any([other in x for x in self])
    return contains
```

karllark:
> If I am the only one affected, there is no need, as I already patched it downstream. Not sure about other users. @karllark , are you affected by this change?

I don't think this affects me.  I haven't used this capability to my knowledge.

rosteen:
For what it's worth, this also breaks [Jdaviz](https://github.com/spacetelescope/jdaviz) in a few places where we were using `unit.physical_type.title` to populate axis labels. We can fix it, but I wouldn't be surprised if there are other places throughout the ecosystem explicitly treating `physical_type` as a string.

mhvk:
Hmm, two reports when not even released makes me think that perhaps we should add something like
```
def __getattr__(self, attr):
    self_str_attr = getattr(str(self), attr, None)
    if self_str_attr is not None:
        warnings(... DeprecationWarning)
        return self_str_attr
    return super().__getattribute__()  # Raise AttributeError
```

With the deprecation error mentioning that if one relied on string attributes, one could just do `str(physical_type).attr` (which would be backwards compatible).

Or perhaps this is still rare enough that we can just put it in the docs (and the what's new entry)?

astrofrog:
Yes I agree that we should transparently make string methods work on the new physical type class, with a deprecation warning. I'm pretty sure I've also relied on physical_type being a string in some of my code.

namurphy:
I just created #11617 on addressing this.  Thank you for raising these issues, and my apologies for breaking things!

astrofrog:
Another downstream package whose tests are broken with the current astropy latest: https://github.com/radio-astro-tools/spectral-cube/pull/709 - is there any way to fix this? Could we perhaps make sure the hash is the same as the string hash?

astropy/units/physical.py
```diff
@@ -1,133 +1,530 @@
...
-"""
+"""Defines the physical types that correspond to different units."""
 
+import numbers
 
 from . import core
 from . import si
 from . import astrophys
 from . import cgs
-from . import imperial
```
pllim, line 17: Looking at #11975 , it is unclear to me now why this line was removed...

namurphy: Oh my.  This was a huge mistake on my part. I apologize for that. I have no idea how or why I let that slip in there....

maxnoe: Probably a tool warned "unused import"?

PR #11210, opened 2021-01-05
Allow Quantity to be initialized with a Distribution

mhvk:
This is really a spin-off from trying to get a general `Masked` class to work in #11127, where I wanted to ensuring that `Quantity(masked, unit)` gave a `MaskedQuantity`. But the changes I made suggested  `Quantity(distribution, unit)` should also give a `QuantityDistribution` - which indeed worked! I split the required changes out of #11127, since it is likely much easier to review here (and being able to do `distribution * unit` is nice!).

### DIFF
--- a/astropy/uncertainty/core.py
+++ b/astropy/uncertainty/core.py
@@ -4,6 +4,7 @@
 """
 Distribution class and associated machinery.
 """
+import builtins
 
 import numpy as np
 
@@ -275,14 +276,22 @@ class ArrayDistribution(Distribution, np.ndarray):
     # Override view so that we stay a Distribution version of the new type.
     def view(self, dtype=None, type=None):
         if type is None:
-            if issubclass(dtype, np.ndarray):
-                type = dtype
-                dtype = None
-            else:
-                raise ValueError('Cannot set just dtype for a Distribution.')
+            if not (isinstance(dtype, builtins.type)
+                    and issubclass(dtype, np.ndarray)):
+                raise ValueError('cannot view Distribution with just a new dtype.')
 
-        result = self.distribution.view(dtype, type)
-        return Distribution(result)
+            type = dtype
+            view_args = (type,)
+
+        else:
+            view_args = (dtype, type)
+
+        if issubclass(type, Distribution):
+            return super().view(*view_args)
+
+        else:
+            result = self.distribution.view(*view_args)
+            return Distribution(result)
 
     # Override __getitem__ so that 'samples' is returned as the sample class.
     def __getitem__(self, item):
--- a/astropy/uncertainty/core.py
+++ b/astropy/uncertainty/core.py
@@ -275,23 +275,30 @@ class ArrayDistribution(Distribution, np.ndarray):
 
     # Override view so that we stay a Distribution version of the new type.
     def view(self, dtype=None, type=None):
-        if type is None:
-            if not (isinstance(dtype, builtins.type)
-                    and issubclass(dtype, np.ndarray)):
-                raise ValueError('cannot view Distribution with just a new dtype.')
+        """New view of array with the same data.
 
+        Like `~numpy.ndarray.view` except that the result will always be a new
+        `~astropy.uncertainty.Distribution` instance.  If the requested
+        ``type`` is a `~astropy.uncertainty.Distribution`, then no change in
+        ``dtype`` is allowed.
+
+        """
+        if type is None and (isinstance(dtype, builtins.type)
+                             and issubclass(dtype, np.ndarray)):
             type = dtype
-            view_args = (type,)
+            dtype = None
 
-        else:
-            view_args = (dtype, type)
+        view_args = [item for item in (dtype, type) if item is not None]
 
-        if issubclass(type, Distribution):
+        if type is None or (isinstance(type, builtins.type)
+                            and issubclass(type, Distribution)):
+            if dtype is not None and dtype != self.dtype:
+                raise ValueError('cannot view as Distribution subclass with a new dtype.')
             return super().view(*view_args)
 
-        else:
-            result = self.distribution.view(*view_args)
-            return Distribution(result)
+        # View as the new non-Distribution class, but turn into a Distribution again.
+        result = self.distribution.view(*view_args)
+        return Distribution(result)
 
     # Override __getitem__ so that 'samples' is returned as the sample class.
     def __getitem__(self, item):
--- a/astropy/uncertainty/tests/test_distribution.py
+++ b/astropy/uncertainty/tests/test_distribution.py
@@ -1,9 +1,11 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 import numpy as np
+from numpy.testing import assert_array_equal
 import pytest
 
 from astropy import units as u
+from astropy.coordinates import Angle
 from astropy.uncertainty.core import Distribution
 from astropy.uncertainty import distributions as ds
 from astropy.utils import NumpyRNGContext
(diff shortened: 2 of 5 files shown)


astropy/units/quantity.py
```diff
@@ -373,7 +373,7 @@ def __new__(cls, value, unit=None, dtype=None, copy=True, order=None,
                     copy = False  # copy will be made in conversion at end
 
         value = np.array(value, dtype=dtype, copy=copy, order=order,
-                         subok=False, ndmin=ndmin)
+                         subok=True, ndmin=ndmin)
```
mhvk, line 376: Note: this could also be `subok=subok` so that `Quantity` initialization itself always returns a pure `Quantity` - on the other hand, it is not clear that getting a Quantity with a structured dtype out of a Distribution is useful for anything.

But I can change this if requested - most important I feel is that `distribution * unit` gives a `QuantityDistribution`.

mhvk:
@eteq, @astrofrog - would one of you have time to review this (or suggest someone else)? I'd like to try to get the `Masked` PR in a state where it touches as little else as possible, so that it is easier to review... And the suggestion here seems good on its own.

mhvk:
@eteq, @astrofrog - Gentle ping - quite simple PR and review would be nice...

astrofrog (commented):
I'll defer to @eteq for a proper review, but reading over the docs I realized we never actually spell out what ``QuantityDistribution`` is, do we?

astropy/uncertainty/core.py
```diff
@@ -276,11 +276,17 @@ class ArrayDistribution(Distribution, np.ndarray):
     def view(self, dtype=None, type=None):
         if type is None:
             if issubclass(dtype, np.ndarray):
-                type = dtype
-                dtype = None
+                if issubclass(dtype, Distribution):
+                    return super().view(dtype)
+                else:
+                    result = self.distribution.view(dtype)
+                    return Distribution(result)
```
nstarman, line 283: Nitpick and feel free to ignore. I'm just thinking of McCabe complexity. I endeavor to structure code to minimize the complexity score and reducing the number of return statements is an easy couple points.

astropy/uncertainty/tests/test_distribution.py
```diff
@@ -15,30 +17,42 @@
...
+    def setup_class(self):
+        self.rates = np.array([1, 5, 30, 400])[:, np.newaxis]
+        self.parr = np.random.poisson(self.rates, (4, 1000))
+        self.parr_t = np.random.poisson(self.rates.squeeze(), (1000, 4))
+
+    def test_numpy_init(self):
+        # Test that we can initialize directly from a Numpy array
+        Distribution(self.parr)
+
+    def test_numpy_init_T(self):
```
nstarman, line 31: Is the point here that we're instantiating from a view object? or is there something special about the transpose?

astropy/units/quantity.py
nstarman: Does this potentially open a problem if doing NDData * unit?

nstarman (commented):
Nifty 👍

astropy/uncertainty/tests/test_distribution.py
mhvk: Here, I am really just collecting the function tests that @eteq put in originally in a class...  I think the original reason is that a view as a new `.dtype` gets taken and one better be sure that that is possible.  It would be better to actually check that the output has the right shape and so on, but maybe I can be excused because this is existing code :smiley_cat:

astropy/units/quantity.py
mhvk: Since `NDData` is not an `ndarray` subclass, it will likely break regardless... (though perhaps `nddata` overrides `__mul__` - it is one piece of astropy I have not really interacted with...).

nstarman: It doesn't appear to. Just had a quick peak.

astropy/uncertainty/core.py
mhvk: Numpy's strange choice to allow one to omit `dtype` and just pass in a class remains a bit annoying... I've made it slightly simpler (and found a bug on the way, so good!).

astropy/uncertainty/tests/test_distribution.py
nstarman: GitHub needs a :shipit: (shipit) reaction. The rocket just doesn't suffice.

astropy/units/quantity.py
nstarman: That might be a good PR though... ``NDData * unit = NDData(..., unit=old_unit * unit)``

astropy/uncertainty/tests/test_distribution.py
```diff
@@ -15,30 +17,42 @@
...
+
+    def test_quantity_init_T(self):
+        # Test that we can initialize directly from a Quantity
+        pq = self.parr_t << u.ct
+        Distribution(pq.T)
+
+    def test_quantity_init_with_distribution(self):
+        # Test that we can initialize a Quantity from a Distribution.
+        pd = Distribution(self.parr)
+        qpd = pd << u.ct
```
eteq, line 51: This is really smooth! As an editorial note I can't help but think we need to get the `<<` syntax more widely known...

astropy/uncertainty/tests/test_distribution.py
```diff
@@ -374,3 +388,49 @@ def test_distr_noq_to_value():
...
+    assert isinstance(qd2, Distribution)
+    assert_array_equal(qd2.distribution, qd.distribution)
+    qd3 = ad.view(qd.dtype, qd.__class__)
+    assert not isinstance(qd3, Angle)
+    assert isinstance(qd3, u.Quantity)
+    assert isinstance(qd3, Distribution)
+    assert_array_equal(qd3.distribution, qd.distribution)
+
+    with pytest.raises(ValueError, match='just a new dtype'):
+        ad.view(np.dtype('f8'))
```
eteq, line 436: This is confusing to me. Is it that this works for a regular Angle but *not* one that uses `Distribution`? Or it doesn't work for *any* angle? If it's the former I think it's probably a bug (which we can fix later to not hold up this PR, but should make an issue and add a comment here) whereas if it's the latter this should probably go in the regular `Angle` tests.

docs/changes/uncertainty/11210.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+``Distribution`` instances can now be used as input to ``Quantity`` to
+initialize ``QuantityDistribution``.  Hence, ``distribution * unit``
```
eteq, line 2: Should this instead be
```suggestion
initialize ``QuantityDistribution``.  Hence, ``distribution << unit``
```
?

eteq (approved):
LGTM broad-brush!  I have a few minor inline comments, but nothing major!

I also noticed @astrofrog's https://github.com/astropy/astropy/pull/11210#pullrequestreview-615842789 - @astrofrog, were you thinking @mhvk should do that in this PR, or that we should do that separately?

astrofrog:
Follow up :)

astropy/uncertainty/tests/test_distribution.py
mhvk: Indeed!  Though I think the multiplication remains more intuitive and for most purposes the extra copy really doesn't matter.

nstarman: does ``pd <<= u.ct`` work in these instances to avoid a copy?

astropy/uncertainty/tests/test_distribution.py
mhvk: Indeed, it works for `Angle` but cannot very easily work for any `Distribution` (since it has a structured `dtype`) - also just don't quite know what to do with things that change the number of elements (say, `.view('f4')`) - in practice, this basically never happens...

Anyway, I'll add a comment.

astropy/uncertainty/tests/test_distribution.py
mhvk: `pd << u.ct` already does not make a copy (it calls `Quantity(pd, u.ct, copy=False)`). `pd <<= u.ct` does not do anything different, really, except that `pd` now points to the new `QuantityDistribution` object.

docs/changes/uncertainty/11210.bugfix.rst
mhvk: Both did not work before, in the sense that one did not get a `QuantityDistribution` out (but instead a `Quantity` with a structured dtype, which is less than useful:
```
In [7]: Distribution(np.ones((10,10)))
Out[7]: 
NdarrayDistribution([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]]) with n_samples=10

In [8]: Distribution(np.ones((10,10))) * u.m
Out[8]: 
<Quantity [([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],),
           ([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],)] m>
```

With this PR:
```
In [3]: Distribution(np.ones((10,10))) * u.m
Out[3]: 
<QuantityDistribution [[1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]] m with n_samples=10>
```

astropy/uncertainty/tests/test_distribution.py
mhvk: Now added a comment and restructured the .view code as well.

docs/changes/uncertainty/11210.bugfix.rst
mhvk: Now clarified.

Issue #10023, opened 2020-03-10
Add option to use first identifed reader from registry


aragilar:
This comes from https://github.com/astropy/specutils/issues/171 - currently specutils has a priority system built in to its wrapper of astropy.io.registry (it does dig into the internals of astropy.io.registry, but that's a separate thing). However, were I to do `Spectrum1d.read("test.fits")`, if there were multiple readers which could handle the file (the case I hit was the SDSS loader and the IRAF loader both are found to work), astropy/specutils would claim the format was ambiguous. If there was a option in `read` to specify to load the first reader found (which will be the one with the higher priority via specutils priority system), such as `use_best_format` or similar, `Spectrum1d.read("test.fits", use_best_format=True)` would load the file as an SDSS spectrum.

I'm happy to create a PR adding this, but I'd like to have some agreement on the API before creating it.

PR #11214, opened 2021-01-06
Idea for adding priorities to registry

aragilar:
This is one idea for how to handle priorities between readers/writers, as discussed in https://github.com/astropy/specutils/issues/171, https://github.com/astropy/specutils/issues/584 and https://github.com/astropy/astropy/issues/10023.

This does not yet have tests (happy to write some, but I think exactly what API is being provided to reader/writer developers needs to be discussed).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,7 +58,6 @@ astropy.io.registry
 - Readers and writers can now set a priority, to assist with resolving which
   format to use. [#11214]
 
-
 astropy.io.votable
 ^^^^^^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,13 @@ astropy.io.fits
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
+astropy.io.registry
+^^^^^^^^^^^^^^^^^^^
+
+- Readers and writers can now set a priority, to assist with resolving which
+  format to use. [#11214]
+
+
 astropy.io.votable
 ^^^^^^^^^^^^^^^^^^
 
--- a/astropy/io/registry.py
+++ b/astropy/io/registry.py
@@ -599,11 +599,46 @@ def _get_valid_format(mode, cls, path, fileobj, args, kwargs):
                               "The available formats are:\n"
                               "{}".format(format_table_str))
     elif len(valid_formats) > 1:
-        raise IORegistryError(
-            "Format is ambiguous - options are: {}".format(
-                ', '.join(sorted(valid_formats, key=itemgetter(0)))))
+        return _get_highest_priority_format(mode, cls, valid_formats)
 
     return valid_formats[0]
+
+
+def _get_highest_priority_format(mode, cls, valid_formats):
+    """
+    Returns the reader or writer with the highest priority. If it is a tie, or
+    no priorities set, error.
+    """
+    if mode == "read":
+        funcs = {format: get_reader(format, cls) for format in valid_formats}
+    elif mode == "write":
+        funcs = {format: get_writer(format, cls) for format in valid_formats}
+
+    has_priority = False
+    best_formats = []
+    current_priority = 0
+    for format, func, in funcs:
+        priority = getattr(func, "priority", None)
+        if priority is not None:
+            has_priority = True
+        else:
+            priority = 0
+        if priority == current_priority:
+            best_formats.append(format)
+        elif priority > current_priority:
+            best_formats = [format]
+            current_priority = priority
+
+    if has_priority:
+        if len(best_formats) == 1:
+            return funcs[best_formats[0]]
+        raise IORegistryError(
+            "Format priorities are tied - best formats are: {}".format(
+                ', '.join(sorted(valid_formats, key=itemgetter(0)))))
+
+    raise IORegistryError(
+        "Format is ambiguous - options are: {}".format(
+            ', '.join(sorted(valid_formats, key=itemgetter(0)))))
 
 
 class UnifiedReadWrite:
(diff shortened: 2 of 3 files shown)


pllim:
Hello and thanks! Due to the WIP nature of this PR, I have converted it to draft status. Feel free to click "Ready for review" once the other pieces are in place.

aragilar:
I was looking for some discussion of the design e.g. currently it uses priorities if any reader/writer has a priority, but maybe it should be a thing that is opted into by classes; or where/how should priorities be stored. I marked this as WIP to flag that is shouldn't be merged, but I would like reviews/opinions.

pllim:
Thanks for the clarification! Perhaps @taldcroft could advise or know who else to ping here.

taldcroft:
I think the idea is OK here. The implementation of attaching a `priority` attribute to the loader functions seems unorthodox / awkward. Since the current `_readers` and `_writers` registries are hidden, there is probably nothing preventing making the values there be a tuple `(function, priority)` instead of just `function`.  (Famous last words, but worth a try).

Then the `register_reader/writer` functions can get a new kwarg `priority=0`.

aragilar:
The priority on the loader functions comes from specutils, so I was trying to make their current system work, I don't we need to be constrained by what has been done in specutils, that could always be patched.

aragilar:
@taldcroft I've made the values for the reader/writer dict tuples as suggested, I think there should be no issues (the exception output users may see may changes slightly though, if that's considered a breaking change).

If this seems reasonable to people, I'll add tests and get the CI passing (once I've done that I'll make this not WIP/draft).

taldcroft:
Seems basically OK to me. If possible I'd like @astrofrog to weigh in on the idea of changing what's store in the private registry dicts to make sure I'm not missing anything.

astropy/io/registry.py
```diff
@@ -599,13 +611,40 @@ def _get_valid_format(mode, cls, path, fileobj, args, kwargs):
...
+    current_priority = 0
+    func = None
+    for format in valid_formats:
+        priority = format_dict[(format, cls)][1]
+
+        if priority == current_priority:
+            best_formats.append(format)
+        elif priority > current_priority:
+            best_formats = [format]
+            current_priority = priority
```
taldcroft, line 639: Need to set `func` here.

aragilar: Oops, I'll fix that.

astropy/io/registry.py
```diff
@@ -273,9 +281,13 @@ def register_writer(data_format, data_class, function, force=False):
         Whether to override any existing function if already present.
         Default is ``False``.
     """
+    if priority is None:
+        priority = getattr(func, "priority", None)
```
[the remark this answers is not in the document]
aragilar: I added this as a fallback for older specutils versions, happy to drop this if we're not going to support them though (setting the priority on the function was not going to be part of the official API anyway).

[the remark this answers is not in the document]
taldcroft: Yeah, I'd rather take it out and not have some specutils-only easter egg in there. :smile:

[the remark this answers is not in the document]
aragilar: Cool, I'll drop it then.

aragilar:
I'll add fixes based on the comments, add tests, and then move this to not-draft.

aragilar:
I've added the tests plus the suggestions from the reviews, let me know if there's any additional docs/changelog changes I need to add (I'm guessing this would go in dev branch, so doesn't need a change to `CHANGES.rst`?).

pllim:
Re: Change log -- maybe under "new feature" for 4.3

pllim:
Maybe also user facing doc but let's wait and see what @taldcroft says. Thanks!

CHANGES.rst
```diff
@@ -52,6 +52,13 @@ astropy.io.fits
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
+astropy.io.registry
+^^^^^^^^^^^^^^^^^^^
+
+- Readers and writers can now set a priority, to assist with resolving which
+  format to use. [#11214]
+
+
```
taldcroft, line 61: Remove the extra space.

astropy/io/registry.py
```diff
@@ -599,13 +607,46 @@ def _get_valid_format(mode, cls, path, fileobj, args, kwargs):
...
+            # this case better, instead maximally deprioritise the format.
+            priority = - np.inf
+
+        if priority == current_priority:
+            best_formats.append(format)
+        elif priority > current_priority:
+            best_formats = [format]
+            current_priority = priority
+
+    if len(best_formats) == 1:
```
taldcroft, line 643: Probably nicer to test for the failure condition and raise an exception, instead of the short-circuit return here.

astropy/io/registry.py
```diff
@@ -599,13 +607,46 @@ def _get_valid_format(mode, cls, path, fileobj, args, kwargs):
...
+        return best_formats[0]
+    raise IORegistryError(
+        "Format ambiguous, priorities are tied - best formats are: {}".format(
```
taldcroft, line 646: This exception message is basically giving the same idea as the previous one, but the different choice of words will break tests for other packages that might be defining their own readers. So I'd suggest just emitting the same exception message as before. (I.e. - whenever a change "breaks" existing astropy tests then you have to think about whether it will break other package's tests).

taldcroft (changes requested):
@aragilar - thanks, this basically looks good. These are just a few easy fixes and this should be good for merge. I don't think there is a need for updates to the narrative docs.

CHANGES.rst
aragilar: Done

astropy/io/registry.py
aragilar: Done

astropy/io/registry.py
aragilar: Sure, I've switched back to the old text.

taldcroft:
Merged, thanks @aragilar !!

PR #11217, opened 2021-01-06
Fix bug in Table init of TableAttribute's

taldcroft:
### Description

This pull request is to fix this bug reported by @Cadair on slack:
```
In [1]: from astropy.table import *                                             
In [2]: class MyTable(Table): 
   ...:     attr = TableAttribute() 
   ...:                                                                         
In [3]: t = MyTable(attr=2)                                                     
In [4]: t2 = MyTable(t, attr=3)                                                 
In [5]: t2.attr  # Should be 3
Out[5]: 2
```

The implementation of the fix got a bigger scope than I'd like, but in a good way. The table initialization code always had an early `return` for the special case of an empty table like `Table()`. I was never happy about it but had not previously seen a nice way to make this case fit in with the others. Today I found a simple way that makes the code processing take the same path for all initialization options. No new tests but the existing testing should be pretty robust.

### DIFF
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -817,7 +817,7 @@ class Table:
                         dtype = [dtype[name] for name in names]
                     except Exception:
                         raise ValueError('dtype was specified but could not be '
-                                        'parsed for column names')
+                                         'parsed for column names')
             # names is guaranteed to be set at this point
             init_func = self._init_from_list
             n_cols = len(names)
--- a/astropy/table/table.py
+++ b/astropy/table/table.py
@@ -717,13 +717,17 @@ class Table:
         default_names = None
 
         # Handle custom (subclass) table attributes that are stored in meta.
-        # These are defined as class attributes using the MetaAttribute
-        # descriptor.  Any such attributes get removed from kwargs here.
+        # These are defined as class attributes using the TableAttribute
+        # descriptor.  Any such attributes get removed from kwargs here and
+        # stored for use after the table is otherwise initialized. Any values
+        # provided via kwargs will have precedence over existing values from
+        # meta (e.g. from data as a Table or meta via kwargs).
+        meta_table_attrs = {}
         if kwargs:
             for attr in list(kwargs):
                 descr = getattr(self.__class__, attr, None)
                 if isinstance(descr, TableAttribute):
-                    setattr(self, attr, kwargs.pop(attr))
+                    meta_table_attrs[attr] = kwargs.pop(attr)
 
         if hasattr(data, '__astropy_table__'):
             # Data object implements the __astropy_table__ interface method.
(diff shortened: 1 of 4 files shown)


astropy/table/table.py
```diff
@@ -799,18 +803,21 @@ def __init__(self, data=None, masked=False, names=None, dtype=None,
         elif data is None:
             if names is None:
                 if dtype is None:
-                    if meta is not None:
-                        self.meta = deepcopy(meta) if copy else meta
```
taldcroft, line 803: This was previously repeating code from below and then jumping out early, making the code harder to follow.

astropy/table/table.py
```diff
@@ -799,18 +803,21 @@ def __init__(self, data=None, masked=False, names=None, dtype=None,
...
+                    names = []
+                else:
+                    try:
+                        # No data nor names but dtype is available.  This must be
+                        # valid to initialize a structured array.
+                        dtype = np.dtype(dtype)
+                        names = dtype.names
+                        dtype = [dtype[name] for name in names]
+                    except Exception:
+                        raise ValueError('dtype was specified but could not be '
```
mhvk, line 819: Probably good to write a few tests that cover unparsable dtype and one that does not have fields.

astropy/table/table.py
```diff
@@ -1163,6 +1176,11 @@ def _init_from_list(self, data, names, dtype, n_cols, copy):
         """Initialize table from a list of column data.  A column can be a
         Column object, np.ndarray, mixin, or any other iterable object.
         """
+        # Special case of initializing an empty table like `t = Table()`. No
```
mhvk, line 1179: Out of curiosity, where does it break to just let things go on with zero columns? Might it just take a `data = []` somewhere in the initializer to have everything else work?

mhvk (commented):
Agreed this has ended up as a nice improvement! Very small in-line comments.

astropy/table/table.py
taldcroft: Nothing breaks, at least from the test perspective. It just ends up running a bunch of code that does not need to run.  But more to the point I was being conservative in this bug fix patch to not change behavior. So putting in this test in `_init_from_list` made it easier for me to believe that the code is not doing anything differently than before. OTOH this new line is making the code path for `Table([])` different (albeit simpler and faster).

I'm always scared when hacking at core bits of `Table`, so that `if n_cols == 0` seems more conservative for the common use case of `Table()`, while `Table([])` is more ususual.

But I can go either way with convincing.

mhvk: I was mostly curious whether the rest was able to handle no columns, and it is good to know that it is! I might have gone the other way, but can see your argument too, so happy to go with this as is!

taldcroft:
@mhvk - thanks for the quick review. I added the tests you asked for (I think).

mhvk (approved):
Yes, that covers it (and beyond with the 'hello' version!). Merging...

PR #11224, opened 2021-01-08
Make classproperty and lazyproperty thread-safe

### LINKED ISSUE #11221: Lazy properties are not threadsafe
### Description
The `@lazyproperty` decorator and `@classproperty` decorator with `lazy=True` have race conditions when accessing the cache, which can lead to sporadic errors when using astropy from multithreaded code e.g. dask. One such case is with the Unit parser.

The Python 3.8 `cached_property` (mentioned in #9036) is thread-safe, and looking at the implementation it seems to use a double-checked locking pattern so the approach shouldn't hurt performance. Let me know if you'd like me to make a PR.

### Expected behavior
I expect to be able to parse unit strings safely from multi-threaded code.

### Actual behavior
Running the script below sometimes leads to this exception (I run the script in a loop in the shell and it triggers within a second):
```
  File "/home/bmerry/src/astropy/astropy/units/format/generic.py", line 562, in _do_parse
    return cls._parse_unit(s, detailed_exception=False)
  File "/home/bmerry/src/astropy/astropy/units/format/generic.py", line 485, in _parse_unit
    raise ValueError()
ValueError

  File "/home/bmerry/src/astropy/astropy/units/format/generic.py", line 565, in _do_parse
    return cls._parser.parse(s, lexer=cls._lexer, debug=debug)
  File "/home/bmerry/src/astropy/astropy/extern/ply/yacc.py", line 333, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/home/bmerry/src/astropy/astropy/extern/ply/yacc.py", line 1201, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/home/bmerry/src/astropy/astropy/extern/ply/yacc.py", line 192, in call_errorfunc
    r = errorfunc(token)
  File "/home/bmerry/src/astropy/astropy/units/format/generic.py", line 443, in p_error
    raise ValueError()
ValueError

  File "/home/bmerry/src/astropy/astropy/units/core.py", line 1850, in __call__
    return f.parse(s)
  File "/home/bmerry/src/astropy/astropy/units/format/generic.py", line 547, in parse
  File "/home/bmerry/src/astropy/astropy/units/format/generic.py", line 570, in _do_parse
    raise ValueError(f"Syntax error parsing unit '{s}'")
ValueError: Syntax error parsing unit 'm / (s)'

  File "./astropy-constants-threading.py", line 29, in <module>
    main()
  File "./astropy-constants-threading.py", line 26, in main
    raise excs[0]
  File "./astropy-constants-threading.py", line 11, in thread_func
    3 * u.Unit('m / (s)') + 1 * (u.km / u.s)
  File "/home/bmerry/src/astropy/astropy/units/core.py", line 1870, in __call__
    raise ValueError(msg)
ValueError: 'm / (s)' did not parse as unit: Syntax error parsing unit 'm / (s)' If this is meant to be a custom unit, define it with 'u.def_unit'. To have it recognized inside a file reader or other code, enable it with 'u.add_enabled_units'. For details, see https://docs.astropy.org/en/latest/units/combining_and_defining.html
```

I'm not totally sure that this is the cause (or is the only cause), but `astropy.units.format.generic.Generic` uses lazy class properties for the lexer and parser, and if several threads try to access the property at the same time, they could each try to build their own copy. I don't know about the lexer, but the comments at the top of astropy/extern/ply/yacc.py suggest that building parsers in parallel is a Bad Idea.

### Steps to Reproduce
Run this code in bash with `while ./astropy-constants-threading.py ; do :; done` until it errors.

```python
#!/usr/bin/env python3

import sys
import threading

import astropy.units as u


def thread_func(excs):
    try:
        u.Unit('m / (s)')
    except Exception as exc:
        excs.append(exc)


def main():
    sys.setswitchinterval(1e-6)
    excs = []
    threads = [threading.Thread(target=thread_func, args=(excs,))
               for i in range(32)]
    for thread in threads:
        thread.start()
    for thread in threads:
        thread.join()
    if excs:
        raise excs[0]


main()
```

### System Details
```
Linux-5.4.0-58-generic-x86_64-with-glibc2.29
[GCC 9.3.0]
Numpy 1.19.1
astropy 4.3.dev439+g4b242e6f5
Scipy 1.5.2
Matplotlib 3.3.0
```

bmerry:
### Description
This pull request is to address thread safety in the decorator utilities.

Fixes #11221.

I've also added unit tests. Testing race conditions is notoriously tricky, so I've made it try 10000 times each. That adds about 1s (for each), which is my attempt at a compromise between not slowing down the tests and trying enough times to trigger the bug. Without the fix it is still not enough to reliably fail, but it should be enough that if someone breaks it in future it will fail sooner or later in CI and prompt someone to look closer.

### DIFF
--- a/astropy/utils/tests/test_decorators.py
+++ b/astropy/utils/tests/test_decorators.py
@@ -8,7 +8,7 @@ import sys
 import pytest
 
 from astropy.utils.decorators import (deprecated_attribute, deprecated, wraps,
-                                      sharedmethod, classproperty,
+                                      sharedmethod, classproperty, lazyproperty,
                                       format_doc, deprecated_renamed_argument)
 from astropy.utils.exceptions import AstropyDeprecationWarning, AstropyUserWarning
 
@@ -567,6 +567,33 @@ def test_classproperty_lazy_threadsafe(fast_thread_switching):
             assert calls == 1
             assert values[0] is not None
             assert values == [values[0]] * workers
+
+
+def test_lazyproperty_threadsafe(fast_thread_switching):
+    """
+    Test thread safety of lazyproperty.
+    """
+    # This test is generally similar to test_classproperty_lazy_threadsafe
+    # above. See there for comments.
+
+    class A:
+        def __init__(self):
+            self.calls = 0
+
+        @lazyproperty
+        def foo(self):
+            self.calls += 1
+            return object()
+
+    workers = 8
+    with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
+        for p in range(10000):
+            a = A()
+            futures = [executor.submit(lambda: a.foo) for i in range(workers)]
+            values = [future.result() for future in futures]
+            assert a.calls == 1
+            assert a.foo is not None
+            assert values == [a.foo] * workers
 
 
 def test_format_doc_stringInput_simple():
--- a/astropy/utils/tests/test_decorators.py
+++ b/astropy/utils/tests/test_decorators.py
@@ -1,7 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import concurrent.futures
 import inspect
 import pickle
+import sys
 
 import pytest
 
@@ -526,6 +528,45 @@ def test_classproperty_docstring():
         foo = classproperty(_get_foo, doc="The foo.")
 
     assert B.__dict__['foo'].__doc__ == "The foo."
+
+
+@pytest.fixture
+def fast_thread_switching():
+    """Fixture that reduces thread switching interval.
+
+    This makes it easier to provoke race conditions.
+    """
+    old = sys.getswitchinterval()
+    sys.setswitchinterval(1e-6)
+    yield
+    sys.setswitchinterval(old)
+
+
+def test_classproperty_lazy_threadsafe(fast_thread_switching):
+    """
+    Test that a class property with lazy=True is thread-safe.
+    """
+    workers = 8
+    with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
+        # This is testing for race conditions, so try many times in the
+        # hope that we'll get the timing right.
+        for p in range(10000):
+
+            class A:
+                @classproperty(lazy=True)
+                def foo(cls):
+                    nonlocal calls
+                    calls += 1
+                    return object()
+
+            # Have all worker threads query in parallel
+            calls = 0
+            futures = [executor.submit(lambda: A.foo) for i in range(workers)]
+            # Check that only one call happened and they all received it
+            values = [future.result() for future in futures]
+            assert calls == 1
+            assert values[0] is not None
+            assert values == [values[0]] * workers
 
 
 def test_format_doc_stringInput_simple():
--- a/astropy/utils/tests/test_decorators.py
+++ b/astropy/utils/tests/test_decorators.py
@@ -3,7 +3,6 @@
 import concurrent.futures
 import inspect
 import pickle
-import sys
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


mhvk (approved):
@bmerry - very nice as well! I'm approving, though it may be good for @embray to have a look as well.

mhvk:
@embray, @bsipocz - for these thread-safety PRs there is a general question whether we should treat them as bugs to be backported. For now, I've left the milestone at 4.3 (consistent with where @bmerry put the changelog entry).

bmerry:
Thanks for the fast reviews. I have one more question: where should the `fast_thread_switching` fixture go so that it can be shared between tests in different modules? Is the top-level `astropy/conftest.py` appropriate?

bmerry:
I've been thinking about this a bit more: it's probably not necessary to take the locker in the setter and deleter, because if one thread is setting/deleting at the same time as another is accessing the property, the user's code has a race condition anyway. We just need to allow concurrent reads to be safe.

bmerry:
> I've been thinking about this a bit more: it's probably not necessary to take the lock in the setter and deleter

I've pushed a change for that.

mhvk:
Yes, not guarding the setter/deleter makes sense. As for the location, I'm not sure, maybe @embray has a suggestion... I'll admit that I've simply imported from another test function elsewhere, but it is obvious that eventually that is just going to be a hopeless mess.

astropy/utils/tests/test_decorators.py
```diff
@@ -528,6 +530,72 @@ def _get_foo(cls): return 1
     assert B.__dict__['foo'].__doc__ == "The foo."
 
 
+@pytest.fixture
+def fast_thread_switching():
+    """Fixture that reduces thread switching interval.
+
+    This makes it easier to provoke race conditions.
+    """
+    old = sys.getswitchinterval()
```
pllim, line 539: Same question as https://github.com/astropy/astropy/pull/11226#discussion_r555412590 . That is, define this in `setup_module` and `teardown_module` in this case.

pllim (commented):
`classproperty` is used in `io.fits` and `coordinates` too, so maybe @saimn and @adrn want to have a look as well.

As for `lazyproperty`, it just have to not break existing code until we can drop Python 3.7 and then switch to using `cached_property` (#9036).

I also wonder if we need a temporary commit to enable running "exotic" architectures and also pyinstaller in this PR (and other PRs on the topic of multi-threading). They are defined in other YAML files inside `.github/workflows`.

astropy/utils/tests/test_decorators.py
bmerry: I don't think it's appropriate to use for the entire module. It will likely slow down the rest of the tests in the module.

I'm also expecting that we'll find more thread safety issues in other modules and write more regression tests, so avoiding duplication of this code seems worthwhile (particularly since I don't yet know what the best value for the magic number is).

bmerry:
I've moved the fast_thread_switching fixture to astropy/conftest.py. There is already another fixture in there (ignore_matplotlibrc) so it seemed like a reasonable place for it.

bmerry:
> I also wonder if we need a temporary commit to enable running "exotic" architectures

What is the motivation for that? Note that the thread safety issues I've been fixing are all in Python code, which means that the threads aren't really running in parallel (they're serialised by the GIL), and hence ISA memory models are unlikely to have any impact.

saimn:
Sounds good to me :+1:

mhvk (approved):
The thing I've never liked much about pytest fixtures in `conftest` is that it becomes quite hard to find out where something is defined. But I don't really have a better suggestion and at least the na me of the fixture does tell what it does... Does anyone else have a better suggestion?  If not, I suggest we merge this - it is nice!

pllim (approved):
Yeah, `conftest.py` seems like a good place as any. 👍 

As for the exotic arch, it's more of a sanity check but if we don't do it here, we'll find out soon enough when the cron runs, and I know who to ping. 😉 

Let's get this in then, so people have plenty of time to test against this in 4.3.dev before feature freeze time.

Thank you, all!

embray:
@bmerry Thanks for this.  `lazyproperty` was of course written many years ago, long before Python added `cached_property`.  As not a lot of people were using astropy in multi-threaded code it was never intended to be thread-safe, but I'm glad you caught this and fixed it.

PR #11226, opened 2021-01-08
Make UfuncHelpers thread-safe

### LINKED ISSUE #11220: UfuncHelpers is not thread-safe
### Description
The UfuncHelpers class is not thread-safe, and this can occasionally lead to strange errors when using astropy from multi-threaded code (e.g. with Dask).

### Expected behavior
Multiple threads should be able to call ufuncs simultaneously.

### Actual behavior
Errors like this sporadically occur:
```
  File "./astropy-ufunc-threading.py", line 13, in thread_code
    erfa.s2c(1 * u.rad, 2 * u.rad)
  File "/home/bmerry/work/sdp/env3/lib/python3.8/site-packages/erfa/core.py", line 19788, in s2c
    c = ufunc.s2c(theta, phi)
  File "/home/bmerry/src/astropy/astropy/units/quantity.py", line 457, in __array_ufunc__
    converters, unit = converters_and_unit(function, method, *inputs)
  File "/home/bmerry/src/astropy/astropy/units/quantity_helper/converters.py", line 157, in converters_and_unit
    ufunc_helper = UFUNC_HELPERS[function]
  File "/home/bmerry/src/astropy/astropy/units/quantity_helper/converters.py", line 86, in __missing__
    raise TypeError("unknown ufunc {}.  If you believe this ufunc "
TypeError: unknown ufunc s2c.  If you believe this ufunc should be supported, please raise an issue on https://github.com/astropy/astropy
```

People have also seen errors about collections mutating while being iterated, which I suspect is closely related.

### Steps to Reproduce
It's a little tricky to reproduce reliably since it relies on thread timing. There's a script below, but I've found the easiest way to make it reliable is to add sleeps into the UfuncHelpers code at strategic points.

```python
#!/usr/bin/env python3

import threading
import time

import astropy.units as u
import erfa


def thread_code(excs):
    try:
        erfa.s2c(1 * u.rad, 2 * u.rad)
    except Exception as exc:
        excs.append(exc)


def main():
    excs = []
    threads = [threading.Thread(target=thread_code, args=(excs,))
               for i in range(32)]
    for thread in threads:
        thread.start()
    for thread in threads:
        thread.join()
    if excs:
        raise excs[0]


main()
```

The races I can immediate see are:
- The `modules` property has a race between checking if there is a `_modules` attribute and returning it. That's probably harmless in most cases since the first registration will likely be done when first importing astropy and before launching threads.
- The call to `list(self.modules.items())` in `__missing__` iterates the modules dictionary to construct the list, and at the same time another thread might mutate it (possibly with `register_module`, but more likely with `import_module` which pops).
- `import_module` removes the module from the list before updating the helper dictionary. If another thread calls `__missing__` in that gap, it won't be able to find a helper (leading to the exception above).
- `__setitem__` modifies both `UNSUPPORTED` and the helper dictionary. There are probably some race conditions if `__missing__` is called in the middle of that process.

### Recommendations

I think the following should address most of the issues. It still depends on Python treating accesses to the core dict class as atomic, which as far as I know is an implementation detail of CPython but not guaranteed anywhere:
1. Add a lock, that is held during all the member functions. That shouldn't affect the common case (a ufunc for which the helper is already registered) because this `__getitem__` won't take the lock.
2. Give the class a constructor and initialise `modules` there, instead of doing it lazily with a property.

Let me know if you'd like a PR.

### System Details

Linux-5.4.0-58-generic-x86_64-with-glibc2.29
[GCC 9.3.0]
Numpy 1.19.1
astropy 4.3.dev439+g4b242e6f5
Scipy 1.5.2
Matplotlib 3.3.0

bmerry:
### Description

- Remove `modules` property that lazily creates the underlying
  `_modules` attribute, in favour of an `__init__` that creates
  `modules` as an attribute.
- Add a lock that is held for all mutations of the object.
- Moved UNSUPPORTED from a class attribute to an instance attribute,
  for consistency. This is a singleton class and it's accessed via
  the `UNSUPPORTED_UFUNCS` global so this shouldn't make any difference
  other than making it easier to isolate unit tests.

Fixes #11220.

### DIFF
--- a/astropy/units/quantity_helper/converters.py
+++ b/astropy/units/quantity_helper/converters.py
@@ -2,6 +2,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Converters for Quantity."""
 
+import threading
+
 import numpy as np
 
 from astropy.units.core import (UnitsError, UnitConversionError, UnitTypeError,
@@ -21,6 +23,11 @@ class UfuncHelpers(dict):
     """
     UNSUPPORTED = set()
 
+    def __init__(self, *args, **kwargs):
+        self.modules = {}
+        self._lock = threading.RLock()
+        super().__init__(*args, **kwargs)
+
     def register_module(self, module, names, importer):
         """Register (but do not import) a set of ufunc helpers.
 
@@ -35,15 +42,9 @@ class UfuncHelpers(dict):
             keyed by those ufuncs.  If the value is `None`, the ufunc is
             explicitly *not* supported.
         """
-        self.modules[module] = {'names': names,
-                                'importer': importer}
-
-    @property
-    def modules(self):
-        """Modules for which helpers are available (but not yet loaded)."""
-        if not hasattr(self, '_modules'):
-            self._modules = {}
-        return self._modules
+        with self._lock:
+            self.modules[module] = {'names': names,
+                                    'importer': importer}
 
     def import_module(self, module):
         """Import the helpers from the given module using its helper function.
@@ -53,8 +54,9 @@ class UfuncHelpers(dict):
         module : str
             Name of the module. Has to have been registered beforehand.
         """
-        module_info = self.modules.pop(module)
-        self.update(module_info['importer']())
+        with self._lock:
+            module_info = self.modules.pop(module)
+            self.update(module_info['importer']())
 
     def __missing__(self, ufunc):
         """Called if a ufunc is not found.
@@ -62,25 +64,30 @@ class UfuncHelpers(dict):
         Check if the ufunc is in any of the available modules, and, if so,
         import the helpers for that module.
         """
-        if ufunc in self.UNSUPPORTED:
-            raise TypeError(f"Cannot use ufunc '{ufunc.__name__}' with quantities")
+        with self._lock:
+            # Check if it was loaded while we waited for the lock
+            if ufunc in self:
+                return self[ufunc]
 
-        for module, module_info in list(self.modules.items()):
-            if ufunc.__name__ in module_info['names']:
-                # A ufunc with the same name is supported by this module.
-                # Of course, this doesn't necessarily mean it is the
-                # right module. So, we try let the importer do its work.
-                # If it fails (e.g., for `scipy.special`), then that's
-                # fine, just raise the TypeError.  If it succeeds, but
-                # the ufunc is not found, that is also fine: we will
-                # enter __missing__ again and either find another
-                # module or get the TypeError there.
-                try:
-                    self.import_module(module)
-                except ImportError:
-                    pass
-                else:
-                    return self[ufunc]
+            if ufunc in self.UNSUPPORTED:
+                raise TypeError(f"Cannot use ufunc '{ufunc.__name__}' with quantities")
+
+            for module, module_info in list(self.modules.items()):
+                if ufunc.__name__ in module_info['names']:
+                    # A ufunc with the same name is supported by this module.
+                    # Of course, this doesn't necessarily mean it is the
+                    # right module. So, we try let the importer do its work.
+                    # If it fails (e.g., for `scipy.special`), then that's
+                    # fine, just raise the TypeError.  If it succeeds, but
+                    # the ufunc is not found, that is also fine: we will
+                    # enter __missing__ again and either find another
+                    # module or get the TypeError there.
+                    try:
+                        self.import_module(module)
+                    except ImportError:
+                        pass
+                    else:
+                        return self[ufunc]
 
         raise TypeError("unknown ufunc {}.  If you believe this ufunc "
                         "should be supported, please raise an issue on "
@@ -92,12 +99,13 @@ class UfuncHelpers(dict):
         # mean that something is not implemented, but this means an
         # extra if clause for the output, slowing down the common
         # path where a ufunc is supported.
-        if value is None:
-            self.UNSUPPORTED |= {key}
-            self.pop(key, None)
-        else:
-            super().__setitem__(key, value)
-            self.UNSUPPORTED -= {key}
+        with self._lock:
+            if value is None:
+                self.UNSUPPORTED |= {key}
+                self.pop(key, None)
+            else:
+                super().__setitem__(key, value)
+                self.UNSUPPORTED -= {key}
 
 
 UFUNC_HELPERS = UfuncHelpers()
--- a/astropy/units/quantity_helper/converters.py
+++ b/astropy/units/quantity_helper/converters.py
@@ -21,10 +21,10 @@ class UfuncHelpers(dict):
 
     Such modules should be registered using ``register_module``.
     """
-    UNSUPPORTED = set()
 
     def __init__(self, *args, **kwargs):
         self.modules = {}
+        self.UNSUPPORTED = set()   # Upper-case for backwards compatibility
         self._lock = threading.RLock()
         super().__init__(*args, **kwargs)
 
--- a/astropy/units/quantity_helper/converters.py
+++ b/astropy/units/quantity_helper/converters.py
@@ -84,7 +84,7 @@ class UfuncHelpers(dict):
                     # module or get the TypeError there.
                     try:
                         self.import_module(module)
-                    except ImportError:
+                    except ImportError:  # pragma: no cover
                         pass
                     else:
                         return self[ufunc]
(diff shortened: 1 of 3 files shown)


bmerry:
I wouldn't merge this exactly as is, because I copied-and-pasted a fixture from #11224. Once that is merged it should probably go to a shared location - I'm not sure what that should be.

bmerry:
I'm not sure why astropy-bot says the changelog entry is missing: I'm pretty sure I added it, but let me know if I missed something.

mhvk (approved):
@bmerry - and thanks a third time! This looks good too. I was a bit hesitant about adding complexity, but in the end this is pretty deep in the bowels and central to quantities, so it seems well worth it. And the comments will help anyone looking in the future.

I'll approve now but will wait with merging until the other PRs have had a few more eyes, and we've decided on whether or not to backport.

astropy/units/quantity_helper/converters.py
```diff
@@ -53,34 +54,40 @@ def import_module(self, module):
...
+                    # Of course, this doesn't necessarily mean it is the
+                    # right module. So, we try let the importer do its work.
+                    # If it fails (e.g., for `scipy.special`), then that's
+                    # fine, just raise the TypeError.  If it succeeds, but
+                    # the ufunc is not found, that is also fine: we will
+                    # enter __missing__ again and either find another
+                    # module or get the TypeError there.
+                    try:
+                        self.import_module(module)
+                    except ImportError:
```
pllim, line 87: If it is impossible to write a test to cover `ImportError` case, `# pragma: no cover` will tell the coverage checker that this is not tested on purpose.
```suggestion
                    except ImportError:  # pragma: no cover
```

astropy/units/quantity_helper/converters.py
```diff
@@ -92,16 +99,17 @@ def __setitem__(self, key, value):
...
+                self.unsupported |= {key}
+                self.pop(key, None)
+            else:
+                super().__setitem__(key, value)
+                self.unsupported -= {key}
 
 
 UFUNC_HELPERS = UfuncHelpers()
-UNSUPPORTED_UFUNCS = UFUNC_HELPERS.UNSUPPORTED
+UNSUPPORTED_UFUNCS = UFUNC_HELPERS.unsupported
```
pllim, line 112: Will replacing `UNSUPPORTED` with lowercase version break people's code? Do people use `UNSUPPORTED` from this class directly? Perhaps it is a smaller price to pay to break some PEP 8 and do a `self.UNSUPPORTED |= ...` and such. 💭

astropy/units/tests/test_quantity_ufuncs.py
```diff
@@ -52,6 +56,21 @@ def test_testwarn(tw):
         tw.f(*tw.q_in)
 
 
+@pytest.fixture
+def fast_thread_switching():
+    """Fixture that reduces thread switching interval.
+
+    This makes it easier to provoke race conditions.
+
+    TODO: this is copied from another branch which adds it elsewhere. It
```
pllim, line 65: Well... Should we work to get that other PR merged first and then rebase this one on top of upstream `master` after that is merged? If so, please turn this PR into draft status and put a note on the top post to indicate that this PR is dependent on that other PR. (GitHub still has no proper linking like JIRA.) 

But that said, is a new fixture really necessary? I can we use `setup_class` and `teardown_class` instead within `TestUfuncHelpers`?

astropy/units/quantity_helper/converters.py
bmerry: It should be theoretically possible to test, but I didn't write this bit of code (it's in the diff because the indentation level changed). Do you still want the pragma added, or would you rather leave it as is to remind someone else to test it?

astropy/units/quantity_helper/converters.py
bmerry: You're right, I'd missed that users could directly access it as UFUNC_HELPERS.UNSUPPORTED or UfuncHelpers.UNSUPPORTED rather than through this alias, even though all the astropy code uses the alias. I'll change it back to the way it was before (uppercase, class attribute) and think up some way to isolate it in the test.

bmerry: Done.

astropy/units/quantity_helper/converters.py
pllim: Adding a "pragma" makes it easier for future devs to find it, as it marks where the line isn't covered, so my vote is to add it. 😄

mhvk: This would be triggered if someone tried to run a ufunc with the same name as a scipy one on a system without scipy... I think adding the pragma is fine...

astropy/units/tests/test_quantity_ufuncs.py
pllim: #11224 is merged, so this is moot now. Please rebase and use your fixture from `conftest.py`. Thanks! 😸

bmerry: I've rebased - the only difference should be that the fast_thread_switching fixture is gone as it was added by #11224.

astropy/units/tests/test_quantity_ufuncs.py
```diff
@@ -88,6 +92,30 @@ def test_removal_addition(self):
...
+            return np.sqrt(*args, **kwargs)
+
+        def register():
+            return {dummy_ufunc: helper_sqrt}
+
+        workers = 8
+        with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
+            for p in range(10000):
+                helpers = UfuncHelpers()
+                # UNSUPPORTED is (for some reason) a class attribute rather than an
```
mhvk, line 106: Hmm, what this solved by your lower-case version?  Sorry to only realize now, but it would seem to make sense to just initialize that in `__init__` as well.

astropy/units/tests/test_quantity_ufuncs.py
```diff
@@ -88,6 +92,30 @@ def test_removal_addition(self):
         assert np.add in qh.UFUNC_HELPERS
         assert np.add not in qh.UNSUPPORTED_UFUNCS
 
+    def test_thread_safety(self, fast_thread_switching):
+        def dummy_ufunc(*args, **kwargs):
```
mhvk, line 96: Would it make sense to just create a new helper instance here, so that we are sure the test does not change the main `UFUNC_HELPERS`?

mhvk (commented):
Some small wonderings. Not essential, but let me know what you think.

pllim:
Ops, looks like there is a conflict in the change log too. 😬

astropy/units/tests/test_quantity_ufuncs.py
bmerry: That is what I do, a few lines further down (`helpers = UfuncHelpers()`).

astropy/units/tests/test_quantity_ufuncs.py
bmerry: Yes, that's why I moved it from class attribute to instance attribute. I moved it back due to concern that it would break code that uses `UfuncHelpers.UNSUPPORTED`, although I've just searched Github and not found any hits for "UfuncHelpers" outside of astropy source code so it may be an unwarranted concern.

Looking at it again, is the `UfuncHelpers` name considered to be public API? It doesn't have an underscore, but it's also not listed in `__all__`.

bmerry:
> Ops, looks like there is a conflict in the change log too. grimacing

I've rebased to fix that. Let me know what you want to do about UNSUPPORTED (class or instance attribute).

astropy/units/tests/test_quantity_ufuncs.py
pllim: I'll defer to @mhvk here for a design decision. While you couldn't find usage on GitHub, doesn't mean people are not using it in their private code. But if we do decide to make break (pun?) for this, then we'll find out soon enough if people complain about it after the next release.

mhvk: My suggestion would be to leave it as upper case, but make it an instance property. Then, if anybody uses `UFUNC_HELPERS.UNSUPPORTED`, it will still work fine.

astropy/units/tests/test_quantity_ufuncs.py
mhvk: Duh...

astropy/units/tests/test_quantity_ufuncs.py
mhvk: p.s. So a bit in between - I don't think it is likely anybody would use the class, while I think it is quite possible they might have used the instance.

bmerry: Ok, I'll change it to an instance attribute tomorrow.

bmerry: Done.

mhvk (approved):
Super, thanks so much!

PR #11228, opened 2021-01-08
Add a get method to table's Row class

Cadair:
I came across wanting this in sunpy so I thought it might be useful to more than just me.

### DIFF
--- a/astropy/table/row.py
+++ b/astropy/table/row.py
@@ -167,6 +167,14 @@ class Row:
     def dtype(self):
         return self._table.dtype
 
+    def get(self, key, default=None):
+        """
+        Extract a value from the row if the key is present otherwise return the value of ``default``
+        """
+        if key in self.colnames:
+            return self[key]
+        return default
+
     def _base_repr_(self, html=False):
         """
         Display row as a single-line table but with appropriate header line.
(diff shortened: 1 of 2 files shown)


pllim:
Need a change log entry and user-facing doc.

mhvk:
I think it makes sense to make the interface of `Row` more dict-like, but @taldcroft should definitely have a look. 

I guess one question is if we should do the same for `Table` itself. (Most other dict methods make much less sense...)

astropy/table/row.py
```diff
@@ -167,6 +167,14 @@ def colnames(self):
     def dtype(self):
         return self._table.dtype
 
+    def get(self, key, default=None):
+        """
+        Extract a value from the row if the key is present otherwise return the value of ``default``
+        """
+        if key in self.colnames:
```
taldcroft, line 174: This probably needs to be more like your version on slack with a try/except, because `row[key]` can accept more than just column names. I.e. column number or a tuple of names work.

taldcroft (changes requested):

taldcroft:
@cadair - if you want this in for 4.3 then there were some changes requested. Otherwise re-milestone to 5.0.

pllim:

PR #11240, opened 2021-01-12
Introduce a special function for unit scale conversion.

mhvk:
This allows one to remove duplication of functionality in `quantity_helpers`.

Note that the commit also removes a test that uses a private method. Contrary to the description of the test, it no longer does anything recommended in the documentation, so I thought it was better to just remove the test.

p.s. This is factored out from a branch introducing structured units and quantities, which are needed to be able to properly deal with ERFA ufuncs. It seemed useful on its own. Note that it is not essential for that branch, so if on balance slightly changing the private method is seen as too disruptive, this can be dropped.

### DIFF
--- a/astropy/units/core.py
+++ b/astropy/units/core.py
@@ -886,7 +886,12 @@ class UnitBase:
             f"{unit_str} and {other_str} are not convertible")
 
     def _get_converter(self, other, equivalencies=[]):
-        other = Unit(other)
+        """Get a converter for values in ``self`` to ``other``.
+
+        If no conversion is necessary, returns ``unit_scale_converter``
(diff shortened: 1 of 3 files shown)


astropy/units/core.py
```diff
@@ -886,15 +886,23 @@ def get_err_str(unit):
             f"{unit_str} and {other_str} are not convertible")
 
     def _get_converter(self, other, equivalencies=[]):
-        other = Unit(other)
```
mhvk, line 889: Removal of this is helpful to ensure the use in ``quantity_helpers`` remains as fast as possible.

adrn (approved):
Seems fine to me! Always good to try to reduce duplication...

PR #10081, opened 2020-03-29
Add TimeUnixTai (unix_tai) time format class

taldcroft:
### Description

This PR adds a new time format `unix_tai` which is roughly Unix time (`unix` format) but with leap seconds included.

Closes #10052

PR #11249, opened 2021-01-14
Change epoch of TimeUnixTAI to 1970-01-01T00:00:00.0 TAI, fixes #11245

### LINKED ISSUE #11245: The unix_tai time format appears to be off by 8 seconds
### Description

The zero point for the new time format `unix_tai` appears to be off by 8 seconds.

### Expected behavior

The new time format `unix_tai` is defined to match the `CLOCK_TAI` Linux real-time clock (and I was looking forward to using it, for that reason).

For a current date  I expect CLOCK_TAI  - CLOCK_UTC = 37 seconds and that is what I observe on a linux system whose leap second table is properly maintained by ntp.
When I try the same computation using Astropy I get 29 seconds.

### Actual behavior

Astropy.time seems to believe that CLOCK_TAI - CLOCK_UTC is 29 seconds for current dates. That is what is shown as an example in your docs and is coded in the unit test, so you clearly have a reason for this. But I don't see how it is compatible with CLOCK_TAI.

### Steps to Reproduce

1: Find a linux system that has a current leap second table (linux is the only OS I know of that supports the CLOCK_TAI real time clock).
2: Use a reasonably recent Python 3 (I use 3.7)
3: Use AstroPy 4.1 or the current dev version

```python
import time
import astropy.time
clock_tai = getattr(time, "CLOCK_TAI", 11)
now_utc_unix = time.time()
now_tai_unix = time.clock_gettime(clock_tai)
astropy_now_utc = astropy.time.Time(now_utc_unix, format="unix")
print(now_tai_unix - now_utc_unix)
print(astropy_now_utc.unix_tai - now_utc_unix)
```

### System Details

Linux-3.10.0-1062.12.1.el7.x86_64-x86_64-with-centos-7.7.1908-Core
Python 3.7.8 | packaged by conda-forge | (default, Jul 23 2020, 03:54:19) 
[GCC 7.5.0]
Numpy 1.19.1
astropy 4.3.dev452+gc3132e70a
Scipy 1.5.2
Matplotlib 3.0.3

maxnoe:
* TimeUnixTAI (name unix_tai) used the epoch ``1970-01-01T00:00:00 UTC``
  since its introduction in astropy 4.2, however, contrary to the
  documented behaviour, this is not the definition of CLOCK_TAI on linux
  and no system using this epoch is known.

* This switches the epoch to ``1970-01-01T00:00:00.0 TAI`` to match the
  intended defintion of being compatible to CLOCK_TAI on linux and also
  is the definition of the now obselete and not yet released
  ``TimeUnixPTP``, which is removed in this commit.

Fixes #11245

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -381,6 +381,12 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- Change epoch of ``TimeUnixTAI`` (``"unix_tai"``) from ``1970-01-01T00:00:00 UTC``
+  to ``1970-01-01T00:00:00 TAI`` to match the intended and documented behaviour.
+  This essentially changes the resulting times by 8.000082 seconds, the initial
+  offset between TAI and UTC. [#11249]
+
+
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
 
@@ -409,6 +415,11 @@ astropy.wcs
 Other Changes and Additions
 ---------------------------
 
+
+- Change epoch of ``TimeUnixTAI`` (``"unix_tai"``) from ``1970-01-01T00:00:00 UTC``
+  to ``1970-01-01T00:00:00 TAI`` to match the intended and documented behaviour.
+  This essentially changes the resulting times by 8.000082 seconds, the initial
+  offset between TAI and UTC. [#11249]
 
 
 4.2 (2020-11-24)
--- a/astropy/time/formats.py
+++ b/astropy/time/formats.py
@@ -690,7 +690,7 @@ class TimeFromEpoch(TimeNumeric):
 
 class TimeUnix(TimeFromEpoch):
     """
-    Unix time: seconds from 1970-01-01 00:00:00 UTC, ignoring leap seconds.
+    Unix time (UTC): seconds from 1970-01-01 00:00:00 UTC, ignoring leap seconds.
 
     For example, 946684800.0 in Unix time is midnight on January 1, 2000.
 
@@ -710,38 +710,53 @@ class TimeUnix(TimeFromEpoch):
 
 class TimeUnixTai(TimeUnix):
     """
-    Seconds from 1970-01-01 00:00:08 TAI (see notes), including leap seconds.
+    Unix time (TAI): SI seconds elapsed since 1970-01-01 00:00:00 TAI (see caveats).
 
-    This will generally differ from Unix time by the cumulative integral number
-    of leap seconds since 1970-01-01 UTC.  This convention matches the definition
-    for linux CLOCK_TAI (https://www.cl.cam.ac.uk/~mgk25/posix-clocks.html).
+    This will generally differ from standard (UTC) Unix time by the cumulative
+    integral number of leap seconds introduced into UTC since 1972-01-01 UTC
+    plus the initial offset of 10 seconds at that date.
+
+    This convention matches the definition of linux CLOCK_TAI
+    (https://www.cl.cam.ac.uk/~mgk25/posix-clocks.html),
+    and the Precision Time Protocol
+    (https://en.wikipedia.org/wiki/Precision_Time_Protocol), which
+    is also used by the White Rabbit protocol in High Energy Physics:
+    https://white-rabbit.web.cern.ch.
 
     Caveats:
 
     - Before 1972, fractional adjustments to UTC were made, so the difference
       between ``unix`` and ``unix_tai`` time is no longer an integer.
     - Because of the fractional adjustments, to be very precise, ``unix_tai``
-      is the number of seconds since ``1970-01-01 00:00:08 TAI`` or equivalently
-      ``1969-12-31 23:59:59.999918 UTC``.  The difference between TAI and UTC
+      is the number of seconds since ``1970-01-01 00:00:00 TAI`` or equivalently
+      ``1969-12-31 23:59:51.999918 UTC``.  The difference between TAI and UTC
       at that epoch was 8.000082 sec.
-    - On the day of a leap second the difference between ``unix`` and ``unix_tai``
-      times increases linearly through the day by 1.0.  See also the
+    - On the day of a positive leap second the difference between ``unix`` and
+      ``unix_tai`` times increases linearly through the day by 1.0. See also the
       documentation for the `~astropy.time.TimeUnix` class.
+    - Negative leap seconds are possible, though none have been needed to date.
 
     Examples
     --------
 
+      >>> # get the current offset between TAI and UTC
       >>> from astropy.time import Time
       >>> t = Time('2020-01-01', scale='utc')
       >>> t.unix_tai - t.unix
-      29.0
+      37.0
 
+      >>> # Before 1972, the offset between TAI and UTC was not integer
       >>> t = Time('1970-01-01', scale='utc')
       >>> t.unix_tai - t.unix  # doctest: +FLOAT_CMP
-      8.200000198854696e-05
+      8.000082
+
+      >>> # Initial offset of 10 seconds in 1972
+      >>> t = Time('1972-01-01', scale='utc')
+      >>> t.unix_tai - t.unix
+      10.0
     """
     name = 'unix_tai'
-    epoch_val = '1970-01-01 00:00:08'
+    epoch_val = '1970-01-01 00:00:00'
     epoch_scale = 'tai'
 
 
--- a/astropy/time/tests/test_basic.py
+++ b/astropy/time/tests/test_basic.py
@@ -1448,9 +1448,9 @@ def test_set_format_basic():
 
 def test_unix_tai_format():
     t = Time('2020-01-01', scale='utc')
-    assert allclose_sec(t.unix_tai - t.unix, 29.0)
+    assert allclose_sec(t.unix_tai - t.unix, 37.0)
     t = Time('1970-01-01', scale='utc')
-    assert allclose_sec(t.unix_tai - t.unix, 8.2e-05)
+    assert allclose_sec(t.unix_tai - t.unix, 8 + 8.2e-05)
 
 
 def test_set_format_shares_subfmt():


maxnoe:
@mhvk @pllim @r-owen

pllim:
As @taldcroft mentioned in #11245, #11143 is slated for 4.3 and not yet released. However, you requested this to be milestoned to 4.2.x, so what other PR is this reverting that was released in 4.2; Can you please clarify? Thanks!

astropy/time/formats.py
```diff
@@ -710,83 +719,48 @@ class TimeUnix(TimeFromEpoch):
...
-      ``1969-12-31 23:59:51.999918 UTC``.  The difference between TAI and UTC
+      ``1969-12-31T23:59:51.999918 UTC``.  The difference between TAI and UTC
       at that epoch was 8.000082 sec.
-    - On the day of a leap second the difference between ``unix`` and ``unix_ptp``
-      times increases linearly through the day by 1.0.  See also the
+    - On the day of a positive leap second the difference between ``unix`` and
+      ``unix_tai`` times increases linearly through the day by 1.0. See also the
       documentation for the `~astropy.time.TimeUnix` class.
+    - While no negative leap seconds were yet introduced into UTC, these are
+      possible in the future, if the length of the astronomical day further
```
mhvk, line 748: This is difficult, as the length of the day increases due to tides of the moon (the poles melting does not help...), though perhaps a fluctuation can cause it...

mhvk (changes requested):
Looks good, thanks! One thing to add is a quick change for the epoch for `unix_tai` listed in `docs/time/index.rst` (in "Time from epoch formats"; I hope I caught everything...)

mhvk:
Oops, failure is real too: the format list in one of the docstrings needs to be adjusted.

maxnoe:
@pllim

## TL;DR
This is a bug fix for a feature introduced in 4.2 which makes an already merged feature planned for 4.3 obsolete.
This is why this is milestoned for 4.2.x but reverts a PR milestoned 4.3.


Summarizing the discussions in #11245 and #11143 the situation and how we got here is as follows:

* #10081 (released with astropy 4.2) introduced the new time format `TimeUnixTAI` with epoch `1970-01-01T00:00:00.0 UTC`.
* #11143 Assuming that the epoch choice was intentional and correct, another new time format `TimeUnixPTP` with epoch  `1970-01-01T00:00:00.0 TAI`, which is different by the initial offset between TAI and UTC of 8.000082 seconds is added, this is not yet released.

We now realized, that the proper epoch for `TimeUnixTAI` should have been  `1970-01-01T00:00:00.0 TAI`. 
This makes `UnixTimePTP` unecessary, so it is removed.

astropy/time/formats.py
maxnoe: Now I am a bit confused, I remember reading an announcment by the IERS that the day shortened so much in 2020 that there are not any positive leap seconds on the horizon anymore. I will reword to make clear that negative leap seconds are unlikely.

maxnoe: https://www.timeanddate.com/time/earth-faster-rotation.html

r-owen: Consider a more neutral statement, such as "Negative leap seconds are also possible, though none have been needed as of 2020-01".

I did not know that the day length has been decreasing and it's very interesting, but it seems out of scope for this documentation.

r-owen (commented):
Thank you for making this change.

taldcroft:
Sorry @maxnoe, you kind of beat me to it, but I was planning to literally revert #11143 by hitting the revert button on GitHub. I think that is cleaner than taking it out as part of this fix.

taldcroft:
Then this fix will be nothing more than changing '1970-01-01 00:00:08' to '1970-01-01 00:00:00' in the right places and fixing tests.

taldcroft:
Doing it this way makes backporting this change possible, at least if I understand things.

maxnoe:
Yes I agree, will rebase.

maxnoe:
Done, this now only contains the changes to `TimeUnixTAI` and the `TimeUnixPTP` PR can be reverted via github.

astropy/time/formats.py
maxnoe: I find it important to stress that there *might* be negative leap seconds

taldcroft:
See #11251.

astropy/time/formats.py
mhvk: Ha, I had missed that completely!  Short-term trends can clearly overwhelm the long-term behaviour....  Anyway, agreed, do keep a statement, but just keep it a bit more general like suggested by @r-owen (the long-term trend better be for days getting longer!).

maxnoe:
@pllim I misread the changelog. `UnixTimeTAI` was already part of 4.1, so we should milestone for the next `4.1.x` release here.

astropy/time/formats.py
```diff
@@ -690,10 +690,19 @@ def _default_scale(self):
 
 class TimeUnix(TimeFromEpoch):
     """
-    Unix time: seconds from 1970-01-01 00:00:00 UTC, ignoring leap seconds.
+    Unix time (UTC): seconds from 1970-01-01 00:00:00 UTC, subject to leap seconds.
```
taldcroft, line 693: The words `subject to` are ambiguous. Please leave this as it was originally.

astropy/time/formats.py
```diff
@@ -690,10 +690,19 @@ def _default_scale(self):
...
 
     For example, 946684800.0 in Unix time is midnight on January 1, 2000.
 
+    In unix time, each UTC day always has exactly 86400 seconds, even on
```
taldcroft, line 697: Please keep PR's in scope.  I think that the new content is actually confusing the situation because of the mix here between POSIX unix time and astropy `unix_time`. In particular with the new text, the meaning of "NOTE: this quantity .." becomes ambiguous. I think the original paragraph is sufficient, so please just revert these new paragraphs.

astropy/time/formats.py
```diff
@@ -710,38 +719,48 @@ class TimeUnix(TimeFromEpoch):
...
+    integral number of leap seconds introduced into UTC since 1970-01-01 UTC.
+
+    This convention matches the definition of the Precision Time Protocol,
+    https://en.wikipedia.org/wiki/Precision_Time_Protocol, which
+    is also used by the White Rabbit protocol in High Energy Physics:
+    https://white-rabbit.web.cern.ch/
+    and the linux CLOCK_TAI (https://www.cl.cam.ac.uk/~mgk25/posix-clocks.html).
+
+    For more information on this time format definition, see:
+    https://www-acc.gsi.de/wiki/Timing/TimingSystemButisInterface#Time
```
taldcroft, line 734: I'm not sure this link is relevant in this context. The "time format definition" from the typical user perspective is quite simple and stated clearly in the first line. The link provides a lot of information on a complex protocol which uses TAI, but I didn't learn anything more about the time format definition.

astropy/time/formats.py
```diff
@@ -710,38 +719,48 @@ class TimeUnix(TimeFromEpoch):
...
 
     Caveats:
 
     - Before 1972, fractional adjustments to UTC were made, so the difference
       between ``unix`` and ``unix_tai`` time is no longer an integer.
     - Because of the fractional adjustments, to be very precise, ``unix_tai``
-      is the number of seconds since ``1970-01-01 00:00:08 TAI`` or equivalently
-      ``1969-12-31 23:59:59.999918 UTC``.  The difference between TAI and UTC
+      is the number of seconds since ``1970-01-01 00:00:00 TAI`` or equivalently
+      ``1969-12-31T23:59:51.999918 UTC``.  The difference between TAI and UTC
```
taldcroft, line 742: Please revert to the original space instead of `T`.

astropy/time/formats.py
```diff
@@ -710,38 +719,48 @@ class TimeUnix(TimeFromEpoch):
...
       at that epoch was 8.000082 sec.
-    - On the day of a leap second the difference between ``unix`` and ``unix_tai``
-      times increases linearly through the day by 1.0.  See also the
+    - On the day of a positive leap second the difference between ``unix`` and
+      ``unix_tai`` times increases linearly through the day by 1.0. See also the
       documentation for the `~astropy.time.TimeUnix` class.
+    - Negative leap seconds are also possible, though none have been needed
+      atleast until 2021-12-30.
```
taldcroft, line 748: Change this to `have been neededa to date`. If there is a negative leap second we'll update this.

astropy/time/formats.py
```diff
@@ -837,7 +856,7 @@ class TimePlotDate(TimeFromEpoch):
       >>> import matplotlib.pyplot as plt
       >>> jyear = np.linspace(2000, 2001, 20)
       >>> t = Time(jyear, format='jyear', scale='utc')
-      >>> plt.plot_date(t.plot_date, jyear)
+      >>> plt.plot_date(t.plot_date, jyear) # doctest: +IGNORE_OUTPUT
```
taldcroft, line 859: Why was this needed?

CHANGES.rst
```diff
@@ -380,6 +380,12 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
+- Change epoch of ``TimeUnixTAI`` (``"unix_tai"``) from ``1970-01-01T00:00:00 UTC``
```
taldcroft, line 383: This should also be put into the API change section.

taldcroft (changes requested):

astropy/time/formats.py
maxnoe: Ignoring is just wrong. UTC is the only timescale that has leap seconds. Subject to may be ambiguous, but it is rightly so because it confers to the reader that it is not as simple as "ignoring" and that needs to understand what is actually happening.

astropy/time/formats.py
maxnoe: This was copied from the PTP docs that were introduced and now reverted. Since the two formats are now the same I thought it should be merged.

astropy/time/formats.py
maxnoe: because `plot_date` returns a `Line` object which lead to a doctest failure.

astropy/time/formats.py
maxnoe: copy paste, fixed

CHANGES.rst
maxnoe: Where would that go? We decided this is a bug fix which should be applied to 4.2.x, there is no API changes section for this, since it is a bugfix release.

maxnoe: Or do you mean we should change the change log entry for astropy 4.2?

astropy/time/formats.py
taldcroft: I was thinking of PTP as the format introduced in the context of these physics libraries you reference, and `unix_tai` as the one for most astronomers. On re-reviewing this as a *change* to the `unix_tai` docs it just seems out of place and not a generally useful link for most users of the format.

astropy/time/formats.py
taldcroft: That makes sense, but makes me wonder why it was passing tests before. Did something else change in the meantime that made this doctest line suddenly start failing?

CHANGES.rst
taldcroft: Normally we don't make API changes in a bug fix release, but this is definitely an API change that needs to be highlighted. So that would mean either creating an API change section, or else deciding that we should not break the normal rules and instead just target 4.3.

Thoughts @mhvk or @pllim or @bsipocz ?

mhvk: I think this is really a bug fix, as we concluded that the code just gave the wrong answer before. It doesn't seem an API change - no "application program interface" was changed after all!

For highlighting this change, perhaps it should go up front? Or in "other changes"?

Though perhaps there is not that much difference from other bug fixes which have happened before without (I think) special mention, which also may have caused real problems:
```
- Fixed use of weights with ``LinearLSQFitter``. [#10687]

- Fix ``with_H0`` equivalency to use the correct direction of
  conversion. [#8292]
```

mhvk: p.s. In the end, a main lesson I am learning from this is to be thankful how well astropy documents things! The fact that it is so difficult to find a precise definition of `CLOCK_TAI` is concerning!

astropy/time/formats.py
mhvk: @maxnoe - did this fail for you locally or in the CI tests? (good to change regardless, but also good to understand indeed why this did not trigger problems before!)

CHANGES.rst
bsipocz: Occasionally there were API changing bug fixes, mostly in lts and mostly for utils stuff that are needed downstream. Either add an API change section or other changes, for examples look for the 2.0.x sections. 
However I would be very cautious making a real API change. If it's a bug then it's not really an API change but a fix and can land in the appropriate section

taldcroft: @mhvk - I think of an API change including the *specification* of a function that causes the code to give a different output for the same input.  It is certainly possible that someone has written code like below, so it would be good to highlight this change.
```
t_tai = Time(tm).unix_tai + 8
...
```
The code gave the correct answer given the specification, but the spec (code "requirements") was wrong. I tend to think of that as API, but maybe it's just me.

astropy/time/formats.py
maxnoe: It failed locally, did not check the CI, I fixed this together with other test failures so I don't know whether this also affected CI or whether something with my local setup was strange. Can change back to check.

CHANGES.rst
maxnoe: The point is, the code before did not match the documented definition. So this is really a fix that brings the code in agreement with the docs.

mhvk: Indeed (to both). I think the main question really is how to ensure this gets seen. Easiest would seem to have a second entry in "Other changes".

astropy/time/formats.py
mhvk: Actually, right now we instruct the test to be skipped: https://github.com/astropy/astropy/blob/ef40111640f9f93215eea87eb793da9c7093de68/astropy/time/formats.py#L35

If with you `IGNORE_OUTPUT`, this works, perhaps we can change it to a matplotlib requirement?
```
__doctest_requires__ = {'TimePlotDate': ['matplotlib']}
```
(Though if you prefer to just leave things as they were, that's fine too!)

CHANGES.rst
maxnoe: The "Known Issues" document seems to mainly handle higher level problems like numpy integration (and some things look lilke they could be removed because they belong to versions of astropy or dependencies that are not supported anymore).

I am not sure how best to proceed.

astropy/time/formats.py
maxnoe: I think this is best for another PR, will revert the change here.

maxnoe:
@taldcroft I removed most of the docs changes and will open another PR about that.

We should get this fixed, so I don't want to discuss about leap second wordings here.

astropy/time/formats.py
taldcroft: In this context "ignoring leap seconds" is equivalent to the definition of unix time which is computing the seconds since the epoch assuming exactly 86400 seconds in every day. The actual unix time on a computer *is* as simple as ignoring, no?


The value of Unix time on a computer is related to the UTC time name components (year, yday, hour, min, sec) according to the expression:
```
tm_sec + tm_min*60 + tm_hour*3600 + tm_yday*86 400 + (tm_year-70)*31 536 000 + ((tm_year-69)/4*86 400
```

taldcroft (changes requested):
Just one last doc suggestion, otherwise good. Thanks @maxnoe !

astropy/time/formats.py
```diff
@@ -710,38 +710,45 @@ class TimeUnix(TimeFromEpoch):
...
 class TimeUnixTai(TimeUnix):
     """
-    Seconds from 1970-01-01 00:00:08 TAI (see notes), including leap seconds.
+    Unix time (TAI): SI seconds elapsed since 1970-01-01 00:00:00 TAI (see caveats).
 
-    This will generally differ from Unix time by the cumulative integral number
-    of leap seconds since 1970-01-01 UTC.  This convention matches the definition
-    for linux CLOCK_TAI (https://www.cl.cam.ac.uk/~mgk25/posix-clocks.html).
+    This will generally differ from standard (UTC) Unix time by the cumulative
+    integral number of leap seconds introduced into UTC since 1970-01-01 UTC.
```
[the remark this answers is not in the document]
maxnoe: The sentence is missing the `plus the initial offset of 8 seconds`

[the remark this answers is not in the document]
maxnoe: The correct sentence would be 

* This will generally differ from standard (UTC) Unix time by the cumulative
   integral number of leap seconds introduced into UTC since 1972-01-01 UTC plus the initial offset of 10 seconds at that date. 
* Between 1970-01-01 and 1972-01-01 there were also fractional second adjustments made to UTC, resulting in an offset of `8.000082` seconds at the unix epoch of `1970-01-01 00:00:00 UTC`.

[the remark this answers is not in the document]
maxnoe: > . There have been 29 leap seconds since 1970, not 37.

No, 27 leap seconds. The additional 2 seconds come from fractional adjustments between 1970 and 1972.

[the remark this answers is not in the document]
maxnoe: updated, please have another look

astropy/time/formats.py
```diff
@@ -710,38 +710,48 @@ class TimeUnix(TimeFromEpoch):
...
-    Seconds from 1970-01-01 00:00:08 TAI (see notes), including leap seconds.
+    Unix time (TAI): SI seconds elapsed since 1970-01-01 00:00:00 TAI (see caveats).
 
-    This will generally differ from Unix time by the cumulative integral number
-    of leap seconds since 1970-01-01 UTC.  This convention matches the definition
-    for linux CLOCK_TAI (https://www.cl.cam.ac.uk/~mgk25/posix-clocks.html).
+    This will generally differ from standard (UTC) Unix time by the cumulative
+    integral number of leap seconds introduced into UTC since 1972-01-01 UTC
+    plus the initial offset of 10 seconds at that date.  To date, 27 positive
+    leap seconds were introduced into UTC, resulting in the current offset of
```
taldcroft, line 718: This might become wrong in the future, so I'd just cut this `To date` sentence.

maxnoe: Removed and added comments and another example for the offset in 1972.

taldcroft (approved):
Thanks @maxnoe !

taldcroft:
ci/circleci: image-tests-mpldev failure looks unrelated so merging this.

PR #11251, opened 2021-01-14
Revert "Merge pull request #11143 from maxnoe/unix_time_ptp"

taldcroft:
This reverts commit ec983207505a2c1dcaefdb3b32fe0d9f30a2b716, reversing
changes made to 960314366be1bfdc3b2e7488fde0d2f780562f68.

### Description

This pull request is to revert #11143 as part of fixing #11245.

I did this with the following since the GitHub `Revert` button failed:
```
git revert -m 1 ec98320
```

But maybe @pllim knows the best way to do this?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -82,9 +82,6 @@ astropy.tests
 astropy.time
 ^^^^^^^^^^^^
 
-- Add new time format ``unix_ptp`` (Precision Time Protocol) which provides
-  the number of SI seconds since ``1970-01-01T00:00:00 TAI``. [#11143]
-
 astropy.timeseries
 ^^^^^^^^^^^^^^^^^^
 
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -476,9 +476,9 @@ class TimeBase(ShapedLikeNDArray):
         that could be used for initialization.  These can be listed with::
 
           >>> list(Time.FORMATS)
-          ['jd', 'mjd', 'decimalyear', 'unix', 'unix_tai', 'unix_ptp', 'cxcsec',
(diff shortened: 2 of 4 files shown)


pllim:
I don't think I have reverted in this repo before, but if the diff looks good to you, then it worked. 😸

pllim:
p.s. Is the diff for the change log missing?

mhvk:
Looks good apart from what @pllim noted: the changelog entry somehow was missed. How bizarre!

maxnoe:
He reverted the commit,  not the pr. The changelog every is in another commit, since it requires knowing the pr number.

pllim:
Hmmm, I thought `-m 1 ec98320` would revert the whole PR merge...

pllim:
That PR only has 1 commit and `git revert 36c0f603f9916100837406c0e3a4934c3685f462` doesn't grab the change log either. 🤯

taldcroft:
> The changelog every is in another commit, since it requires knowing the pr number.

Slightly OT, but check out this gem to get the PR number (*most* of the time) before you make the PR.
https://github.com/astropy/astropy-procedures/blob/master/next_pr_number.py

taldcroft:
That's really really weird. I confirm also that reverting the single commit doesn't catch the change log as well. Maybe there was some kind of conflict and it just silently ignores it??

Anyway I manually fixed the changelog with another commit.

mhvk (approved):
You're surely right about there being some conflict, but very odd that `git` reversion does not at least warn about that! Anyway, it looks all good now, so approving.

maxnoe:
The test failure seems unrelated. It would be easier to merge this before #11249, because then I don't have to fix the tests for comparing unix with unix_ptp

pllim:
Merged!

PR #11255, opened 2021-01-17
Ensure SkyOffsetFrame can again be transformed back to base frame

### LINKED ISSUE #11254: Removal of OrderedDescriptor has broken SkyOffsetFrame when using SkyCoord
The removal of `OrderedDescriptor` (#11099, merged today) has broken `SkyOffsetFrame` when using `SkyCoord`:

```python
>>> import astropy.units as u
>>> from astropy.coordinates import SkyCoord, SkyOffsetFrame, ICRS

>>> origin = SkyCoord(10*u.deg, 0*u.deg, frame='icrs')
>>> offset_frame = SkyOffsetFrame(origin=origin)
>>> coord = SkyCoord(3*u.deg, 4*u.deg, frame=offset_frame)
>>> coord
<SkyCoord (SkyOffsetICRS: rotation=0.0 deg, origin=<ICRS Coordinate: (ra, dec) in deg
    (10., 0.)>): (lon, lat) in deg
    (3., 4.)>

>>> coord.transform_to('icrs')
<ipython-input-21-d349054966e0> in <module>
----> 1 coord.transform_to('icrs')

c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\sky_coordinate.py in transform_to(self, frame, merge_attributes)
    625             frame_kwargs.pop(attr)
    626 
--> 627         return self.__class__(new_coord, **frame_kwargs)
    628 
    629     def apply_space_motion(self, new_obstime=None, dt=None):

c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\sky_coordinate.py in __init__(self, copy, *args, **kwargs)
    321             for attr in skycoord_kwargs:
    322                 # Setting it will also validate it.
--> 323                 setattr(self, attr, skycoord_kwargs[attr])
    324 
    325             if info is not None:

c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\sky_coordinate.py in __setattr__(self, attr, val)
    809             super().__setattr__('_' + attr, val)
    810             # Validate it
--> 811             frame_transform_graph.frame_attributes[attr].__get__(self)
    812             # And add to set of extra attributes
    813             self._extra_frameattr_names |= {attr}

c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\attributes.py in __get__(self, instance, frame_cls)
    102                 out = getattr(instance, self.secondary_attribute, self.default)
    103 
--> 104         out, converted = self.convert_input(out)
    105         if instance is not None:
    106             instance_shape = getattr(instance, 'shape', None)  # None if instance (frame) has no data!

c:\users\ayshih\onedrive\code\astropy\astropy\coordinates\attributes.py in convert_input(self, value)
    445         if value is None:
    446             return None, False
--> 447         elif isinstance(value, self._frame):
    448             return value, False
    449         else:

TypeError: isinstance() arg 2 must be a type or tuple of types
```

Using frames directly instead of `SkyCoord` works fine:
```python
>>> coord.frame.transform_to(ICRS())
<ICRS Coordinate: (ra, dec) in deg
    (13., 4.)>
```

It looks like it's related to the additional processing that `SkyCoord` does for frame attributes, and the fact that this line:
https://github.com/astropy/astropy/blob/ef40111640f9f93215eea87eb793da9c7093de68/astropy/coordinates/builtin_frames/skyoffset.py#L58

is no longer working as intended (it's supposed to overwrite the created frame's `origin` frame attribute from one with `frame=None` to one with `frame=framecls`).  The created frame's `origin` frame attribute doesn't have `frame=framecls`:
```python
>>> offset_frame.frame_attributes['origin']._frame is None
True
```

mhvk:
This Meta-class creation was missed in #11099 because there was no test that showed it was a problem. Now got rid of the whole meta-class and added an appropriate test.

Fixes #11254

### DIFF
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -250,8 +250,6 @@ class BaseCoordinateFrame(ShapedLikeNDArray):
         default_repr = getattr(cls, 'default_representation', None)
         default_diff = getattr(cls, 'default_differential', None)
         repr_info = getattr(cls, 'frame_specific_representation_info', None)
-        frame_attrs = getattr(cls, 'frame_attributes', None)
-
         # Then, to make sure this works for subclasses-of-subclasses, we also
         # have to check for cases where the attribute names have already been
         # replaced by underscore-prefaced equivalents by the logic below:
@@ -275,22 +273,20 @@ class BaseCoordinateFrame(ShapedLikeNDArray):
         cls._create_readonly_property('frame_specific_representation_info',
                                       copy.deepcopy(repr_info))
 
-        # Set the frame attributes
+        # Set the frame attributes. We follow frames in reverse order to keep
+        # insertion order and also allow subclasses to override attributes
+        # already set by their parent while keeping order.
         # TODO: Should this be made to use readonly_prop_factory as well or
         # would it be inconvenient for getting the frame_attributes from
         # classes?
-        if frame_attrs is None:
-            frame_attrs = {}
-        else:
-            frame_attrs = frame_attrs.copy()
-
-        for basecls in (cls,) + cls.__bases__:
+        frame_attrs = {}
+        for basecls in ((cls,) + cls.__bases__)[::-1]:
             if not issubclass(basecls, BaseCoordinateFrame):
                 continue
 
             for k, v in basecls.__dict__.items():
                 if isinstance(v, Attribute):
-                    frame_attrs.setdefault(k, v)
+                    frame_attrs[k] = v
 
         cls.frame_attributes = frame_attrs
 
--- a/astropy/coordinates/baseframe.py
+++ b/astropy/coordinates/baseframe.py
@@ -273,20 +273,23 @@ class BaseCoordinateFrame(ShapedLikeNDArray):
         cls._create_readonly_property('frame_specific_representation_info',
                                       copy.deepcopy(repr_info))
 
-        # Set the frame attributes. We follow frames in reverse order to keep
-        # insertion order and also allow subclasses to override attributes
-        # already set by their parent while keeping order.
+        # Set the frame attributes. We first construct the attributes from
+        # superclasses, going in reverse order to keep insertion order,
+        # and then add any attributes from the frame now being defined
+        # (if any old definitions are overridden, this keeps the order).
+        # Note that we cannot simply start with the inherited frame_attributes
+        # since we could be a mixin between multiple coordinate frames.
         # TODO: Should this be made to use readonly_prop_factory as well or
         # would it be inconvenient for getting the frame_attributes from
         # classes?
         frame_attrs = {}
-        for basecls in ((cls,) + cls.__bases__)[::-1]:
-            if not issubclass(basecls, BaseCoordinateFrame):
-                continue
+        for basecls in reversed(cls.__bases__):
+            if issubclass(basecls, BaseCoordinateFrame):
+                frame_attrs.update(basecls.frame_attributes)
 
-            for k, v in basecls.__dict__.items():
-                if isinstance(v, Attribute):
-                    frame_attrs[k] = v
+        for k, v in cls.__dict__.items():
+            if isinstance(v, Attribute):
+                frame_attrs[k] = v
 
         cls.frame_attributes = frame_attrs
 
--- a/astropy/coordinates/builtin_frames/skyoffset.py
+++ b/astropy/coordinates/builtin_frames/skyoffset.py
@@ -52,9 +52,6 @@ def make_skyoffset_cls(framecls):
          # defaults of SkyOffsetFrame set by BaseCoordinateFrame.
          '_default_representation': framecls._default_representation,
          '_default_differential': framecls._default_differential,
-         # Similarly, we need to ensure the frame attributes are recalculated,
-         # otherwise the default from SkyOffsetFrame will be used.
-         'frame_attributes': None,
          '__doc__': SkyOffsetFrame.__doc__,
          })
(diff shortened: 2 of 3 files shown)


astropy/coordinates/builtin_frames/skyoffset.py
```diff
@@ -43,36 +43,20 @@ def make_skyoffset_cls(framecls):
...
-            # Only 'origin' is needed here, to set the origin frame properly.
-            members['origin'] = CoordinateAttribute(frame=framecls, default=None)
-
-            # This has to be done because FrameMeta will set these attributes
-            # to the defaults from BaseCoordinateFrame when it creates the base
-            # SkyOffsetFrame class initially.
-            members['_default_representation'] = framecls._default_representation
-            members['_default_differential'] = framecls._default_differential
-
-            newname = name[:-5] if name.endswith('Frame') else name
```
mhvk, line 66: This was, I think, trying too hard to even account for generic subclasses of `SkyOffsetFrame`. It is not actually useful since below a subclass of `SkyOffsetFrame` is created...

astropy/coordinates/builtin_frames/skyoffset.py
```diff
@@ -43,36 +43,20 @@ def make_skyoffset_cls(framecls):
...
-
-            return super().__new__(cls, newname, bases, members)
-
-    # We need this to handle the intermediate metaclass correctly, otherwise we could
-    # just subclass SkyOffsetFrame.
-    _SkyOffsetFramecls = SkyOffsetMeta('SkyOffsetFrame',
-                                       (SkyOffsetFrame, framecls),
-                                       {'__doc__': SkyOffsetFrame.__doc__})
+    # Create a new SkyOffsetFrame subclass for this frame class.
+    name = 'SkyOffset' + framecls.__name__
```
mhvk, line 47: It is possible to generalize this, but then we need to pass on `cls` from `__new__` below as well.

mhvk:
Hmm, something went wrong with __repr__, with suddenly `rotation=0` appearing. So, maybe something that I did is not quite right... any suggestions?

astropy/coordinates/baseframe.py
```diff
@@ -275,22 +273,20 @@ def __init_subclass__(cls, **kwargs):
...
         # TODO: Should this be made to use readonly_prop_factory as well or
         # would it be inconvenient for getting the frame_attributes from
         # classes?
-        if frame_attrs is None:
-            frame_attrs = {}
-        else:
-            frame_attrs = frame_attrs.copy()
-
-        for basecls in (cls,) + cls.__bases__:
+        frame_attrs = {}
```
mhvk, line 282: @embray - please check you agree with this change in logic. I got failures because any new `SkyOffsetFrame` class overrides `SkyOffsetFrame.origin` and then it either was not properly inserted in `frame_attributes` or, if I reset that one (first commit), it changed the order. I think not relying on a previous calculation of `frame_attributes` is probably better practice anyway.

mhvk:
OK, all tests passed now.

mhvk:
@pllim - numpy dev is OK again!

astropy/coordinates/baseframe.py
embray: I admit I don't really understand the problem here since I don't know what a SkyOffsetFrame is or why it was broken.  What you did here makes sense and is actually close to what I think was my original approach to the problem.  I just didn't see that there was a difference and was trying to avoid two loops.  It didn't occur to me that there could be base classes with metaclasses that would do something different with frame_attributes.

IIUC that's what the problem was here, in which case your solution makes sense.

astropy/coordinates/builtin_frames/skyoffset.py
embray: Okay, I didn't know about this, and seeing it now kind of answers my previous question about what the problem was.  But if you remove this metaclass are the previous changes needed?  Or do you think the previous changes are still good to have in case another case like this metaclass arises?

astropy/coordinates/builtin_frames/skyoffset.py
```diff
@@ -43,36 +43,17 @@ def make_skyoffset_cls(framecls):
...
+    _SkyOffsetFramecls = type(
+        name, (SkyOffsetFrame, framecls),
+        {'origin': CoordinateAttribute(frame=framecls, default=None),
+         # The following two have to be done because otherwise we use the
+         # defaults of SkyOffsetFrame set by BaseCoordinateFrame.
+         '_default_representation': framecls._default_representation,
+         '_default_differential': framecls._default_differential,
+         '__doc__': SkyOffsetFrame.__doc__,
+         })
```
embray, line 56: :+1: this is simpler.

embray (approved):
My existing comments are more rhetorical and for edification than anything else.  LGTM.

astropy/coordinates/builtin_frames/skyoffset.py
mhvk: Yes, the changes in `baseframe` are still needed - I did this before the other ones. It seems a tricky business merging the attributes, keeping order, and possibly overriding ones...

astropy/coordinates/builtin_frames/skyoffset.py
mhvk: I had actually hoped to somehow use `__init_subclass__` for this, but it seems at that point it is too late to define a descriptor on the class (it doesn't seem to get bound). Anyway, this is fairly clean.

mhvk:
Thanks! A few answers just for interest's sake.

PR #11256, opened 2021-01-18
Add QDP data file handling

### LINKED ISSUE #8382: IO support for reading/writing QDP files
One thing that has been bugging me for quite a long time is that astropy does not support qdp files which are quite common in the X-ray community.  It is not difficult to write a small procedure to read them in, but it is just one of those things that keep coming back and while I can read nearly all files using astropy, I can't read the qdp files. For that I need to import a different routine. And yet, these are very commonly used in the X-ray community. I would like to suggest to add a reader/ perhaps writer for those kind of files as well.

matteobachetti:
### Description

This pull request is to address Issue #8382, adding QDP data handling.

Fixes #8382


astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+    file_line = 0
+    table_list = []
+    initial_comments = ""
+    comment_text = ""
+    err_specs = {}
+    colnames = None
+    if input_colnames is None or len(input_colnames) == ncol:
+        colnames = interpret_err_lines(
+            err_specs, ncol, input_colnames=input_colnames
+        )
```
embray, line 280: I don't really understand what this is doing.  Is it initializing the `err_specs` data structure?

astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+
+    for key, group in groupby(contents):
+        n_lines = len(list(group))
```
embray, line 283: Here you seem to be re-parsing parts of each line; I think the regexp I provided earlier could help here; with some refactoring you could parse each line just once.

I don't understand why the `groupby` and the slicing of `lines`.  Part of that is maybe my ignorance of the file format (are comments, commands, and data always grouped together consecutively in the file)?  That's not what it looks like from the example files, so it's no clear to me why this is done this way as opposed to looping over the lines directly.

astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+    Returns
+    -------
+    tables : List of `Table` objects
+        List containing all the tables present inside the QDP file
+    """
+
+    contents, ncol = analyze_qdp_file(qdp_file)
+
+    with open(qdp_file) as fobj:
+        lines = list(fobj.readlines())
```
embray, line 269: `.readlines()` returns a list to begin with on most file-like objects.

astropy/io/misc/tests/test_qdp.py
```diff
@@ -0,0 +1,117 @@
...
+    53000.123456 2.37847222222222e-05    -2.37847222222222e-05   0.726155        0.583890
+    55045.099887 1.14467592592593e-05    -1.14467592592593e-05   2.410935        1.393592
+    NO NO NO NO NO
+    ! WT -- hardness ratio
+    !MJD            Err (pos)       Err(neg)        Rate            Error
+    53000.123456 2.37847222222222e-05    -2.37847222222222e-05   -0.292553       -0.374935
+    55045.099887 1.14467592592593e-05    -1.14467592592593e-05   0.000000        -nan
+    """
+
+    fd, path = tempfile.mkstemp()
```
embray, line 31: you can use the tmpdir or tmp_path fixture from pytest: https://docs.pytest.org/en/stable/tmpdir.html

embray (changes requested):
This is helpful for understanding how this file format is interpreted and the code is very well formatted.  I think it can be simplified though.

astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+import numpy as np
+import warnings
+from astropy.table import Table
+from astropy.io import registry as io_registry
+
+
+def is_qdp(origin, filepath, fileobj, *args, **kwargs):
+    if filepath is not None:
+        return filepath.endswith(('.qdp'))
+    return False
```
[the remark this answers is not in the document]
matteobachetti: It's basically just a text file with free-format column data, with or without instructions for PLT, with or without comments. It's really difficult to distinguish it from other free-format column data :(

astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+        line = line[:comment_in_line]
+
+    probe = line.split()[0]
+    n = len(line.split())
+
+    if probe.startswith("!"):
+        return "comment"
+
+    # Full line of NOs -> new table
+    if probe.strip("NO ") == "":
```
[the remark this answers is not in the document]
matteobachetti: I'm impressed by the `_line_type_re`, but if I may, I had a hard time following that up :). 
I thought about using re's but decided otherwise because I wanted something easy to read and fix. No strong opinions.

Just one comment about the `_decimal_re`: does it understand numbers in scientific format?

astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+    try:
+        float(probe)
+        return f"data,{n}"
+    except ValueError:
+        pass
+
+    raise ValueError(f"Unrecognized QDP line: {line}")
+
+
+def get_type_from_list_of_lines(list_of_lines):
```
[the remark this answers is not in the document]
matteobachetti: Done

astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+    all_data_specs = [c for c in contents if c.startswith("data")]
+
+    # Verify that all data specs are the same (i.e. all data lines
+    # contain the same number of entries)
+    if len(list(set(all_data_specs))) != 1:
+        raise ValueError("Inconsistent number of columns")
+
+    ncol = int(all_data_specs[0].replace("data,", "").strip())
+
+    return contents, ncol
```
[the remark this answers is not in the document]
matteobachetti: Cool, done

astropy/io/misc/qdp.py
```diff
@@ -0,0 +1,451 @@
...
+
+    Other parameters
+    ----------------
+    input_colnames : List of strings
+        Name of data columns (defaults to ['col1', 'col2', ...]), _not_
+        including error columns.
+
+    Returns
+    -------
+    colnames : List
```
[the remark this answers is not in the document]
matteobachetti: Done

astropy/io/misc/qdp.py
matteobachetti: Basically, groupby allows to associate all the tables in the file to separate chunks, that I then read all together instead of parsing the file only once and wait for a change of data type. Since most QDP files are small (thousands of lines at most), parsing the file twice has usually no large overhead. I agree it is not the most efficient way :)

astropy/io/misc/qdp.py
matteobachetti: Oh, I thought it was an iterator. Fixed!

astropy/io/misc/tests/test_qdp.py
matteobachetti: In this case, I would prefer using what the Python standard library gives me instead of a dependency that might change its behavior in the future, unless there is a compelling reason. After all, I just want a temporary file, and anything pytest can give me -as far as I understand- complicates the code instead of simplifying it

astropy/io/misc/qdp.py
matteobachetti: It is reading err_specs and understanding if some columns represent the uncertainty of others. Main column are named col1, col2, ... or whatever the user gives (e.g. MJD, Rate). Error column are named col0_err (or MJD_err) for symmetric error, col0_perr and col0_nerr for positive/negative error

matteobachetti: I do this once at the start of the file, assuming that err_specs is empty, for the case where the file does not contain the READ [ST]ERR commands. This gets overwritten when these commands are found.

astropy/io/misc/qdp.py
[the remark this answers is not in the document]
matteobachetti: However, I tried all doctests and this solution doesn't pass them. It fails at the very first:
```
103     Examples
104     --------
105     >>> line_type("READ SERR 3")
UNEXPECTED EXCEPTION: ValueError('Unrecognized QDP line: READ SERR 3')
```
As per the "N O N NOON OON O" line... I hope nobody is so evil to write something like that :D

matteobachetti:
I don't understand what is happening in those failed tests. It looks like the logger got confused, but only in one test case :O

pllim (commented):
Probably need user facing doc too.

cc @PaulKuin

astropy/io/misc/qdp.py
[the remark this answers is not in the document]
matteobachetti: Ok, I expanded your regular expression and now it works in all cases, including scientific notation!

astropy/io/misc/qdp.py
matteobachetti: Ok, done with a single pass. Also, I fixed a couple of possible bugs along the way, and interpreted NOs as missing values instead of NaNs

matteobachetti:
@embray: I think I fixed most fixable things :). I left a couple more requests for clarifications or minor "rebuttals"
@pllim: I added something to the main Read/Write docs, feel free to point out more documents to update

taldcroft (commented):
@matteobachetti - thanks, this will be a nice addition! 

One thing that is going to complicate matters is that this should really be made part of `io.ascii`. The QDP format is very much an ASCII table format so it should live in that subpackage and support the user doing `ascii.read('data.qdp', format='qdp', names=['x', 'y', 'z'])`. The machinery in `io.ascii` is a bit orthogonal to the way you've written this so far. If you just take the first step of moving it into `io.ascii` I can help with getting it to fit in.

matteobachetti:
Hi @taldcroft, I started moving things around. Alas, while the `io.misc` machinery is more or less straightforward, I still haven't understood how `io.ascii` works. I need some help to plug my functions in the correct places :)

taldcroft:
>  I need some help to plug my functions in the correct places.

Yes. The `io.ascii` machinery is very general and extensible but not very easy to understand. (I'd make it simpler if I wrote it today...). The easiest is if I just play around with it once you have done the basic changes, since I cannot from the top of my head tell you the path. 

It looks like you moved the code around. Are there other changes you are planning?

matteobachetti:
> It looks like you moved the code around. Are there other changes you are planning?

Not until I understand a little more :). I drafted a couple of classes based on the functionality in `basic.py`, but I need to understand some very basic stuff (For example: what delimiter should I give in `QDPsplitter` for "any number of spaces"?)

taldcroft:
> For example: what delimiter should I give in QDPsplitter for "any number of spaces"?

The delimiter `' '` implies any number of spaces. This is delegated to the standard library `csv.reader`.

matteobachetti:
@taldcroft I don't understand how to make `read` accept the `table_id` keyword

taldcroft:
@matteobachetti - I pushed a commit that seems to do the trick. 

One comment, can you use `names` instead of `input_colnames` for API consistency with `io.ascii`?  An analog is the `no_header` format where supplying `names` will define the column names.

matteobachetti:
Thanks! I'll change the keyword to `names` as requested (and I guess there will be some more API fixes like this to do)

taldcroft:
There are a few flake8 issues that should be cleaned up:
```
(astropy) ➜  astropy git:(add_qdp) flake8 astropy/io/ascii/qdp.py --ignore=E402,W503,F541 --max-line-length=100 
astropy/io/ascii/qdp.py:9:1: F401 'itertools.groupby' imported but unused
astropy/io/ascii/qdp.py:14:1: F401 'astropy.io.registry as io_registry' imported but unused
astropy/io/ascii/qdp.py:29:101: E501 line too long (131 > 100 characters)
astropy/io/ascii/qdp.py:443:5: F401 'astropy.io.ascii' imported but unused
```

CHANGES.rst
```diff
@@ -177,6 +177,8 @@ astropy.io.fits
 astropy.io.misc
 ^^^^^^^^^^^^^^^
 
+- Added QDP data file handling [#11256]
```
taldcroft, line 180: This should be in `io.ascii` and can be a bit more verbose, perhaps "Added support for reading and writing ASCII tables in QDP (Quick and Dandy Plotter) format."

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,552 @@
...
+    header_class = QDPHeader
+    data_class = QDPData
+
+    def __init__(self, table_id=0, input_colnames=None):
+        super().__init__()
+        self.table_id = table_id
+        self.input_colnames = input_colnames
+
+    def read(self, qdp_file):
+        return read_table_qdp(qdp_file, table_id=self.table_id,
```
taldcroft, line 545: Another key point for API consistency is to allow the input to be a filename, a string with the data, a list of strings or a file object. This is normally done using the following:
```
        # Get a list of the lines (rows) in the table
        self.lines = self.inputter.get_lines(table, newline=newline)
```
Then the core table parsing should always expect a list of lines. I see you have support for providing the table data as a string, though this is not obvious from the docstrings.

To go 100% in the style of `io.ascii`, the standalone routines like `read_table_qdp` would not be considered part of the public API and then you would rewrite that function and all the downstream ones to accept only a list of lines.

Another option is to adapt `get_lines_from_file` to support the standard input options.

docs/io/unified.rst
```diff
@@ -202,6 +202,7 @@ ascii.fixed_width_no_header    Yes          :class:`~astropy.io.ascii.FixedWidth
                  ascii.ipac    Yes          :class:`~astropy.io.ascii.Ipac`: IPAC format table
                 ascii.latex    Yes    .tex  :class:`~astropy.io.ascii.Latex`: LaTeX table
             ascii.no_header    Yes          :class:`~astropy.io.ascii.NoHeader`: Basic table with no headers
+                  ascii.qdp    Yes    .qdp   QDP_: Quick and Dandy Plotter files
```
taldcroft, line 205: Should be a `:class: ..` like the others where the class docstring has the reference info.

docs/io/unified.rst
```diff
@@ -911,6 +912,15 @@ By default, ``serialize_method`` for Time columns is equal to
 
 .. _table_io_hdf5:
 
+QDP
```
taldcroft, line 915: It does not get a separate section here since it is "just" another ASCII format.

matteobachetti:
@taldcroft next: I'm trying to supersede the write interface, but I don't understand how to do that. I commented out my definition of write so that tests pass, but at the moment the interface isn't able to write proper QDP files (with SERR/TERR instructions)

docs/io/unified.rst
taldcroft: Also include this new format in `astropy/docs/io/ascii/index.html`  around line 260.

matteobachetti:
@taldcroft, I just noticed that you reviewed the code while I was making the last modifications. I'll work on your comments. Thanks!

CHANGES.rst
matteobachetti: Done

docs/io/unified.rst
matteobachetti: Done

docs/io/unified.rst
matteobachetti: Done

astropy/io/ascii/tests/test_qdp.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+
+    path = str(tmpdir.join('test.qdp'))
+
+    with open(path, "w") as fp:
+        print(example_qdp, file=fp)
+
+    table0 = read_table_qdp(fp.name, names=["MJD", "Rate"], table_id=0)
+    assert table0.meta["initial_comments"].lstrip().startswith("Swift")
+    assert table0.meta["comments"].lstrip().startswith("WT -- hard data")
```
taldcroft, line 38: Comments should be a list instead of string. See https://docs.astropy.org/en/stable/io/ascii/read.html#comments-and-metadata.

astropy/io/ascii/tests/test_qdp.py
```diff
@@ -0,0 +1,135 @@
...
+    table2 = read_table_qdp(fp.name, names=["MJD", "Rate"], table_id=2)
+    assert table2.meta["initial_comments"].lstrip().startswith("Swift")
+    assert table2.meta["comments"].lstrip().startswith("WT -- hardness")
+    assert np.isclose(table2["MJD_nerr"][0], -2.37847222222222e-05)
+
+
+def test_roundtrip(tmpdir):
+    example_qdp = """
```
taldcroft, line 46: For testing it is nice to include a synthetic file with "simple" values. This makes it easier to verify that it is fully working. For example in my testing of this I used this:
```
example_qdp = """
    ! Initial comment line 1
    ! Initial comment line 2
    READ TERR 1
    READ SERR 3
    ! Table 0 comment
    !MJD            Err (pos)       Err(neg)        Rate            Error
    53000.5   0.25  -0.5   1  1.5  3.5 2
    54000.5   1.25  -1.5   2  2.5  4.5 3
    NO NO NO NO NO
    ! Table 1 comment
    !MJD            Err (pos)       Err(neg)        Rate            Error
    54000.5   2.25  -2.5   3  3.5  5.5 5
    55000.5   3.25  -3.5   4  4.5  6.5 nan
    """

dat = ascii.read(example_qdp, format='qdp', table_id=1, names=['a', 'b', 'c', 'd'])
dat

<Table length=2>
   a     a_perr  a_nerr    b       c     c_err     d   
float64 float64 float64 float64 float64 float64 float64
------- ------- ------- ------- ------- ------- -------
54000.5    2.25    -2.5     3.0     3.5     5.5     5.0
55000.5    3.25    -3.5     4.0     4.5     6.5     nan
```
You can leave this realistic Swift file if you like, but you should include something like my example and then test that every single value in both tables is expected, and that all the columns have the right name and dtype.

taldcroft:
More changes (I made these locally for testing):
```
diff --git a/astropy/io/ascii/__init__.py b/astropy/io/ascii/__init__.py
index e10998ac1..9e1b79a19 100644
--- a/astropy/io/ascii/__init__.py
+++ b/astropy/io/ascii/__init__.py
@@ -35,6 +35,7 @@ from .latex import Latex, AASTex, latexdicts
 from .html import HTML
 from .ipac import Ipac
 from .daophot import Daophot
+from .qdp import QDP
 from .sextractor import SExtractor
 from .fixedwidth import (FixedWidth, FixedWidthNoHeader,
                          FixedWidthTwoLine, FixedWidthSplitter,
diff --git a/astropy/io/ascii/qdp.py b/astropy/io/ascii/qdp.py
index d6af72700..a62d2d6c0 100644
--- a/astropy/io/ascii/qdp.py
+++ b/astropy/io/ascii/qdp.py
@@ -536,7 +536,6 @@ class QDP(basic.Basic):
 
     """
     _format_name = 'qdp'
-    _io_registry_format_aliases = ['qdp']
     _io_registry_can_write = True
     _io_registry_suffix = '.qdp'
     _description = 'Quick and Dandy Plotter'
```
The `io_registry_format_aliases` should not be used in this case. It does require that for reading with `Table`, the `format` is given as `ascii.qdp` instead of just `qdp` but this is the preferred syntax for clarity.

astropy/io/ascii/tests/test_qdp.py
```diff
@@ -0,0 +1,135 @@
...
+    NO 1.14467592592593e-05    -1.14467592592593e-05   0.000000        NO
+    """
+
+    path = str(tmpdir.join('test.qdp'))
+    path2 = str(tmpdir.join('test2.qdp'))
+
+    with open(path, "w") as fp:
+        print(example_qdp, file=fp)
+    with pytest.warns(AstropyUserWarning) as record:
+        table = read_table_qdp(fp.name, names=["MJD", "Rate"],
```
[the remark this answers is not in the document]
matteobachetti: Right. done

astropy/io/ascii/tests/test_qdp.py
matteobachetti: Done

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,558 @@
...
+        print(" ".join(values), file=fobj)
+
+    if filename is None:
+        print(fobj.getvalue())
+        return fobj.getvalue().split("\n")
+
+
+class QDPSplitter(core.DefaultSplitter):
+    """
+    Split on comma for QDP tables
```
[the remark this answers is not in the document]
matteobachetti: Done

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,558 @@
...
+    Traceback (most recent call last):
+        ...
+    ValueError: Unrecognized QDP line...
+    """
+    line = line.strip()
+    if not line:
+        return 'comment'
+    match = _line_type_re.match(line)
+    if match is None:
+        raise ValueError(f'Unrecognized QDP line: {line}')
```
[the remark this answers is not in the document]
matteobachetti: It's in the doctest above

astropy/io/ascii/tests/test_qdp.py
matteobachetti: Should be done

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,558 @@
...
+
+    This file format can contain multiple tables, separated by a line full
+    of ``NO``s. Comments are exclamation marks, and missing values are single
+    ``NO`` entries.
+    Headers are just comments, and tables distributed by various missions
+    can differ greatly in their use of conventions. For example, light curves
+    distributed by the Swift-Gehrels mission have an extra space in one header
+    entry that makes the number of  inconsistent with the number of columns
+
+    ..
```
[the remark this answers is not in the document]
matteobachetti: Done

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,558 @@
...
+class QDP(basic.Basic):
+    """QDP table.
+
+    This file format can contain multiple tables, separated by a line full
+    of ``NO``s. Comments are exclamation marks, and missing values are single
+    ``NO`` entries.
+    Headers are just comments, and tables distributed by various missions
+    can differ greatly in their use of conventions. For example, light curves
+    distributed by the Swift-Gehrels mission have an extra space in one header
+    entry that makes the number of  inconsistent with the number of columns
```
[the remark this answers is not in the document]
matteobachetti: Done

matteobachetti:
@taldcroft thanks a lot for your thorough review until now, I think I addressed all of your comments :)

matteobachetti:
There is a failure I don't understand. Aren't warnings allowed?

taldcroft:
@matteobachetti - uncaught warnings are not allowed in testing, they are converted to errors as you've seen. Using https://docs.pytest.org/en/stable/warnings.html you need to catch them and test that the expected warning got emitted.

matteobachetti:
But I do test it, in test_qdp.py. This error comes from the generic test in test_write.py. I guess this assumes that ascii readers don't emit warnings with default values, but in this case, this reader is made for files that can contain multiple tables, and it has to warn the user if the user does not specify which table to read.

taldcroft:
@matteobachetti - you can put in some special-case handling for the `qdp` format to update `kwargs` so the test passes (if indeed it fully supports writing masked data). I.e. add the appropriate `table_id` (if I have gotten things right).

matteobachetti:
@taldcroft I don't think the failure has to do with my PR, so whenever you want, this is ready for review.

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+    elif isinstance(qdp_file, str):
+        with open(qdp_file) as fobj:
+            lines = fobj.readlines()
+    elif isinstance(qdp_file, Iterable):
+        lines = qdp_file
+
+    return lines
+
+
+# def _analyze_qdp_file(qdp_file):
```
taldcroft, line 152: Remove this commented-out function if it's not needed.

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+#     lines = _get_lines_from_file(qdp_file)
+#     return _get_type_from_list_of_lines(lines)
+
+
+def _interpret_err_lines(err_specs, ncols, names=None):
+    """Give list of column names from the READ SERR and TERR commands
+
+    Parameters
+    ----------
+    err_specs : dict, {'serr': [n0, n1, ...], 'terr': [n2, n3, ...]}
```
taldcroft, line 177: The top line should just be `err_specs : dict` (i.e. only specify the type) with the details in the next line.

See https://github.com/astropy/astropy/pull/11118/files for examples of where this is coming from.

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+        Error specifications for symmetric and two-sided errors
+    ncols : int
+        Number of data columns
+
+    Other parameters
+    ----------------
+    names : list of strings
```
taldcroft, line 184: `list of str`  (again https://github.com/astropy/astropy/pull/11118/files).

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+
+def _get_tables_from_qdp_file(qdp_file, input_colnames=None, delimiter=None):
+    """Get all tables from a QDP file
+
+    Parameters
+    ----------
+    qdp_file : str
+        Input QDP file name
+
+    Other parameters
```
taldcroft, line 268: Docstring doesn't match function signature.

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+    ----------------
+    names : list of strings
```
taldcroft, line 270: `list of str`

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+        Name of data columns (defaults to ['col1', 'col2', ...]), _not_
+        including error columns.
+
+    Returns
+    -------
+    tables : list of `Table` objects
```
taldcroft, line 276:
```
list of `~astropy.table.Table`
```

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+    qdp_file : str
+        Input QDP file name
+
+    Other parameters
+    ----------------
+    table_id : int, default 0
+        Number of the table to be read from the QDP file. This is useful
+        when multiple tables present in the file. By default, the first is read.
+
+    names : list of strings
```
taldcroft, line 421: Docstring parameters should have the same order as in the function. Also, `list of str`

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+        Name of data columns (defaults to ['col1', 'col2', ...]), _not_
+        including error columns.
+
+    delimiter : str
+        Any delimiter accepted by the `sep` argument of str.split()
+
+    Returns
+    -------
+    tables : list of `Table` objects
```
taldcroft, line 430:
```
list of `~astropy.table.Table`
```

astropy/io/ascii/qdp.py
```diff
@@ -0,0 +1,578 @@
...
+    Data that uses the :class:`astropy.io.ascii.basic.CsvSplitter`
+    """
+    splitter_class = QDPSplitter
+    fill_values = [(core.masked, 'NO')]
+    comment = "!"
+    write_comment = None
+
+
+class QDP(basic.Basic):
+    """QDP table.
```
taldcroft, line 538: This main docstring should be expanded to illustrate the available parameters and usage. This will be the only way people know about the `header_id`, `err_specs` etc. For a good example see the following that shows a full example of the file format and how to read. You can also include a write example. https://github.com/astropy/astropy/blob/1ed1d945aadbdb9548e37eccfcc28115e3d03ca8/astropy/io/ascii/daophot.py#L315

taldcroft (changes requested):
@matteobachetti - I took another pass and there are just docs and minor test issues to be fixed. One key thing is that you need to rebase on current master and update this to reflect the new change log system. See emails to astropy-dev (ANN: astropy plans to switch to towncrier for changelog management), in particular the one from Simon that talks about how to fix existing PRs. Hopefully the rebase will fix the CI failure as well.

Welcome to astropy development, it's never quite easy!  :smile:

matteobachetti:
@taldcroft should be done, including the rebase. I'm not sure the CHANGES.rst file is correct after rebase though.

pllim:
We don't use `CHANGES.rst ` anymore, see https://github.com/astropy/astropy/blob/main/docs/changes/README.rst

pllim:
@saimn , is squashing necessary here to change the changelog ?

matteobachetti:
The failures should be unrelated. The failing test is about a config file, and unit tests were passing up until 04db56e (the only failure was from readthedocs, bad formatting of a docstring)

pllim:
That is weird (the config failure). Luckily, that test is getting removed in #10877 . Unluckily, it failing is blocking the rest of the jobs for you.

saimn:
> @saimn , is squashing necessary here to change the changelog ?

No, just revert the changes in `CHANGES.rst` (`git checkout master -- CHANGES.rst`)

matteobachetti:
@pllim @taldcroft I renamed my fork's `master` branch to `main`, rebased this branch, and eliminated the changes to `CHANGES.rst` as suggested. Thanks for your help :)

matteobachetti:
Due to the trial-and-error path I followed and the initial proposed addition to `io.misc` then moved to `io.ascii`, there are many commits in this PR. I might give a stab at reducing the number of commits by rebasing interactively one last time, if you think it's useful (or you might just squash and merge when satisfied with the changes)

taldcroft:
Yes squashing the commits would be good, maybe even down to just one.

matteobachetti:
@taldcroft I tried to preserve the main "building blocks" of this PR, while discarding the initial io.misc interface. It's a quite big rebase, hope it works :D

taldcroft (approved):
@matteobachetti - looks great now, thanks for all the hard work and perseverance!

PR #11258, opened 2021-01-19
Add unit_aliases context manager

### LINKED ISSUE #11246: Context manager for handling "bad" units
Lots of existing data has unit strings which are problematic in some way: plural rather than singular (e.g. "Angstrom" vs "Angstroms"), alternate spellings (e.g. "Angstrom" vs "Angstroem" or "meter" vs "metre", although the latter should probably be allowed), or alternate shortened forms (e.g. "Ang" for "Angstrom"). The options for handing this are not ideal: users could create custom units to allow parsing the strings, though this would cause these problematic strings to propagate; try to do something with the ` UnrecognizedUnit` to correct the units (though this does tend to produce a large number of warnings, which is not ideal); or based on prior experience/knowledge hard code the units (which makes their code less robust to changes/new inputs).

An idea I had would be to have a context manager that allows setting temporary aliases—this avoid the need to define new units, and reduce warning fatigue, and prevent the propagation of bad unit strings. Usage would be something like:
```python
import astropy.units as u
from some_project import DataClass

with u.unit_aliases({"Angstroms": u.AA}):
    data = DataClass.read("some_file.fits")
```

aragilar:
Closes #11246.

I'm not sure the test file I picked is the best (it's one of the hosted astropy ones, a local one would likely be better), so if people have suggestions for files/other sources which would be good test cases (especially ones with more than one alias required), feel free to push them to the branch.

Also, I couldn't find the test file for the unit registries, so I created a new one.

I haven't added any prose docs/examples to this, any suggestions where such examples would fit in best?


astropy/units/core.py
```diff
@@ -262,6 +264,38 @@ def add_enabled_equivalencies(self, equivalencies):
         equivalencies = _normalize_equivalencies(equivalencies)
         self._equivalencies |= set(equivalencies)
 
+    @contextmanager
+    def unit_aliases(self, aliases):
+        """
+        Context manager for temporarily adding aliases for units (e.g. to handle
```
pllim, line 270:
```suggestion
        Context manager for temporarily adding aliases for units (e.g., to handle
```

pllim (commented):
Thanks! Needs a change log and user facing doc with examples as well.

Making the tests requiring remote data is unnecessary. I propose you store the header in memory as test module variable or something, as all you need for unit parsing is the header object with one `BUNIT` entry (beyond what is mandatory for a simple FITS header). **OR** to make it even simpler, pass in the value of the BUNIT instead of accessing FITS header (e.g., `u.Unit(Angstromsss) == ...`).

pllim:
p.s. You might also want to address codecov comments about lines not covered by tests.

astropy/units/core.py
```diff
@@ -447,6 +481,26 @@ def add_enabled_equivalencies(equivalencies):
...
+
+    Parameters
+    ----------
+    aliases : dict (or other mapping) of str:unit pairs
+        The aliases to temporarily use. The keys must be the aliases stored
+        as strings, which must not have been used for any existing units (a
+        ValueError is raised otherwise). The values must be the astropy unit
+        that the alias will be mapped to.
+    """
+    registry = get_current_unit_registry()
```
mhvk, line 500: So, this would become:
```
    # get a context with a new registry, which is a copy of the current one
    context = _UnitContext(get_current_unit_registry())
    # in this new current registry, enable the further units requested
    get_current_unit_registry().add_aliased_units(aliases)
    return context
```

astropy/units/tests/test_registry.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+import astropy.io.fits as fits
+import astropy.units as u
+from astropy.utils.data import get_pkg_data_filename
+
+
+@pytest.mark.remote_data
```
mhvk, line 8: I do like the very explicit test, but think this one may make more sense in `io.fits` (inside `units`, I'd prefer not to use anything else from astropy).  Aside: if it is from the astropy server, you can use `remote_data(source="astropy")`. Or maybe better to get just the header as @pllim suggests (the name suggests this is a large file!).

I'd then add some very simple tests just using `u.Unit('counts')`, like @pllim suggested.

For the location, you could just use `test_units.py`. That file is where `add_enabled_units` is tested as well - though "tested" is a bit of an overstatement with just one or two tests! And `set_enabled_units` is not tested explicitly at all...

Alternatively, since `test_units.py` is very long and there are rather few registry tests in `test_units.py` anyway, you could move those tests here.

mhvk (commented):
That looks nice! Some comments inline, mostly about using the function, not the method, as a context manager.

You pose a very good question about where to document this... Looking at our docs, we have a section about "Unrecognized Units" inside "String Representations of Units", and it just describes what happens, not how to solve it. My sense would be to make that its own section "Dealing with Unrecognized Units" and then have a part about how to define an alias. Ideally, we then cross-reference this from, e.g., the fits pages...  

I fear this will be more work than coding it up...  But it will also be very nice. Indeed, amply nice enough that I'm going to add the "what's new needed" label...

mhvk:
Ah, I didn't have to - @pllim already added the label!

mhvk:
@aragilar - this is quite a nice addition. Will you have time to finish it up?  Feature freeze for 4.3 is coming up...

aragilar:
I've expanded the section on Unrecognized Units, and added a changelog entry. I probably won't have a great amount of time before feature freeze to add more documentation (or other changes), but feel free to push changes.

mhvk:
@aragilar - thanks for the docs! I had another look and fixed the bugs. Another thing I realized is that it would be nicer if the name matched a bit better what the other methods of the registry had, so I changed it to `add_enabled_aliases`. Also, most often one probably does not want to add, but just set some aliases, so I added a `set_enabled_aliases` method as well.

p.s. I shortened your changelog fragment, since I think this needs a what's-new -- at least if I can get it to work when reading FITS files!

mhvk:
@adrn - as I contributed to this PR, it is probably good to have another set of eyes for approval...

docs/units/format.rst
```diff
@@ -188,15 +188,19 @@ following formats:
...
+
+Dealing with Unrecognized Units
+===============================
 
 Since many files found in the wild have unit strings that do not
 correspond to any given standard, `astropy.units` also has a
 consistent way to store and pass around unit strings that did not
-parse.
+parse.  In addition, it provides tools for transforming non-standard,
+legacy or misspelt unit strings into their standardised form,
```
adrn, line 200:
```suggestion
legacy or misspelt unit strings into their standardized form,
```
(for consistency, we use US spelling in the documentation https://docs.astropy.org/en/stable/development/style-guide.html#spelling)

docs/units/format.rst
```diff
@@ -221,10 +225,71 @@ this behavior:
   - ``'silent'``: return an `~astropy.units.UnrecognizedUnit`
     instance.
 
+By adding additional unit aliases  for the misspelt units with
```
adrn, line 228:
```suggestion
By adding additional unit aliases for the misspelt units with
```

docs/units/format.rst
```diff
@@ -221,10 +225,71 @@ this behavior:
...
+`~astropy.units.set_enabled_aliases` (e.g. 'Angstroms' for 'Angstrom'), and new units
```
adrn, line 229:
```suggestion
`~astropy.units.set_enabled_aliases` (e.g., 'Angstroms' for 'Angstrom') as demonstrated below, and defining new units
```

docs/units/format.rst
```diff
@@ -221,10 +225,71 @@ this behavior:
...
+via `~astropy.units.def_unit` and `~astropy.units.add_enabled_units`, we can use
+``parse_strict='raise'`` to rapidly find issues with the units used, while also
+being able to read in older datasets where the unit usage may have been less
+standard.
+
+.. tip::
+
+    Not all formats currently support defining aliases. If you are using a
+    unit format that is not supported, you can convert the unit string back
+    into a Python string, and reparse the string with the default format::
```
adrn, line 239: Hm, I found the phrasing here a little confusing.

astropy/units/core.py
```diff
@@ -261,6 +272,48 @@ def add_enabled_equivalencies(self, equivalencies):
         equivalencies = _normalize_equivalencies(equivalencies)
         self._equivalencies |= set(equivalencies)
 
+    def set_enabled_aliases(self, aliases):
+        """
+        Set aliases for units, e.g., to handle alternate spellings.
```
adrn, line 277:
```suggestion
        Set aliases for units.
        
        This is useful for handling alternate spellings of unit names, or misspelled unit names.
```

astropy/units/core.py
```diff
@@ -261,6 +272,48 @@ def add_enabled_equivalencies(self, equivalencies):
...
+
+        Parameters
+        ----------
+        aliases : dict of str, Unit
+            The aliases to set. The keys must be the string aliases, and values
+            must be the `astropy.units.Unit` that the alias will be mapped to.
```
adrn, line 283:
```suggestion
            must be the `astropy.units.Unit` that the alias will be mapped to. For example, ``{"parsnek": u.parsec}``.
```
(actual example might be too much of a joke but I think it would be good to have some small example here?

astropy/units/core.py
```diff
@@ -261,6 +272,48 @@ def add_enabled_equivalencies(self, equivalencies):
...
+
+        Raises
+        ------
+        ValueError
+            If the alias already defines a different unit.
```
adrn, line 288: Does this get raised? Maybe a copy-paste error from `add_`?

astropy/units/core.py
```diff
@@ -261,6 +272,48 @@ def add_enabled_equivalencies(self, equivalencies):
...
+        """
+        self._reset_aliases()
+        self.add_enabled_aliases(aliases)
+
+    def add_enabled_aliases(self, aliases):
+        """
+        Add aliases for units, e.g., to handle alternate spellings.
```
adrn, line 295:
```suggestion
        Add aliases for units.
        
        This is useful for handling alternate spellings of unit names, or misspelled unit names.
```

astropy/units/core.py
```diff
@@ -261,6 +272,48 @@ def add_enabled_equivalencies(self, equivalencies):
...
+
+        Parameters
+        ----------
+        aliases : dict of str, Unit
+            The aliases to add. The keys must be the string aliases, and values
+            must be the `astropy.units.Unit` that the alias will be mapped to.
```
adrn, line 301: Same example as above?

astropy/units/core.py
```diff
@@ -446,6 +499,73 @@ def add_enabled_equivalencies(equivalencies):
     return context
 
 
+def set_enabled_aliases(aliases):
+    """
+    Set aliases for units, e.g., to handle alternate spellings.
```
adrn, line 504: Hm, I guess same comment as above.

astropy/units/core.py
```diff
@@ -446,6 +499,73 @@ def add_enabled_equivalencies(equivalencies):
...
+    # get a context with a new registry, which is a copy of the current one
+    context = _UnitContext(get_current_unit_registry())
+    # in this new current registry, enable the further equivalencies requested
+    get_current_unit_registry().set_enabled_aliases(aliases)
+    return context
+
+
+def add_enabled_aliases(aliases):
+    """
+    Add aliases for units, e.g., to handle alternate spellings.
```
adrn, line 536: And once more...

adrn (changes requested):
Mostly documentation checks so far. I think the code and tests look fine to me!

astropy/units/core.py
mhvk: It can get raised if one is trying to define an alias using a string that already exists as a regular unit

adrn: Ah, right

docs/units/format.rst
mhvk: OK, done.

docs/units/format.rst
mhvk: Rephrased.

docs/units/format.rst
mhvk: Fortunately, I found a different method that means all formats are now supported, so this can go. Will push a revised PR shortly.

astropy/units/core.py
mhvk: Done

astropy/units/core.py
mhvk: I ended up thinking the sentence was mostly useful for the functions, much less so in the method (which the user essentially never sees...)

astropy/units/core.py
mhvk: I'd tend to just leave it to the more commonly used functions, which I did give an example.

mhvk:
@adrn - I rather changed the implementation, to have the actual format parsers become aware of the aliases. This seemed the only way to ensure it would work also in FITS, CDS, VOUnit, and OGIP. Conceptually, now if the parser does not recognize an individual unit string, it will check the aliases (in all formats; possibly not quite logical, since most do not use the registry, but certainly easiest...).

So, this needs a re-review, I fear... I added some comments in-line.

astropy/units/format/ogip.py
```diff
@@ -351,14 +351,6 @@ def p_error(p):
 
         return parsing.yacc(tabmodule='ogip_parsetab', package='astropy/units')
 
-    @classmethod
```
mhvk, line 354: This is inherited directly from generic, so was superfluous to start with.

astropy/units/format/vounit.py
```diff
@@ -98,23 +98,30 @@ def parse(cls, s, debug=False):
...
     @classmethod
-    def _parse_unit(cls, unit, detailed_exception=True):
-        if unit not in cls._units:
-            if cls._explicit_custom_unit_regex.match(unit):
-                return cls._def_custom_unit(unit)
+    def _get_unit(cls, t):
+        try:
+            super()._get_unit(t)
+        except ValueError:
+            if cls._explicit_custom_unit_regex.match(t.value):
```
mhvk, line 105: Moved from the original `_parse_unit` - so that any custom units are defined *after* possible aliases are checked. Arguably perhaps even a bit more logical.

mhvk:
OK, solved a stupid bug and added some tests inside `io.fits` to show that it also works there.

astropy/units/core.py
```diff
@@ -261,6 +269,59 @@ def add_enabled_equivalencies(self, equivalencies):
...
+        """
+        for alias, unit in aliases.items():
+            if alias in self._registry and unit != self._registry[alias]:
+                raise ValueError(
+                    f"{alias} already means {self._registry[alias]}, so "
+                    f"cannot be used as an alias for {unit}.")
+            if alias in self._aliases and unit != self._aliases[alias]:
+                raise ValueError(
+                    f"{alias} already is an alias for {self._aliases[alias]}, so "
+                    f"cannot be used as an alias for {unit} too.")
```
adrn, line 319: Removing the word "too" in the 2nd error message, I think these two if clauses can be "or"'d and have just one error to raise? I like the phrasing of the second better ("already is an alias for" vs. "already means")

docs/units/format.rst
```diff
@@ -221,10 +225,64 @@ this behavior:
   - ``'silent'``: return an `~astropy.units.UnrecognizedUnit`
     instance.
 
+By either adding additional unit aliases for the misspelt units with
+`~astropy.units.set_enabled_aliases` (e.g., 'Angstroms' for 'Angstrom'; as
+demostrated below), or defining new units via `~astropy.units.def_unit` and
```
adrn, line 230:
```suggestion
demonstrated below), or defining new units via `~astropy.units.def_unit` and
```

docs/units/format.rst
```diff
@@ -221,10 +225,64 @@ this behavior:
...
+
 Examples
 --------
 
-.. EXAMPLE START: Handling Unrecognized Units
+.. EXAMPLE START: Define Aliases for Units
+
+To set unit aliases, pass `~astropy.units.set_enabled_aliases` a
+dictionary mapping the misspelt string to an astropy unit. The following code
+snippet shows how to set up Angstroem -> Angstrom::
```
adrn, line 243:
```suggestion
dictionary mapping the misspelt string to an astropy unit. The following code
snippet shows how to set up an alias for a hypothetical nonstandard unit 
"Angstroem" to map to the standard unit Angstrom::
```

docs/units/format.rst
```diff
@@ -221,10 +225,64 @@ this behavior:
...
+You can also set multiple aliases up at once or add to existing ones::
+
+    >>> u.set_enabled_aliases({"Angstroem": u.Angstrom, "Angstroms": u.Angstrom})
+    <astropy.units.core._UnitContext object at 0x...>
+    >>> u.add_enabled_aliases({"angstroem": u.Angstrom})
+    <astropy.units.core._UnitContext object at 0x...>
+    >>> u.Unit("Angstroem") == u.Unit("Angstroms") == u.Unit("angstroem") == u.Angstrom
+    True
+
+The aliases can be reset by passing an empty directory::
```
adrn, line 261:
```suggestion
The aliases can be reset by passing an empty dictionary::
```

adrn (approved):
New implementation looks good! Some suggested documentation changes.

astropy/units/core.py
mhvk: I had that initially but then the f-string becomes annoying, since one has to switch between `self._registry` and `self._aliases`...  I'll just delete "too"...

docs/units/format.rst
mhvk: Done

docs/units/format.rst
mhvk: Ha! That's an odd one. Fixed.

mhvk:
@adrn - thanks for the review. Since it is all docs, I'll push with `[skip ci]`

mhvk:
@aragilar - thanks for starting this, and apologies for running away with your PR...  But it will now be in 4.3!

adrn:
Failure unrelated, and previously approved

aragilar:
@mhvk That's cool, there's even support for the other formats, which is even better!

Issue #11215, opened 2021-01-06
astropy handles spectral wcs units with logs incorrectly


aragilar:
It looks either wcslib or astropy's wrapping of it do not handle the units on the wcs object correctly, and instead override the spectral units to metres, when `"WAVE-LOG"` is used (note the change to metres in both the output of `pixel_to_world` and on the `w.wcs` object):

```python
from astropy.wcs import WCS
import numpy as np
from astropy.units import Unit
w0 = 3.57880000000000E+00
w1 = 1.00000000000000E-04
crval = 10 ** w0
cdelt = crval * w1 * np.log(10)
cunit = Unit('Angstrom')
ctype = "WAVE-LOG"
w = WCS(naxis=1)
w.wcs.crval[0] = crval
w.wcs.cdelt[0] = cdelt
w.wcs.ctype[0] = ctype
w.wcs.cunit[0] = cunit
print(w)
# OUT: WCS Keywords
# OUT: Number of WCS axes: 1
# OUT: CTYPE : 'WAVE-LOG'  
# OUT: CRVAL : 3791.4034418330666  
# OUT: CRPIX : 0.0  
# OUT: PC1_1  : 1.0  
# OUT: CDELT : 0.8730029046691138  
# OUT: NAXIS : 0  0
print(w.wcs.cunit)
# OUT: ['Angstrom']
print(w.pixel_to_world(np.arange(10)))
# OUT: [3792.27644474 3793.14944764 3794.02245055 3794.89545345 3795.76845636
# OUT:  3796.64145926 3797.51446217 3798.38746507 3799.26046798 3800.13347088] m
print(w.wcs.cunit)
# OUT: ['m']
### 
```

### System Details
```
>>> import platform; print(platform.platform())
Linux-5.10.0-1-amd64-x86_64-with-glibc2.31
>>> import sys; print("Python", sys.version)
Python 3.9.1 (default, Dec  8 2020, 07:51:42)
[GCC 10.2.0]
>>> import numpy; print("Numpy", numpy.__version__)
Numpy 1.19.5
>>> import astropy; print("astropy", astropy.__version__)
astropy 4.2
```

pllim:
I don't think `cunit` can take `astropy.units.Unit`, can it?

There was a different issue also related to `wcs` and units at #11173 .

mhvk:
[...]

PR #11259, opened 2021-01-19
Warn about usage of Wcsprm

aragilar:
This is in response to #11215, hopefully helping users avoid getting weird results when they don't call `set()`.

### DIFF
--- a/astropy/wcs/docstrings.py
+++ b/astropy/wcs/docstrings.py
@@ -2232,6 +2232,29 @@ However, it does recognize free-format character (NOST 100-2.0,
 Sect. 5.2.1), integer (Sect. 5.2.3), and floating-point values
 (Sect. 5.2.4) for all keywords.
 
+
+.. warning::
+    There are two possible ways of writing to the attributes on this class::
+
+        Wcsprm.crval = [crval]
+        Wcsprm.cdelt = [cdelt]
+        Wcsprm.ctype = [ctype]
+        Wcsprm.cunit = [cunit]
+
+    or::
+
+        Wcsprm.crval[0] = crval
+        Wcsprm.cdelt[0] = cdelt
+        Wcsprm.ctype[0] = ctype
+        Wcsprm.cunit[0] = cunit
+
+    The former should be strongly preferred as it calls additional helpers which
+    ensure that the class is in the correct state. If the latter is used,
+    `Wcsprm.set()` *MUST* be called once the attributes are set (and, depending
+    on the changes, in between setting attributes), otherwise invalid results
+    may be returned when using the object with WCS functions and methods.
+
+
 Parameters
 ----------
 header : An `astropy.io.fits.Header`, string, or `None`.


pllim:
Thanks! I tentatively set milestone to LTS, but subpackage maintainer or release manager can feel free to move the milestone if backport to LTS isn't possible or relevant.

astropy/wcs/docstrings.py
```diff
@@ -2232,6 +2232,29 @@ def RA_DEC_ORDER(indent=0):
 Sect. 5.2.1), integer (Sect. 5.2.3), and floating-point values
 (Sect. 5.2.4) for all keywords.
 
+
+.. warning::
+    There are two possible ways of writing to the attributes on this class::
```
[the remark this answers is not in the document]
mcara: In addition, even if you look at this as an _acceptable way_ of _setting_ attributes (when it actually is a modification of the object returned by a getter), it would not apply to all attributes in this class. For example, "second way" would not work on `Wcsprm.name` or `Wcsprm.velosys`. Therefore, I think this note should be more specific about parameters to which it applies.

astropy/wcs/docstrings.py
```diff
@@ -2232,6 +2232,29 @@ def RA_DEC_ORDER(indent=0):
...
+
+        Wcsprm.crval = [crval]
+        Wcsprm.cdelt = [cdelt]
+        Wcsprm.ctype = [ctype]
+        Wcsprm.cunit = [cunit]
+
+    or::
+
+        Wcsprm.crval[0] = crval
```
mcara, line 2246: and modify this example as 
```suggestion
        Wcsprm.crval[0] = crval1
```

astropy/wcs/docstrings.py
```diff
@@ -2232,6 +2232,29 @@ def RA_DEC_ORDER(indent=0):
...
+        Wcsprm.cdelt[0] = cdelt
+        Wcsprm.ctype[0] = ctype
+        Wcsprm.cunit[0] = cunit
+
+    The former should be strongly preferred as it calls additional helpers which
+    ensure that the class is in the correct state. If the latter is used,
+    `Wcsprm.set()` *MUST* be called once the attributes are set (and, depending
+    on the changes, in between setting attributes), otherwise invalid results
+    may be returned when using the object with WCS functions and methods.
```
mcara, line 2255: Again, I do not believe this second approach should be recommended under any circumstances because it may produce unexpected results in future releases since it relies on inner workings of the `Wcsprm` class.

By the way, the need to call `Wcsprm.set()` is somewhat mentioned in [its documentation](https://docs.astropy.org/en/stable/api/astropy.wcs.Wcsprm.html#astropy.wcs.Wcsprm.set):

> Some attributes that are based on other attributes (such as lattyp on ctype) may not be correct until after set is called.

mcara (changes requested):
I left some comments below. Fundamentally, I do not think this second method should even be mentioned as a possible way of setting class attributes. I think you should limit the scope of this PR and simply warn users how to set _some_ attributes in the _only one correct way_.

aragilar:
I do think we need to mention `Wcsprm.cunit[0] = cunit` and similar, to tell users not to do it, as most of the code I saw did use the indexing method, which is why I initially wrote my code the way I did. I'm happy to be more explicit about not using indexing.

mcara:
> I do think we need to mention `Wcsprm.cunit[0] = cunit` and similar, to tell users not to do it...

What about something along the line of explaining the correct way and then explain that `attr[0] = ...` is wrong and should never be used.

aragilar:
@mcara I've updated the wording to be more explicit, do you think the new wording better fits the aim of getting users to not use slicing?

mcara:
Thank you! However, I still think that current wording and numerous examples are too much for a _warning_ and, I believe, a warning would be more efficient if it were brief. For example, I think there are too many examples. We do not need to list all possible attributes. One attribute should suffice. May I suggest something like this:

```
[Many of the attributes of this class require additional processing when modifying underlying C structure. When needed, this additional processing is implemented in attribute setters. Therefore, ] For mutable attributes, one should always set the attribute rather than a slice of its current value (or its individual elements) since the latter may lead the class instance to be in an invalid state.
[For example, attribute `crpix` of a 2D WCS' `Wcsprm` object `wcs` should be set as `wcs.crpix = [crpix1, crpix2]` instead of `wcs.crpix[0] = crpix1; wcs.crpix[1] = crpix2`.]
```
I put in square bracket text that I would consider optional but that could add some detail. What do you think?

aragilar:
@mcara That seems reasonable, I'll update the text.

mcara:
@aragilar Two more things:

1. Except maybe for `Wcsprm`, all other back-quotes should be double back-quotes such as:

```
``wcs.crpix[0] = crpix1; wcs.crpix[1] = crpix2``
```
Without this, I am sure one would get plenty of warnings when building documentation. See https://sublime-and-sphinx-guide.readthedocs.io/en/latest/inline.html#code-text and https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#roles This is to avoid Sphinx trying to create references - see errors in the readthedocs build:
```
/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/11259/lib/python3.8/site-packages/astropy/wcs/__init__.py:docstring of astropy.wcs.Wcsprm:20: WARNING: py:obj reference target not found: wcs
/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/11259/lib/python3.8/site-packages/astropy/wcs/__init__.py:docstring of astropy.wcs.Wcsprm:20: WARNING: py:obj reference target not found: wcs.crpix = [crpix1, crpix2]
/home/docs/checkouts/readthedocs.org/user_builds/astropy/envs/11259/lib/python3.8/site-packages/astropy/wcs/__init__.py:docstring of astropy.wcs.Wcsprm:20: WARNING: py:obj reference target not found: wcs.crpix[0] = crpix1; wcs.crpix[1] = crpix2]
```

2. After you do this, please squash all commits. See https://docs.astropy.org/en/stable/development/workflow/development_workflow.html#how-to-squash for guidance and keep in mind that in that documentation `astropy` is used instead of what customarily is called `upstream`.

Before squashing and pushing, try to build documentation locally (for example, `cd docs` and `build html`).

mcara:
Thanks @aragilar !

PR #11263, opened 2021-01-21
Small performance refactor of sigma clipping

mhvk:
While looking at the code in the context of #11219, I noticed that the tracking of the number of rejected pixels could be done a bit more efficiently. Since most time is spent in calculating medians, etc., it does not have a very large effect, but even 5-10% is probably worthwhile. Plus the code is a bit simpler.


larrybradley (approved):
Thanks, @mhvk!

PR #11264, opened 2021-01-21
Define TESTED_VERSIONS for the next version of pytest_astropy_header

saimn:
pytest_astropy_header will no more define Astropy by default in TESTED_VERSIONS, so let's add it here. This is compatible with the current version of pytest_astropy_header.

### DIFF
--- a/astropy/conftest.py
+++ b/astropy/conftest.py
@@ -11,12 +11,14 @@ import tempfile
 import warnings
 
 try:
-    from pytest_astropy_header.display import PYTEST_HEADER_MODULES
+    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
 except ImportError:
     PYTEST_HEADER_MODULES = {}
+    TESTED_VERSIONS = {}
 
 import pytest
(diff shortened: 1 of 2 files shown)


saimn:
@pllim - I set the milestone to 4.0.5, since if we backport this we don't need to pin pytest_astropy_header for the older version.

pllim:
Re: pinning -- That is true.

Though I am unsure how to handle users only upgrading `pytest-astropy-header` without also upgrading `astropy` and then run tests -- perhaps this care is rare enough that memo to dev mailing list is sufficient?

pllim (approved):
```
Running tests with Astropy version 4.3.dev529+ga2a90fa1d.
Running tests in astropy docs.
```

LGTM. Thanks!

saimn:
> Though I am unsure how to handle users only upgrading pytest-astropy-header without also upgrading astropy and then run tests -- perhaps this case is rare enough that memo to dev mailing list is sufficient?

At worst they won't see the "Running tests with Astropy version ..." line, so I think it's fine ;)

PR #11269, opened 2021-01-22
WIP: Add GPS time scale

### LINKED ISSUE #11265: GPS time unexpected isot representation
### Description

GPS time is correctly defined as `1980-01-06T00:00:19 TAI` as its epoch.  Many text formats using GPS time, such as RINEX messages, have defined the ISO or ISO-like text datetimes for GPS to be `TAI - 19s` (leap seconds at the `1980-01-06T00:00:00 UTC` epoch).  With GPS time being specified via the format argument and not having a GPS scale since it follows TAI, the isot representation is not as expected by these standard formats.

### Expected behavior

```python
In [1]: t = Time(0, format='gps')

In [2]: t.isot
Out[2]: '1980-01-06T00:00:00.000'

In [3]: t.tai.isot
Out[3]: '1980-01-06T00:00:19.000'
```
As below under actual behavior, I would not expect the `t.isot` and `t.tai.isot` values to be the same.  I'm not sure if this is a similar issue to #11245 or not, but seems like it could be.  Not sure if this would need effectively a GPS time scale in addition to the current TAI to make it work.  For now, the only work around I've found is explicitly removing the 19 leap second offset for the `.isot` in a wrapper method.  Am I missing something key or is this the only way to handle this at the moment, and should this functionality be added? Obviously using `t.utc.isot` is not the correct answer as that would add all of the leap seconds since the GPS epoch for current datetimes.

```python
In [1]: (Time(0, format='gps') + TimeDelta(-19, format='sec')).isot
Out[1]: '1980-01-06T00:00:00.000'
```

### Actual behavior

```python
In [85]: t = Time(0, format='gps')

In [86]: t.isot
Out[86]: '1980-01-06T00:00:19.000'

In [87]: t.tai.isot
Out[87]: '1980-01-06T00:00:19.000'

In [88]: t.utc.isot
Out[88]: '1980-01-06T00:00:00.000'

In [89]: t.gps
Out[89]: 0.0

In [90]: t.utc
Out[90]: <Time object: scale='utc' format='gps' value=0.0>
```

### System Details

Linux-4.19.0-13-amd64-x86_64-with-debian-10.7
Python 3.7.3 (default, Jul 25 2020, 13:03:44) 
[GCC 8.3.0]
Numpy 1.16.2
astropy 4.0.1.post1
Scipy 1.5.2
Matplotlib 3.0.2

taldcroft:
### Description

This pull request is to add a new GPS time scale `gps_scale` which is TAI - 19 sec. This is basically a go-back and change of course from #1879.

This is the first instance of adding a new custom time scale that is not handled by ERFA. At first glance it isn't too hard, though of course if we keep doing this then the transform graph will start getting longer and longer. But for this first one just enumerating the transforms was easy enough.

I do wish we could have a do-over on #1164 and call the time format `gps_sec` (ala `cxcsec`) and have `gps` for the name of the scale. So it goes.

This is a work in progress pending discussion. To do:

- [ ] Docs
- [ ] Tests
- [ ] Changelog

Fixes #11265

### DIFF
--- a/astropy/time/core.py
+++ b/astropy/time/core.py
@@ -39,12 +39,18 @@ __all__ = ['TimeBase', 'Time', 'TimeDelta', 'TimeInfo', 'update_leap_seconds',
            'ScaleValueError', 'OperandTypeError']
 
 
-STANDARD_TIME_SCALES = ('tai', 'tcb', 'tcg', 'tdb', 'tt', 'ut1', 'utc')
+STANDARD_TIME_SCALES = ('gps_scale', 'tai', 'tcb', 'tcg', 'tdb', 'tt', 'ut1', 'utc')
 LOCAL_SCALES = ('local',)
 TIME_TYPES = dict((scale, scales) for scales in (STANDARD_TIME_SCALES, LOCAL_SCALES)
                   for scale in scales)
 TIME_SCALES = STANDARD_TIME_SCALES + LOCAL_SCALES
-MULTI_HOPS = {('tai', 'tcb'): ('tt', 'tdb'),
+MULTI_HOPS = {('gps_scale', 'tcb'): ('tai', 'tt', 'tdb'),
+              ('gps_scale', 'tcg'): ('tai', 'tt'),
+              ('gps_scale', 'tdb'): ('tai', 'tt'),
+              ('gps_scale', 'tt'): ('tai',),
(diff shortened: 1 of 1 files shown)


pllim:
Should we move the milestone given the upcoming feature freeze and the WIP state of this PR?

pllim:

Issue #10079, opened 2020-03-27
Quantity converts interger array to float array with copy=False and thus copies


maxnoe:
### Expected behavior

Creating a Quantity from another array with copy=False should keep data type and not copy the data.

### Actual behavior

Creating a quantity from an integer array with `copy=False` converts the input data to float
and thus copies.

### Steps to Reproduce
```python
import numpy as np
import astropy.units as u
a = np.zeros(10, dtype=int)
q = u.Quantity(a, copy=False)
print(q.dtype, np.shares_memory(q, a))
```
gives:
```
float64 False
```

pllim:
Maybe a workaround for now is to explicitly cast dtype?
```python
>>> import numpy as np 
>>> import astropy.units as u                                                                   
>>> a = np.zeros(10, dtype=int)                                                                 
>>> q = u.Quantity(a, copy=False, dtype=int)                                                    
>>> print(q.dtype, np.shares_memory(q, a))                                                      
int64 True
```

`Quantity`  has always cast int to float, AFAIK. To change that now would break some code downstream.

mhvk:
Hmm, that's somewhat tricky, and reading the docstring, the current behaviour is probably closest to how it is described: that there is an implicit dtype of `float` (or complex as needed). But your explicit example does not feel quite right.

Before going ahead and consider changing, though, we should think about implications. For instance, I can quite easily see a user doing something like `np.arange(10) << u.s` (mostly because I do that all the time myself!) - here, the `<<` implies `copy=False` (and that is in fact how it is implemented), but is it really a good idea to let the integers go through?  Almost certainly not what was intended...

One possible solution might be to go ahead and make `copy=False` not change the dtype, but have an additional possibility (e.g., `copy=None`??) for the case where we do want to copy if the dtype is not float/complex already.  Obviously, with that, there is a risk that it would break current (slightly poorly written) code, possibly giving quite hard to debug errors.

Of course, yet another alternative is just to make the current behaviour more explicit in the docstring, and just give the work-around that @pllim suggested (or to have a new flag for copy, say `copy='never'`).

Right now, my tendency would be to go for trying to document better, as I think one generally really should not use integers with units, and hence it is OK to make getting integer quantities a bit trickier if that avoids risks elsewhere. What do you think?

MaxNoe:
> Almost certainly not what was intended...

I tend to disagree. Everyone *should* be aware that `np.arange` as dtype integer by default. It has an dtype argument. 

[...]

PR #11273, opened 2021-01-26
Add warning about integer casting in Quantity

dstansby:
Personally I think this is surprising enough behaviour that is warrants a prominent warning. xref https://github.com/astropy/astropy/issues/10079

### DIFF
--- a/astropy/units/quantity.py
+++ b/astropy/units/quantity.py
@@ -272,6 +272,8 @@ class Quantity(np.ndarray):
     Quantities can also be created by multiplying a number or array with a
     :class:`~astropy.units.Unit`. See https://docs.astropy.org/en/latest/units/
 
+    Unless the ``dtype`` argument is explicitly specified, integer
+    or (non-Quantity) object inputs are converted to `float` by default.
     """
     # Need to set a class-level default for _equivalencies, or
     # Constants can not initialize properly


astropy/units/quantity.py
```diff
@@ -266,7 +266,12 @@ class Quantity(np.ndarray):
     TypeError
         If the unit provided is not either a :class:`~astropy.units.Unit`
         object or a parseable string unit.
-
+        
+    Warnings
```
pllim, line 270: I think `.. warning::` at the top might give you a nice red box. But if this is clear enough, I guess that is okay too. Will have to see how RTD renders this.

Are you able to address the PEP 8 and squash? Thanks!

pllim:
Can we close #10079 after merging this?

astropy/units/quantity.py
dstansby: Sorry, that's what I get by using the web interface 😆 . Will fix in a bit.

dstansby:
> Can we close #10079 after merging this?

FWIW I would advocate to deprecating and eventually removing automatically casting `int` to `float` in `Quantity` so one can create a Quantity without copying intergers, unless there's good reasons (beyond that's just the way it is) for forcing all `Quantity`s to be floating point. So I don't think this fixes #10079, but if desired #10079 could be closed as `wontfix` or similar.

mhvk (commented):
I'm not sure a warning is the best way forward. It is generally discouraged by our style guide (as warnings are often confusing to new users - as I think would be the case here) and more generally it might be better to instead try to explain more clearly that `Quantity` is meant to represent something physical, and that, therefore, it float by default. Now, obviously, this would be more useful if I actually had a good suggestion on how to do that...

mhvk:
@dstansby - yes, there is a good reason: suppose that we just kept integers, then
```
# q = 1 * u.m  would now do:
q = u.Quantity(1, 'm', dtype=int) 
q += 1*u.cm
# TypeError: Arguments cannot be cast safely to inplace output with dtype=int64
q /= 10
# TypeError: No loop matching the specified signature and casting was found for ufunc true_divide
```
I think this would cause a *lot* more bug reports from unsuspecting users...

Overall, I think using integer `Quantity` is a power-user case (and one I would discourage even then).

dstansby:
> ```
> # q = 1 * u.m  would now do:
> q = u.Quantity(1, 'm', dtype=int) 
> q += 1*u.cm
Can't you automatically cast to a float here, much like `1 + 1.0` gives `2.0` instead of raising an error?

> ```
> q /= 10
> # TypeError: No loop matching the specified signature and casting was found for ufunc true_divide
> ```
Again, I don't see why this would error, can't one either cast to a float or use normal integer division rules?

mhvk:
@dstansby - maybe it is simply the text for `copy` that should be changed? Right now, that is just copied from `np.array` and even mentions "explicitly given `dtype`" which is actually wrong...

Also, thinking again, I think it is fine to have text similar to what you have here, but just put it under `Notes` rather than make it a warning - a red box in the documentation still feels like overkill to me.

mhvk:
For the in-place: the problem becomes clearer if `q` is an array, what should `q[:2] /= 10` do? At least in this part I think we are stuck with the numpy interface.

astropy/units/quantity.py
```diff
@@ -384,6 +386,11 @@ def __new__(cls, value, unit=None, dtype=None, copy=True, order=None,
 
         # by default, cast any integer, boolean, etc., to float
         if dtype is None and value.dtype.kind in 'iuO':
+            if not copy:
+                warnings.warn(f'Casting {value.dtype} to float, which will '
+                              'implicitly make a copy of the data. '
+                              'This can be prevented by explicitly setting the '
+                              'dtype argument.')
```
dstansby, line 393: Thoughts on raising a warning here? I think it's good if a user has explicitly specified `copy=False` to warn them if a copy is being made.

mhvk: If we go that route, we'd need some way to avoid the warning for things like `np.arange(10) << u.m`...

dstansby: It should be fine to do casting in `__lshift__` before handing to `__class__(self, other, copy=False, subok=True)`.

Do you think having the warning is useful? If so I can go ahead and modify `__lshift__` (and I think there's a couple of other places that need checking too)

mhvk: I'm not keen in duplicating the `Quantity` casting behaviour inside a method on `UnitBase` - it is non-trivial (see further down). 

I think if we go this route, I'd prefer a different `copy` option (perhaps `None` for raise instead of converting). See https://github.com/astropy/astropy/issues/10079#issuecomment-605454494

Maybe it is best to keep this separate, though: I think a change in the docstring to clarify the current behaviour is good regardless!

dstansby:
Fab, I've slimmed down to adding a comment to the Notes.

mhvk (approved):
Thanks, let's get this in! We can think about a possible way to warn separately.

